PR #55, opened 2022-01-03
Update README.md

nilsmartel:
Given the previous statement, I can only hope, that this is what was meant. Godspeed developing this library, great work so far

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,7 +2,7 @@
 
 Most of these examples are run through webview so you don't need the Dioxus CLI installed to preview the functionality.
 
-These examples are fully-fledged micro apps. They can be ran with the `cargo run --example XYZ`
+These examples are fully-fledged micro apps. They can be ran with the `cargo run --features desktop --example XYZ`
 
 | Example                                             | What it does                                | Status |
 | --------------------------------------------------- | ------------------------------------------- | ------ |


jkelleyrtp:
> Given the previous statement, I can only hope, that this is what was meant. Godspeed developing this library, great work so far

Thanks!

I haven't figured out how to get the "desktop" feature enabled by default when running examples, so you have to manually specify the `desktop` feature. Do you know if there's a way to always force the `desktop` feature? Maybe something in `.cargo/config`?

nilsmartel:
I am rather unfamiliar with these settings.

Though I think for something like a GUI library it makes sense to turn on all features by default and manually turning them of can be done as an optimization. 

Is it possible to structure code like this in general?

jkelleyrtp:
> I am rather unfamiliar with these settings.
> 
> Though I think for something like a GUI library it makes sense to turn on all features by default and manually turning them of can be done as an optimization.
> 
> Is it possible to structure code like this in general?

https://github.com/rust-lang/cargo/issues/4663

I'm not too sure if it's possible. At least, cargo doesn't support that, so we'd need to create a new "examples" crate in the workspace that has all those features enabled. So instead of having to pass the ```--features desktop``` flag, you'd need to CD into the examples directory to run any of the examples.

I know repositories that do both.

Though, we might be able to patch .cargo/config.toml with a custom runner that always has those features.

nilsmartel:
I'd prefer to to enable some/all features by default, it seems to be possible: https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature

I think it makes "just getting started" with this crate tremendously easier and disabling features is still a useful fine-tuning step.

This would also make running the examples easier.


Option 2:

Maybe create a custom run command?
We are _not allowed_ to overwrite `cargo run --example`, but (the way I understood it) we're allowed to create a new command like 
`cargo example X` <=> `cargo run --example --features desktop X`

jkelleyrtp:
Update!

I did *not* know you could do this, but you can infact put the current crate as its dev dependency with features enabled.

```rust
[dev-dependencies]
dioxus = { path = ".", features = ["desktop", "ssr", "router"] }
```
I'm going to close your PR and then make a new one that implements this change and simplifies the running process.

Good suggestion!

PR #61, opened 2022-01-04
Fix errors in the suggested index.html template

chris-morgan:
What was there was an abomination that would not behave at all properly. The changes I’ve made are extremely conservative. Myself, I’d get rid of the `<html>`, `<head>`, `</head>`, `<body>`, `</body>` and `</html>` lines as superfluous, and remove the space inside the div, which I certainly hope isn’t load-bearing.


jkelleyrtp:
Thank you!

PR #74, opened 2022-01-06
fix: calculator example

mrxiaozhuox:
The Example `calculator` cannot be use, so I have to fixed the bug, now it's can be run to calculate data.

Some details have yet to be refined.

### DIFF
--- a/examples/calculator.rs
+++ b/examples/calculator.rs
@@ -18,12 +18,56 @@ fn main() {
     });
 }
 
-fn app(cx: Scope) -> Element {
-    let cur_val = use_state(&cx, || 0.0_f64);
-    let operator = use_state(&cx, Option::default);
-    let display_value = use_state(&cx, String::new);
+fn clac_val(val: String) -> f64 {
 
-    let input_digit = move |num: u8| display_value.modify().push_str(num.to_string().as_str());
+    let mut result = 0.0_f64;
+    let mut temp = String::new();
+    let mut operation = "+".to_string();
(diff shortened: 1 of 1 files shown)


examples/calculator.rs
```diff
@@ -18,23 +18,67 @@ fn main() {
     });
 }
 
+fn clac_val(val: String) -> f64 {
```
autarch, line 21: This should be `calc_val`.

mrxiaozhuox: I changed it.

mrxiaozhuox:
all input will display, and if click some button like: `%` `-/+` `=` then the program will calculate this expression and create a new `value-expression`. so the user can know calculate process, not just display a `0`.

> The original demo code didn't work.

jkelleyrtp:
Thank you so much! I can't believe how broken the original was 😳

Can you clean up/remove the comments and move calc_value after the UI code? Just want to make sure it's polished. Once it's cleaned up a bit, just ping me and we can merge it.

mrxiaozhuox:
> Thank you so much! I can't believe how broken the original was 😳
> 
> Can you clean up/remove the comments and move calc_value after the UI code? Just want to make sure it's polished. Once it's cleaned up a bit, just ping me and we can merge it.

OK: I have formatted this code, and move the `calc_val` after UI code.

jkelleyrtp:
Looks great! Thanks for doing this, the calculator works well :)

PR #89, opened 2022-01-07
examples: simplfy running them by enabling features automatically

jkelleyrtp:
cc #55 

This turns on the desktop, router, and ssr features by default for examples.

  @nilsmartel can you make sure this functions the way you expect it to?

```
git pull
git checkout jk/simplify-example-run
cargo run --example readme
```


nilsmartel:
I will check this out over the weekend. 
That like of code looks like it came straight out of a book of forbidden charms and black magic. 

It's brilliant, given it works

nilsmartel:
works splendid

PR #94, opened 2022-01-08
Login Form Example

JtotheThree:
Creating a login form example to demonstrate a common example using a form to login and make an async call to an external api.

A couple of caveats:
Only works on web, but to keep it consistent as router changes are added in left it as dioxus::desktop.
This works on the current version of the router on web, but is bound to change as the router is updated.

I can add something quick to the guide as well, it looks like a lots changed in there. Any suggestions on where to add it?

### DIFF
--- a/examples/login_form.rs
+++ b/examples/login_form.rs
@@ -1,0 +1,89 @@
+//! This example demonstrates the following:
+//! Futures in a callback, Router, and Forms
+
+use dioxus::events::*;
+use dioxus::prelude::*;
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
Thank you for the PR! Sorry it's still outstanding - we need to get the router working locally so it doesn't just crash.

It depends on Gloo which doesn't fallback to anything if not on wasm - which we need to fix!

jkelleyrtp:
Thank you!!

Issue #96, opened 2022-01-08
`element` disabled attribute


mrxiaozhuox:
Cannot use `disabled` attribute like:

```
// cx.props.disabled = true
button {
  disabled: "{cx.props.disabled}",
  "Button",
}
```

the `button` state always be `disabled`.

but you can use this code to do it:
```
let is_disabled = if disabled { "disabled" } else { "" };
rsx!{
   "{is_disabled}": "true"
}
```

mrxiaozhuox:
```rust
button {
    // ...
    disabled: Bool,
    // ...
}
```

PR #97, opened 2022-01-08
Feat: arbitrary custom attribute names with formatting

jkelleyrtp:
This lets you implement boolean attributes with your own custom logic.

cc #97 #96 

The "disabled" example is a great showcase of when this ability is useful:

```rust
    let disabled = use_state(&cx, || false);

    let button_disabled = if *disabled { "disabled" } else { "nnn" };

    cx.render(rsx! {
        div {
            button {
                onclick: move |_| disabled.set(!disabled.get()),
                "click to " [if *disabled {"enable"} else {"disable"} ] " the lower button"
            }

            button {
                "{button_disabled}": "true",
                "lower button"
            }
        }
    })
```

### DIFF
--- a/examples/custom_attributes.rs
+++ b/examples/custom_attributes.rs
@@ -1,0 +1,27 @@
+//! Use custom formatting in attributes to set attributes
+
+use dioxus::prelude::*;
+
+fn main() {
+    dioxus::desktop::launch(app);
+}
+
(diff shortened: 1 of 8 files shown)


jkelleyrtp:
We will support a feature like this with attribute merging. I don't think this PR is the right way to solve problems like this.

PR #98, opened 2022-01-08
feat: `disabled` attribute support

mrxiaozhuox:
#97 is not a good solution for this problem: #96

I change the code can be use like:

```
button {
    disabled: "",
    "Test"
}
```

if the `disabled` is a empty str, `disabled=false` else `disabled=true`

update[feat]: `dioxus-web` && `dioxus-desktop`


mrxiaozhuox:
#97 error:

```
InvalidCharacterError: The string contains invalid characters.
```

PR #101, opened 2022-01-08
CI: Enable clippy

alexkirsz:
This PR enables clippy and applies a few fixes to make it lint successfully. See additional comments below.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
           override: true
       - uses: Swatinem/rust-cache@v1
       - run: sudo apt-get update
-      - run: sudo apt install libwebkit2gtk-4.0-dev libappindicator3-dev libgtk-3-dev libappindicator3-dev
+      - run: sudo apt install libwebkit2gtk-4.0-dev libappindicator3-dev libgtk-3-dev
       - uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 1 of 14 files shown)


.github/workflows/main.yml
```diff
@@ -55,21 +55,21 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: Swatinem/rust-cache@v1
+      - run: sudo apt-get update
+      - run: sudo apt install libwebkit2gtk-4.0-dev libappindicator3-dev libgtk-3-dev libappindicator3-dev
```
alexkirsz, line 70: Clippy would eventually fail when checking the code looking for the 'pango' package, so I added all the dependencies we need for compiling on ubuntu-latest.

packages/core/src/scopes.rs
```diff
@@ -745,6 +745,7 @@ impl ScopeState {
     ///     use_hook(|| Rc::new(RefCell::new(initial_value())))
     /// }
     /// ```
+    #[allow(clippy::mut_from_ref)]
```
alexkirsz, line 748: I'm not qualified to determine whether the mut_from_ref usage here is sound so I allowed it.

packages/router/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(warnings)]
```
alexkirsz, line 1: Since this package is very much WIP, I didn't see the point in allowing or fixing each instance individually. We'll need to remove this once the package is closer to ready.

packages/ssr/src/lib.rs
```diff
@@ -38,6 +38,7 @@ impl SsrRenderer {
     }
 }
 
+#[allow(clippy::needless_lifetimes)]
```
alexkirsz, line 41: I believe the below could be 
```rust
pub fn render_lazy(f: LazyNodes<'_, '_>) -> String {
```
But since we name the lifetimes in the Safety comment below, I allowed it.

jkelleyrtp (approved):

.github/workflows/main.yml
alexkirsz: Done!

PR #106, opened 2022-01-10
docs: improve components and elements

jkelleyrtp:


docs/guide/src/elements/vnodes.md
```diff
@@ -81,13 +81,13 @@ let name = "Bob";
 rsx! ( "hello {name}" )
 ```
 
-Unfortunately, you cannot drop in arbitrary expressions directly into the string literal. In the cases where we need to compute a complex value, we'll want to use `format_args!` directly. Due to specifics of how the `rsx!` macro (we'll cover later), our call to `format_args` must be contained within curly braces *and* square braces.
+Unfortunately, you cannot drop in arbitrary expressions directly into the string literal. In the cases where we need to compute a complex value, we'll want to use `format_args!` directly. Due to specifics of how the `rsx!` macro (we'll cover later), our call to `format_args` must be contained within  square braces.
```
autarch, line 84: I think you mean curly braces, not square.

ajstrand:
I read through the new docs and I really have nothing to add. I think a lot of the additions you made help introduce folks to a React/component based way of thinking.

jkelleyrtp:
Merged what I had for now, will come back later and finish what's missing.

PR #118, opened 2022-01-11
Fix typo in docs

b12f:
This breaks the page currently; https://dioxuslabs.com/guide/setup.html


jkelleyrtp:
You're right! Good catch - sorry about that. Will redeploy in a bit.

PR #119, opened 2022-01-11
docs: allow build core_reference/children example

yuniruyuni:
It needs lifetime parameters If you want to build it so I have added lifetime parameters.


jkelleyrtp:
You're right! Good catch! Thank you!

PR #121, opened 2022-01-12
feat: unify web and desktop interpreter with the same typescript codebase

### LINKED ISSUE #22: internals: unify webview and web interpreter
Currently the web and webview share different interpreter code - it would be nice to move to either a completely wasm-based interpreter or a pure JavaScript/TypeScript interpreter. The appropriate steps have been taken - JS code refactored out and event system is now configured to always use serialized events.

jkelleyrtp:
This PR unifies the web and desktop interpreter so we only have one codebase to maintain for all JS targets.

It also moves all Dom manipulation into JS which has some light performance improvements.

It also sets us up a TypeScript build so we can strongly type the interpreter code now - which has fixed a few bugs.

closes #22 

Would also make any future changes to the interpreter code easier to maintain since they're only in one place


jkelleyrtp:
Something I have discovered here is that wasmbindgen generates an extra directory of modules (in addition to the .wasm file). If you're using trunk or wasm-pack, then I don't think this is an issue. However, it's a bit strange to see not *just* your .wasm file when generating a Dioxus app, but also the interpreter too.

PR #122, opened 2022-01-13
Add `Chinese` Language `Readme` and add some info into README.md

mrxiaozhuox:
In this PR I add a `Chinese` Readme


mrxiaozhuox:
@jkelleyrtp the link problem is fixed.

PR #127, opened 2022-01-16
fix: allow eventhandler to derive default

### LINKED ISSUE #125: `EventHandler` default trait
I think the `EventHandler` must have a `Default` tarit, because I think its optional.

jkelleyrtp:
Fixes #125 

Also allows handlers to properly clean up their memory (an oversight)


mrxiaozhuox (commented):
LGTM

PR #128, opened 2022-01-16
feat: add "optional" flag for props

jkelleyrtp:
Adds the flag "optional" to props that combines default and strip_option

### DIFF
--- a/examples/optional_props.rs
+++ b/examples/optional_props.rs
@@ -1,3 +1,5 @@
+#![allow(non_snake_case)]
+
 //! Example: README.md showcase
 //!
 //! The example from the README.md.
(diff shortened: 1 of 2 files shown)


mrxiaozhuox:
so the `optional`  modifiers have not been merged?
now just can use `default`

jkelleyrtp:
> so the `optional` modifiers have not been merged?
> now just can use `default`

Yes - "optional" is just "default" + "strip_option" which you can do today. This updates the code so the docs match up

mrxiaozhuox:
I think we need declare the version you can use in document, because in document have many feature is not complete. we need tell them is unfinished.

mrxiaozhuox (commented):
LGTM

PR #132, opened 2022-01-17
feat: split out rsx into its own crate

jkelleyrtp:
This PR moves separates out the rsx macro and its infrastructure to allow use in an rsx autoformatter and accompanying vscode extension.

This will be part of our 0.1.8 release train alongside the router, optional props, better docs, and more.

Right now, the crate is not released on crates.io but is being relied upon in the fork of the CLI


jkelleyrtp:
... more to come

PR #133, opened 2022-01-18
The desktop app will default open link in browser

mrxiaozhuox:


mrxiaozhuox:
wait please, I will add a `remove-browser-open` attribute

mrxiaozhuox:
ok

packages/desktop/src/index.js
```diff
@@ -399,6 +401,34 @@ class Interpreter {
...
+              } else {
+                window.location = target.getAttribute("browser-href");
+              }
+
+            })
+
+          } else {
+            node.setAttribute(name, value);
+          }
+
```
jkelleyrtp, line 431: This is a good start, but I'd like to see this code moved into the event handler instead of setattribute.

We use listener multiplexing - so there's only one `click` handler for the whole app. Adding more handlers to the app manually might interfere with our global handling.

Instead of having this code *here*, we need it to either go in the Rust side where we receive events from webview, or in the `AddEventListener` code where we create the global listeners.

Additionally, instead of the "browser-open" attribute, we should just look at the URL itself and make the decision automatically. If it starts with a protocol (ie wry:// or dioxus:// - or anything that's not http:// or https://) then we want the url to move through the handler infrastructure.

Also - any `onclick` in Desktop should have "default prevented" - that way we don't need to mess with `window.location`. By calling "prevent default", the history will be preserved properly, so we can still provide links into a desktop app from the outside.

To recap:
- Move rpc.call into the "add event listener"
- Always call "prevent-default" for `onclick` events in desktop
- Parse URL instead of "browser-open" flag

packages/desktop/src/lib.rs
```diff
@@ -187,6 +187,25 @@ pub fn launch_with_props<P: 'static + Send>(
...
+                                    if arr[0].is_object() {
+                                        let temp = arr[0].as_object().unwrap();
+                                        if temp.contains_key("href") {
+                                            let url = temp.get("href").unwrap().as_str().unwrap();
+                                            match webbrowser::open(url) {
+                                                Ok(_) => {}
+                                                Err(e) => {
+                                                    log::trace!("Open browser error: {:?}", e)
+                                                }
+                                            }
```
jkelleyrtp, line 204: Use the err destructuring pattern here.
```
if let Err(e) =  webbrowser::open(url) {
    log::trace!("Open browser error: {:?}", e)
}
```

packages/desktop/src/lib.rs
```diff
@@ -187,6 +187,25 @@ pub fn launch_with_props<P: 'static + Send>(
...
                                 let _ = proxy.send_event(UserWindowEvent::Update);
                             }
+                            "browser_open" => {
+                                let data = req.params.unwrap();
+                                log::trace!("Open browser: {:?}", data);
+                                if data.is_array() {
+                                    let arr = data.as_array().unwrap();
+                                    if arr[0].is_object() {
+                                        let temp = arr[0].as_object().unwrap();
+                                        if temp.contains_key("href") {
```
jkelleyrtp, line 197: I'm concerned about all the unwrapping and bare indexing. Can you refactor this to use `.map` or `if let Some(_) = `?

There's still a few unwraps laying around that cause issues in other parts of Dioxus that should be `if let Some(_)`

examples/link.rs
```diff
@@ -0,0 +1,28 @@
...
+            }
+            p {
+                a {
+                    href: "http://dioxuslabs.com/",
+                    "browser-open": "false",
+                    onclick: |_| {
+                        println!("Hello Dioxus");
+                    },
+                    "custom event link",
+                }
```
jkelleyrtp, line 24: instead of "browser-open", let's just hijack "prevent_default" . That way the "default" behavior in desktop is to open links remotely.



so the second example should be 

```rust
                a {
                      href: "http://dioxuslabs.com/",
                      prevent_default: "true",
                      onclick: |_| {
                          println!("Hello Dioxus");
                      },
                      "custom event link",
                  }
```

jkelleyrtp (commented):
This is a good start, but I'd like to see this code moved into the event handler instead of setattribute.

We use listener multiplexing - so there's only one click handler for the whole app. Adding more handlers to the app manually might interfere with our global handling.

Instead of having this code here, we need it to either go in the Rust side where we receive events from webview, or in the AddEventListener code where we create the global listeners.

Additionally, instead of the "browser-open" attribute, we should just look at the URL itself and make the decision automatically. If it starts with a protocol (ie wry:// or dioxus:// - or anything that's not http:// or https://) then we want the url to move through the handler infrastructure.

Also - any onclick in Desktop should have "default prevented" - that way we don't need to mess with window.location. By calling "prevent default", the history will be preserved properly, so we can still provide links into a desktop app from the outside.

To recap:

- Move rpc.call into the "add event listener"
- Always call "prevent-default" for onclick events in desktop
- Parse URL or "prevent_default" attribute instead of "browser-open" flag

examples/link.rs
mrxiaozhuox: got it!

packages/desktop/src/index.js
```diff
@@ -356,6 +357,23 @@ class Interpreter {
...
         }
 
+        if (event.type == "submit") {
+          event.preventDefault();
+        }
+        
+        if (event.type == "click") {
+          event.preventDefault();
+          if (should_prevent_default !== `onclick`) {
+            console.log(event.target.getAttribute("href"));
```
jkelleyrtp, line 367: Just remove this console log and then everything looks pretty good.

I would also make a check that "href" exists on the `<a>` tag so we don't send null and panic the Rust code.

jkelleyrtp (changes requested):
Add a check that href isn't null/undefined and then remove the console.log and we're good to go :)

jkelleyrtp:
Looks great :) Thank you!

PR #138, opened 2022-01-19
fix: `dog-app` breed change

mrxiaozhuox:
change the breed but the image will not reload.


jkelleyrtp:
Thanks! We should make a new use_future hook that does this automatically.

PR #146, opened 2022-01-22
Add gap and row_gap to style_trait_methods

FruitieX:
`gap` is quite useful now that it's supported in flex layouts by all major browsers.


jkelleyrtp:
Thank you! :)

PR #147, opened 2022-01-22
Pass form values as a hashmap on `oninput`/`onsubmit`

### LINKED ISSUE #117: event `onsubmit` field `value` problem
`onsubmit` FormData value is just can get the `form` element attr: `value`, but is not for the all `form-data`, like:
```
<form id="x_form" value="XXX_VALUE">
    <input name="test-1" />
    <input name="test-2" />
</form>
```

for the submit, we need get all form in this element, like `test-1` and `test-2` value, but its just can get the value of "XXX_VALUE"

jkelleyrtp:
This PR closes #117 , adding the "values" field to oninput/onsubmit


examples/form.rs
```diff
@@ -0,0 +1,24 @@
...
+fn main() {
+    dioxus::desktop::launch(app);
+}
+
+fn app(cx: Scope) -> Element {
+    cx.render(rsx! {
+        div {
+            h1 { "Form" }
+            form {
+                oninput: move |ev| println!("{:?}", ev.values),
```
mrxiaozhuox, line 17: why we use thie oninput event? I think there have a special event: `onsubmit` for form.

jkelleyrtp: Just updated to showcase how both onsubmit and oninput support "values"

mrxiaozhuox (approved):
Nice!!

PR #150, opened 2022-01-24
Feat: add console_error_panic_hook by default for wasm

jkelleyrtp:
Most users who want to build a Dioxus app *also* want stacktraces on panics. This PR just adds that by default, improving the onboarding DX for users so they can actually see *why* their app panicked.


mrxiaozhuox (approved):
Nice!

PR #158, opened 2022-01-25
Connect the onchange listener for the Router

jkelleyrtp:
This PR wires up onchange listeners for the Router (at the top level) so you can easily connect things like modifying the title/head.

Also ccing @autarch since you spent a while on the router


autarch (approved):
This looks reasonable to me, though I'm not sure I understand every bit of it.

jkelleyrtp:
> This looks reasonable to me, though I'm not sure I understand every bit of it.

We're just sending updates into the update queue and regenerating the root. Dioxus knows to memoize children via their pointers, so the operation is still cheap.

PR #164, opened 2022-01-27
New use_effect hook

jkelleyrtp:


jkelleyrtp:
This is being superseded by #161 which is upgrading both use_coroutine and use_future.

use_future is the new "effectual" hook, and brings with it use_effect. use_effect is basically just use_future that doesn't re-render the component when its finished.

PR #168, opened 2022-01-28
fix: use_route should subscribe to changes to the route

jkelleyrtp:
This PR makes it so the `use_route` command subscribes the component to changes to the URL.

### DIFF
--- a/packages/router/src/hooks/use_route.rs
+++ b/packages/router/src/hooks/use_route.rs
@@ -8,6 +8,7 @@ use crate::RouterService;
 /// This struct provides is a wrapper around the internal router
 /// implementation, with methods for getting information about the current
 /// route.
+#[derive(Clone)]
 pub struct UseRoute {
     router: Rc<RouterService>,
 }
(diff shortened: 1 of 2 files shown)


packages/router/src/hooks/use_route.rs
```diff
@@ -74,10 +74,28 @@ impl UseRoute {
...
-    UseRoute { router }
+pub fn use_route(cx: &ScopeState) -> &UseRoute {
+    &cx.use_hook(|_| {
+        let router = cx
+            .consume_context::<RouterService>()
+            .expect("Cannot call use_route outside the scope of a Router component");
+
+        router.subscribe_onchange(cx.scope_id());
+
+        UseRouteInner {
```
autarch, line 85: I think calling this `UseRouteInner` is confusing when it's not contained in a `UseRoute`. Maybe the existing `UseRoute` needs to be renamed to Inner and this one should be `UseRoute`?

autarch (changes requested):
Just a question about naming.

packages/router/src/hooks/use_route.rs
jkelleyrtp: Hmmmm maybe something like `UseRouteLIstener` since its only job is to be dropped when the component is unmounted?

autarch: That makes sense.

Issue #160, opened 2022-01-26
Some questions about handling events


bingryan:
hello,i'm writing an app using dioxus, but I have some difficulties:

1. How to handle close windows event

```
button { class: "bg-red-500 m-1 w-3 h-3 rounded-full",
    onclick: move |evt| {
        // TODO:  app exit
         println!("close windows:{:?}",evt);
    }
}
```

2.  How to handle system events when the app is hidden.

I can only handle events of a certain tag now, but sometimes  have to deal with system some shortcut keys

3.  Change CSS properties on event(click), looks like:

```
$('#image').click(function() {
  $('#foo').addClass('myClass');
});
```

t1m0t:
You'll have something like this:
```rust
let is_closed = use_state(&cx, move || false);

cx.render(rsx!(if *is_closed {
    rsx!("")
} else {
    rsx!(button {
        class: "bg-red-500 m-1 w-3 h-3 rounded-full",
        onclick: move |_| { is_closed.set(true) }
    })
}))
```

bingryan:
> You'll have something like this:
> 
> ```rust
> let is_closed = use_state(&cx, move || false);
> 
> cx.render(rsx!(if *is_closed {
>     rsx!("")
> } else {
>     rsx!(button {
>         class: "bg-red-500 m-1 w-3 h-3 rounded-full",
>         onclick: move |_| { is_closed.set(true) }
>     })
> }))
> ```

thankyou @t1m0t , but how to make current app exit when  the state `is_closed = true`

mrxiaozhuox:
> thankyou @t1m0t , but how to make current app exit when the state `is_closed = true`

you can call the `exit` to exit your app.

bingryan:
> > thankyou @t1m0t , but how to make current app exit when the state `is_closed = true`
> 
> you can call the `exit` to close your app.

i'm write the logic :`Dialog Box With Maximize, Minimize & Close Button`

```

div { class: "p-1 flex items-center justify-center",
    button { class: "bg-red-500 m-1 w-3 h-3 rounded-full",
        onclick: move |evt| {
             println!("Close:{:?}",evt);
        }
    }
    button { class: "bg-yellow-500 m-1 w-3 h-3 rounded-full",
        onclick: move |evt| {
             println!("Minimize:{:?}",evt);
        }
    }
    button { class: "bg-green-500 m-1 w-3 h-3 rounded-full",
        onclick: move |evt| {
[...]

PR #180, opened 2022-01-30
`DesktopContext` Wrap & Inject

mrxiaozhuox:
`DesktopContext` provide some api to control window, like `drag_window` `minimized` `maximized`.

### DIFF
--- a/examples/borderless.rs
+++ b/examples/borderless.rs
@@ -1,5 +1,4 @@
 use dioxus::prelude::*;
-use dioxus_desktop::desktop_context::DesktopContext;
 
 fn main() {
     dioxus::desktop::launch_cfg(app, |cfg| {
@@ -8,16 +7,13 @@ fn main() {
 }
 
 fn app(cx: Scope) -> Element {
-    let desktop = cx.consume_context::<DesktopContext>().unwrap();
-
-    let drag = desktop.clone();
-    let close = desktop.clone();
+    let desktop = dioxus::desktop::desktop_context::use_desktop_context(&cx);
 
     cx.render(rsx!(
         link { href:"https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css", rel:"stylesheet" }
         header {
             class: "text-gray-400 bg-gray-900 body-font",
-            onmousedown: move |_| drag.drag_window(),
+            onmousedown: move |_| desktop.drag_window(),
             div {
                 class: "container mx-auto flex flex-wrap p-5 flex-col md:flex-row items-center",
                 a { class: "flex title-font font-medium items-center text-white mb-4 md:mb-0",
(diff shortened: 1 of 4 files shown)


mrxiaozhuox:
#160 I think this PR can help you.

examples/borderless.rs
```diff
@@ -0,0 +1,42 @@
...
+    dioxus::desktop::launch_cfg(app, |cfg| {
+        cfg.with_window(|w| w.with_title("BorderLess Demo").with_decorations(false))
+    });
+}
+
+fn app(cx: Scope) -> Element {
+    let desktop = cx.consume_context::<DesktopContext>().unwrap();
+
+    let drag = desktop.clone();
+    let close = desktop.clone();
```
jkelleyrtp, line 14: It would be better if we had a hook called "use_desktop_context()"

```rust
fn use_desktop_context(cx: &ScopeState) -> &Rc<DesktopContext> {
    cx.use_hook(|_| cx.consume_context::<DesktopContext>()).as_ref().unwrap()
}
```

That way desktop context comes out as an &Rc which can be borrowed into the callbacks without issue.

jkelleyrtp (changes requested):
Looks great! Really awesome feature and great job on the code. Very excited to have this demo.

One change:

To make it easier for people to get the desktop context, lets provide a hook that does it for them. Maybe something like `use_window`.

jkelleyrtp:
Just a nit - can we standardize some language on the methods?

The methods added are
- drag window
- minimize
- maximize
- focus


Should each of these methods be `xxx_window`? like `minimize_window`?

If the hook was `use_window` then calling `drag_window` would be redundant. Does the hook provide context just for the current window or all windows (eventually)? Let's make sure whatever we go with, our naming is consistent, keeping in mind that we want to support multiwindow in the near future without breaking the APIs used here.

mrxiaozhuox:
> Just a nit - can we standardize some language on the methods?
> 
> The methods added are
> 
> * drag window
> * minimize
> * maximize
> * focus
> 
> Should each of these methods be `xxx_window`? like `minimize_window`?
> 
> If the hook was `use_window` then calling `drag_window` would be redundant. Does the hook provide context just for the current window or all windows (eventually)? Let's make sure whatever we go with, our naming is consistent, keeping in mind that we want to support multiwindow in the near future without breaking the APIs used here.
> 
> I think it would be clearer to specify that this hook applies to the current window (ie use_window gets the current window context) and then have the methods be just `drag`, `minimize`, `maximize`, and `focus`, since you'll likely be calling `window.drag()` or `window.minimize()` rather than `desktop.minimize()`. It seems that `desktop.minimize()` minimize _all_ the windows.

ok

jkelleyrtp (approved):
Looks great! Good job on the code!

PR #181, opened 2022-01-30
Fix the license field in Cargo.toml to be valid SPDX

autarch:
There's a specific format for this field, because it's intended to be machine-readable.


mrxiaozhuox (approved):

autarch:
I also threw in the license files, since that's what the Rust core does.

jkelleyrtp:
Looks good, thanks! I made sure to read through it :)

Issue #186, opened 2022-01-31
Improved CI


jkelleyrtp:
## Specific Demand

CI should be doing a lot more heavy lifting. bors might be too heavy for us right now, but we should at least configure things like

- [x] better test suite
- [ ] visual CI
- [ ] performance tracking
- [ ] automatic app screenshots
- [ ] automatic example deployments
- [x] automatic docsite deployments
- [x] run all examples

t1m0t:
I believe we should distinguish between build and runtime tests. Build tests is easily catchable whereas runtime tests should be run in a browser environment.
In this [doc](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html) we have some hints on how to perform these. 

disclaimer: inspired by yew ci/di pipeline

I would then propose this example for testing runtime:
```rust
use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
// import necessary dioxus libs
use dioxus_core::prelude::*;
use dioxus_core_macro::*;
use serde::{Deserialize, Serialize};

wasm_bindgen_test_configure!(run_in_browser);

// for each test case
#[wasm_bindgen_test]
fn some_desc_of_the_test() {
    fn main() {
        console_error_panic_hook::set_once();
        wasm_logger::init(wasm_logger::Config::new(log::Level::Debug));
        dioxus_web::launch(APP);
    }

    static APP: Component = |cx| {
        cx.render(rsx! {
           // some code
        })
    };

    // defining some components

    main();
    // perform some tests with gloo_utils or other for instance in this section:
    // assert(...);
}
```
The tests should be located in a test folder in each packages along with a `Makefile.toml` with this for instance:
```
[tasks.test]
extend = "core::wasm-pack-base"
command = "wasm-pack"
args = [
    "test",
    "@@split(DIOXUS_TEST_FLAGS, )",
    "--",
    "--features",
    "${DIOXUS_TEST_FEATURES}",
]
```

In the `Cargo.toml` of the package, we should include at minima:
```
[dev-dependencies]
wasm-bindgen-test = "0.3"
```

In the root repo, we should have a `Makefile.toml` with for instance:
```
[config]
min_version = "0.32.4"
default_to_workspace = false

[env]
CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
CARGO_MAKE_CLIPPY_ARGS = "-- --deny=warnings"

[config.modify_core_tasks]
private = true
[...]

Issue #185, opened 2022-01-31
Automated UI/UX testing


jkelleyrtp:
## Specific Demand

We should have a suite of automated tests and a code coverage report so everything you can do with Dioxus is thoroughly tested.

Areas of testing:
- [x] Router
- [ ] Hooks
- [ ] Shared State
- [x] Core
- [x] Desktop
- [x] Web
   - [x] hydration
- [x] SSR
- [x] TUI
- [x] Readme/Docs are up to date

autarch:
For the Router ...

* Route matching against paths
* Nested routes - `Route` inside a `Route`
* All the methods on the `UseRoute` struct - query, segment methods, current_location
* The onchange subscriber stuff

t1m0t:
I believe we should distinguish between build and runtime tests. Build tests is easily catchable whereas runtime tests should be run in a browser environment.
In this [doc](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html) we have some hints how to perform these. 

disclaimer: inspired by yew ci/di pipeline

I would then propose this example for testing runtime:
```rust
use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
// import necessary dioxus libs
use dioxus_core::prelude::*;
use dioxus_core_macro::*;
use dioxus_html as dioxus_elements;
use dioxus_router::*;
use serde::{Deserialize, Serialize};

wasm_bindgen_test_configure!(run_in_browser);

// for each test case
#[wasm_bindgen_test]
fn some_desc_of_the_test() {
    fn main() {
        console_error_panic_hook::set_once();
        wasm_logger::init(wasm_logger::Config::new(log::Level::Debug));
        dioxus_web::launch(APP);
    }

    static APP: Component = |cx| {
        cx.render(rsx! {
            Router {
                onchange: move |route| log::info!("route changed to {}", route),
                Route { to: "/", Home {} }
                Route { to: "blog"
                    Route { to: "/", BlogList {} }
                    Route { to: ":id", BlogPost {} }
                }
            }
        })
    };

    fn Home(cx: Scope) -> Element {
        cx.render(rsx! {
            div {
                h1 { "Home" }
            }
        })
    }

    fn BlogList(cx: Scope) -> Element {
        cx.render(rsx! {
            div {

            }
        })
    }

    fn BlogPost(cx: Scope) -> Element {
        let id = use_route(&cx).segment::<usize>("id")?;

[...]

PR #190, opened 2022-02-01
wip: Makefile setup

t1m0t:
Added configuration files to perform `cargo make tests` so that it will run the tests in all packages where the root Makefile has a link with and allow runtime tests inside a browser.


.vscode/settings.json
```diff
@@ -4,4 +4,5 @@
     "desktop",
     "router"
   ],
+  "editor.formatOnSave": false
```
jkelleyrtp, line 7: https://stackoverflow.com/questions/44831313/how-to-exclude-file-extensions-and-languages-from-format-on-save-in-vscode

let's do it just for the toml files

t1m0t: corrected

t1m0t:
wip for #185 and #186

jkelleyrtp (changes requested):
Everything looks good! But, what is the docker submodule doing here? Do we need it?

It doesn't seem to be used anywhere.

t1m0t:
> Everything looks good! But, what is the docker submodule doing here? Do we need it?
> 
> It doesn't seem to be used anywhere.

Yes for local testing before pushing. It is useful to execute tests locally on someone's machine, so that we are sure no errors would happen in git workflows. It saves us a lot of push in the remote repo that wouldn't work. It is not mandatory but it can be useful for anyone who wants to tests his changes before pushing it.

t1m0t:
> Everything looks good! But, what is the docker submodule doing here? Do we need it?
> 
> It doesn't seem to be used anywhere.

fixed it

jkelleyrtp (approved):
Thanks!!

PR #194, opened 2022-02-02
More conservative node garbage collection

jkelleyrtp:
We got scopes right in #182 but nodes are struggling in some cases. This PR fixes that.


jkelleyrtp:
The approach is #195 is way better

PR #200, opened 2022-02-04
fix: remove code generation entirely

### LINKED ISSUE #173: todomvc.rs checkbox didn't work
Hi there, have you noticed the examples/todomvc.rs seemed not to function, each item could not be checked, thus there is no way to make a item to be completed.

jkelleyrtp:
It's a nice idea to have tsc generate our js binding files, but it leads to weird changes in PRs and noticeably slows down compilation.

This PR moves the binding code into the jsinterpreter crate so we can service both desktop and web from the same crate and same single `.js` file.

It seems that other projects that use bindings like this just write all their bindings in JS with heavy linting, so that's what we're going to follow until it becomes impossible.

Also, as a side-effect, we've fixed #173


t1m0t:
we should rerun the tests (workflow has been fixed)

Jessidhia:
Amusingly the [v0.2 release notes](https://dioxuslabs.com/blog/release-020/) mention the interpreter is "now (...) written in TypeScript", but that seems to have no longer been the case for over a month :innocent:

PR #203, opened 2022-02-04
Feat: abstract the router on a per-platform basis and add docs

jkelleyrtp:
This PR adds docs to the router and makes it possible to work locally.

The local router is implemented simply as a list of strings with a configurable max stack.

I would like to add deep linking through something like "ondeeplink" but that won't make it into this PR unfortunately.


jkelleyrtp:
Found a fun bug in desktop where the order of edits applied to the desktop dom was wrong. 

The router has been a fantastic stress test of Dioxus :)

Issue #205, opened 2022-02-04
Improve Makefile tests


t1m0t:
The test process defined by the command `cargo make tests` is taking too much time and should be optimized.

PR #206, opened 2022-02-04
Improve test speed in the CI

t1m0t:
I modified the Makefile a bit so that it doesn't compile on every crate. Instead, it builds first then runs the tests (eventually build what is required for the tests). With this in place, tests in CI take much less time to proceed.
Added also a separate test in the Makefile will test with proper configuration for runtime tests.


t1m0t:
related to #205

PR #209, opened 2022-02-05
Update ZH_CN.md

mrxiaozhuox:
translate some new content, and change document link.


jkelleyrtp:
We should try to internationalize the guide/reference in this repo rather than linking out to another website, but it's fine for now.

mrxiaozhuox:
> We should try to internationalize the guide/reference in this repo rather than linking out to another website, but it's fine for now.

you mean add i18n for the doc site ?

PR #214, opened 2022-02-06
Code coverage setup

t1m0t:
Setting up code coverage locally with tarpaulin and adapt ci workflow


t1m0t:
#185

jkelleyrtp:
Is this ready for merge?

t1m0t:
There is still a lack of support for wasm so the coverage indicator is not perfect but still it is better than nothing. We should keep track of this issue in another ticket otherwise it is ok for merging imo.

PR #218, opened 2022-02-08
fix: update the sample

activeguild:
The sample in the guide and the sample in READE.md were outdated APIs, so we fixed them.


jkelleyrtp:
Thanks!!

PR #230, opened 2022-02-11
fix some uncaught error during runtime test

t1m0t:


jkelleyrtp:
Just noticed the failure. Thanks for fixing it :)

PR #231, opened 2022-02-11
Transparent window means transparent WebView

blemelin:
Hi,

Small contribution here. When creating a transparent Window, shouldn't it also set the `WevView` to be transparent automatically ? This is what this PR is about.

Unfortunately, I am wondering if someone somewhere would need the window to be transparent, but not the `WevView`. If that's the case, then this solution won't do...but really, why would someone need that kind of feature ?

Have a nice day!

### DIFF
--- a/packages/desktop/src/lib.rs
+++ b/packages/desktop/src/lib.rs
@@ -177,6 +177,7 @@ pub fn launch_with_props<P: 'static + Send>(
 
                 let mut webview = WebViewBuilder::new(window)
                     .unwrap()
+                    .with_transparent(cfg.window.window.transparent)
                     .with_url("dioxus://index.html/")
                     .unwrap()
                     .with_rpc_handler(move |_window: &Window, req: RpcRequest| {


jkelleyrtp:
Thanks!

mchaliadzinau:
`.with_transparent(true)` does not works for me on MacOS 13.0

I've tried with-decorations true and false, no difference.
By checking inspector, I see that html element is not even covering full view, so AFAIK this should not be issue with styles.

```
use dioxus::prelude::*;
use dioxus_desktop::{tao::dpi::PhysicalPosition, use_window, LogicalSize, WindowBuilder};

fn main() {
    dioxus_desktop::launch_cfg(app, make_config());
}

    let window = use_window(cx);

    cx.render(rsx! {
        div {
            width: "100%",
            height: "100%",
            border: "1px solid black",
            "This is an overlay!"
        }
    })
}

fn make_config() -> dioxus_desktop::Config {
    dioxus_desktop::Config::default()
        .with_window(make_window())
        .with_custom_head(
            r#"
<style type="text/css">
    html, body {
        height: 100px;
        border: 1px solid yellow;
        margin: 0;
        overscroll-behavior-y: none;
        overscroll-behavior-x: none;
        overflow: hidden;
        background-color: #f0f8ff4d;
    }
    #main, #bodywrap {
        height: 100%;
        margin: 0;
        overscroll-behavior-x: none;
        overscroll-behavior-y: none;
        background-color: #f0f8ff4d;
    }
</style>
"#
            .to_owned(),
        )
}

fn make_window() -> WindowBuilder {
    WindowBuilder::new()
        .with_transparent(true)
        .with_decorations(true)
        .with_resizable(true)
        .with_always_on_top(true)
        .with_position(PhysicalPosition::new(300, 300,))
        .with_max_inner_size(LogicalSize::new(1280, 720))
}

```

May be I'm missing smth else...

vyeevani:
Not sure if you've tried this, but one thing that worked for me in macOS was making sure that wry/transparent was enabled in features (it's not by default). This fixed the example that you posted here showing it not working.

mchaliadzinau:
> Not sure if you've tried this, but one thing that worked for me in macOS was making sure that wry/transparent was enabled in features (it's not by default). This fixed the example that you posted here showing it not working.

Thank you for the comment.
I've already made switch to egui because of that but I would love to try dioxus once more.

PR #236, opened 2022-02-12
Clean up desktop's index.html

asaaki:
Initially I wanted to delete only the duplicated `<html>` opening tag, but then decided to make it a fully valid HTML document.

Passes https://validator.w3.org/nu/#textarea check.

### DIFF
--- a/packages/desktop/src/index.html
+++ b/packages/desktop/src/index.html
@@ -1,22 +1,15 @@
 <!DOCTYPE html>
 <html>
-
-<html>
-
-<head>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" charset="utf-8" />
-</head>
-
-
-<body>
-    <div id="main">
-    </div>
-</body>
-
-<script>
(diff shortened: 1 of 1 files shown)


packages/desktop/src/index.html
```diff
@@ -1,22 +1,15 @@
...
-</body>
-
-<script>
-    import("./index.js").then(function (module) {
-        module.main();
-    });
-</script>
-
+    <head>
+        <title>Dioxus app</title>
```
asaaki, line 4: This is not strictly necessary for dioxus, but turns this file into a fully valid document.
It get's overwritten by dioxus or the user's application configuration anyway.

mrxiaozhuox: I don't know should we add this `title` tag. but maybe it doesn't matter.

mrxiaozhuox:
looks good, you have format the HTML code.

jkelleyrtp:
Thank you!

PR #237, opened 2022-02-12
feat: enable multiwindow support for desktop apps

jkelleyrtp:
This PR uses the work done by @mrxiaozhuox to enable multiwindow desktop apps through the window context API. To achieve this, we send along props and a root component to launch a new window with. This means windows cannot accept new children, but they *can* accept `'static` props. I'd like to add non-static prop support at some point but core itself doesn't support this.

In the future, we can support multiwindow more natively with subtrees, but for now, we can only do VirtualDom linking.

I'm not entirely sure how cross-app contexts will work in this architecture since most contexts believe they are the only context in an app.

Anyways, this will be an awesome feature to support through components or hooks.


jkelleyrtp:
Temporarily closing

PR #240, opened 2022-02-12
Convert web_sys Events to html Data

oovm:
Convert web_sys Events to html data with From trait can improve reuse rate and readability.


oovm:
My main purpose is to reuse this part of the logic when working with third-party libraries like `web_sys`, `gloo`, etc.

https://github.com/oovm/dioxus-hooks/blob/dev/projects/dioxus-use-cursor/src/hooks/use_cursor/mod.rs#L30-L48

If `From` is used, it must be implemented in `html` due to orphan rules

jkelleyrtp:
I've finally decided to merge this, though we might change some of the behavior in the future with some sort of compat crate that implements Into rather than From

oovm:
It's been a long time, I don't have time to rebase, just squash it

Issue #213, opened 2022-02-06
Default Desktop-App Icon


mrxiaozhuox:
## Specific Demand

I think we can add a default desktop icon, like the dioxus logo.

PR #242, opened 2022-02-12
Add default desktop icon

oovm:
image

**Tested on windows only**

Related: #213


packages/desktop/Cargo.toml
```diff
@@ -32,6 +32,7 @@ dioxus-html = { path = "../html", features = ["serialize"], version = "^0.1.6" }
 webbrowser = "0.5.5"
 mime_guess = "2.0.3"
 dioxus-interpreter-js = { path = "../interpreter", version = "^0.0.0" }
+image = "0.24.0"
```
jkelleyrtp, line 35: Is there a way to implement this feature without relying on the `image` crate?

Can we do the version once and then just load the converted bytes in at runtime?

I ask because image takes up some (quite a bit?)x compile time.

oovm: we can dump the bytes

jkelleyrtp:
Looks great, thank you :)

PR #221, opened 2022-02-08
feat: upgrade wry to 0.13.1 

jkelleyrtp:
This PR upgrades wry to support the latest and greatest ayatana features for our fellow deb11 users to upgrade.

Fixes #196 

To enable ayatana support, you'll need to manually add dioxus desktop to your crate in addition to dioxus, and enable its ayatana feature.

```toml
dioxus = # regular stuff
dioxus-desktop = { version = "*", features = ["ayatana"] }
```

TODO: this should be documented somewhere or possibly enabled from very feature flags at the dioxus level

PR #243, opened 2022-02-13
Update to wry 0.13; support IPC, devtool, and feature flags

asaaki:
First of all: Thank you for this amazing project!
I had already slightly more fun with this than with tauri itself, even though it stands on the shoulders of it (wry, tao).

I know it's a lot at once, but the wry 0.13 upgrade led me down a rabbit hole.

What changes:
- general dependency updates as much as possible (not only wry)
- switch from RPC to IPC handler, fix message (de)serialization
- modularize some code pieces
- adjust module/function/struct/enum visibility where it made sense (not all need to be pub)
- general cleanups

If something is not to your liking or you want something split up differently, I can try to cut separate PRs.

Note: I tested it with some examples and a local project, all on a Win10 machine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,13 @@ web = ["dioxus-web"]
 desktop = ["dioxus-desktop"]
 router = ["dioxus-router"]
 
+devtool = ["dioxus-desktop/devtool"]
+fullscreen = ["dioxus-desktop/fullscreen"]
+transparent = ["dioxus-desktop/transparent"]
+
+tray = ["dioxus-desktop/tray"]
+ayatana = ["dioxus-desktop/ayatana"]
+
 # "dioxus-router/web"
 # "dioxus-router/desktop"
 # desktop = ["dioxus-desktop", "dioxus-router/desktop"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,15 +57,15 @@ members = [
 ]
 
 [dev-dependencies]
-futures-util = "0.3.17"
+futures-util = "0.3.21"
 log = "0.4.14"
 num-format = "0.4.0"
 separator = "0.4.1"
-serde = { version = "1.0.131", features = ["derive"] }
+serde = { version = "1.0.136", features = ["derive"] }
 im-rc = "15.0.0"
-anyhow = "1.0.51"
-serde_json = "1.0.73"
+anyhow = "1.0.53"
+serde_json = "1.0.79"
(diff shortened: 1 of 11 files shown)


mrxiaozhuox:
I think upgrade to wry 0.13 has been implemented in another PR: #221

mrxiaozhuox:
modularize looks great. But I think some `feature flag` is unnecessary.

jkelleyrtp (approved):
Looks great!! Thank you for diving into this!

asaaki:
Note on the feature flags:

According to wry's documentation, the following flags are MacOS-specific and only enable private APIs, which might only be used during development anyway:
- devtool
- transparent
- fullscreen

That does not mean, those flags actually toggle real behaviour for anyone else; if you want to use devtools on Windows or Linux, you always can, as well as transparency and fullscreen.

To be honest, those flags are named misleading and confusing, they should probably have platform prefixes to communicate, that they are **not** general purpose feature toggles.

The reason I included them here is that I do not want to block MacOS developers from using the private APIs if they so desire.

-----

`devtool`

That one was interesting (at least on Windows).
wry 0.13 changed quite a bit on the behaviour, probably maybe due to the underlying switch of windows/webview2 related crates.
In my testing there was no entry on the context menu anymore, and also the Ctrl+Shift+i shortcut did not open.
That might be confusing at first, but considering that you do not want to have the devtool in a release, that is already quite nice.

So to actually open the devtool, one has to trigger that.
That's why I had to include the function on DesktopContext, so people can programmatically use it in their apps.

```rust
fn AppWithDevTool(cx: Scope) -> Element {
  let window = dioxus::desktop::use_window(&cx);

  // snip

  cx.render(rsx! {
    button {
      onmousedown: |e| e.cancel_bubble(),
      onclick: move |_| { window.devtool() },
      "Open DevTool"
    }
  })
}
```

I guess this should be documented probably. 😅 

-----

Bonus: disabling the context menu all together (as it's not very useful when no devtool entry is there):

```js
// add this JS snippet somewhere in your app
window.addEventListener('contextmenu', e => e.preventDefault());
```

Kudos: https://github.com/tauri-apps/wry/issues/30#issuecomment-831896232

jkelleyrtp:
> To be honest, those flags are named misleading and confusing, they should probably have platform prefixes to communicate, that they are not general purpose feature toggles.

We'll probably remove them from the top-level dioxus crate, meaning you need to manually bring in the dioxus-desktop crate with features enabled to dig into platform-specific configurations. And perhaps prefix them with `macos` or something similar.

PR #254, opened 2022-02-15
Make all packages which require futures-channel ask for the same version

autarch:
If they aren't the same then Cargo cannot resolve a working version for some reason.


jkelleyrtp:
Hmmm, I wonder if our cached lockfiles were finding a valid version - none of my projects had this issue. Anyways - it's what the Cargo team suggests.

PR #257, opened 2022-02-16
Feat: NodeBuilder API for ergonomically building UI without the rsx! macro

jkelleyrtp:
This PR adds a builder API to the HTML crate. It's mostly intended for three things:

- A new target for the rsx! macro to compile to
- A way of building nodes when rsx! has limitations 
- A way of building nodes when you don't like the rsx! macro

With this PR, you can cleanly design reactive web-bsaed UIs in Rust with absolutely no macros  - which is pretty awesome!

Progress:
- [x] Listeners
- [x] All HTML elements
- [x] All SVG elements
- [x] All global attributes
- [x] All SVG/trait attributes
- [ ] Support for components
- [ ] Migrate rsx! macro to the NodeBuilder API (much better IDE support)
- [x] Listeners

Eventually, we want to be able to convert between rsx! and rust code.

This is what it looks like:

```rust
        div(&cx)
            .class("a")
            .draggable(false)
            .id("asd")
            .accesskey(format_args!("{anything}")
            .class("sm-lg text-red etc etc")
            .contenteditable(true)
            .data(false)
            .dir(false)
            .dangerous_inner_html(false)
            .onclick(move |_| {
                println!("clicked");
            })
            .onclick(move |_| {
                println!("clicked");
            })
            .onclick(move |_| {
                println!("clicked");
            })
            .attr("name", "asd")
            .children([
                match true {
                    true => div(&cx).build(),
                    false => div(&cx).class("asd").build(),
                },
                match 10 {
                    10 => div(&cx).build(),
                    _ => div(&cx).class("asd").build(),
                },
            ])
            .fragment((0..10).map(|i| {
                div(&cx)
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .class("val")
                    .build()
            }))
            .fragment((0..10).map(|i| div(&cx).class("val").class("val").class("val").build()))
            .fragment((0..20).map(|i| div(&cx).class("val").class("val").class("val").build()))
            .fragment((0..30).map(|i| div(&cx).class("val").class("val").class("val").build()))
            .fragment((0..40).map(|i| div(&cx).class("val").class("val").class("val").build()))
            .children([
                match true {
                    true => div(&cx).build(),
                    false => div(&cx).class("asd").build(),
                },
                match 10 {
                    10 => div(&cx).build(),
                    _ => div(&cx).class("asd").build(),
                },
            ])
            .build()
```

```rust

trait ComponentBuilder<P> {
    fn render(self, cx: &ScopeState, props: P) -> Element;
}

impl<P, F> ComponentBuilder<P> for F
where
    F: for<'a> Fn(Scope<'a, P>) -> Element<'a>,
{
    fn render(self, cx: &ScopeState, props: P) -> Element {
        todo!()
    }
}
```

## Edit: More Learnings

- It looks like the best approach here is the "newtype" pattern.
- Custom elements can be implemented however people want. But the best approach is to implement the builder trait and then wrap it with your own methods.
- The builder API can exist within the core crate
- A `fragment` element exists to fulfill the role of dynamically generated lists
- Children can only be specified in a fixed size array, which should make builder code just as efficient as non builder code.
- Autocomplete in IDE can only be achieved through codegeneration
- Autocomplete and autoformatting is really nice to have :-)
- The builder pattern can actually be cleaner than rsx! it some instances
- The rsx! macro actually gets better with the builder pattern as its base


MuhannadAlrusayni:
I would love to use this! and I think this would give us a way to do generic components (#321) with letting `rsx!()` blocking us (at less when I use NodeBuilder API)

jkelleyrtp:
> I would love to use this! and I think this would give us a way to do generic components (#321) without letting `rsx!()` blocking us (at less when I use NodeBuilder API)

This PR is "very breaking" but definitely a better direction. I think the ergonomics for this are fixed, fortunately

PR #258, opened 2022-02-16
fix unnecessary div

HansTrashy:
This PR removes the `div` element that was wrapped around children of the router. This could confuse people and throw css styles off.


jkelleyrtp:
Can you run `cargo fmt` on the code? Or enable autoformatting in your editor? Looks like it's not passing CI?

jkelleyrtp:
Thank you!!!

PR #271, opened 2022-02-20
Disable reload context menu in production

jkelleyrtp:
Webview will never get custom context menus (the thing when you right click), so it's up to app authors to implement their own context menus. A library for making that easier will be very helpful to the Dioxus ecosystem.

Anyways, this PR adds an initialization script in release mode that disables the context menu altogether, since having a right click menu that *just* shows reload is probably not super helpful to most users.

### DIFF
--- a/packages/desktop/src/lib.rs
+++ b/packages/desktop/src/lib.rs
@@ -174,7 +174,25 @@ pub fn launch_with_props<P: 'static + Send>(
                 }
 
                 if cfg!(debug_assertions) {
+                    // in debug, we are okay with the reload menu showing and dev tool
                     webview = webview.with_dev_tool(true);
+                } else {
+                    // in release mode, we don't want to show the dev tool or reload menus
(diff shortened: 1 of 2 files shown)


packages/desktop/src/lib.rs
```diff
@@ -174,7 +174,25 @@ pub fn launch_with_props<P: 'static + Send>(
                 }
 
                 if cfg!(debug_assertions) {
+                    // in debug, we are okay with the reload menu showing and dev tool
```
mrxiaozhuox, line 177: we will keep menu in debug mode ?

jkelleyrtp: We want to provide the `inspect element` tool in debug which means we need to keep the menu around. So, yes.

mrxiaozhuox: oh!

mrxiaozhuox (approved):

PR #278, opened 2022-02-23
feat: canoncialize assets for macOS, Windows, and Linux

jkelleyrtp:
This PR adds the right relative pathing for assets when bundled through cargo-bundle.

It's never been this easy to include an image in your Rust `.app`! :)

I don't have any other machine than a mac box so I can't test alternate implementations.

https://crates.io/crates/directories

- [x] macOS

These targets do not provide a straightforward way of bundling assets. I've opted to make it the asset dir configurable instead.
- [ ] windows
- [ ] linux (rpm)
- [ ] linux (deb)


jkelleyrtp:
We will add windows/linux later, or anyone apsiring to figure it out can take care of it.

It seems that when building with cargo-wix, you can choose the resource directory. Same with linux. Either we'll work with the tauri config and get theirs working by default or come up with our owm.

Issue #281, opened 2022-02-24
Document how to use desktop feature "ayatana" in "Getting started" Tutorial


jkur:
When following you tutorial on building a desktop app on Linux the first example fails to build on Debian Stable(Bullseye) because of the remove libappindicator3-dev packages.

The dioxus-desktop package has the flag "ayatana" to enable building with libayatana-appindicator3, but it is not documented anywhere.

My solution was to clone this repo and change the dependencies in the dioxus-desktop Cargo.toml:

`wry = { version = "0.13.1" , features = ["ayatana"]}`

But how would you configure the flag for your dioxus project without relying on the cloned source repo?
Should be documented in the Tutorial. I would think that most Debian Developer Machines are at least on bullseye now.

My Cargo.toml in the project looks like:

```
[package]
name = "somedesktop-ui"
version = "0.1.0"
edition = "2021"

[dependencies]
dioxus = { path="../dioxus/", features = ["desktop"] }
```

jkur:
Maybe this would be a solution:
#282

jkelleyrtp:
I like the solution you made a PR for. We should document the feature flags somewhere in the readme of either dioxus main or the dioxus desktop repo.

PR #282, opened 2022-02-24
Add options for compilation on systems that lack libappindicator3

jkur:
Possible fix for #281


jkelleyrtp:
Thank you so much for this! Looks good.

PR #294, opened 2022-03-06
Added active class to router link

maccesch:
There should be a global way of setting the active class as well but I don't know how to properly do that. Please let me know what I can do.

### DIFF
--- a/packages/router/src/components/link.rs
+++ b/packages/router/src/components/link.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use crate::{RouterCore, use_route};
+use crate::{use_route, RouterCore};
 use dioxus_core as dioxus;
 use dioxus_core::prelude::*;
 use dioxus_core_macro::{format_args_f, rsx, Props};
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I'll be honest, I didn't know that "active" was a thing :-)

I think the implementation makes sense - but does that also mean that all "Link" components subscribe to the current route? I guess that makes sense?

Anyways - I think you might've made these changes on a previous commit (I rebased the local router PR before/after a merge) and the method to get current location/path has changed a bit.

Otherwise - looks good and is the approach I would suggest!

maccesch:
Thanks for rebasing!
Yes, I’ve done work with Vue.js and Nuxt.js and there it works like this. 
I guess if we really want to provide ‘active’ there’s no way around subscribing to the route, right?

maccesch:
Updated

jkelleyrtp (approved):
Looks great! I think we might (later) update the router service to provide some of the parsing/validation logic internally for some extra caching. But - routes change so infrequently that we don't gain that much performance anyways.

Thanks for adding this!

maccesch:
@jkelleyrtp Hey, could you run the workflows again? Should work now

jkelleyrtp:
Thank you!!

PR #302, opened 2022-03-10
tui bugfixes and text modifier elements

Demonthos:
- Fix ctrl c not quitting (new crossterm version interprets ctrl + key as ctrl + capital key)
- adds margin for borders
- adds all keys
- add text property elements: b, strong, u, ins, del, i, em, and mark


jkelleyrtp:
Awesome - thank you!!

PR #309, opened 2022-03-13
active_class prop for Router

maccesch:
In order to not have to modify the `active_class` on every link there is now a prop with the same name on `Router`.

Useful when you're using a CSS framework like Bulma where the active nav links are expected to have the class `"is-active"`.

Also: `use_route` can now also be used outside of `Route` components.


jkelleyrtp:
Much better, thank you :)

PR #313, opened 2022-03-15
feat: add eval support for desktop

jkelleyrtp:
This simple PR adds eval support for desktop context.

Values are currently not returned, so this is like a "one way eval".

We can add a better two-way eval or dedicated messaging pipeline later.


mrxiaozhuox (approved):
Cool!! But I think `log::error!` is not useful for a desktop app.

jkelleyrtp:
Closing in favor of #318

PR #318, opened 2022-03-17
Eval stuff

overlisted:
1. Adds this function to `dioxus_desktop` and `dioxus_web`:
```rs
pub fn use_eval<S: std::string::ToString>(cx: &ScopeState) -> &dyn Fn(S)
```
2. Re-exports them in the root crate
3. Adds an example


overlisted:
Very WIP! @jkelleyrtp what do you think? I should probably add an example but I'm not sure how to make examples that compile on wasm. And I'm generally not very familiar with wasm-bindgen.

overlisted:
And another question: Should I somehow move the re-export to the `hooks` package? That would make more sense but I was lazy

tqwewe:
Should `S` be `std::string::ToString` rather than `std::fmt::Display` to better show intention?

overlisted:
@tqwewe done!

packages/desktop/src/desktop_context.rs
```diff
@@ -185,5 +192,16 @@ pub(super) fn handler(
...
+            .expect("eval shouldn't panic"),
     }
 }
+
+/// Get a closure that executes any JavaScript in the WebView context.
+pub fn use_eval<S: std::string::ToString>(cx: &ScopeState) -> impl Fn(S) + '_ {
+    let desktop = use_window(&cx);
+
+    move |script| desktop.eval(script)
+}
```
jkelleyrtp, line 207: We haven't really been returning `impl Fn` from any of ours hooks - mostly because they can't be shared between two handlers. I think that makes sense here too - eval might need to be used from two handlers. I would suggest - to be more inline with how we return other functions, to return `&dyn Fn`. To be fancier, we could return `&Rc<dyn Fn>` for use in async contexts.

jkelleyrtp:
Do you think this is ready?

jkelleyrtp:
Looks good! We might tweak a thing here or there before releasing but I like it :)

PR #320, opened 2022-03-19
Capture correct radio button value in a form input event

naturalethic:
Solution implemented is to return an optional value from the target element, which for all cases except radio buttons is a Some.  For radios, only the checked element will return a Some.  This prevents the last radio in the group from always overwriting the group value.


jkelleyrtp:
Looks good - though I will follow this PR up with some changes that make the same fix for desktop.

jkelleyrtp (approved):

Issue #306, opened 2022-03-12
extremely responsive futures don't work


elexing:
**Problem**

when the response time of future is less than 552 millis in my computer(macbook air m1),  custom element cannot render normally.

**Steps To Reproduce**

- Have the below code

``` rust
use dioxus::prelude::*;


fn main() {
    dioxus::desktop::launch_cfg(App);
}

#[allow(non_snake_case)]
fn App(cx: Scope) -> Element {
    cx.render(rsx! (
        Demo{}
    ))
}

#[allow(non_snake_case)]
fn Demo(cx: Scope) -> Element{
    let fut = use_future(&cx, (), |_|async move{
        std::thread::sleep(std::time::Duration::from_millis(1000));
        10
    });
    cx.render(match fut.value() {
        Some(value) => {
            let content = format!("content : {:?}", value);
            rsx!(div{"{content}"})
        },
        None => rsx!(div{"computing!"})
    })

}
```

- if sleep time is less than 552 millis, such as 550ms, it will always show "computing"
- if sleep time is not less than 552 millis, it will switch to "10" quickly.
- 

**Expected behavior**
it should switch to '10' if the response time of future is great than 20 millis.

**Screenshots**

**Environment:**
 - Dioxus version: [v0.2.3]
 - Rust version: [1.46.0]
 - OS info: [MacOS]
 - App platform: [`desktop`]

dranikpg:
`std::thread::sleep` is a blocking function. You shouldn't be using that in async code

dranikpg:
Checked this issue out. It has nothing to do with time. Simultaneous updates are processed in the wrong order.

```rust
fn Demo(cx: Scope) -> Element{
    let fut = use_future(&cx, (), |_| async move { "done" });
    let val = match fut.value() {
        Some(val) => val,
        None => "waiting..."
    };
    cx.render( rsx! { div { "state: {val}" } })
}
```
This code shows "waiting..." forever, however all the futures stuff runs correctly.
The render changes queue [pushes](https://github.com/DioxusLabs/dioxus/blob/268cc5bd393f49b3950eb05a978f1b72041c68c6/packages/desktop/src/controller.rs#L70) **and** [pops](https://github.com/DioxusLabs/dioxus/blob/268cc5bd393f49b3950eb05a978f1b72041c68c6/packages/desktop/src/controller.rs#L100) from its back, whereas it should pop from its front to preserve order.

[...]

PR #326, opened 2022-03-21
Replace Vec with VecDeque for desktop edit queue

dranikpg:
Hi!

[This issue](https://github.com/DioxusLabs/dioxus/issues/306) is caused by edit events being handled in the wrong order on the desktop backend.

If I understand correctly, the author might have assumed that all events between two dom work deadlines are independent. However this is certainly not the case for updates before first paint. If this is the only exception, [the first edit](https://github.com/DioxusLabs/dioxus/blob/268cc5bd393f49b3950eb05a978f1b72041c68c6/packages/desktop/src/controller.rs#L57) could be processed somehow differently without resorting to a deque. Nonetheless, I assume the deque doesn't introduce any overhead so this would be in general the cleanest solution.

### DIFF
--- a/packages/desktop/src/controller.rs
+++ b/packages/desktop/src/controller.rs
@@ -1,7 +1,7 @@
 use crate::desktop_context::{DesktopContext, UserWindowEvent};
 use dioxus_core::*;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
     sync::Arc,
     sync::{atomic::AtomicBool, Mutex},
 };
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I found an issue that breaks the bug that were trying to fix originally. Instead of switching back to a vecdeque, I figured we could just drain the queue instead of popping it to get the right order. 

The two spots for issues come when the modifications occur in the same stream as the initial edit stream to build the page. The modification stream is pushed in backwards, which places the initial edit stream in a weird place.

The VecDeque fixed that case but stumbled on the ordering of the edits themselves.

I think #330 fixes it - can you test a few case and see if it works?

dranikpg:
It does for my cases 👍🏻 . Sorry for not digging deeper to find more issues 😬

PR #331, opened 2022-03-26
fix: diff_lazynodes bug adding children

### LINKED ISSUE #332: bug: diff_lazynodes add children missing PushRoot
**Problem**

```rust
use dioxus_core::DomEdit::*;
use dioxus::prelude::*;

#[test]
fn add_nested_elements() {
    #[allow(non_snake_case)]
    fn Base(cx: Scope) -> Element {
        rsx!(cx, div {})
    }
    let vdom = VirtualDom::new(Base);
    let (_create, change) = vdom.diff_lazynodes(
        rsx! {
            div{}
        },
        rsx! {
            div{
                div{}
            }
        },
    );
    assert_eq!(
        change.edits,
        [
            PushRoot { root: 1 },
            CreateElement { root: 2, tag: "div" },
            AppendChildren { many: 1 },
        ]
    );
}
```

**Steps To Reproduce**

- call diff_lazynodes on a element without children and a element with childern

**Expected behavior**
diff_lazynodes to return a PushRoot to change the parent element before AppendChildren.

**Environment:**
 - Dioxus version:  master
 - Rust version: 1.59.0
 - OS info: Windows
 - App platform: Working on tests, desktop feature was active

**Questionnaire**

Demonthos:
Fixes  #332
When adding children to a element which previously had none, the mutations did not add the PushRoot instruction before adding children.
This PR adds the instruction.

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -423,8 +423,8 @@ impl<'b> DiffState<'b> {
         match (old.children.len(), new.children.len()) {
             (0, 0) => {}
             (0, _) => {
-                let created = self.create_children(new.children);
                 self.mutations.push_root(root);
+                let created = self.create_children(new.children);
                 self.mutations.append_children(created as u32);
             }
             (_, _) => self.diff_children(old.children, new.children),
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
This needs to "PopRoot" after the root has been added but I'll just follow this PR up with that change. Thanks!

PR #333, opened 2022-03-27
Fix form value collecting from form element instead of input element

naturalethic:


naturalethic:
That title isn't perhaps clear.  I noticed that for TextAreas, I was getting a value of all the text within the form itself.  In the three non-input type value matchers below the main one, I noticed that it was matching against the form element itself.  So the first two would fail to match, and it would default to taking all the text in the third one from the form element.

PR #343, opened 2022-04-08
Include optional public-url flag in build instruction within getting …

TimboTambo:
…started docs


jkelleyrtp:
This is good - probably a little verbose but no problem with including a little more info.

PR #349, opened 2022-04-11
Update mobile.md

erlend-sh:
Also worth noting that Android support is further along now, so maybe this section of the Guide should be updated to reflect that:

> It is possible to get Dioxus running on Android and rendered with WebView, but the Rust windowing library that Dioxus uses - tao - does not currently supported Android.

https://github.com/tauri-apps/wry/pull/510


mrxiaozhuox (approved):

jkelleyrtp:
Thanks!

PR #355, opened 2022-04-16
fix: render_lazy can't handle components

jkelleyrtp:
This PR changes render_lazy to use the virtual dom's create_nodes method which progress component lifecycles. This is important for render_lazy to be able to handle components.

Be warned though - there's no concept of "suspense" with Dioxus-core, so all futures will end up as empty nodes in render_lazy.

Fixes #355

### DIFF
--- a/examples/ssr.rs
+++ b/examples/ssr.rs
@@ -22,6 +22,16 @@ fn main() {
         })
     );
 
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
This API is changing in 0.3, making this PR moot.

PR #358, opened 2022-04-17
Fix query parsing

mirichan:
The router was silently throwing away queries not prefixed with an extra `?`.

The documentation for `url::Url::query` indicates that it will not include the `?` delimiter ahead of the URL-encoded form string, and `strip_prefix` returns `None` if the prefix doesn't match.


jkelleyrtp (approved):
Thank you! Great catch

PR #362, opened 2022-04-21
add benchmark, headless mode, and shutdown context to tui

Demonthos:
This PR adds a benchmark to measure the performance of small updates to elements in the tui render. To allow the benchmark to run in CI, It also adds a headless mode and a context that allows the user program to shut down. It also adds the builder pattern and non-exhaustive to config, so that in the future adding more fields to config will not be a braking change.


jkelleyrtp:
Looks good - thank you :)

PR #363, opened 2022-04-21
Removed alerts when right click is disabled.

blemelin:
This PR removes the JavaScript alerts shown when the user makes a right click but they are disabled (by default, they are disabled on release).

Is the `alert` on purpose or was it just forgotten ? Since alerts are *kinda* considered bad practice, I think they should be removed. Feel free to close this PR if I'm mistaken.


jkelleyrtp:
They should not be there at all - thank you!

PR #364, opened 2022-04-23
Feat: Extends props attribute with features to inject attributes and event handlers declaratively

Nejat:
# _*** UPDATED - Added CSS Syntax support for selectors ***_

## Adds `inject_as`, `selector`, `root` features to `props` attribute, to enable automatic injection of properties into an `rsx!` definition.

The basic principle is to be able to declaratively inject values of a custom component onto tags or other components in the `rsx!` definition of that custom component.

### `inject_as` - maps name

* `inject_as` is an optional attribute name mapper of the value to inject, _i.e._<br>
`my_class` => `class` or <br>
`clicky` => `onclick`

### `selector` injects attribute values or event handlers:

* `selector` is a `;` separated list of pseudo `css` like selectors that indicate where to inject the value<br>

   *  `:root` - root element, only if root level has only one element
   *  `:any` - any element in level
   *  `tag | component` - _i.e., div | MyCustomComponent
   *  `:` - similar to `css` pseudo selector, marks the declaration of a child selector
   * `:first` - first child in elements list
   * `:last` - last child in elements list
   * `:even` - even children in elements list
   * `:odd` - odd children, in elements list
   * `:only` - only child, in elements list
   *  `:[nth[,nth]]` - comma delimited list of `nth` children
   *  `:last[nth[,nth]]` - comma delimited list of `nth` children from end
   *  `:[Nn±B]` - every N, _i.e._ `div:[3n]` _every 3rd element that is_ `div`
   *  `:[..nth]` - the first `nth` children
   *  `:[nth..]` - all children from `nth` onward
   *  `@` - similar to `:`, but marks the declaration child-of-type selector
   * `@first` - first child-of-type in elements list
   * `@last` - last child-of-type in elements list
   * `@even` - even children-of-type in elements list
   * `@odd` - odd children-of-type, in elements list
   * `@only` - only child of type, in elements list
   *  `@[nth[,nth]]` - comma delimited list of `nth` children-of-type
   *  `@last[nth[,nth]]` - comma delimited list of `nth` children-of-type from end
   *  `@[Nn]` - every N, _i.e._ `div:[3n]` _every 3rd_ `div`
   *  `@[..nth]` - the first `nth` children-of-type
   *  `@[nth..]` - all children-of-type from `nth` onward
   *  `>` - similar to `css` child selector
   *  `;` - is a divider for subsequent selectors, _values can be injected more than once_

###  `root` injects attribute values or event handlers only to a `root` element:

`root` is equivalent to, therefore a shortcut of `selector = ":root"`

### one addition to `rsx!` syntax

To make the automatic inject possible the `rsx!` needs to know the custom context type as well as custom context instance name. The syntax has been extended to take the form `name:type;`, _i,e., `rsx! { cx: Component; ... }`_

This requirement for injection does not preclude or interfere with the original definition of `rsx!`, it just extends it for this use case.

## Manual Example

```rust
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default)]
    my_class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
    cx.render(rsx! {
        div {
            onclick: move |evt| cx.props.onclick.call(evt),
            a {
                "class": "{cx.props.my_class}"
            }
            &cx.props.children
        }
        div {
            onclick: move |evt| cx.props.onclick.call(evt),
            "class": "{cx.props.my_class}"
        }
    })
}

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct Component2Prop<'a> {
    #[props(default)]
    class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component2<'a>(cx: Scope<'a, Component2Prop<'a>>) -> Element {
    cx.render(rsx! {
        div {
            Component {
                my_class: "{cx.props.class}",
                onclick: move |evt| cx.props.onclick.call(evt),
                &cx.props.children,
            }
            div {}
        }
        div {}
    })
}
```

## Declarative Equivalent

```rust
use dioxus::events::MouseEvent;
use dioxus::prelude::*;

use dioxus_sugar::attributes;

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default, inject_as = "class", selector= "div:[1]; div:first > a")]
    my_class: &'a str,
    #[props(default, selector= "div")]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
    rsx! {
        cx: ComponentProps;
        div {
            a {}
            &cx.props.children
        }
        div {}
    }
}

#[derive(Props)]
pub struct Component2Prop<'a> {
    #[props(default, inject_as = "my_class", selector = "div > Component")]
    class: &'a str,
    #[props(default, selector = "div > Component")]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component2<'a>(cx: Scope<'a, Component2Prop<'a>>) -> Element {
    rsx! {
        cx: Component2Prop;
        div {
            Component {
                &cx.props.children,
            }
            div {}
        }
        div {}
    }
}
```

## Motivation

The idea was born out of supporting [Dioxus-Sugar](https://github.com/Nejat/dioxus-sugar) which declaratively includes valid html attributes and event handlers for custom component properties. `Dioxus-Sugar` adds the attribute and event handler definitions to the custom component properties and this feature injects those and other custom properties into the custom component definition. Which lets me finish my ultimate reason for building any of these `Dioxus-Bulma` :)

## Supports CSS Selector and a more concise "rusty" equivalent

this is a list of supported selectors, the left side is the new css syntax and the right side is the original more concise "rusty" syntax

```
:not(p)						:not[p] 

element:nth-child(n)				element:[n]

element:nth-child(an ± b)			element:[an ± b]

element:nth-child(even)				element:even

element:nth-child(odd)				element:odd

element:nth-last-child(n)			element:last[n]

element:first-child				element:first

element:last-child				element:last

element:only-child				element:only


element:nth-of-type(n)				element@[n]

element:nth-of-type(an ± b)			element@[an ± b]

element:nth-of-type(even)			element@even

element:nth-of-type(odd)			element@odd

element:nth-last-of-type(n)			element@last[n]

element:first-of-type				element@first

element:last-of-type				element@last

element:only-of-type				element@only


PLUS Non-CSS

:root

:any						*


element:nth-child(n1..n2)			element:[n1..n2]

element:nth-child(..n)				element:[..n]

element:nth-child(n..)				element:[n..]


element:nth-of-type(n1..n2)			element@[n1..n2]

element:nth-of-type(..n)			element@[..n]

element:nth-of-type(n..)			element@[n..]
```


Nejat:
@jkelleyrtp @Demonthos 
I had to close the last PR, because it was off of the wrong branch, and I could not make updates to the PR.
I have made some syntax changes and added more docs to above comments.

Would it make sense to put this logic behind an "Injection" feature?

Demonthos:
Taking another look at this, and this is amazing! I think it would be good as part of the default features as long as the compile times aren't significantly longer.

rMazeiks:
(im not a maintainer, just bumped into this and have some thoughts)

This looks cool and shiny, but i think it's worth noting that there are some drawbacks:

- Introduces coupling of components props to the rendered HTML. When changing the HTML, the props may need to be updated.
- Rendering now happens in 2 places, this can become hard to follow & debug. What happens when `inject_as` conflicts with the rsx?
- The additional complexity (new attributes, new css-like syntax)

This is just my opinion, but personally i find the explicit, "manual" version of the code simpler and easier to follow – rendering only happens in 1 place and concerns are well-separated. And it doesn't seem too verbose either.

Of course since this feature is optional, the effect of the downsides is somewhat minimized. However, it still contributes to the overall complexity (and arguably feature bloat) -> steeper learning curve, more stuff/docs to maintain.

Demonthos:
> * Introduces coupling of components props to the rendered HTML. When changing the HTML, the props may need to be updated.

Good point, I don't see any reason for this to be specific to the attributes, but I think generally having a css-like solution that works with custom elements would be a good thing. This is also agnostic of renders, and more scoped then css which would make it easier to maintain both for dioxus as it grows to more renders, and for larger projects using dioxus.

Could this be moved into the rsx!{} macro, with the attributes passed explicitly? Then it could be used with use_state, and allow for any custom logic, and would keep everything rendering related in the same place.

Something like this for the example:
```rust
use dioxus::events::MouseEvent;
use dioxus::prelude::*;

use dioxus_sugar::attributes;

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default)]
    my_class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
    rsx! {
        cx: ComponentProps;
        inject(attribute = "class", selector = "div:[1]; div:first > a", value = cx.props.my_class),
        inject(attribute = "onclick", selector = "div", value = cx.props.onclick),
        div {
            a {}
            &cx.props.children
        }
        div {}
    }
}

#[derive(Props)]
pub struct Component2Prop<'a> {
    #[props(default)]
    class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component2<'a>(cx: Scope<'a, Component2Prop<'a>>) -> Element {
    rsx! {
        cx: Component2Prop;
        inject(attribute = "my_class", selector = "div > Component", value = cx.props.class),
        inject(attribute = "onclick", selector = "div", value = cx.props.onclick),
        div {
            Component {
                &cx.props.children,
            }
            div {}
        }
        div {}
    }
}
```

I think the framework gains a lot by sticking close to css and html. It might be better to keep the original css syntax instead of mapping it to make learning dioxus with a web background easier.

Nejat:
>  ... I'm not a maintainer ...

my apologies, for some reason I thought you were, my bad

Nejat:
thank you for thoughtful feedback, let me try to address these points, by first describing how I got here (I'm already two projects removed, but I was determined) ...

I started out by wanting to create a Dioxus Bulma wrapper library, first in a series of wrappers; bootstrap, etc. ultimately hoping to find a pattern and create a tool for creating custom css ui frameworks ...

As I created Bulma components, I found the need for these components to behave more like their underlying html counter parts and wrapping things in components hid all of those attributes and more importantly events. It quickly became apparent to me that mapping each of those attributes and events manually was going to be tedious, error prone and a maintenance nightmare. 

So, what do you do when you have a lot of boiler plate code? yep that's right you write generators. That's where the idea for Dioxus-Sugar came from, it would have knowledge of all HTML-5 tags, attributes and events, and also keep up with future changes, and would allow you to inject HTML 5 attributes or attributes of a tag into your custom properties.

_from my example above_
```rust
use dioxus_sugar::attributes;

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default, inject_as = "class", selector= "div:[1]; div:first > a")]
    my_class: &'a str,
    #[props(default, selector= "div")]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}
```
Check out the `attributes` macro, it adds all of the html attributes of a `div`, excluding `class`, `name` & `id`, as optional properties for the custom component.'s `ComponetProps`. That's when it dawned on me that now mapping those properties into the custom component was the next manual hurdle. 

Originally, I attempted to create a macro to use inside of the `rsx!` macro, but quickly learned expanding macros inside of other macros does not work in the necessary order and more importantly it would need to understand the inner working of `rsx!` which would make it too strongly couple even if it were possible. So, I set out to see how I can solve my problem by contributing something possibly useful for others too. 

That is why my PR is so extensive as I wanted it to be a more general tool. What would really make it a more generally useful tool, is if injection also worked for dynamically created content, this implementation does not cover that use case.

>
> (im not a maintainer, just bumped into this and have some thoughts)
> 
> This looks cool and shiny, but i think it's worth noting that there are some drawbacks:
> 
> * Introduces coupling of components props to the rendered HTML. When changing the HTML, the props may need to be updated.
>

True, but 
- props are already coupled, they are just manually added to the `rsx!`
- the changes are local and very visible, i.e. no magic, plus compiler errors would guide you

>
> * Rendering now happens in 2 places, this can become hard to follow & debug. What happens when `inject_as` conflicts with the rsx?
>

do you mean component design? because this is a compile time process, rendering is runtime

- design can happen in two places, though it is conceivable to do all your attribute mappings with injection if desired
- and if you needed to map it to multiple places in your component it would be more efficient to describe it once 
- if it conflicts you will get a compile error, for what it's worth I did my best to map errors, but that can be iterated upon

>
> * The additional complexity (new attributes, new css-like syntax)
>

- complexity sometimes comes with flexibility and productivity 
- and when coupled with a tool like what I've written in Dioxus-Sugar
  - you attain productivity gains with much less boilerplate 
  -  also, not to mention much less maintenance
  - and even less bugs from manually implementing everything it provides automatically
  - and because of the web-reference crate I created, it even follows html-5 standard, I plan to enhance this to scrap the actual draft, so future updates will be easy to incorporate

> 
> This is just my opinion, but personally i find the explicit, "manual" version of the code simpler and easier to follow – rendering only happens in 1 place and concerns are well-separated. And it doesn't seem too verbose either.
> 

Very valid and as you say below this is completely optional and for now probably best for crate creators. 
This thought exercise makes me want to hide this behind an optional feature.

>
> Of course since this feature is optional, the effect of the downsides is somewhat minimized. However, it still contributes to the overall complexity (and arguably feature bloat) -> steeper learning curve, more stuff/docs to maintain.
>

Looking forward to hearing more of your thoughts

Nejat:
>
> I think the framework gains a lot by sticking close to css and html. It might be better to keep the original css syntax instead of mapping it to make learning dioxus with a web background easier.
>

I really like this idea and I can investigate it further, but it's a completely different use case. Please read my previous comment, as what I designed would not benefit from this approach, so for my use case I still need the functionality of injecting with attribute macros.

Additionally, I can also provide a syntax that is closer to css compatibility, but even that wouldn't be 100%

Nejat:
I have updated this PR with new CSS Syntax support option.
All of the parsing and match logic is thoroughly tested.
The only feature not implemented is the `:not(elm)` selector

@jkelleyrtp - really hoping this gets accepted, thanks

Nejat:
Once again closing this PR because of an irreversible mistake that would pollute the `master` branch with too many commits.

Please refer to https://github.com/DioxusLabs/dioxus/pull/439 for the third attempt to creating this PR ;)

Demonthos:
> Once again closing this PR because of an irreversible mistake that would pollute the `master` branch with too many commits.
> 
> Please refer to #439 for the third attempt to creating this PR ;)

If too many commits is the issue you can use [get squash](https://www.git-tower.com/learn/git/faq/git-squash/) to combine multiple commits, and then force push it to your branch next time. :)

Nejat:
> > Once again closing this PR because of an irreversible mistake that would pollute the `master` branch with too many commits.
> > Please refer to #439 for the third attempt to creating this PR ;)
> 
> If too many commits is the issue you can use [get squash](https://www.git-tower.com/learn/git/faq/git-squash/) to combine multiple commits, and then force push it to your branch next time. :)

thanks, the problem also was there were too many merges from upstream to keep up with changes, which made things too messy to do that. i thought long and hard before taking this path

PR #366, opened 2022-04-24
Liveview Axum Integration + Example updates

WIGGLES-dev:
I like Axum, and wanted to try to make a Liveview toy app. I have implemented a connector in the same style as the warp connector (it is pretty much copy paste).

I have also changed the examples to suppress VScode IDE errors when having dioxus open in the same workspace as a seperate application (I moved the cfg statement inside the main function so that VScode wouldn't complain that there was no main function).

I have also added the menu element to Dioxus html

### DIFF
--- a/packages/html/src/elements.rs
+++ b/packages/html/src/elements.rs
@@ -429,6 +429,11 @@ builder_constructors! {
     mark {};
 
     /// Build a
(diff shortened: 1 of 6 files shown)


jkelleyrtp:
Looks great - thank you so much! LiveView is pretty slick and I'm excited to get axum support.

When checks pass I will merge it as is.

jkelleyrtp (approved):

PR #370, opened 2022-04-24
concept: move rsx out to its own crate

jkelleyrtp:
Trying this again... the first time I tried this we ran into errors with proc_macro_error. Not getting the same errors again, thankfully.


jkelleyrtp:
#425 will need to come in and will likely conflict with master if I merge this... so I'm going to let that PR finish the work that was started ehre.

PR #377, opened 2022-04-26
FEAT: Allow starting live view apps with props

WIGGLES-dev:
I added a new method the the adapters and changed the code to allow users to start a live view app with props for the root component.


jkelleyrtp:
Looks great! Can you `cargo fmt` so checks pass and I can merge?

jkelleyrtp (approved):

PR #378, opened 2022-04-26
fix: unmarking component as dirty in silent write

koptan:
Quick fix for Fermi crate to avoid re-render in silent write.


jkelleyrtp:
Thank you!

PR #382, opened 2022-04-28
Fix cfg conditions for wasm

DusterTheFirst:
In the src/lib.rs file, there were `cfg` conditional compilation flags around the imports of `use_eval` which were written as

```rs
    #[cfg(all(target = "wasm", feature = "web"))]
```
which were meant to only enable the following code on the wasm target, but these were setup incorrect. The correct way to check if on wasm is the following:
```rs
    #[cfg(all(target_arch = "wasm32", feature = "web"))]
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,10 +6,10 @@ pub mod hooks {
     #[cfg(feature = "hooks")]
     pub use dioxus_hooks::*;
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Thank you :)

jkelleyrtp (approved):

PR #387, opened 2022-04-30
Renderers are now packages, not features.

jkelleyrtp:
This PR changes how we do packages since things have started to become unwieldy when adding functionality to crates like the router, fermi, and 3rd party renderers.

Now, all renderers are their own package and need to be added to Cargo.toml as such:

```toml
[dependencies]
dioxus = "0.3"
dioxus-router = "0.3"
dioxus-web = "0.3"
```

Instead of `dioxus::web::launch` you'll do `dioxus_web::launch`.

For releases, we'll using the minor version to keep things in lock step, but freely bump minor versions as new updates are pushed. It should've been this way since the beginning but I was experimenting. This also gets rid of the rsx-prelude which was a patch until we had a better solution.

The benefits:
- 3rd party crates can be released without bumping dioxus as a whole
- rsx prelude is now easier to deal with in 3rd party crates
- tests and examples can now be moved into the 3rd party crates without conflicts
- vscode and cargo no longer yells at me when I do circular imports

The stuff that remains under "dioxus" includes:
- core
- core-macro
- html
- hooks
- rsx
- rsx-interpreter (until we've moved to a more flexible hot reloading system)


packages/router/src/components/router.rs
```diff
@@ -40,81 +38,13 @@ pub struct RouterProps<'a> {
...
-                    }
-
-                    for route in svc.ordering.borrow().iter().rev() {
-                        regen_route(*route);
-                    }
-                }
-            }
-        });
-
-        cx.provide_context(svc)
```
rMazeiks, line 117: deleting this seems sus and unrelated, is it intentional?

rMazeiks (commented):
I don't understand where a bunch of changes are comming from; this is supposed to be just a re-structuring right? Might wanna check what Git has done here in the diff. Anyway, the other stuff looks good!

packages/router/Cargo.toml
```diff
@@ -38,22 +38,18 @@ serde = { version = "1", optional = true }
 serde_urlencoded = { version = "0.7.1", optional = true }
 
 [features]
-default = ["query"]
+default = ["query", "hot-reload"]
```
Demonthos, line 41: I'm hesitant to put this in the default features right now. It gives access to change the html if anyone has access to ./_dioxus/hot_reload. For most cases this shouldn't be an issue, but for site hosting websites for example notify it could pose issues. It also almost definitely has worst performance right now (that should improve with subtree memorization).

packages/router/examples/simple.rs
```diff
@@ -37,10 +37,10 @@ fn BlogList(cx: Scope) -> Element {
     cx.render(rsx! { div { "Blog List" } })
 }
 
-fn BlogPost(cx: Scope) -> Element {
```
Demonthos, line 40: Commenting this out causes the test to stop compiling because it is referenced in line 25

packages/router/tests/ssr_router.rs
```diff
@@ -13,16 +13,16 @@ fn generates_without_error() {
     assert_eq!(out, "<nav>navbar</nav>default<!--placeholder-->");
 }
 
-fn app(cx: Scope) -> Element {
```
Demonthos, line 16: Commenting this out causes the test to stop compiling. We could add ssr as a dev dependency to the router.

Demonthos (changes requested):
This seems much more organized. 🙂
A few things I noticed that don't fit specific lines:
- There are some dioxus::{plaform} imports left in the docs/guide/examples
- The tui_benchmark should probably be moved from the dioxus to the tui crate
- I think we can remove rsx-prelude now

packages/dioxus/examples/core_reference/antipatterns.rs
```diff
@@ -0,0 +1,185 @@
+//! Example: Antipatterns
```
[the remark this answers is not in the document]
jkelleyrtp: merge madness :) I need to figure out which ones slipped through

packages/router/Cargo.toml
jkelleyrtp: Whoops, got included on accident. Thanks for pointing it out.

PR #388, opened 2022-04-30
change with_ctrl_c_quit to without_ctrl_c_quit for tui config

Demonthos:
The default config has ctrl_c_quit enabled, but the method with_ctrl_c_quit sets it to enabled again. It should disable it instead.


jkelleyrtp:
Whoops - thank you!

PR #391, opened 2022-05-04
Offset coordinates for mouse events

### LINKED ISSUE #389: Getting mouse position relative to target element
## Specific Demand

I am trying to build an interactive element that creates a new Thingy wherever the user clicks. Since the container element can be positioned anywhere on the page, I would need the mouse coordinates relative to the container, not just the window. It seems that the client/page coordinates are relative to the window, and the screen coordinates are relative to the screen.

In JavaScript, the relative coordinates can be calculated by subtracting the bounding rectangle coordinates of the target element from the client coordinates of the mouse. However, I'm not aware of any way of querying the page layout in Dioxus.

Is there any way of getting the target element-relative coordinates of a mouse event?

My use-case involves clicking to create new Thingys and dragging those Thingys around within the container.

## Implement Suggestion

For the most flexibility, it would be cool to be able to make arbitrary queries about the DOM elements and their layout (imperatively, with something like the React `ref` API but in Rust) – but I do realize this level of DOM/Rust interop is difficult, especially with the need to support multiple targets. So a temporary workaround of simply including `parent_x, parent_y` coordinates would be really helpful too for many use cases!

rMazeiks:
- Document `MouseData` fields (copypasted from MDN)
- Add fields for `offset_x` and `offset_y`
- Adds an example showcasing different coordinate systems.

I tested this on desktop (ubuntu), web, and TUI still works. Didn't test SSR, but hopefully, probably it should be the same thing.

Note: on TUI, all of the coordinates are equal to the same `x, y` values. This is technically incorrect, but I see no easy way of calculating the offset here, so I did the same for `offset_x/y` coordinates. I left a note about this in a comment.

Thoughts?

Closes #389.


jkelleyrtp (approved):
This is awesome! Thank you for adding this!

Edit: for TUI, row/col is roughly x/y. Each character is 1px which makes TUI a bit strange. This solution is good for now - TUI will need some more polish as we near maturity.

Edit2: Liveview interpreter should eventually reuse the same interpreter as Desktop.

PR #397, opened 2022-05-06
only trigger onmousemove if the position has changed (tui)

Demonthos:
onmousemove was not checking if the mouse has moved. onmousepress and onmousereleased were switched


jkelleyrtp:
Whooops - thank you!

Demonthos:
Yeah, sorry about that. The way I had the example set up, it didn't differentiate from mouse up and down so I missed it 🤦‍♂️

PR #400, opened 2022-05-06
add query context (tui)

Demonthos:
Some elements need to know the size of the current element. For example the slider element in the input element needs to know how wide the object is to handle mouse input.


jkelleyrtp:
Can you rebase this and then merge it?

jkelleyrtp (approved):

jkelleyrtp:
Also, I'd like to implement some of these native widgets in the renderer in a way that doesn't require platform-specific contexts to do things like measurement.

Perhaps it's time to actually implement refs?

PR #401, opened 2022-05-07
Fix: AttributeValue bug; tests; CI

rMazeiks:
Fixes:
- String attributes incorrectly rendered as `[Object object]`
- Couple of failing tests
- Missing dependency in Github CI

With the new `AttributeValue`, attributes were serialized with values like `{Text: "value"}`, resulting in `[Object object]` when the interpreter applied them.

This simple fix restores the original behavior at least for Text values, by serializing them without the tag (just `"value"`).

I am not sure if this is a correct/complete approach, and I don't even know what the expected behavior would be for more complex variants such as `Vec3Float`. But at least the simple Text case is working again now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
           override: true
       - uses: Swatinem/rust-cache@v1
       - run: sudo apt-get update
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Thank you - I merged AttributeValue stuff knowing there would be some complications. I appreciate you fixing it!

PR #403, opened 2022-05-09
Update useref.rs defualt -> default

rowinbot:
Hi, I was going through the hooks files and spot this small typo.

Thanks again for such an amazing framework. 💯


jkelleyrtp:
Thank you!!

PR #405, opened 2022-05-11
Add some new window api for desktop

mrxiaozhuox:
- Print: open print modal
- SetZoomLevel: change webview zoom level

### DIFF
--- a/examples/window_zoom.rs
+++ b/examples/window_zoom.rs
@@ -1,0 +1,23 @@
+use dioxus::desktop::use_window;
+use dioxus::prelude::*;
+
+fn main() {
+    dioxus::desktop::launch(app);
+}
+
+fn app(cx: Scope) -> Element {
+    let window = use_window(&cx);
+
+    let level = use_state(&cx, || 1.0);
+
+    window.set_zoom_level(*level.get());
+    cx.render(rsx! {
+        input {
+            r#type: "number",
+            value: "{level_display}",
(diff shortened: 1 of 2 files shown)


examples/window_zoom.rs
```diff
@@ -0,0 +1,23 @@
...
+fn main() {
+    dioxus::desktop::launch(app);
+}
+
+fn app(cx: Scope) -> Element {
+    let window = use_window(&cx);
+
+    let level = use_state(&cx, || 1.0);
+
+    window.set_zoom_level(*level.get());
```
rMazeiks, line 13: looks like this sets the zoom level on every re-render?

in this example its kinda fine, but people might copy-paste it into their own work and that won't be great

may i suggest calling this inside the event handling closure?

examples/window_zoom.rs
```diff
@@ -0,0 +1,23 @@
...
+    cx.render(rsx! {
+        input {
+            r#type: "number",
+            value: "{level}",
+            oninput: |e| {
+                level.set(e.value.parse::<f64>().unwrap_or_default())
```
rMazeiks, line 19: default for f64 is 0; is this what you want?

you could also `unwrap_or(1.)`

rMazeiks (commented):
looks quite useful!

jkelleyrtp (approved):
Looks good - thanks for adding this! :)

PR #406, opened 2022-05-12
Rustify events api for WheelData and KeyboardData

rMazeiks:
As a follow-up to #398, this makes the API of WheelData and KeyboardData more rusty:

- Deprecate direct use of fields and deprecated web APIs
- Idiomatic accessors
- Deprecate use of our custom `KeyCode` in favor of keyboard_types types

Breaking change: introduces new private field, `code`, in KeyboardData (no longer possible to construct with literal syntax). The "alternative", `key_code`, is deprecated in the web standards and should have never been used in the first place imo. I think there's no sane way we can introduce the proper code without adding it as a new field.

Also, `locale` was a useless string equal to one of "", "locale" or "not implemented", on most platforms. I could not find any reference to it in the web docs; i think it's only applicable to some other even types. Anyway, I've removed it. Hopefully no one will cry "nooo i was using that" :)


jkelleyrtp:
Sorry, I think we merged stuff on top of this and two things are conflicting.

Do you mind resolving merge conflicts?

rMazeiks:
Welp, I finally merged the conflicts :)

I tested on Linux, and everything seems fine. However, we have found in the past that crossterm may have some platform inconsistencies/bugs, so it would be great if someone could check this on Windows/Mac (especially given that I rewrote some of the keyboard handling code as part of the refactor).

For this purpose, I have created the `all_events` and `tui_all_events` examples (the former replacing the previous event examples, which were a bit all over the place). Open it up, trigger some mouse/keyboard/wheel/focus events and check that things make sense in the log below.

image

packages/tui/src/hooks.rs
```diff
@@ -140,14 +142,18 @@ impl InnerInputState {
...
                     .as_ref()
-                    .filter(|k2| k2.0.key == k.key && k2.1.elapsed() < MAX_REPEAT_TIME)
+                    // heuristic for guessing which presses are auto-repeating. not necessarily accurate
+                    .filter(|k2| k2.0.key() == k.key() && k2.1.elapsed() < MAX_REPEAT_TIME)
                     .is_some();
-                k.repeat = repeat;
-                let new = k.clone();
-                self.last_key_pressed = Some((new, Instant::now()));
+                if is_repeating {
+                    let new = k.clone();
```
Demonthos, line 152: We need to set last_key_pressed every time not just when it is repeating because the key/instant needs to be set to start the timer the first time the event is received.

Demonthos:
Windows terminal is mostly working, the repeat attribute seems to be broken. With the patch from crossterm master everything else is working great!
Very Cool!

packages/tui/src/hooks.rs
rMazeiks: Good catch, thanks!

jkelleyrtp (approved):

jkelleyrtp:
Awesome!!!! Thank you for this!!

Issue #371, opened 2022-04-25
Overflow is not working


George-Miao:
**Problem**

Long text is not hidden or trimmed properly according to `overflow` attribute, which will cause out-of-boundry panic.

**Steps To Reproduce**

Code:

```rust
fn main() {
    dioxus_tui::launch(|cx| {
        cx.render(rsx!{
            div {
                "overflow": "hidden",
                "------------------------------------------------------------------------------------------------------------------"
            }
        })
    });
}
```

**Expected behavior**

Text should be trimmed or wrapped when length is greater than container's width.

**Screenshots**

![image](https://user-images.githubusercontent.com/38549822/165022309-fbb48e33-782b-429f-a48d-23c85f9e6ce4.png)

**Environment:**
 - Dioxus version: [`0.2.3`]
 - Rust version: [`rustc 1.61.0-nightly (8d60bf427 2022-03-19)`]
 - OS info: [`Linux`]
 - Terminal: [`Alacritty`]
 - App platform: [`tui`]

**Questionnaire**

- [X] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Demonthos:
Good catch. It looks like the text rendering in https://github.com/DioxusLabs/dioxus/blob/master/packages/tui/src/render.rs#L66 ignores the size of the layout right now. Stretch2 handles overflow, we just need to trim the text if it goes out of bounds. I think the width might need to be explicitly set to 100% for the div right now, but that is a bug in stretch2?

PR #412, opened 2022-05-17
Allow Multiple dependencies for states in the RealDom (native core)

Demonthos:
Depends on #https://github.com/rust-lang/rust/pull/96709 for GATs unless someone has a workaround

Refactor and generalize the incremental computation system for the RealDom to allow for mixed parent and child dependent state, and multiple dependencies for NodeDepState. The new trait for NodeDepState requires generic associated types to allow borrowing the elements of a tuple.

Why do we need this?
It allows some property of a node to depend on both the state of it's children and it's parent. Specifically this would make text wrapping, and overflow possible (#371). The layout of a parent node depends on the the layout of the children, and the layout of a text node depends on the text-wrap property of the parent.

- [x] #https://github.com/rust-lang/rust/pull/96709 or workaround
- [x] update documentation
- [x] allow multiple dependencies
- [x] allow mixed parent/child dependencies


packages/native-core/src/state.rs
```diff
@@ -79,6 +80,51 @@ pub trait ParentDepState {
 /// This state that is upadated lazily. For example any propertys that do not effect other parts of the dom like bg-color.
 /// Called when the current node's node properties are modified or a parrent's [PushedDownState] is modified.
 /// Called at most once per update.
+/// NodeDepState is the only state that can accept multable dependancies, but only from the current node.
```
rMazeiks, line 83: multable dependancies -> multiple dependencies?

packages/native-core/src/state.rs
```diff
@@ -92,6 +138,7 @@ pub trait NodeDepState {
     ) -> bool;
 }
 
+/// Do not implement this trait. It is only meant to be derived and used through [RealDom].
```
rMazeiks, line 141: also consider `#[doc(hidden)]` if this is an internal implementation detail

rMazeiks: (it's unclear to me; perhaps it's meant to be used but not implemented? if so, disregard comment)

packages/native-core/src/state.rs
Demonthos: oops

Demonthos:
I think GATs would be a cleaner solution, but this is possible with an associated type on the trait.

jkelleyrtp:
This seems like a rather complex system that might be worth refactoring in the future to be less trait-heavy so we don't run into issues with GATS/associated types since that gets pretty gnarly. 

I dug into Taffy recently and was happy with a their ECS approach. We might be able to move some of the function values into an ECS-type structure and get a similar system but with less complexity.

http://github.com/dioxusLabs/taffy

But, we don't have a public facing crate with compatibility to worry about, so there's no harm in getting to a working point and then refactoring or cleaning up some structures later.

jkelleyrtp (approved):

PR #421, opened 2022-05-23
fix broken doc links

Demonthos:
Some doc links are broken or outdated. This fixes them.

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -9,7 +9,7 @@
 //! ## Notice:
 //!
 //! The inspiration and code for this module was originally taken from Dodrio (@fitzgen) and then modified to support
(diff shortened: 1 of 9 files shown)


packages/core/src/diff.rs
```diff
@@ -44,13 +44,13 @@
 //! into a promise-like value. React will then work on the next "ready" fiber, checking back on the previous fiber once
 //! it has finished its new work. In Dioxus, we use a similar approach, but try to completely render the tree before
 //! switching sub-fibers. Instead, each future is submitted into a futures-queue and the node is manually loaded later on.
-//! Due to the frequent calls to [`yield_now`] we can get the pure "fetch-as-you-render" behavior of React Fiber.
```
Demonthos, line 47: I was not around at the time this existed, but it seems like work_with_deadline is the closest equivalent?

jkelleyrtp:
Thank you!!

Issue #416, opened 2022-05-21
onfocus event  can not be triggered


duan-zhou:
**Problem**
```rust
use dioxus::prelude::*;
use gloo::console::log;

fn app(cx: Scope) -> Element {
    cx.render(rsx!( div {
        input {
            name: "input_test",
            value: "test",
            r#type: "text",
            onfocus: |_| {log!("focus")},
        }
    }))
}

fn main() {
    dioxus::web::launch(app);
}
```
this is my code, when focus on `input` element, no log print, but if change it to `onchange`, it can work.
I find the event would be attached to `main` div when with Firefox inspector
![image](https://user-images.githubusercontent.com/65211298/169666389-83a95f0c-a2b7-4a4e-8093-3d108c344a3b.png)
but in other html example, it is attached to `input` element directly.
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onfocus
![image](https://user-images.githubusercontent.com/65211298/169666466-2cc38b67-da15-418e-9ae9-4aa7563a52d4.png)

I'm not sure that whether the event position result in this problem.

**Environment:**
 - Dioxus version: [e.g. v0.24, `master`]
 - Rust version: [e.g. 1.61, `stable`]
 - OS info: [e.g. win10, Firfox 100]
 - App platform: [e.g. `web`]

rMazeiks:
Hi,

Thanks for the report, I can confirm this is a problem on desktop as well.

I suspect the issue is that the [focus event](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event) does not bubble, but Dioxus only listens for events at the top level.

As a workaround, you can use the `focusin` event, which is pretty much the same, except it works:

```
onfocusin: |_| {println!("focus")}
```

(this is because `focusin` does bubble)

However, this issue with `onfocus` would probably have to be fixed somehow. Should Dioxus warn against/disallow `onfocus`? Or allow it but use `focusin` under the hood? Not sure... this is starting to feel hacky.

duan-zhou:
thanks,  I found the `onfocusin` solution in the former issue, I tried and found that it can compilied but has runtime error like this
![image](https://user-images.githubusercontent.com/65211298/169730522-d25b080a-0b34-4296-9730-56750d6e8abe.png)

rMazeiks:
Duplicate of #248 and #304

PR #422, opened 2022-05-23
Fix web missing event types

rMazeiks:
Fixes bug where using the `focusin` event on the web caused a panic.

This issue was discussed in #416 (but does not fix use of `onfocus`). Tested on the web, `focusin` now works.


rMazeiks:
To be clear, this does *not* fix #416 despite a mistake in a commit message indicating so. I hope github's automagic does not get in the way :/

mrxiaozhuox (approved):

jkelleyrtp:
Much better, thank you for adding these.

PR #425, opened 2022-05-26
Hot Reloading Rsx

Demonthos:
Enable hot reloading for the rsx macro.

### How?

- If the hot reload feature is enabled, the rsx macro parses the rsx and determines what variables are captured, and any iterators or components that are built.
- It then passes that context to a interpreter which looks for the text of the rsx macro within a global context with the code position as a key ([using panic's location](https://doc.rust-lang.org/std/panic/struct.Location.html#method.caller)).
- This approach is renderer independent

### Why?

- Faster iteration times

### Limitations

- Components, iterators, listeners, and expressions cannot be hot reloaded
- Change to what variables are captured cannot be hot reloaded
- It requires the text editor or cli to track the line/character of the rsx macro as it moves.
- Nested rsx calls cannot be hot reloaded

- [x] capture anything that cannot be interpreted
- - [x] variables
- - - [x] handle formatting options
- - [x] expressions in attributes
- - [x] components
- - [x] iterators
- - [x] listeners
- - [x] cx (if supplied)
- [x] interpret rsx with the captured context
- - [x] elements
- - [x] components
- - [x] iterators
- - [x] text
- - [x] listeners
- - [x] keys
- [x] provide a way to edit the rsx index in different programs
- - [x] desktop with [interprocess](https://docs.rs/interprocess/latest/interprocess/index.html)
- - [x] web with WebSockets
- [x] better error handling
- [x] update elements when the rsx is set
- [x] documentation
- [x] better error reporting

Demo with https://github.com/DioxusLabs/cli/pull/43:

https://user-images.githubusercontent.com/66571940/174206798-1b73e42a-0b36-4bce-83c4-aa7d875ec800.mp4


Demonthos:
Before opening this PR I have some notes/painpoints:
- The interpreter assumes that formatted expression will resolve to the same string every time. Eg: in "{x}-{x}" is assumes the first {x} will be the same as the second {x}. This allows adding a new {x} without recompiling.
- It also assumes the expressions used in the rsx call do not depend on the order.
I can't think of a reasonable example where either of these would matter, but if it is an issue I can fix it.
- It passes/stores code as a String and re-parses it when it is used because syn's representation is not Send.

jkelleyrtp:
You put an absolute ton of work into this!!!

There's a lot of code here, but I do have some thoughts. In reality I might end up merging this branch as-is and massaging it once it's in master.

In particular, I think (hope?) we can resolve scopes/rsx calls without needing to probe the user's code directly. I don't really have a great design for this yet, but I think part of it might involve hashing each rsx! body and then lining up the source against the macro once the server is running. As I said, I don't quite have a great design yet, so I can't propose anything better. In theory - before we release hot reloading in the next big release - I want to enable it without any feature flags to be turned on/off.

I'm still reviewing it, but given the size of the PR, it might end up making more sense merging it and then massaging it once it's in. Especially since you started with the `jk/rsx-refactor` fork and completed that body of work... and I want to implement a few features in the post-rsx-refactor world.

Again - this is so awesome and I'm very excited to build apps with hot reloading working! :)

Demonthos:
Thanks for the feedback!

> In theory - before we release hot reloading in the next big release - I want to enable it without any feature flags to be turned on/off.

I would argue against removing the feature, but it could be enabled by default in debug mode.
1. I don't see a way to get hot reloading working without some amount of runtime overhead, but I could be missing something. I haven't benchmarked it but I am assuming pre-compiling the rsx will be faster than interpreting it.
2. There also could be some security concerns. The desktop implementation has a open ipc port and the web implementation will try to connect to a new WebSocket. Both of them would allow changing the rsx and indirectly the html and js. That could allow any malicious desktop program to watch for user inputs, impersonate the organization that created the program, etc.

> I dove into this some more, and I think we might be able to get scopes resolved from rsx! calls using the file!, row!, and col! macros.
> https://doc.rust-lang.org/std/macro.file.html

The file!, row!, and col! macros seem similar to panic's location tracker but with better documentation and more well defined behavior! I didn't know they existed. I changed that and also added env!("CARGO_MANIFEST_DIR") to find the absolute path instead of the path relative to the current crate.

jkelleyrtp:
The more I think about it, the more I agree with this approach - at least for the near term.

In the future I want to add some sort of internal cache in the VirtualDom for rsx! calls. This metadata doesn't really exist currently - the only metadata we have right now is the name of components. In dev mode, we want all this metadata exposed to provide good tooling, but we want this metadata hidden when the app is built in release mode.

Whenever I add rsx! call caching and static-ify of rsx! call bodies, then we can reconsider how deeply integrated hot reloading is in the core/web/desktop libraries.

In theory, each virtualdom registers itself with a remote interpreter, rsx! calls are cached on the interpreter's side, and the hot reloading system only needs access to the JS-interpreter (via injection from a dev server) since the JS interpreter provides enough information to match up rsx! calls to the corresponding source. There's a slight runtime overhead for this, but the performance gains of caching rsx! calls would outweigh it.

However, this body of work requires a lot of knowledge about the vdom implementation and frankly should be done after this is merged.

Everything in the PR looks good, but I might play with a little bit of the code style after merging it. Is https://github.com/DioxusLabs/cli/pull/43 still in draft or would it be out of draft after we add this?

PR #430, opened 2022-06-01
docs: examples of mistakes

madcodelife:


rMazeiks (approved):
I wonder if we can now remove the `ignore` from the top of code blocks – that way, doctests will be run to check if the code compiles during `cargo test`, hopefully preventing mistakes like this.

If not, that's fine, I'm just sayin..

jkelleyrtp:
Thank you!

PR #431, opened 2022-06-01
Fix outdated guide section (React-like foo, set_foo)

AOx0:
I was reading the guide and I crossed upon a pair of examples that were not valid. So I updated them.


rMazeiks (approved):

jkelleyrtp:
Thank you!!

PR #434, opened 2022-06-09
fix(desktop): makes release build working

CGQAQ:
https://github.com/tauri-apps/wry/blob/454bf18084cd5f8879f2982d1c107719aa25a9d9/src/webview/mod.rs#L474-L477

open_devtools requires feature `devtools` or debug build

which this line of code used. so we need add this feature
https://github.com/DioxusLabs/dioxus/blob/fe1279a1c591a9ba57feb82a5162e076c940817b/packages/desktop/src/desktop_context.rs#L215

### DIFF
--- a/packages/desktop/Cargo.toml
+++ b/packages/desktop/Cargo.toml
@@ -20,7 +20,7 @@ serde = "1.0.136"
 serde_json = "1.0.79"
 thiserror = "1.0.30"
 log = "0.4.14"
-wry = { version = "0.16.0", features = ["devtools"] }
+wry = { version = "0.16.0" }
 futures-channel = "0.3.21"
 tokio = { version = "1.16.1", features = [
     "sync",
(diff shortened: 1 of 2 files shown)


rMazeiks:
Hi, thanks for the PR!

I can see why you made this change, but unfortunately, I don't think devtools should be available in release builds. Devtools should just be used for debugging, in development builds.

I think it would be better if the usage of devtools was disabled altogether in release builds – that would fix this issue as well. So, for example, [this line](https://github.com/DioxusLabs/dioxus/blob/0ac5e76d98f63de3b7e6593dcb6720532b8e5cde/packages/desktop/src/desktop_context.rs#L215) should have conditional compilation: in release builds, it should `log::warn!("Devtools are disabled in release builds")`.

In case the developer wishes to re-enable release devtools, we can put this behind a feature flag. But by default, I think devtools should be 100% disabled, just like in Wry.

CGQAQ:
Thanks for replying, @rMazeiks , I think you are right and I changed it. If anything else needs to be changed please let me know.

rMazeiks (approved):

jkelleyrtp:
Looks great. Thank you @rMazeiks and @CGQAQ !

PR #436, opened 2022-06-09
remove style attributes correctly for the js interpreter

### LINKED ISSUE #435: Style attributes are not removed with the web interpreter
**Problem**

In the web interpreter, style attributes are treated the same as normal attributes when removing attributes, which fails.

**Steps To Reproduce**
1. Run:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus::web::launch(app);
}

    let clicked = use_state(&cx, || false);
    if *clicked.get() {
        cx.render(rsx! {
            div {
                "This has been clicked"
            }
        })
    } else {
        cx.render(rsx! {
            div {
                background_color: "red",
                width: "100px",
                height: "100px",
                onmouseup: |_| {
                    clicked.set(true);
                }
            }
        })
    }
}
```
2. click on the red div

**Expected behavior**

The attribute that was removed should be removed in the html.

**Environment:**
 - Dioxus version: master
 - Rust version: 1.61.0
 - OS info: Windows
 - App platform: web, and desktop

**Questionnaire**

ealmloff:
If the namespace of an attribute being removed is style use style.removeProperty instead of removeAttributeNS.

fixes #435

### DIFF
--- a/packages/interpreter/src/interpreter.js
+++ b/packages/interpreter/src/interpreter.js
@@ -140,7 +140,9 @@ export class Interpreter {
   RemoveAttribute(root, field, ns) {
     const name = field;
     const node = this.nodes[root];
-    if (ns !== null || ns !== undefined) {
+    if (ns == "style") {
+      node.style.removeProperty(name);
+    } else if (ns !== null || ns !== undefined) {
       node.removeAttributeNS(ns, name);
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
Thank you!

PR #438, opened 2022-06-10
Use Taffy instead of stretch2 for tui and native core

Demonthos:
Switch to taffy where future development will take place.

Major changes:
- it outputs relative layout instead of absolute layout
- it doesn't pretend to support direction
- it doesn't pretend to support overflow


jkelleyrtp (approved):

jkelleyrtp:
Thanks! <3

PR #439, opened 2022-06-11
feat: declaratively add props to component using css like selectors

Nejat:
# _*** UPDATED - Added CSS Syntax support for selectors ***_

## Adds `add_as`, `selector`, `root` features to `props` attribute, to enable automatic injection of properties into an `rsx!` definition.

The basic principle is to be able to declaratively add values of a custom component onto elements or other components in the `rsx!` definition of that custom component.

### `add_as` - maps name

* `add_as` is an optional value that defines the target attribute name to add prop value as, _i.e._<br>
`my_class` => `class` or <br>
`clicky` => `onclick`

### `selector` adds attribute values or event handlers:

* `selector` is a `;` separated list of pseudo `css` like selectors that indicate where to add the value<br>

   *  `:root` - root element, only if root level has only one element
   *  `:any` - any element in level
   *  `tag | component` - _i.e., div | MyCustomComponent
   *  `:` - similar to `css` pseudo selector, marks the declaration of a child selector
   * `:first` - first child in elements list
   * `:last` - last child in elements list
   * `:even` - even children in elements list
   * `:odd` - odd children, in elements list
   * `:only` - only child, in elements list
   *  `:[nth[,nth]]` - comma delimited list of `nth` children
   *  `:last[nth[,nth]]` - comma delimited list of `nth` children from end
   *  `:[Nn±B]` - every N, _i.e._ `div:[3n]` _every 3rd element that is_ `div`
   *  `:[..nth]` - the first `nth` children
   *  `:[nth..]` - all children from `nth` onward
   *  `@` - similar to `:`, but marks the declaration child-of-type selector
   * `@first` - first child-of-type in elements list
   * `@last` - last child-of-type in elements list
   * `@even` - even children-of-type in elements list
   * `@odd` - odd children-of-type, in elements list
   * `@only` - only child of type, in elements list
   *  `@[nth[,nth]]` - comma delimited list of `nth` children-of-type
   *  `@last[nth[,nth]]` - comma delimited list of `nth` children-of-type from end
   *  `@[Nn]` - every N, _i.e._ `div:[3n]` _every 3rd_ `div`
   *  `@[..nth]` - the first `nth` children-of-type
   *  `@[nth..]` - all children-of-type from `nth` onward
   *  `>` - similar to `css` child selector
   *  `;` - is a divider for subsequent selectors, _values can be added more than once_

###  `root` adds attribute values or event handlers only to a `root` element:

`root` is equivalent to, therefore a shortcut of `selector = ":root"`

### one addition to `rsx!` syntax

To make the automatic injection possible the `rsx!` needs to know the custom context type as well as custom context instance name. The syntax has been extended to take the form `name:type;`, _i,e., `rsx! { cx: Component; ... }`_

This requirement for injection does not preclude or interfere with the original definition of `rsx!`, it just extends it for this use case.

## Manual Example

```rust
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default)]
    my_class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
    cx.render(rsx! {
        div {
            onclick: move |evt| cx.props.onclick.call(evt),
            a {
                "class": "{cx.props.my_class}"
            }
            &cx.props.children
        }
        div {
            onclick: move |evt| cx.props.onclick.call(evt),
            "class": "{cx.props.my_class}"
        }
    })
}

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct Component2Prop<'a> {
    #[props(default)]
    class: &'a str,
    #[props(default)]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component2<'a>(cx: Scope<'a, Component2Prop<'a>>) -> Element {
    cx.render(rsx! {
        div {
            Component {
                my_class: "{cx.props.class}",
                onclick: move |evt| cx.props.onclick.call(evt),
                &cx.props.children,
            }
            div {}
        }
        div {}
    })
}
```

## Declarative Equivalent

```rust
use dioxus::events::MouseEvent;
use dioxus::prelude::*;

use dioxus_sugar::attributes;

#[attributes(optional(div), exclude(class, name, id))]
#[derive(Props)]
pub struct ComponentProps<'a> {
    #[props(default, add_as = "class", selector= "div:[1]; div:first > a")]
    my_class: &'a str,
    #[props(default, selector= "div")]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
    rsx! {
        cx: ComponentProps;
        div {
            a {}
            &cx.props.children
        }
        div {}
    }
}

#[derive(Props)]
pub struct Component2Prop<'a> {
    #[props(default, add_as = "my_class", selector = "div > Component")]
    class: &'a str,
    #[props(default, selector = "div > Component")]
    onclick: EventHandler<'a, MouseEvent>,
    children: Element<'a>,
}

#[allow(non_snake_case)]
pub fn Component2<'a>(cx: Scope<'a, Component2Prop<'a>>) -> Element {
    rsx! {
        cx: Component2Prop;
        div {
            Component {
                &cx.props.children,
            }
            div {}
        }
        div {}
    }
}
```

## Motivation

The idea was born out of supporting [Dioxus-Sugar](https://github.com/Nejat/dioxus-sugar) which declaratively includes valid html attributes and event handlers for custom component properties. `Dioxus-Sugar` adds the attribute and event handler definitions to the custom component properties and this feature adds those and other custom properties into the custom component definition. Which lets me finish my ultimate reason for building any of these `Dioxus-Bulma` :)

## Supports CSS Selector and a more concise "rusty" equivalent

this is a list of supported selectors, the left side is the new css syntax and the right side is the original more concise "rusty" syntax

```
:not(p)						:not[p] 

element:nth-child(n)				element:[n]

element:nth-child(an ± b)			element:[an ± b]

element:nth-child(even)				element:even

element:nth-child(odd)				element:odd

element:nth-last-child(n)			element:last[n]

element:first-child				element:first

element:last-child				element:last

element:only-child				element:only


element:nth-of-type(n)				element@[n]

element:nth-of-type(an ± b)			element@[an ± b]

element:nth-of-type(even)			element@even

element:nth-of-type(odd)			element@odd

element:nth-last-of-type(n)			element@last[n]

element:first-of-type				element@first

element:last-of-type				element@last

element:only-of-type				element@only


PLUS Non-CSS

:root

:any						*


element:nth-child(n1..n2)			element:[n1..n2]

element:nth-child(..n)				element:[..n]

element:nth-child(n..)				element:[n..]


element:nth-of-type(n1..n2)			element@[n1..n2]

element:nth-of-type(..n)			element@[..n]

element:nth-of-type(n..)			element@[n..]
```


Nejat:
To correct mistakes in my prior PR, I'm new to public PRs, I re-forked, updated and re-created this PR.

To read some of the original discussions please refer to [#364](https://github.com/DioxusLabs/dioxus/pull/364)  

Since I created a new fork, I also lost my commit history, so I archived the [original fork](https://github.com/Archived-Repo-History/dioxus) if you are interested.

Thanks

jkelleyrtp:
Thank you for all this work! Sorry it sat here so long - just got back from vacation and a bunch of other stuff.

I have a few thoughts and some hesitations with bringing this big of a feature set into the primary Props macro.

My main hesitation is that it's a big feature, following a CSS-focused pattern, and it complicates the rsx! and props implementations quite a bit.

Is there a way we can implement the same selectors or sugar macro to allow auto-casting of elements onto components without having to invent the "Custom Context" paradigm? Maybe there's a `cx.render_with(props.filter(), rsx! { ` we could invent that would allow any form of extension/context? Or something like `cx.render(rsx!{}).apply(props.filters())` to transform an existing VNode at runtime? Could a wrapper macro achieve similar functionality, but at compile time?

Could we get a similar functionality by supporting an advanced spreading system? I'm not 100% convinced yet that the CSS selector system is *that* much simpler than spreading in attributes. 

Ultimately we'd either want a more general purpose extension system, or some way of facilitating this type of work in the form of 3rd party crates, ideally not paying a runtime performance penalty.

Nejat:
> Thank you for all this work! Sorry it sat here so long - just got back from vacation and a bunch of other stuff.

Thank you for taking the time to look at this, I have added my thoughts, comments and questions below ...

> I have a few thoughts and some hesitations with bringing this big of a feature set into the primary Props macro.
> 
> My main hesitation is that it's a big feature, following a CSS-focused pattern, and it complicates the rsx! and props implementations quite a bit.

From the onset it has been my intention to follow the structure of the library, tread lightly with minimum changes to the public interface and keep the logic as self-contained as possible, that is after I realized my first attempt to be completely third-party was not possible. Which is why I decided to contribute to the project.

Unfortunately, the complexity comes from the requirements of the injection logic to be work, `rsx!` had to be aware of the `Props` context to be able to identify which props were declared to be injected.

Can you explain why you believe it complicates `rsx!` and props? Maybe I can find a less complicated approach based on your viewpoint.

Would hiding this behind a `Feature` gate make this less complicated?

> Is there a way we can implement the same selectors or sugar macro to allow auto-casting of elements onto components without having to invent the "Custom Context" paradigm? Maybe there's a `cx.render_with(props.filter(), rsx! { ` we could invent that would allow any form of extension/context? Or something like `cx.render(rsx!{}).apply(props.filters())` to transform an existing VNode at runtime? Could a wrapper macro achieve similar functionality, but at compile time?

I think runtime injection would be nice, as it would support a dynamic tree, but should be optional imo. I prefer the idea of an outer wrapper macro, but how would that work? I can't think of a way to design it without the outer macro knowing and ultimately being tied to the internal workings of the `rsx!` which would make it too fragile. _Throughout the different iterations of the many implementations this code went through, I found at one point the logic "knew too much" of inner logic, until I fixed it._

> Could we get a similar functionality by supporting an advanced spreading system? I'm not 100% convinced yet that the CSS selector system is _that_ much simpler than spreading in attributes.

Can you explain what you mean by an "advanced spreading system"? I am willing to work on it if it makes things better.

> Ultimately we'd either want a more general purpose extension system, or some way of facilitating this type of work in the form of 3rd party crates, ideally not paying a runtime performance penalty.

This would be great, and this is the first approach I considered, but falling short on any concrete ideas or use cases I followed the current path hoping it was generic enough to be useful for a larger audience.

I honestly think this functionality is going to be very important in larger projects where productivity and lots of boilerplate functionality are going to create a great deal of friction, so a solution inline with projects aspirations is something I'm willing to look into, I just need other ideas. 

Once again thanks for taking the time.

Nejat:
I had not coded in React and as I was thinking about your response, I decided to look through tutorials and read through some documentation to get a better understanding of it capabilities. Here are my thoughts from this ...

> I think my primary concern here is the little language that's being used for injection. There's a lot of HTML-esque patterns we follow, but at least the core implementation and macros are completely agnostic to HTML/CSS. The rsx! macro - both in UX and code - knows nothing about HTML/CSS or anything particularly special. All it knows is how to transform the rsx! syntax into Rust code. I'm worried that coupling this tiny language to the rsx!/Props macros just doesn't fit the ethos of the separation between core and web that we're going for.

I actually thought about this while I was implementing this feature and reasoning about why I felt it would be an important addition. The main thing I came away with was that Dioxus is backend agnostic, and that it's virtual dom could be applied to any backend technology, web, os native, tui, etc. 

_i.e. xaml, which much like html is an xml based gui document graph, and much like html has something similar and in some way more powerful css alternative, called styles._ 

What I encountered while writing a Dioxus Bulma Components library, was that for the components to behave as their html counter parts, buttons, input etc., they would need to implement a lot of the base component's attributes and event handlers. Besides having to write a lot of boring boilerplate code, it occurred to me that it would also be very error prone, especially if you didn't know all of the attributes and events. 

My strategy was threefold,  scrap the definitions of html elements, their attributes and event handler from source of truth, build a derive macro to add them to the component properties `struct` and implement a way to automatically apply those properties with pretested attributes and event handlers.

- The first solution was [Web Reference](https://github.com/Nejat/web-reference), a queryable web reference library.
- The second solution was [Dioxus Sugar](https://github.com/Nejat/dioxus-sugar), a library of proc macros for applying standard html element attributes and event handlers to Dioxus `Props` `structs` with compile time error checking of valid html definitions. _not 100% feature complete_
- The third solution originally attempted to inject attributes with custom macros externally, but when I learned that this is not possible with rust macros, they did not expand properly, it made me wish there was native plugin supportin Dioxus, lacking that I looked into this approach.

The thought of writing all of those automatically generated attributes by hand seemed counter intuitive. The problem grew in my mind when I considered my ultimate goal, which was to create several other libraries, such as Materials, Bootstrap, etc. Which made me think this would be useful for any component's library developer developing for the Html backend.

While I agree with and celebrate Dioxus' goal of being backend rendering agnostic, I think pairing HTML to the semantics of CSS is a view that limits the usefulness of the most powerful part of CSS, it's selector functionality. They are not strongly coupled and the CSS selector language can be applied to many problems, including this one.

My initial implementation of the selector was more rust like and terse, see the chart above, I only added the more verbose css syntax based on comments in the prior PR [#364](https://github.com/DioxusLabs/dioxus/pull/364). As implemented them I saw their usefulness for people coming from web development.

I think, if I can convince you of this feature, the inject and it's selector dialects can all be feature gated to be opt in only.

> That being said...
> 
> If I had to implement this in the way that is agnostic but fits.... I would
> 
> * Keep the "extend a component with an element" syntax (in a generic way)
> * Use the spread operator to keep all rendering in the same body.
> 
> ```rust
> #[attributes(optional(div), exclude(class, name, id))]
> #[derive(Props)]
> pub struct ComponentProps<'a> {
>     #[props(default, add_as = "class")]
>     my_class: Attribute<'a>,
> 
>     #[props(default)]
>     onclick: EventHandler<'a, MouseEvent>,
> 
>     children: Element<'a>,
> }
> 
> #[allow(non_snake_case)]
> pub fn Component<'a>(cx: Scope<'a, ComponentProps<'a>>) -> Element {
>     cx.render(rsx! {
>         div { 
>             ..props.my_class,
>             ..props.onclick,
>             a {}
>             &cx.props.children
>         }
>         div { ..props.onclick }
>     })
> }
> ```

From my understand of what I read about the spread operator in React documentation, they are used with complete objects, at least the few examples I found. Based on your example, spread would offer little benefit over writing attributes individually. Additionally, `add_as` wouldn't work in compile time.

> The parallel in Rust is stuff like macro-rules. They have their own little syntaxes, aren't really obvious on first brush, but they do carry their weight. I'm not sure if the little DSL here carries its own weight, especially if the alternative is just placing the attributes directly into the code in the appropriate spots. I think I'd be more enthusiastic if a feature like what you've proposed already existed in React (or elsewhere in the JS ecosystem), but all I've seen is the spread syntax, which is where I think Dioxus is currently falling short.

Because React is HTML only this issue seems to be solved in a different manner. I found references to two functions [`React.useRef`](https://github.com/mui/material-ui/blob/c7ea9ae2b069e5a9bad39d648903866bdaa9f6db/packages/mui-base/src/ButtonUnstyled/ButtonUnstyled.tsx#L61) and [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html) that seem to define underlying functionality and refer to the underlying element. But again I'm not 100% on this either.

---

In the end, I really took to Dioxus very easily and feel it will be a strong product when it's feature complete. This comes down to how well it was designed and how ergonomic it is, it's why I tried very hard to find the most idiomatic Rust/Dioxus approach I can come up with from the beginning, so for that reason I respect any final decisions from the team.

Thanks

Issue #304, opened 2022-03-11
ondoubeclick/onblur events doesn’t work


hanpei:
**Problem**

some event listener did not work on web platform,
e.g "ondoubleclick" / "onblur" / "onfocusout",
but "onclick" / "oninput" works well.

```
 label {
            ondoubleclick: move |e| {
                info!("double click");
                editing.set(*id);
            },
            "{title}",
        }

input {
         onblur: move |_| {
              editing.set(0);
              info!("onblur")
          },
}
```

**Environment:**
 - Dioxus = { version="0.2.3", features=["web"] }
 - Rust version: [e.g. 1.59.0, `stable `]
 - OS info: [e.g. MacOS]
 - App platform: [e.g. `web`]

mpfaff:
> You can use `onfocusin` and `onfocusout` instead to get the same effect.

`onfocusout` panics for me:

```
panicked at 'unsupported event type', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/dioxus-web-0.2.1/src/dom.rs:378:13
```

rMazeiks:
@mpfaff, I've made https://github.com/DioxusLabs/dioxus/pull/422 to address the issue with `onfocusin` and `onfocusout` on the web. Until it gets merged (maintainer on vacation), you can use my fork if you trust me :smile: :

```
[patch.crates-io]
dioxus = { git = "https://github.com/rMazeiks/dioxus.git", branch="fix-web-missing-event-types" }
```

PR #452, opened 2022-06-22
Fix nonbubbling web events

### LINKED ISSUE #248: Mouseenter/mouseleave does not work properly with event delegation
**Problem**

Due to how we use listener multiplexing, all mouseenter events actually trigger on the root element rather than the target element. This means we cannot determine which element was the cause of the event and we cannot actually call the event. 

For most cases you should use mouseover because it's essentially mouseover but with proper bubbling.

**Steps To Reproduce**
We would expect this example to produce both mouseentered/mouseover events, but we only get mouse over events. If we attach JS handlers through arbitrary attribute support, then we can get logs from JS.

```rust
use dioxus::{events::onmouseover, prelude::*};

fn main() {
    dioxus::desktop::launch(app);
}

    cx.render(rsx! {
        div {
            background: "red",
            onmouseenter: |_| {
                println!("mouse entered");
            },
            "onmouseenter": "console.log('mouse entered')",
            onmouseleave: |_| {
                println!("mouse leave");
            },
            "onmouseleave": "console.log('mouse left')",
            onmouseover: move |_| {
                println!("mouse over");
            },
            onmouseout: move |_| {
                println!("mouse out");
            },
            "hi"
        }
    })
}
```
**Expected behavior**

Mouseenter should behave properly.

**Environment:**
 - Dioxus version: [e.g. v0.17, `master`]
 - Rust version: [e.g. 1.43.0, `nightly`]
 - OS info: [e.g. MacOS]
 - App platform: [e.g. `web`, `desktop`]

**Questionnaire**

How does React/Infero/Preact handle this case with event delegation?

Demonthos:
Handles bubbling events in the virtual dom and web interpreter. It uses the global event handler map if the event bubbles otherwise it uses creates an event handler that is specific to the element.
fixes #248, #304, and #416


jkelleyrtp:
Looks good!


My only nit: in the listener map I think it'd be better to have add/remove as single methods with a conditional on the bubbling inside the methods themselves.

jkelleyrtp (approved):

PR #455, opened 2022-06-24
fix tui event tests

Demonthos:
Updates the tests with the focus system (tab to focus the element before triggering the keydown event). This also uses a future that is tracks the number of times it has been polled instead of using a timer, which should make failing tests faster and generally make the test more sound.


jkelleyrtp (approved):
Much better - the poll mechanics could be used elsewhere in our testing. Definitely an improvement over JS where the specifics of the event loop are fuzzy WRT user code.

PR #458, opened 2022-06-25
change the hot_reload feature to hot-reload

mrxiaozhuox:
I think use `hot-reload` is better


Demonthos:
I agree hot-reload is more consistent with native-core features in dioxus, could you change the hot_reload feature in the other crates (core-macro, decktop, web, etc) to match?

Demonthos (approved):
looks good

PR #459, opened 2022-06-25
Clean up rsx imports for components and tests

Demonthos:
There is some duplication in the imports when tests or components need to use the rsx macro. This PR extracts the necessary imports into a package. This makes changing the rsx macro easier and adds support for hot_reloading for the router.


jkelleyrtp:
This is fine for now to get master compiling again, but I think it would be ultimately worthwhile to not need the rsx-prelude package.

Demonthos:
> This is fine for now to get master compiling again, but I think it would be ultimately worthwhile to not need the rsx-prelude package.

I was lazy squashing fixing hot reloading router compatibility, fixing keys, and refactoring into one, but I could easily pull out specific parts if you want me to.
All of this could be done without the rsx-prelude package, by just copy pasting the imports.

PR #462, opened 2022-06-28
fix path separator for hot reloading

Demonthos:
The path to the file in hot reloading currently produces a different result on unix and windows because they have different path separators, but when building for wasm it fails to detect the os.
Solution: Keep the crate and file path separate

### DIFF
--- a/packages/rsx_interpreter/src/lib.rs
+++ b/packages/rsx_interpreter/src/lib.rs
@@ -24,7 +24,8 @@ lazy_static! {
 // the location of the code relative to the current crate based on [std::panic::Location]
 #[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize)]
 pub struct CodeLocation {
-    pub file: String,
+    pub crate_path: String,
+    pub file_path: String,
     pub line: u32,
     pub column: u32,
 }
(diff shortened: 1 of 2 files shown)


packages/rsx_interpreter/src/lib.rs
```diff
@@ -83,14 +84,12 @@ macro_rules! get_line_num {
     () => {{
         let line = line!();
         let column = column!();
-        let file = file!();
+        let file_path = file!().to_string();
+        let mut crate_path = env!("CARGO_MANIFEST_DIR").to_string();
```
rMazeiks, line 88: Now that they're separate things, we don't require any formatting. So I think we can avoid allocating and just store them as `&'static str`? (blazing fast 🚀🚀)

Also the `mut` seems unnecessary

Demonthos: This seemed like a good idea, but I don't think it works. The macro will produce a &'static str, and we can get serde to work with it, but the cli cannot produce &'static str when it detects changes in the code, so I'm going to revert this change.

rMazeiks: Ah true, I didn't know the CLI constructs it as well

PR #466, opened 2022-06-29
check that an node is an element before removing listeners

### LINKED ISSUE #465: Removing text nodes fails
**Problem**

The js interperter panics when a text node is removed.

**Steps To Reproduce**

```rust
use dioxus::prelude::*;

fn main() {
    dioxus::web::launch(app);
}

    let state = use_state(&cx, || false);
    if *state.get() {
        cx.render(rsx! {
            div {}
        })
    } else {
        state.set(true);
        cx.render(rsx! {
            div {
                "hello"
            }
        })
    }
}
```

**Expected behavior**

The text should appear and then immediately disappear.

**Environment:**
 - Dioxus version: master
 - Rust version: 1.61.0
 - OS info: Windows
 - App platform: web

**Questionnaire**

Demonthos:
When a text node is removed the interpreter trys to get the id attribute to remove any listeners, but that only works on elements, so it panics.
fixes #465

### DIFF
--- a/packages/interpreter/src/interpreter.js
+++ b/packages/interpreter/src/interpreter.js
@@ -15,41 +15,43 @@ class ListenerMap {
     this.root = root;
   }
 
-  createBubbling(event_name, handler) {
-    if (this.global[event_name] === undefined) {
-      this.global[event_name] = {};
-      this.global[event_name].active = 1;
-      this.global[event_name].callback = handler;
-      this.root.addEventListener(event_name, handler);
-    } else {
-      this.global[event_name].active++;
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
Seems a little bit like codesmell to have these checks on every element remove.

I'm going to dig in and see if we can rid the need of the nonbubbling listener map altogether. That will mean we leak (maybe?) one callback max and reuse it for every nonbubbling listener. 

I believe react is going to completely remove event delegation so our bubbling listener map could be replaced entirely with the one callback approach.

Fine for now, but maybe worth our time to look into it?

PR #468, opened 2022-06-29
Fix Diffing Equal Length Listeners

### LINKED ISSUE #467: Changing listeners fails
**Problem**

Adding and removing a listener causes a panic.

**Steps To Reproduce**

```rust
use dioxus::prelude::*;

fn main() {
    dioxus::web::launch(app);
}

    let state = use_state(&cx, || false);
    if *state.get() {
        cx.render(rsx! {
            div {
                onmouseout: |_| state.set(false),
                "goodbye"
            }
        })
    } else {
        cx.render(rsx! {
            div {
                onmouseover: |_| state.set(true),
                "hello"
            }
        })
    }
}
```

**Expected behavior**

- Hover the text
- It should change to goodbye
- Leave the text
- It should change back to hello

**Environment:**
 - Dioxus version: 0.2.4
 - Rust version: 1.61.0
 - OS info: Windows
 - App platform: Web, but this issue is on all platforms

**Questionnaire**

Demonthos:
fixes #467

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -402,12 +402,12 @@ impl<'b> DiffState<'b> {
 
         if old.listeners.len() == new.listeners.len() {
             for (old_l, new_l) in old.listeners.iter().zip(new.listeners.iter()) {
+                new_l.mounted_node.set(old_l.mounted_node.get());
                 if old_l.event != new_l.event {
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
Whoops... Thank you :)

PR #474, opened 2022-06-30
Rewrite Readme in example directory

### LINKED ISSUE #471: fix links in examples readme.md
just came to see some examples and have got "404 page not found" (The basics) straight away, not good.

rMazeiks:
The table was mostly useless, incomplete and with dead links, so I re-wrote it (and introduced some categories)

Closes #471 

---

While going through the examples, I noticed a few things:

inputs.rs - does not render different kinds of input. Is that intentional?

framework_benchmark - blazing slow (i have to wait seconds for the rows to be inserted). is that normal?

order.rs - purpose unclear despite comment

rsx_compile_fail.rs - confusing name, comment and purpose

svg.rs - doesn't appear to work
svg_basic.rs - doesn't appear to work

---

Also, the `core_reference` folder contains a bunch of duplicates, and cannot be run. While the documentation states that this is intentional, this makes it seem kinda useless, both for users learning Dioxus and for testability. I propose to get rid of this and move any useful examples out of there. This will also ensure that they compile during tests. Thoughts on this?


jkelleyrtp:
Thank you so much ❤️
This has just been one of those things that felt like a chore but brought down the overall polish.

----
- inputs.rs is a messy file and was added a long time ago to make sure all of the possible input types are covered. It really needs to be a test but I don't know how to test the webview properly
- framework_benchmark is significantly faster in release mode, but still slower than its WASM counterpart. The overhead of the webview is relatively high and we're pretty naive with serialization/deserialization on that boundary.
- order.rs should removed. It exists for a similar purpose to inputs.rs as an adhoc test
- I'm not sure what's up the svg examples. There's been some changes to the interpreter recently that might've broken them. They definitely worked for the 0.2.4 release.
----
The core reference folder should really be merged into the reference mdbook, especially since a number of them are meant not to compile.

jkelleyrtp (approved):

Issue #352, opened 2022-04-13
Many documentation pages are not rendering


csi-lk:
**Problem**

Many sections of the documentation are not rendering eg.

https://dioxuslabs.com/reference/guide/components.html
https://dioxuslabs.com/reference/guide/props.html
https://dioxuslabs.com/reference/reference/anti.html
https://dioxuslabs.com/reference/reference/inline.html

They show the heading but not the content

**Steps To Reproduce**

- Go to the docs on Brave or Chrome
- Click on a few of the reference of advanced guides

**Expected behaviour**

I can read the documentation

**Screenshots**

![image](https://user-images.githubusercontent.com/2224266/163097016-c47c91a4-7829-46b5-92c3-7616cc35eff5.png)

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

rMazeiks:
I think they just haven't been written yet.

e.g. see:

https://github.com/DioxusLabs/dioxus/blob/master/docs/reference/src/guide/components.md - empty

I think the empty pages were created to sketch out a structure for the docs, but I agree it's a bit confusing.

Feel free to write some of the missing docs if you want! You can learn a lot by looking at existing docs, examples, source code, and experimentation.

csi-lk:
Oh I didn't realise there was a "reference" section and a "guide" section, I just clicked on one and assumed they were the whole docs

Totally makes sense, I won't have time to start working on Dioxus for a few weeks still but i'll try to contribute some docs back then

Closing, cheers

erichulburd:
@rMazeiks It is quite confusing as a developer new to the project to see blank pages on the official reference and guide pages. I couldn't tell if there was perhaps a bug in how the page was rendering or Dioxus didn't support my browser.

Seems like there should be a link to a community page about contributing documentation on those empty pages, or they should only show up in "developer" mode.

PR #476, opened 2022-06-30
Delete blank docs

rMazeiks:
This was causing confusion (#352) and most of the "work in progress" pages haven't been touched since January.

Also deleted outdated example.


jkelleyrtp (approved):
Thanks!

PR #481, opened 2022-07-01
panic on errors in hot-reload if no error handler socket is connected

Demonthos:
Issue: If a application has the hot-reload feature enabled, but did not run through dioxus serve, the application will silently fail on errors.

Solution: When an error is produced warn the user the websocket was not connected and note that they should use hot reloading through dioxus serve --hot-reload, and panic with the error message.


jkelleyrtp:
This is fine provided that hot reload is not enabled by default. 

Again, I'd like to see if we can get the metadata required hot reloading integrated natively without features as to avoid the complexity blowup.

jkelleyrtp (approved):

PR #482, opened 2022-07-01
allow custom attributes to be captured for hot reloading

### LINKED ISSUE #479: Hot reload: Elements with custom attributes fail to render
This works fine:

```
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    dioxus::desktop::launch(App);
}

    cx.render(rsx! {
        div {
            "Hello, world!"
        }
    })
}
```

Adding a custom attribute makes the element disappear (it's not even in the DOM):

```
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    dioxus::desktop::launch(App);
}

    cx.render(rsx! {
        div {
            "customAttr": "whatev",
            "Hello, world!"
        }
    })
}
```

Disturbingly, I saw no error messages.

If I use the same syntax with an existing attribute (e.g. `class`) it works fine.

Demonthos:
Hot reloading currently only supports attributes from the html crate. This adds support for custom attributes.

```
div {
    "customAttr": "whatev",
    "customAttr": 0,
}
```

closes #479

### DIFF
--- a/packages/rsx_interpreter/src/captuered_context.rs
+++ b/packages/rsx_interpreter/src/captuered_context.rs
@@ -1,11 +1,13 @@
+use std::collections::HashSet;
+
 use dioxus_core::{Listener, VNode};
 use dioxus_rsx::{
     BodyNode, CallBody, Component, ElementAttr, ElementAttrNamed, IfmtInput, Segment,
 };
 use quote::{quote, ToTokens, TokenStreamExt};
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Both this and #473 are good to go, but you might have issues with merge conflicts. Feel free to merge both and massage anything once they're in!

PR #484, opened 2022-07-04
fix web imports and extract hot reload handlers into seperate files

Demonthos:
Extracts hot reload handlers into separate files to make things more readable and fix the web-sys console import being imported even when the hot-reload feature is off and the web-sys console feature is disabled.


jkelleyrtp (approved):
Much better. I'm going to explore some smarter testing strategies in CI to catch these issues we've been having.

PR #486, opened 2022-07-04
Rework Guide

rMazeiks:
- Merge Reference and Guide
- Also merge some examples from `examples/core_reference`
- Move unused content into a __unused directory (commented out stuff was really getting in the way)
- Assume the reader knows Rust (less handholding on basics like modules and borrow checker. there's no way we gonna teach rust from scratch anyway)
- Most code samples made into compilable examples and included as snippets. That way we can CI this thing to make sure syntax is not out of date.
- As a consequence, updated some code samples

I (think I) improved the organization quite a bit. It's now structured in logical chapters, and no concepts are referenced before being introduced. It should serve well both as a guide (learning Dioxus A to Z) and as a reference (checking back on things later).

I also deleted quite a bit of stuff I found irrelevant/distracting. I hope nobody minds. This included:
- In a lot of places, a concept was explained by introducing the "bad" way to solve a problem first, and then improving it. I didn't like this because it's 3x longer to read (also confusing and hight mental overhead) and doesn't really provide any value. I prefer to directly show the solution, and mention any things *not* to do as an afterthought.
- Some introductions/conclusions that didn't really contain any info. They felt like noise to me.
- I cut down on some history lessons and theory discussion to make the guide more concise and consistent. Perhaps we can later write a dedicated chapter for all the reasoning behind Dioxus and how it compares to other frameworks etc., but that's not really what most people will read the guide for.
- I did my best to move all this removed content to the __unused directory

I was basically optimizing for people that know Rust, and want to learn Dioxus as fast as possible.

btw this is the new structure

image


rMazeiks:
Basically hopefully good to go. Maybe the chapter on writing custom renderers and publishing the app haven't gotten as much love as they deserve, maybe i'll come back to them one day. Right now I'm tired of writing docs XD

jkelleyrtp:
Do you mind rebasing so the merge conflicts go away?

I'm still going through the material and will likely make some edits post merge instead of holding this PR up :)

jkelleyrtp (approved):

PR #488, opened 2022-07-05
Subtree memorization / reactive templates

Demonthos:
Allows templates to be created to skip diffing on static parts of the html. The templates can be changed when the program is running to allow for hot reloading. The dynamic parts of the templates contain a index that maps dynamic nodes/attributes in the template to the captured dynamic nodes/attributes. This allows dynamic parts to be moved for hot reloading.

The templates allow for fine-grained diffing within a node. It will only diff attributes within a node that are dynamic, and only the dynamic parts of a formatted text node.

- [x] Create DomEdits for templates
- [x] Create structure for templates
- - [x] Allow static/dynamic attributes in the same element
- - [x] Allow Dynamically created or Statically created versions of templates
- - [x] Allow skipping to any point in the template based on what part was edited
- - [x] Handle volatile values in templates
- [x] Create templates with the rsx macro
- [x] Diff templates
- [x] Handle templates in native-core
- [x] Handle templates in web interpreter
- - [x] Render static templates
- - [x] Handle updating templates
- [x] Handle templates in desktop
- [x] Use templates in hot reloading
- [x] documentation
- [x] Use keys in the template to provide a key for the template?
- [x] Integrate with ssr
- [x] Integrate with liveview

Future work:
- Collect templates at compile time in a separate html file to avoid creation cost at runtime. The calls to create the templates would be replaced with something like document.getElementById("dioxus-template-#").


jkelleyrtp:
This is awesome!

I think it might be worth writing down some of the architecture and approach if you have anything in mind.

Demonthos:
Here is what I am thinking for the architecture:

### Diouxs side:

1. A new vnode variant VNode::NodeRef that contains a reference to a template and the dynamic context to hydrate the template.
2. Each dynamic part of the template contains a index that points to the dynamic part to use so that the templates created by the hot reloading cli can move the dynamic parts around.

### Renderer side:

New DomEdits:
1. CreateTemplate{template_id}
     CreateTemplates starts the creation of a template. A template node is created through normal DomEdit messages, but must be allocated on a stack that is different from the normal vnode stack because the memory is managed differently.
2. FinishTemplate{}
     FinishTemplate finishes the template creation.
3. CreateTemplateRef{id, template_id}
     Create a reference to a template. When the reference is on the top of the stack, any ids in messages refer to ids in the template stack instead of the vnode stack

### Open questions:

1. How should the templates be stored to allow for heap allocated and &'static variants?
  a. The only solution I can think of is an enum with static and allocated variants, but it would result in a lot of duplication.
2. Is there a better way to structure the DomEdits?

Demonthos:
The web implementation is mostly complete (🎉), so I thought it might be a good time for more notes:

I forked and updated the js framework benchmark to work with this branch, and we preform better than react for all but one of the performance related benchmarks:
image

This also improves memory usage compared to the 0.2 results, but it is probably the area we most need to improve:
Screenshot 2022-08-13 130144

A very breaking change to attributes:
This PR changes attributes to be represented as an associated constant instead of a function, to allow them to be used at compile time. For the most part this shouldn't effect existing user code, but it would break custom renderers.

Demonthos:
Untitled-2022-09-24-1015

Demonthos:
Everything appears to work. I checked all the examples, and they work correctly. Ssr, tui, liveview, web, desktop, and hot-reloading all seem to work.

I don't love the way the mutations are structured right now, but they work. If there is any better ideas, I'm open to suggestions. I think shifting to a mutations trait instead of struct might be nice so that different platforms can store them differently. Eg: The ffi friendly mutations could be stored differently form the native-core mutations meant for use in rust, which is stored differently from the binary encoding used for web, etc.

Here is the final performance results:

Screenshot 2022-09-24 225434

jkelleyrtp (approved):
Looks good! There are quite a bit of warnings by clippy and some code that does strange things (like dropping references). 

Try enabling clippy on save to see some of what I'm talking about.

I'm going to follow this up with some PRs to fix some of the warnings which will be a bit more helpful as review material.

Otherwise, looks great. Might even be worth submitting the new benchmarks to js-framework-benchmark.

trevyn:
@Demonthos Note that `cargo install --git https://github.com/DioxusLabs/cli` (the only current recommended way to install the CLI) is now broken with:

```
  no matching package named `dioxus-rsx-interpreter` found
```

PR #456, opened 2022-06-24
Add PT-BR translations

amindWalker:
<div align='center'>

# Dioxus Translated Documentation

</div>

Hi there, I'm building a Rust Knowledge Base Center and I'm using **Rust** for both **backend** and **frontend**.

I was using _Yew.rs_ before I discovered about **Dioxus** 2 days ago while I was googling Yew's documentation... and I have to say, that was the best find of the day! Dioxus is so much more elegant than Yew imo. **I'll build my frontend with Dioxus now!**

I got really excited and I wanna help to spread the word and the fastest way for me right now is by translating (PT-BR in this case) all the documentation. I wish I could help with code but you know... I'm still learning Dioxus😅. But later, after enough hacking with Dioxus, I'm planning to help with examples and with code as well.

I don't know which Markdown files are used for the mdbooks so I decided to translate every one of them with relevant content in it.

I've forked the project to `diff` with my translated files. For every translated file I appended **_"\_pt-br.md"_** for easy tracking.

_**I've made a progress table summary:**_

# Repo Description README

| Original  | Brazilian Portuguese | Translated |
| :-------: | :------------------: | :--------: |
| README.md |   README_pt-br.md    |     ✔      |

# Guide

|    Original    | Brazilian Portuguese | Translated |
| :------------: | :------------------: | :--------: |
|    final.md    |    final_pt-br.md    |     ✔      |
| hello_world.md | hello_world_pt-br.md |     ✔      |
|   README.md    |   README_pt-br.md    |     ✔      |
|   ROADMAP.md   |   ROADMAP_pt-br.md   |     ✔      |
|    setup.md    |    setup_pt-br.md    |     ✔      |
|   SUMMARY.md   |   SUMMARY_pt-br.md   |     ✔      |

<details>
<summary>Guide Section</summary>
<ul>

## Advanced Guides

|        Original        |     Brazilian Portuguese     | Translated |
| :--------------------: | :--------------------------: | :--------: |
|      00-index.md       |      00-index_pt-br.md       |     ✔      |
| 06-subscription-api.md | 06-subscription-api_pt-br.md |     ✔      |
| 10-concurrent-mode.md  | 10-concurrent-mode_pt-br.md  |     ✔      |
|    11-arena-memo.md    |    11-arena-memo_pt-br.md    |     ✔      |
|  12-signals copy.md'   |        12-signals.md         |     ✔      |
|     13-subtrees.md     |     13-subtrees_pt-br.md     |     ✔      |
[...]

PR #494, opened 2022-07-09
Internationalization: move guide to en folder and preserve its history.

jkelleyrtp:
This PR extracts the movement of the guide into an "en" folder, preserving its history so #456 can be executed in a non-lossy way.


jkelleyrtp:
TODO:

Use a flavor of github actions that supports building the guide with the forked mdbook branch
```
        # install mdbook --git https://github.com/Ruin0x11/mdBook.git --branch localization --rev e74fdb10725d30917ce1b6b39c968730133ba7ee --force
```

PR #504, opened 2022-07-22
allow the last attribute in a component to contain formatting

### LINKED ISSUE #502: How to build Redirect routing path dynamically?
When building Redirect, the static path is ok:
```
cx.render(rsx! {
    Redirect { to: "/abc" }
})
```

But when joining pieces of String to create dynamic path, like:
```
let id = "id1";
let url = format!("/abc:{}", id).as_str();
cx.render(rsx! {
    Redirect { to: url }
})
```
It will alert the following error:
```
cannot return value referencing temporary value
returns a value referencing data owned by the current functionrustc[E0515](https://doc.rust-lang.org/error-index.html#E0515)
```

How to create a dynamic routing path?
Thanks.

Demonthos:
fixes #502 
The current method of parsing components checks if the attribute has a trailing comma before checking for formatting. This PR allows the last attribute to not have a trailing comma, but still contain formatted segments.

### DIFF
--- a/packages/rsx/src/component.rs
+++ b/packages/rsx/src/component.rs
@@ -234,13 +234,15 @@ impl Parse for ComponentField {
             let content = ContentField::ManExpr(input.parse()?);
             return Ok(Self { name, content });
         }
-
-        if input.peek(LitStr) && input.peek2(Token![,]) {
-            let t: LitStr = input.fork().parse()?;
-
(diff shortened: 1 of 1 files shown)


mrxiaozhuox (approved):

jkelleyrtp:
Thank you!

PR #510, opened 2022-07-28
Added initial_url to the router component

norman784:
It seems that there was no way, nor workaround to set an active url when doing ssr, so I added a new prop to the router component.

### DIFF
--- a/packages/router/src/components/router.rs
+++ b/packages/router/src/components/router.rs
@@ -29,8 +29,8 @@ pub struct RouterProps<'a> {
     /// By default set to `"active"`.
     pub active_class: Option<&'a str>,
 
-    /// Set the active url.
-    pub active_url: Option<String>,
+    /// Set the initial url.
+    pub initial_url: Option<String>,
 }
 
 /// A component that conditionally renders children based on the current location of the app.
@@ -46,7 +46,7 @@ pub fn Router<'a>(cx: Scope<'a, RouterProps<'a>>) -> Element {
             RouterCfg {
                 base_url: cx.props.base_url.map(|s| s.to_string()),
                 active_class: cx.props.active_class.map(|s| s.to_string()),
-                active_url: cx.props.active_url.clone(),
+                initial_url: cx.props.initial_url.clone(),
             },
         ))
     });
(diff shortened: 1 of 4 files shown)


packages/router/src/components/router.rs
```diff
@@ -28,6 +28,9 @@ pub struct RouterProps<'a> {
     /// This is useful if you don't want to repeat the same `active_class` prop value in every Link.
     /// By default set to `"active"`.
     pub active_class: Option<&'a str>,
+
+    /// Set the active url.
```
rMazeiks, line 32: A bit more detail in the docs would be nice here...

Is this the initial URL? If so, `initial_url` might be a better name. As well as a note on what the default value would be.

Or is it a fixed URL that, when specified, cannot be changed with the push_route functions?

rMazeiks (approved):
Looks mostly good to me! Though it would be good if someone with more knowledge of the Router could take a look at this.

Thanks!

packages/router/src/components/router.rs
norman784: I'll update the docs and yes, `initial_url` is a better name.

This is intended to specify the initial url when working specifically with `ssr`, but I'm sure it will be helpful also for mobile and desktop, but I didn't tested those scenarios.

rMazeiks (approved):

norman784:
@mrxiaozhuox are you available to review my PR?

rMazeiks:
Note: there's a new, very big PR #522, a complete re-write of the router. I really like that PR, but unfortunately, merging it would make this one obsolete. Sorry!

(Let's see what the maintainers think of it – it's possible they won't want to make such breaking changes, and the new router would have to become a separate package – then, this PR is unaffected)

Not sure if the new router has something equivalent toe `initial_url`. If not, this code will probably be hard to integrate in it, and most likely will need to be re-written.

norman784:
I though his work was stale, he just submitted yesterday his PR, I already checked it a while ago and his way on setting the initial url was more involved, but let's wait till his PR gets merge then.

jkelleyrtp (approved):

jkelleyrtp:
Thanks!

I'm going to work on helping #522 get in but that'll take a bit longer than getting this one in now.

PR #515, opened 2022-07-30
allow duplicate variables in format_args_f

Demonthos:
currently this fails to compile:
```rust
let x = 0;
rsx!{
  "{x} {x}"
}
```
This is because the format_args_f macro transforms it into
...
format_args!("{x} {x}", x = x x = x)
...
which contains duplicates that are not allowed in format_args.
This PR deduplicates the assignments.


mrxiaozhuox (approved):

mrxiaozhuox:
cool!

jkelleyrtp:
Thanks!

PR #516, opened 2022-07-31
Update Docs CI with mdbook translation support 

### LINKED ISSUE #514: Update Docs CI with mdbook translation support
Deploying docs with CI is broken. We are using https://github.com/rust-lang/mdBook/pull/1306 to support translations, but the action we use does not support this.

**Questionnaire**

ghostnumber7:
Fixes #514

mdbook is build from source until the feature is released, then we can go back to previous action.
Example paths where updated to point to the right place also :)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -24,7 +24,7 @@ jobs:
       # NOTE: Delete when the previous one is enabled
       - name: Setup mdBook
         run: |
-          cargo install --git https://github.com/Ruin0x11/mdBook.git --branch localization --locked mdbook
+          cargo install mdbook --git https://github.com/Ruin0x11/mdBook.git --branch localization --rev e74fdb1
 
       - name: Build
         run: cd docs &&
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,10 +15,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      # NOTE: Comment out when https://github.com/rust-lang/mdBook/pull/1306 is merged and released
+      # - name: Setup mdBook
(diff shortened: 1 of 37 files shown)


.github/workflows/docs.yml
```diff
@@ -15,10 +15,16 @@ jobs:
...
+      #   with:
+      #     mdbook-version: "0.4.10"
+
+      # NOTE: Delete when the previous one is enabled
       - name: Setup mdBook
-        uses: peaceiris/actions-mdbook@v1
-        with:
-          mdbook-version: "0.4.10"
+        run: |
+          cargo install --git https://github.com/Ruin0x11/mdBook.git --branch localization --locked mdbook
```
Demonthos, line 27: If I understand correctly, I don't think --locked does anything here because the build data isn't persistent, so the cargo.lock is regenerated every time. We could use the [--rev](https://doc.rust-lang.org/cargo/commands/cargo-install.html#install-options) flag to lock to a specific commit.

Demonthos (commented):
There were some [concerns](https://github.com/rust-lang/mdBook/pull/1306#issuecomment-1179580044) about this branch being behind the master branch, but it is ahead of the branch we currently use, so this should be good. Once the download is locked to a commit.

.github/workflows/docs.yml
ghostnumber7: I added the `locked` part because it was complaining that the repo has 2 binaries and did not know which one to install. `locked` solved that, and also tried with `-p` but it's not supported by `cargo install` it seams. Maybe there is another way more correct to achieve the same, but don't know about it

ghostnumber7: Nevermind, saw my mistake and added the `--rev` part

Demonthos (approved):
Looks good, Thank you!

PR #528, opened 2022-08-13
Improve documentation for the native core and native core macro crates

Demonthos:
Adds more internal comments and documentation to the native core and native core macro crates. It also renames some private functions/types to improve readability.

### DIFF
--- a/packages/native-core-macro/src/lib.rs
+++ b/packages/native-core-macro/src/lib.rs
@@ -14,6 +14,7 @@ use syn::{
     parse_macro_input, parse_quote, Error, Field, Ident, Token, Type,
 };
 
+/// Sorts a slice of string literals at compile time.
 #[proc_macro]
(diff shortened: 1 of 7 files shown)


packages/native-core-macro/src/lib.rs
```diff
@@ -22,12 +23,107 @@ pub fn sorted_str_slice(input: TokenStream) -> TokenStream {
 }
 
 #[derive(PartialEq, Debug, Clone)]
-enum DepKind {
+enum DependancyKind {
```
rMazeiks, line 26: typically spelled "dependency"

"dependancy" looks real weird to me, but is apparently a valid alternative spelling...

Demonthos: Oops, good catch

jkelleyrtp:
Awesome! Thank you!

Issue #512, opened 2022-07-29
DesktopConfig.with_event_handler is unused


ColonelThirtyTwo:
`DesktopConfig.with_event_handler` sets the `event_handler` field, but that isn't used anywhere:

```
> grep -r 'event_handler' packages/desktop/
packages/desktop/src/cfg.rs:    pub(crate) event_handler: Option<Box<DynEventHandlerFn>>,
packages/desktop/src/cfg.rs:            event_handler: None,
packages/desktop/src/cfg.rs:    pub fn with_event_handler(
packages/desktop/src/cfg.rs:        self.event_handler = Some(Box::new(handler));
```

Would be nice to install an application-wide keypress handler.

rMazeiks:
If I understand correctly, it's for window events such as "minimize" or "close", etc. I think it should just be removed until such functionality is implemented.

> Would be nice to install an application-wide keypress handler.

You can probably add this on some wrapper for the whole document if needed.

ColonelThirtyTwo:
Well, when I tried to use `onkeydown` on the top level `rsx` macro, it complained about an unexpected token.

rMazeiks:
wrap it in a `div` or similar. only elements can have attributes.

ColonelThirtyTwo:
Will that work even if the window is focused by, say, clicking on the titlebar?

PR #529, opened 2022-08-13
expose DesktopConfig in public interface

### LINKED ISSUE #511: DesktopConfig should be exported
The `DesktopConfig` structure is publicly usable via the `dioxus_desktop::launch_cfg` function, but is not exported, so no documentation is generated for it.

Demonthos:
fixes #511 and #512
It also comments out with_event_handler because it is currently not implemented.

### DIFF
--- a/packages/desktop/src/cfg.rs
+++ b/packages/desktop/src/cfg.rs
@@ -2,23 +2,21 @@ use std::path::PathBuf;
 
 use wry::application::window::Icon;
 use wry::{
-    application::{
-        event_loop::EventLoop,
-        window::{Window, WindowBuilder},
-    },
+    application::window::{Window, WindowBuilder},
     http::{Request as HttpRequest, Response as HttpResponse},
-    webview::{FileDropEvent, WebView},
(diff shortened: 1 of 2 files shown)


rMazeiks (approved):

jkelleyrtp (approved):

John0x:
@Demonthos This is not currently accessible from external crates. Would be mir idiomatic to be able to use it outside the launch_cfg call

Demonthos:
> @Demonthos This is not currently accessible from external crates. Would be mir idiomatic to be able to use it outside the launch_cfg call

I might have missed something here but I think this is exposed on the master version of Dioxus.
See: https://github.com/jkelleyrtp/blazemail/blob/master/src/main.rs#L20 for an example

These changes have not been published yet. The 0.2.4 release does not expose DesktopConfig to external crates so you may be running into issuse with that release.

John0x:
@Demonthos ah, yeah, that could be the case. Sorry, didn't notice that it wasn't released yet. I'll try the master branch later :)

PR #533, opened 2022-08-22
Added dialog tag

norman784:


packages/html/src/elements.rs
```diff
@@ -1006,7 +1009,12 @@ builder_constructors! {
         open: Bool,
     };
 
-
+    /// Build dialog
+    /// [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)
+    /// element.
+    dialog {
+        open: bool,
```
rMazeiks, line 1016: The other attributes in this module seem to use a `Bool` struct, not the `bool` primitive... I don't know why that is, but I think we should keep it consistent.

rMazeiks (approved):

Demonthos (approved):
Thank you!

PR #542, opened 2022-09-04
fix: set false for media event bubbling

### LINKED ISSUE #541: Audio tag events do not fire
**Problem**

`onended`, `ontimeupdate`, `onloadedmetadata` `<audio />` events never fire.

**Steps To Reproduce**

See [main.rs](https://github.com/erichulburd/dioxus-audio-tag-event-bugs/blob/master/src/main.rs) in this simplified app that reproduces the bug: https://github.com/erichulburd/dioxus-audio-tag-event-bugs.

<details>
<summary>tldr</summary>

```rs
    let mut metadata = use_state(&cx, || 0);
    let mut timeupdate = use_state(&cx, || 0);
    let mut ended = use_state(&cx, || 0);

    cx.render(rsx!(
        h1 { "Audio Event Bugs" }
        table {
            thead {
                tr {
                    th { "Event" }
                    th { "Count" }
                }
            }
            tbody {
                tr {
                    td {
                        "onloadedmetadata"
                    }
                    td { "{metadata}" }
                }
                tr {
                    td {
                        "ontimeupdate"
                    }
                    td { "{timeupdate}" }
                }
                tr {
                    td {
                        "onended"
                    }
                    td { "{ended}" }
                }
            }
        }
        audio {
            src: "/static/sample.mp3",
            preload: "metadata",
            controls: "true",
            onloadedmetadata: move |_| metadata += 1,
            ontimeupdate: move |_| timeupdate += 1,
            onended: move |_| ended += 1,
        }
    ))
}
```

</details>

**Expected behavior**

When the audio tag loads metadata, updates time, or ends, we want the events to fire. In the sample app, this would show up in the HTML table that contains counts of each event.

**Screenshots**

-

**Environment:**
 - Dioxus version: 0.2.4
 - Rust version:

```shell
% rustc --version
rustc 1.62.1 (e092d0b6b 2022-07-16)
% cargo --version
cargo 1.62.1 (a748cf5a3 2022-06-08)
% rustup --version
rustup 1.25.1 (bb60b1e89 2022-07-12)
```

 - OS info: Apple M1 Max, macOS Monterey v12.3
 - App platform: `web`

**Questionnaire**

erichulburd:
Closes #541 

This simply sets media events false in `pub fn event_bubbles`. I've confirmed this works in https://github.com/erichulburd/dioxus-audio-tag-event-bugs/tree/with_applied_fix

However, I'm unsure about your integration test setup here. How and where should I add tests?

Also, I was unable to find a definitive source of truth for event bubbling. MDN marks _some_ events as bubbling (eg [canplay](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canplay_event)), but not others (eg [onloadedmetadata](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadedmetadata_event)).

I would assume there is some specification out there that we can reference in a docstring, but I was unable to find it.


rMazeiks (commented):
Rust part looks good to me! (Didn't check docs for each event individually, but I'd be surprised if any of these bubbled.

What about [interpreter.js](https://github.com/DioxusLabs/dioxus/blob/master/packages/interpreter/src/interpreter.js#L664-L831)? Not 100% sure how it's used but #452 introduced the mapping there too. In any case, would be good to make sure this works on all platforms (web & desktop)

Demonthos:
The source of truth I used for which events bubble is [the dom events wikipedia article](https://en.m.wikipedia.org/wiki/DOM_events) (it's not the most reliable source, but it had the most complete data I could find in a easy to read format).

We don't currently have any scaffolding for browser based testing, but I agree it would be useful to add.

Demonthos (approved):
Thanks! I'm going to look into some automated browser testing for this part of the code in the next few days.

PR #543, opened 2022-09-05
feat: Support Bytes type in AttributeValue

marc2332:
I am not yet familiarized with the core codebase so I am not sure why this is needed, but looks like I need it, otherwise it would just panic.

Context: I am working on my own renderer and I am implementing an image element with an attribute that receives the content of the image in form of bytes.

I tried it with these changes and it worked :)

### DIFF
--- a/packages/core/src/arbitrary_value.rs
+++ b/packages/core/src/arbitrary_value.rs
@@ -65,7 +65,7 @@ impl<'a> std::fmt::Display for AttributeValue<'a> {
             AttributeValue::Vec4Float(_, _, _, _) => todo!(),
             AttributeValue::Vec4Int(_, _, _, _) => todo!(),
             AttributeValue::Vec4Uint(_, _, _, _) => todo!(),
-            AttributeValue::Bytes(_) => todo!(),
+            AttributeValue::Bytes(a) => write!(f, "{:?}", a),
             AttributeValue::Any(_) => todo!(),
         }
     }


packages/core/src/arbitrary_value.rs
```diff
@@ -65,7 +65,7 @@ impl<'a> std::fmt::Display for AttributeValue<'a> {
             AttributeValue::Vec4Float(_, _, _, _) => todo!(),
             AttributeValue::Vec4Int(_, _, _, _) => todo!(),
             AttributeValue::Vec4Uint(_, _, _, _) => todo!(),
-            AttributeValue::Bytes(_) => todo!(),
+            AttributeValue::Bytes(a) => write!(f, "{:?}", a),
```
rMazeiks, line 68: `:?` is for Debug formatting, which is usually only used for developer-facing messages, traces, etc. I think it would be more correct to just use Display formatting as above, if possible.

rMazeiks (commented):
(This comment is not about the PR itself but)

> I am implementing an image element with an attribute that receives the content of the image in form of bytes.

Does this mean that these bytes are diffed every time the component is re-rendered? Not 100% sure this is great for performance - but then, I'm no expert either.

packages/core/src/arbitrary_value.rs
marc2332: I thought the same but It didn't let me, I'll take a look again later

marc2332:
> (This comment is not about the PR itself but)
> 
> > I am implementing an image element with an attribute that receives the content of the image in form of bytes.
> 
> Does this mean that these bytes are diffed every time the component is re-rendered? Not 100% sure this is great for performance - but then, I'm no expert either.

I think so, perhaps it's better to ignore the content in the diffing? it wouldn't be possible to dynamically change the image content tho

Demonthos:
> > (This comment is not about the PR itself but)
> > > I am implementing an image element with an attribute that receives the content of the image in form of bytes.
> > 
> > 
> > Does this mean that these bytes are diffed every time the component is re-rendered? Not 100% sure this is great for performance - but then, I'm no expert either.
> 
> I think so, perhaps it's better to ignore the content in the diffing? it wouldn't be possible to dynamically change the image content tho

You could create a wrapper struct that tracks when it has been changed (via Deref/DerefMut) before doing the full PartialEq check, or use the use_ref hook which does something similar automatically.

packages/core/src/arbitrary_value.rs
marc2332: I couldn't find an alternative way. Any suggestions? This works but I agree that using the `Debug` trait for this is weird.

rMazeiks: You're right, I've given this some more thought.

I suspect the issue here is that we shouldn't be implementing `Display` in the first place, since that's not what it's used for. I dug a bit through the codebase, and the only usage I could find was rendering elements to HTML in `TextRenderer::html_render`. It seems to me like the best way here would have been to serialize the HTML with Serde.

Anyway, using Debug here as a hack is fine with me now. I don't really like this code but that's not your fault :) and I have no better suggestions.

rMazeiks (approved):

Demonthos (approved):
This works for now. For a u8 array the output of serde_json, and debug are the same, but I agree with @rMazeiks that we should switch to using the serde derive for ssr in the future.

jkelleyrtp:
I think for some of these types, we will want to defer to any "Any" type that can be downcasted and serialized. Supporting these types natively would be nice - and might be required for high performance stuff - but it will bloat our codebase and make it difficult to work with.

PR #547, opened 2022-09-08
Feat: file drag and drop working properly on desktop

jkelleyrtp:
This PR gets rid of the custom file drag and drop handler from config in favor of a new DragEvent that returns a true list of Files.

There isn't a good cross-platform approach here unfortunately, so we'll have to settle for some wrapper structs with different methods depending on the platform.

https://user-images.githubusercontent.com/10237910/189492691-904e2d76-f3a2-475c-a204-0b1bf3c15684.mov


jkelleyrtp:
Too many conflicts with this, following it up with something new.

PR #548, opened 2022-09-09
feat: use infer instead of outdated mime-guess, bump wry

jkelleyrtp:
Infer is more robust that mime-guess for loading assets from the file system.

### DIFF
--- a/packages/desktop/src/protocol.rs
+++ b/packages/desktop/src/protocol.rs
@@ -68,13 +68,9 @@ pub(super) fn desktop_handler(
                 .body(String::from("Not Found").into_bytes());
         }
 
-        let mime = mime_guess::from_path(&asset).first_or_octet_stream();
-
-        // do not let path searching to go two layers beyond the caller level
-        let data = std::fs::read(asset)?;
-        let meta = format!("{}", mime);
-
-        ResponseBuilder::new().mimetype(&meta).body(data)
+        ResponseBuilder::new()
+            .mimetype(get_mime_from_path(trimmed)?)
+            .body(std::fs::read(asset)?)
     }
 }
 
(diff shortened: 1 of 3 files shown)


packages/desktop/src/protocol.rs
```diff
@@ -110,3 +106,42 @@ fn get_asset_root() -> Option<PathBuf> {
...
+        Some(t) if t == "text/plain" => get_mime_by_ext(trimmed),
+        Some(f) => f,
+        None => get_mime_by_ext(trimmed),
+    };
+
+    Ok(res)
+}
+
+/// Get the mime type from a URI using its extension
+fn get_mime_by_ext(trimmed: &str) -> &str {
```
Demonthos, line 126: Switching to infer for more accurate filetypes makes sense, but I wonder if keeping mime_guess as a backup instead of our own implementation might be better here. It supports [a lot more file types](https://github.com/abonander/mime_guess/blob/master/src/mime_types.rs) than infer, so it might be good to have as a backup.

packages/desktop/src/protocol.rs
```diff
@@ -110,3 +106,42 @@ fn get_asset_root() -> Option<PathBuf> {
...
 }
+
+/// Get the mime type from a path-like string
+fn get_mime_from_path(trimmed: &str) -> Result<&str> {
+    if trimmed.ends_with(".svg") {
+        return Ok("image/svg+xml");
+    }
+
+    let res = match infer::get_from_path(&trimmed)?.map(|f| f.mime_type()) {
+        Some(t) if t == "text/plain" => get_mime_by_ext(trimmed),
```
Demonthos, line 117: Nitpick, but we could use 
```Some("text/plain") => ```
directly here instead of a if guard

Demonthos (approved):
Looks mostly good. A few small comments, not sure if they are worth changing just stuff to think about.

Issue #530, opened 2022-08-13
Support optionally adding an attribute


Demonthos:
## Specific Demand

There is currently no way to optionally add a attribute within the rsx macro. This is needed to complete #394.

## Implement Suggestion

- React only adds the property if it is truthy, we could mimic it, but it doesn't seem very "rusty"...
- We could only add a Option<AttributeValue> when attribute if the option is Some:
```rust
let x = Some("myClass");
rsx!{
  div{
    class: x,
  }
}
```
- We could have a opt in way to pass a Option<AttributeValue> instead that only adds the attribute if the option is Some:
```rust
let x = Some("myClass");
rsx!{
  div{
    class?: x,
  }
}
```

jkelleyrtp:
I like the syntax you proposed here!

TeFiLeDo:
Dioxus could provide a trait for merging multiple occurrences of a single attribute. This trait would only be needed when multiple occurrences are actually detected.

Also, IIRC, macros have a way of requiring trait implementations for specific types. This maybe could be used to provide more helpful error messages, when the merging trait is required.

The merging trait would also need to be implemented for various `std` types, like `Vec` and `String`.

PR #549, opened 2022-09-11
support optionally adding attributes

Demonthos:
- [ ] implement in rsx macro
- [ ] update docs

#530

### DIFF
--- a/packages/autofmt/src/buffer.rs
+++ b/packages/autofmt/src/buffer.rs
@@ -143,8 +143,8 @@ impl Buffer {
         let mut total = 0;
 
         for attr in attributes {
-            if self.current_span_is_primary(attr.attr.flart()) {
-                'line: for line in self.src[..attr.attr.flart().start().line - 1].iter().rev() {
+            if self.current_span_is_primary(attr.attr.start()) {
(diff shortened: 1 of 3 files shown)


Demonthos:
Thinking about this some more...
It might be better to allow creating attributes from a impl IntoIterator<(Attribute, Value)>. This would allow optionally adding props and make it more ergonomic to dynamically create many different attributes. You could have a single &'static [(Attribute, Value)] array and insert it into a element to get a style. Or use a Vec<(Attribute, Value)> to more easily integrate with a library that generates html.

jkelleyrtp:
With the changes in #600 we can do one of two things
- allow all Option<S> in attribute positions without the custom syntax
- Use the question mark syntax to use the same code pathways as #1 but be explicit about it

I kinda think #1 is better than #2 since the parsing is more consistent. We have a new attribute type called "None" - and could even have a list of attributes as you mentioned above. I'm not sure if `None` is wired up to properly remove the attribute from the dom, so if we revisit this PR, then we might need to add that.

Demonthos:
Closing this in favor of a new PR after the rewrite of core.

Option 1 is implemented in #708

Issue #475, opened 2022-06-30
Add "why" a component re-rendered.


jkelleyrtp:
## Specific Demand

React has a monkey patch called why-did-you-render. It's mostly used for debugging, but in theory, it could be used to write more efficient hooks.

https://github.com/welldone-software/why-did-you-render

For example, a component could know that it's being re-rendered because it's actively being mounted, it was forced to by a hook, or a parent was diffed causing a diff of the current component.

## Implement Suggestion

Add a new method to "Scope" that includes reasoning of why a component is being re-rendered.

An enum along the lines of:

```rust
enum WhyDidYouRender {
    // A distant parent is causing a partial reflow of the tree
    Parent { originator: ScopeId },

    // The component is being mounted for the first time
    Mount,

    // A hook or external force directly called this component to be re-rendered
    Explicit
}
```

PR #554, opened 2022-09-13
feat: implement why-did-you-render

jkelleyrtp:
Solves #475 

Needs more tests, but seemingly works.


Exotik850:
Is there a reason this was closed? This would be really useful imo

tkr-sh:
up

PR #564, opened 2022-09-25
feat: simple iterators and conditionals in rsx

jkelleyrtp:
This adds the ability to use very simple iterators and if chains inside of rsx. 

## Iterators
The rule is simple:

whenever a `for` is encountered as the leading token in a BodyNode of rsx!, we parse it as if it were actually an into_iter expression, performing similar desugaring as rustc. This is inline with some other frameworks, but makes working with lists simpler.

The body of the for loop is treated as closure expression, so it must be explicit rather than implicitly RSX.

```rust
    cx.render(rsx!(
        for id in 0..10 {
            rsx! { "{id}" }
        }
    ))
}
```
## If chains 
Generally, `if` chains should be valid today, but they require a weird additional some/none. 

In this PR, we add some sugaring when there is no termination to the if chain, autowrapping as None/Some.

```rust
    cx.render(rsx!(
        if enabled {
            rsx!{ "it's enabled" }
        }
    ))
}
```


jkelleyrtp:
Closing to be wrapped into #600

PR #565, opened 2022-09-25
Feat: Make AttributeValue::Any usable

marc2332:
## Changes:
- Implemented `Display` for `AttributeValue::Any`
- Made `ArbitraryAttributeValue` public outside the `core` crate

Note about the first change: 
Just like https://github.com/DioxusLabs/dioxus/pull/543, I went using `{:?}`, but if you think necessary it can be changed to simply:

```rust
AttributeValue::Any(_) => write!(f, ""),
```

This way it won't actually write anything in the buffer... but that's okay? It works fine both ways, I still don't know why `AttributeValue` needs to implement `Display`.

Thoughts?


jkelleyrtp:
Thanks!!

PR #576, opened 2022-10-01
fix: Remove a Node listeners when it is removed

marc2332:
This will clean up all the listeners of a Node when it's removed.

I noticed while using `get_listening_sorted()` it would sometimes panic because it was unable to find a removed Node that previously had a listener. With this fix, the listeners and the node list should be on sync, so whenever you try to retrieve all the node listeners it should only get Nodes that actually exist.

Note: I haven't actually tested these changes with the new changes from https://github.com/DioxusLabs/dioxus/commit/047ed1e5536d266785240fb3db475ac2fc77b785


marc2332:
Btw, if this actually gets accepted I would appreciate if this PR was labeled with `hacktoberfest-accepted`, this will make the PR count for my [Hacktoberfest](https://hacktoberfest.com/participation/) participation. I would try to make more meaningful contributions in the future anyway

Demonthos (approved):
Looks good. Thanks!

PR #581, opened 2022-10-07
feature: allow futures and streams to be passed directly into rsx

jkelleyrtp:
This PR essentially implements suspense and partial rendering :)

The best improvement we get with this is support in SSR for knowing when an app has finished loading data before passing out the edits.

It also means the VirtualDom can implement streaming edits - a very significant feature for React.

However, we lack a proper React-equivalent scheduler with swim lanes (priority lanes), so there's no concept of "hot and cold" DOM zones.


jkelleyrtp:
Part of #600

PR #582, opened 2022-10-07
optimizations for wasm size

Demonthos:
Moves logging into a feature flag, and removes a few dependencies that only have one usage.
For the js-framework benchmark these changes save the zipped wasm 6 kb (off of 111 kb).


tamasfe:
Is there a specific reason why a new feature is made rather than using the `log` crate's [feature flags](https://docs.rs/log/latest/log/#compile-time-filters)? The users of the library can already disable logs via those.

tamasfe:
I understand that, however there is a de facto standard way of enabling/disabling per-library logging via filters (e.g. in [env_logger](https://docs.rs/env_logger/latest/env_logger/#enabling-logging)).

If there's an issue, `trace` level logs can be enabled downstream for all dependencies without looking up specific feature flags for each library. Likewise, log levels can be disabled at compilation time for all libraries at the same time via the `log` (or `tracing`) crate's flags.

Demonthos:
> I understand that, however there is a de facto standard way of enabling/disabling per-library logging via filters (e.g. in [env_logger](https://docs.rs/env_logger/latest/env_logger/#enabling-logging)).
> 
> If there's an issue, `trace` level logs can be enabled downstream for all dependencies without looking up specific feature flags for each library. Likewise, log levels can be disabled at compilation time for all libraries at the same time via the `log` (or `tracing`) crate's flags.

👍 I was not aware of that, but that makes sense. Fixed

jkelleyrtp:
Thanks!

PR #584, opened 2022-10-08
to_owned!: allow optional comma at the end

saicu:
Allow an optional comma at the end of to_owned!
```
#[test]
fn it_works() {
    let a = 1;
    let b = true;
    to_owned![a, b,];
}
```
I would prefer it this way. Let me know what you think.


jkelleyrtp (approved):
Thank you!

PR #587, opened 2022-10-12
Byte optimized interpreter

Demonthos:
Improves wasm-js boundary by handwriting implementation instead of relying on wasm-bindgen

- [x] make DomEdit a trait to avoid excessive copying
- [x] all edits are sent in one slice of bytes instead of calling individual functions
- [x] It dynamically resizes the size of the id as more number are sent so ids start with a cost of 1 byte and grow up to a cost of 8 bytes
- [x] Instead of sending a &[u8] to js every time, it assigns a global variable to the pointer and length of the slice and tells the interpreter where those globals are stored in memory once at the beginning (so when handle edits is called nothing is passed between memory).
~~It uses a utf8 decoder that is written in js to instead of the one that is built in to avoid native-js interop overhead. (wasm-bindgen uses the built in version).~~
- [x] Pack strings into separate segment, decode them in one pass, and then use substring to separate them (https://jsbench.me/4vl97c05lb/4)
- Platforms
- [x] Web
- [ ] Desktop
- [ ] Liveview

image


Demonthos:
I'm separating this into a [library](https://github.com/demonthos/sledgehammer/). Once the library is more stable, I will open another PR that integrates with it.

PR #588, opened 2022-10-12
replace fxhash crate with rustc-hash

saicu:
The rust performance book recommends using rustc-hash instead of fxhash for maintenance reasons - https://nnethercote.github.io/perf-book/hashing.html

"[rustc-hash](https://crates.io/crates/rustc-hash) provides FxHashSet and FxHashMap types that are drop-in replacements for HashSet and HashMap. Its hashing algorithm is low-quality but very fast, especially for integer keys, and has been found to out-perform all other hash algorithms within rustc. ([fxhash](https://crates.io/crates/fxhash) is an older, less well maintained implementation of the same algorithm and types.)"

Do you see any reasons against this? Let me know what you think!

### DIFF
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -18,7 +18,7 @@ keywords = ["dom", "ui", "gui", "react", "wasm"]
 bumpalo = { version = "3.6", features = ["collections", "boxed"] }
 
 # faster hashmaps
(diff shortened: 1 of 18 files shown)


jkelleyrtp:
This looks good to me :)

jkelleyrtp (approved):

PR #600, opened 2022-10-22
Template architecture, async components, inline iterators, error boundaries, multiple renderers

jkelleyrtp:
This PR is doing some major overhauls to the Dioxus core engine, adding a bunch of new features and generally making Dioxus faster.

- The architecture of Dioxus is being re-centered around the template work #488 
- Components can now be async and will block the rendering of their nearest suspense boundary until ready
- The virtualdom can now order changes for different renderers in the same stream, unlocking multiwindow desktop apps
- unterminated if/then and for loops with rsx bodies (experimental, may revert this)
- internal tasks and futures are way more efficient, using their own dedicated waker 
- SSR now has a cache for all static content it's seen before, making it much faster
- hooks don't need to create a reference. IE `use_state(cx, ` works automatically now

Todo
- [x] SSR Cache
- [ ] Finish cleaning up SSR crate and overhaul its docs and APIs
- [x] Rip out old vnode in favor of template-only vnode
- [x] implement `create` method with templates
- [x] implement async components
- [x] implement suspense boundaries
- [x] a dedicated waker for futures inside the dom that doesn't require Rc
- [x] Sanitize method on mutations to wipe template IDs for testing
- [x] fix pointer stability issues with new bumpslab crate
- [x] get diffing working
- - [x] Basic template <-> template
- - [x] Old template to new template light internal diffing
- - [ ] handle async rollover
- - [ ] handle error boundary rollover
- [x] rework garbage collection
- [ ] wire up subtrees into resume logic
- [x] Element is no longer Option<T> but anyhow::Result<T>
- [ ] Error boundaries and render cancelling

Tests
- [x] We don't hand out null ptrs
- [x] creating text
- [x] creating elements
- [x] creating fragments
- [x] Creating placeholders
- [x] Tasks get polled
- [x] Components pass through their children
- [x] creating suspense
- [ ] Errors get bubbled to a suspense boundary
- [ ] Errored components cancel their render
- [x] Diffing templates with just components
- [x] Dynamic-only templates don't generate CreateTemplate mutations
- [x] Diffing templates with the same ID
- [x] Diffing templates with different IDs should light diff and then replace
- [x] Mounting dynamic roots should not use LoadTemplate
- [x] Unkeyed diffing ordering
- [x] Keyed diffing ordering
- [x] No leaking when the virtualdom is dropped
- [x] Drop safety before running listeners
- [x] Suspended components should not respond to user events
- [ ] Rolling from nodes to async nodes should not wipe the old tree - render gen doesn't advance

Future todo
- rework some docs
- blog posts on the changes
- transition components
- Default error screen? Currently we just panic if an error gets bubbled to the root

Notes
----
Apparently "clear all children" is not fast - clearing each node one-by-one is!


jkelleyrtp:
General status
- core has been rewritten to support the new async features
- still finishing up diff with templates
- still working on events
- instead of polling every task, we only poll tasks that have woken up via their waker
- renderers have yet to be updated with the new system
- SSR has been updated with the new system
- use_future now implements IntoFuture and changes its behavior to no longer re-render the component when it complete unless .value() is called

packages/interpreter/src/interpreter.js
```diff
@@ -262,87 +219,115 @@ export class Interpreter {
...
-        break;
       case "AppendChildren":
-        this.AppendChildren(edit.root, edit.children);
+        this.AppendChildren(edit.m);
+        break;
+      case "AssignId":
+        this.AssignId(edit.path, edit.id);
+        break;
+      case "CreateElement":
+        if (edit.namespace !== null || edit.namespace !== undefined) {
```
Demonthos, line 269: this should be && not ||

jkelleyrtp:
Current status:
- A bunch more tests for templates. We now test both template mutations and edits
- Pulled out Option<Node> to anyhow::Result<Node>. We can't use our own error type if we want to integrate with anyhow's, unfortunately. Anyhow's is nice since you can downcast the anyhow error into your own error, dynamically.
- A bunch of optimizations around "zero sized templates". We don't cache/generate templates on the HTML side if the template is just dynamic nodes, thanks to how we load templates root-by-root. This is a big win for components that pass through other components. 

I want to merge this PR in a somewhat incomplete state to unblock work in the areas that depend on this PR. Namely, native-core, TUI, liveview, etc. I'm getting a little fatigued on this branch so it'd be nice to get something out the door that's at parity with what we have already.

jkelleyrtp:
Punting forward the integration of error boundaries and suspense rollover.

Right now suspense/await works for SSR but if a component re-enters suspense, the code currently just gives up.

I will follow this PR up with a cleaner, more precise PR to solve those two challenges, but for the most part, we're done here.

I'm waiting for #630 to be ready and then we can connect everything together and merge.

packages/desktop/src/desktop_context.rs
```diff
@@ -185,107 +200,104 @@ pub enum UserWindowEvent {
...
-        },
+                desktop.views.push(ui_view);
+            },
+
+            #[cfg(target_os = "ios")]
+            PopView => unsafe {
+                use objc::runtime::Object;
+                use objc::*;
+                assert!(is_main_thread());
+                if let Some(view) = desktop.views.pop() {
```
MarijnS95, line 292: Looks like the `desktop` argument was removed from this function, making `ios` fail to compile: https://github.com/MarijnS95/xbuild/actions/runs/3637588613/jobs/6138997629

jkelleyrtp: Just fixed, this, sorry about that!

MarijnS95: Thanks! Our bad for using a `git` reference but at least we helped pointing it out I guess :tada:

PR #602, opened 2022-10-22
Some fixes for wayland and ios.

dvc94ch:
Updating to webbrowser 0.8.0 fixes compiling for ios, while not crashing if the webview isn't initialized yet fixes an issue encountered on sway/wayland/linux.


jkelleyrtp (approved):

jkelleyrtp:
Looks great, thank you!

PR #612, opened 2022-11-04
Implement push view for ios.

dvc94ch:
So this would allow drawing an ios screen ontop of the webview that could be exited later. Is this a feature we can add to dioxus?


jkelleyrtp:
This would be a welcome feature! It would need to be feature gated by platform cfg, of course.

dvc94ch:
this was tested and it works.

PR #618, opened 2022-11-10
fix desktop dependency `wry`

TeFiLeDo:
`dioxus-desktop` currently depends on `wry` via git.

This is an issue because:
1. The specified branch (`release`) doesn't seem to exist within the [wry repository](https://github.com/tauri-apps/wry).
2. IIRC crates.io doesn't allow git dependencies.
3. The specified version is already available via crates.io

This PR fixes all of the above by changing the dependency back to crates.io.


Demonthos (approved):
Looks good, Thanks!

PR #637, opened 2022-12-03
router rewrite (again)

TeFiLeDo:
As mentioned on discord, I rewrote my other router rewrite (#522).

# Comparison to the current router
- completely new API, relying on rust instead of `rsx!`
- over 130 tests
- about 50 doc tests (in addition to the 130 above)
- nested routes
- parameter routes (with a filtering variant)
- named navigation
- SSR & sitemap generation support

# Comparison to #522 
- the routing core is no longer dependant on dioxus and could be used for other UI crates
- things that were previously identified by a `&'static str` now use `TypeId`
- named targets are constructed via a function instead of `From` impls

# Tracking
- [x] routing core
- [x] dioxus integration
   This should be fairly easy, I just need to figure out the things that changed in dioxus since #522. Most of the logic is in the core already.
- [x] web history integration
   I'll rewrite what's in #522, but the logic is already there.
- [x] documentation 
   This'll take a bit longer, but I should be able to reuse most of the work I did for #522.
- [x] examples
   By this I mean updating the examples outside the router that rely on it.

# Dioxus integration
- I plan to keep the dioxus components and hooks mostly as they were in #522.
- I was told that the `Component` definition in dioxus would go away. This means the `as Component` thing wont be possible any more. Since there still needs to be a conversion from the `fn`s strange type, I plan to add a `comp()` function to do this.
- I have thought about replacing the `Router` component with a `use_router` hook. I'm not entirely sure whether that is a good idea though. **Pro:** it would be obvious that the router won't react to changing props. **Con:** It is a bit less obvious what is inside the router and what isn't.


TeFiLeDo:
This should now be on par with #522 feature-wise. However, due to the bug I reported in #653, I haven't been able to test if it actually works.

Next steps are:
1. Add documentation comments and eventually add `#![forbid(missing_docs)]` (already done in the core).
2. Write some tests to see if the dioxus parts actually work
3. Maybe add a few examples

After that I'll continue on to the third and forth points I listed in the PR description

TeFiLeDo:
I implemented scroll restoration and was able to avoid the bug I mentioned above.

I also made it optional. However, because of that the router can no longer default to `WebHistory`, even when the `web` feature is enabled. This is because of side effects, specifically setting `history.scrollRestoration = "manual"`.

If this is a problem, I think it should be possible to find a solution. However, for now I'll mark the browser history integration as done and continue on to the documentation and examples.

TeFiLeDo:
This is ready for review now.

I now about one issue I've already written about on [discord](https://discord.com/channels/899851952891002890/952687176737501194/1053264770427535390). It should be solved before merging, either on `master` or in this PR. I'd also like to test the examples manually once that is done.

TeFiLeDo:
I just fixed the CI by reading the `web_router.rs`. I just copied it from before my first change and updated it. I don't know how the test actually works.

TeFiLeDo:
I think the failed checks aren't a result of this PR, but came from `master`.

jkelleyrtp:
Yeah, I wanted to fix the bug but there were too many tests to alter. Will circle around on that

TeFiLeDo:
One other thing that should be considered before merging:

Several hooks and components need access to the routers current state. For that reason, they all share an [`async_rwlock::RwLock`](https://docs.rs/async-rwlock/1.3.0/async_rwlock/struct.RwLock.html). However, this only has an asynchronous `read` function, so currently the code busy waits using the synchronous `try_read` function.

This is obviously not ideal, but I don't know if the other options are better:

1. It could be with `std::sync::RwLock`, which wouldn't busy wait (I assume), but also has no async access, which the router uses in the task the router spawns by default.
2. An alternative to `async-rwlock` could be used, that provides async and sync read functions. The only such crate I could find is `tokio`, which IIRC is not used by `dioxus-web`.
3. I believe you added support for async components to dioxus. Should I convert the components to async, or would that hurt usability/limit them to specific targets/…? Also, should I convert the hooks to be async?

TeFiLeDo:
@jkelleyrtp any chance this will get merged?

Demonthos:
> Several hooks and components need access to the routers current state. For that reason, they all share > an [async_rwlock::RwLock](https://docs.rs/async-rwlock/1.3.0/async_rwlock/struct.RwLock.html). However, this only has an asynchronous read function, so currently the code busy waits using the synchronous try_read function.
> 
> This is obviously not ideal, but I don't know if the other options are better:
> 
> It could be with std::sync::RwLock, which wouldn't busy wait (I assume), but also has no async access, which the router uses in the task the router spawns by default.
> An alternative to async-rwlock could be used, that provides async and sync read functions. The only such crate I could find is tokio, which IIRC is not used by dioxus-web.

It looks like [async_lock](https://docs.rs/async-lock/latest/async_lock/struct.RwLock.html#method.read) has a similar RwLock that has a blocking variant which could be used here.

> I believe you added support for async components to dioxus. Should I convert the components to async, or would that hurt usability/limit them to specific targets/…? Also, should I convert the hooks to be async?

Because the RwLock shouldn't block for very long, I don't think we should use async variants here. Async hooks would force the users component to be async. I think async components also have a small overhead because they first load in a placeholder and then load the final component (also not sure about SSR compatablility).

PR #640, opened 2022-12-06
Optimize Web Template Implementation

Demonthos:
This branch is built on #600 and #630.

- Uses [sledgehammer-bindgen](https://crates.io/crates/sledgehammer_bindgen) to batch operations to js.
- - The Js is about one third the size of the previous interperter
- Uses [smallbox](https://andylokandy.github.io/smallbox/doc/smallbox/index.html) to replace lazy nodes.
- Uses Fxhash wherever possible.
- Template ids are first hashed in rust before being passed to js.

Templates-v3:
Screenshot 2022-12-06 220444

This branch without rif-raf:
Screenshot 2022-12-06 210056

With rif-raf:
Screenshot 2022-12-06 194801

The Dioxus overhead benchmark shows a improvement of ~20%


jkelleyrtp:
Errrr.... I don't know why this was closed + deleted. If you have a local version

jkelleyrtp:
I'm torn on whether or not to include rif-raf. We don't have any renderers that make use of this yet. Plus, since ric-raf is implemented using async, it relies on the browser's macroticks and microticks, even for actions that should be immediate.

In react, all "immediate" events like clicks, typing, etc are processed in the handler that fired the update since it needs to allow bubble cancellation. With how ric-raf is implemented today, we have to wait for the next macrotick before moving the serialized event to the VirtualDom, clearing showing the lag in the charts above.

For this PR, let's turn off ric-raf until we can re-enable it with synchronous event handling.

jkelleyrtp:
I think CI is failing because it's trying to test sledgehammer on windows. There might be some feature flag or cfg-flag that needs to be set to enable/disable that.

jkelleyrtp (approved):
Sooooo happy to have gotten rid of LazyNodes. Looks great! Thank you!

Issue #631, opened 2022-11-25
Node Refs


Demonthos:
## Specific Demand

- Focusing elements
- Scrolling to specific elements
- Accessing native element types (web-sys elements, etc)
- #517

## Implement Suggestion

# API

- We could instead have an onmounted event that is run once when a component is mounted that provides access to all of it's api.
```rust
let maybe_element_ref = use_state(&cx, None);
rsx!{
  div{
    // this is run once on mount
    onmount: |element_ref|{
      // allows escaping the closure by assigning an optional state
      maybe_element_ref.set(Some(element_ref));
      // executed immediately
      element_ref.scroll_left = element_ref.scroll_left() + 1;
      #[cfg(target_arch = "wasm32)]
      let _: web_sys::Node = element_ref.native();
    },
    onclick: |_|{
      maybe_element_ref.unwrap().focus();
    }
  }
}
```
- We should abstract as much about the platform as possible, but for platform specific details we could have a native element type for each platform available through a ```.native``` function.

# Internals

However we provide the element ref, it needs to allow passing around to different elements and event handlers. It also need to persist moves on renderers so that moving an element in a list will not invalidate the node ref.

We could add a Renderer trait that allows the VDom to synchronously query the renderer and make the VDom generic over the renderer. The ElementRef would then contain a reference to the renderer and the ElementId of the node.

PR #647, opened 2022-12-08
Return values from use_eval

Demonthos:
This allows returning serde_json::Value from javascript executed in use_eval. This solves the issue of querying data from the desktop renderer and acts as a precursor to #631. 

The desktop implementation needs to asynchronously return the result. The return type of both the desktop and web implementation implement IntoFuture so that they are interchangeable.

The web implementation has a .get() method to get the value without awaiting a future.


jkelleyrtp (approved):
Looks good! Pretty cool to get it working on desktop. Too bad we can't get them in sync.

PR #648, opened 2022-12-09
fix custom attribute value support

Demonthos:
Implements diffing on remaining attribute values and adds Any support in native core

It would be nice to accept a generic T as a value and then coerce it into Any, but I don't think that is possible without negative impls.


jkelleyrtp (approved):
This looks good!

PR #649, opened 2022-12-10
fix: replace fxhash with rustc-hash in core

saicu:
It seems that https://github.com/DioxusLabs/dioxus/pull/600 has overwritten https://github.com/DioxusLabs/dioxus/pull/588 in the core package, this PR changes it again.


jkelleyrtp:
Thanks!

PR #651, opened 2022-12-10
Fix hot reloading

Demonthos:
Reimplements hot reloading

todo:
- [x] create trait for namespaces to implement that allows for hot reloading outside of the default html namespace
- [x] build Template<'staitc> from CallBody
- [x] create helper to diff CallBodys
- [x] web
- [x] desktop
- [x] companion PR to integrate with cli


jkelleyrtp:
Much appreciated! Thank you!

packages/html/src/global_attributes.rs
```diff
@@ -25,2273 +71,2275 @@ macro_rules! trait_methods {
...
-pub trait GlobalAttributes {
-    trait_methods! {
-        @base
 
-        /// Prevent the default action for this element.
-        ///
-        /// For more information, see the MDN docs:
-        /// <https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault>
-        prevent_default: "dioxus-prevent-default";
+trait_methods! {
```
jkelleyrtp, line 75: Can we keep it so we're running the impls inside the trait declared outside the macro?

This should keep the diff smaller and be a little bit more obvious from the outside what's happening.

Demonthos: We could do that but then the map_global_attributes and map_svg_attributes functions would be associated with the trait

jkelleyrtp: Oh, I understand now. Sorry about that, I will merge an equivalent to master so the diff goes down.

jkelleyrtp:
Gonna play with some of the organization a bit to cut down on dependency chains, but looks good! I hope to follow this up with a PR that turns on hot reloading for desktop by default.

jkelleyrtp (approved):

PR #652, opened 2022-12-10
fix fmt escape sequences in static attribute/text nodes

Demonthos:
fixes attributes or text nodes with ```{{``` or ```}}```


packages/rsx/src/lib.rs
```diff
@@ -244,8 +250,11 @@ impl<'a> DynamicContext<'a> {
             }
 
             BodyNode::Text(text) if text.is_static() => {
-                let text = text.source.as_ref().unwrap();
-                quote! { ::dioxus::core::TemplateNode::Text{ text: #text } }
+                if let Some(text) = text.to_static() {
```
jkelleyrtp, line 253: can we just unwrap here instead of the let some unreachable?

Demonthos: done

jkelleyrtp (approved):

PR #662, opened 2022-12-19
allow changing the root name for desktop

Demonthos:
This implements changing the root name in the desktop renderer similar to what the web renderer does. This is useful for multiple Dioxus instances running in the same app and fixes conflicts with another main element.


jkelleyrtp:
Thanks!

PR #663, opened 2022-12-19
remove references to renderer features that do not exist

Demonthos:
There are some places in the docs that refence renderer features on the dioxus crate that no longer exist. This PR removes those references
followup to #387


azzamsa:
> Thanks! We should migrate more of the docs over for 0.3 sooner rather than later :)

Yes. I want to test out v0.3 but the docs even in nightly is still outdated. Thanks for Demonthos for this.

PR #671, opened 2022-12-20
feat: complete the rollover of aborted and async render states

jkelleyrtp:
This is the final blocker for the suspense work to get 0.3 released. The VirtualDom now intelligently knows how to jump between ready and async states.

The one thing I'm curious about is the `None` return type. Should we give it any semantic value? There's utility in having components that don't render anything, which is what it's currently used for. In 0.2, a `None` return type means that the component will defacto return a placeholder. I'm considering in this relase changing it so that `None` means `abort` and if you abort rendering early, the tree won't actually disappear. IE, you can memoize a component during render.

The one challenge with that is that it's not true memoization as that chunk of the tree is now in an invalid state since we have no choice but to wipe borrowed props and listeners before running a scope. The elements will stay but their listeners are gone and they can't respond to user input.


jkelleyrtp:
Merging to master to fix the issues with returning None elements. Still not 100% done with async rollover.

PR #690, opened 2022-12-29
use lightning css to parse Taffy layout attributes

Demonthos:
This switches from a manual parsing implementation for taffy attributes in native-core to lightning-css. This PR has support for all layout attributes Taffy currently supports and should make it easier to add more attributes in the future.

### DIFF
--- a/packages/native-core/src/layout_attributes.rs
+++ b/packages/native-core/src/layout_attributes.rs
@@ -1,6 +1,6 @@
 /*
-- [ ] pub display: Display,
-- [x] pub position_type: PositionType,  --> kinda, taffy doesnt support everything
+- [ ] pub display: Display, ----> taffy doesnt support all display types
+- [x] pub position_type: PositionType,  --> taffy doesnt support everything
 - [x] pub direction: Direction,
 
 - [x] pub flex_direction: FlexDirection,
@@ -9,7 +9,7 @@
 - [x] pub flex_shrink: f32,
 - [x] pub flex_basis: Dimension,
 
-- [x] pub overflow: Overflow, ---> kinda implemented... taffy doesnt have support for directional overflow
+- [x] pub overflow: Overflow, ---> taffy doesnt have support for directional overflow
 
 - [x] pub align_items: AlignItems,
 - [x] pub align_self: AlignSelf,
@@ -22,19 +22,21 @@
 - [x] pub position: Rect<Dimension>,
 - [x] pub border: Rect<Dimension>,
 
-- [ ] pub size: Size<Dimension>, ----> ??? seems to only be relevant for input?
+- [ ] pub size: Size<Dimension>, ----> seems to only be relevant for input?
 - [ ] pub min_size: Size<Dimension>,
 - [ ] pub max_size: Size<Dimension>,
 
-- [ ] pub aspect_ratio: Number,
+- [ ] pub aspect_ratio: Number, ----> parsing is done, but taffy doesnt support it
 */
 
 use lightningcss::{
     properties::{align::GapValue, border::BorderSideWidth, Property, PropertyId},
     stylesheet::ParserOptions,
+    traits::Parse,
     values::{
         length::{Length, LengthPercentageOrAuto, LengthValue},
         percentage::DimensionPercentage,
+        ratio::Ratio,
     },
 };
 use taffy::{
@@ -326,6 +328,12 @@ pub fn apply_layout_attributes(name: &str, value: &str, style: &mut Style) {
                 style.size.height = convert_size(height);
             }
             _ => (),
+        }
+        // currently not implemented in lightningcss
+        if name == "aspect-ratio" {
+            if let Ok(ratio) = Ratio::parse_string(value) {
+                style.aspect_ratio = Some(ratio.0 / ratio.1);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
+    {
+        match property {
+            Property::Display(display) => match display {
+                lightningcss::properties::display::Display::Keyword(_) => todo!(),
+                lightningcss::properties::display::Display::Pair(pair) => {
+                    match pair.outside {
+                        lightningcss::properties::display::DisplayOutside::Block => {
+                            style.display = Display::None
+                        }
+                        lightningcss::properties::display::DisplayOutside::Inline => todo!(),
```
nicoburns, line 60: I feel like these `todo()`'s shouldn't be included, as they will cause the code to panic if the relevant style is specified. I think it would be better to silently ignore unsupported styles, or else surface a warning/error in a more graceful way

packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
-        "column-fill" => {}
-        "column-gap" => {}
-        "column-rule" => {}
-        "column-rule-color" => {}
-        "column-rule-style" => {}
-        "column-rule-width" => {}
-        "column-span" => {}
-        "column-width" => {}
-        _ => {}
+fn convert_dimention_percentage(
```
nicoburns, line 348: Typo here s/dimention/dimension.

packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
-                "column" => Column,
-                "column-reverse" => ColumnReverse,
-                _ => Row,
-            };
-        }
-        "flex-basis" => {
-            if let Some(v) = parse_value(value) {
-                style.flex_basis = v;
-            }
+fn convert_length_percentage_or_auto(
```
nicoburns, line 358: Doesn't change anything for now, but just FYI, Taffy 0.3 has `LengthPercentage` and `LengthPercentageAuto` types in addition to `Dimension`.

packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
-        | "transition-timing-function" => apply_transition(name, value, style),
-
-        "vertical-align" => {}
-        "visibility" => {}
-        "white-space" => {}
-        "width" => {
-            if let Some(v) = parse_value(value){
-                style.size.width = v;
+            Property::Left(left) => style.position.left = convert_length_percentage_or_auto(left),
+            Property::Right(right) => {
```
nicoburns, line 94: Consider adding support for the `Inset` property which is shorthand for left, right, top and bottom.

packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
+                        }
+                        lightningcss::properties::align::ContentDistribution::Stretch => todo!(),
+                    },
+                    lightningcss::properties::align::JustifyContent::ContentPosition(
+                        _,
+                        position,
+                    ) => match position {
+                        lightningcss::properties::align::ContentPosition::Center => Center,
+                        lightningcss::properties::align::ContentPosition::Start => todo!(),
+                        lightningcss::properties::align::ContentPosition::End => todo!(),
```
nicoburns, line 223: `Start` and `End` can be mapped to `FlexStart` and `FlexEnd`. They're the same in practice. Taffy 0.3 renamed these variants in Taffy to `Start` and `End`. Same applies for the other align/justify properties. Note: this does apply to `SelfStart` and `SelfEnd`.

packages/native-core/src/layout_attributes.rs
```diff
@@ -19,573 +19,400 @@
...
-                style.size.height = v;
+    if let Ok(property) =
+        Property::parse_string(PropertyId::from(name), value, ParserOptions::default())
+    {
+        match property {
+            Property::Display(display) => match display {
+                lightningcss::properties::display::Display::Keyword(_) => todo!(),
+                lightningcss::properties::display::Display::Pair(pair) => {
+                    match pair.outside {
+                        lightningcss::properties::display::DisplayOutside::Block => {
```
nicoburns, line 57: `Block` should not map to `Display::None`. There is actual parsed value for `Display::None` the `Display::Keyword` variant above.

nicoburns (changes requested):

nicoburns:
This is cool. I suspect we'll want to implement something like this upstream in Taffy at some point (either behind a feature flag or as a standalone crate).

packages/native-core/src/layout_attributes.rs
Demonthos: It appears there is some difference at least in the case of justify-content when flex-direction is row-reverse or column-reverse (see https://codepen.io/rachelandrew/pen/KKmazZR)

packages/native-core/src/layout_attributes.rs
Demonthos: Oops, this was carried over from the old system

packages/native-core/src/layout_attributes.rs
nicoburns: Ah, so *that's* the difference (the \*-reverse flex directions). I'll create an issue in the Taffy repo to support both. The 0.3 previews has renamed the variants to `Start`/`End`, but is still implementing the existing `FlexStart`/`FlexEnd`  behaviour.

jkelleyrtp:
Any idea why windows keeps running out of space on this branch?

Demonthos:
> Any idea why windows keeps running out of space on this branch?

I ran the build locally on my machine and it completed successfully. This PR adds lightning css as a dependency. It could be just heavy enough to push it over the memory limit of CI

PR #700, opened 2023-01-01
fix(Examples/Multi-Window): launch the popup content instead of app.

InfamousVague:
Not sure if my assumption is correct but it seemed this was the intention of the popup fn.


jkelleyrtp:
Thanks!!

PR #703, opened 2023-01-02
feat: simple signals implementation

jkelleyrtp:
Implements a new type of state API using `Copy + 'static` keys. The one downside is that the runtime that backs the signals needs to be leaked. We can reclaim the runtime safely by cycling them into linked list or slab, but can never shrink down to zero.

Currently this is not integrated with templates or diffing, but we can eventually expose hooks for both, allowing for the full signal experience.

This PR doesn't add "create_signal", only "use_signal" which taps into the regular re-render process of Dioxus.

---

Going to merge this in an experimental fashion. Will test it against some examples.

Screenshot 2023-01-01 at 8 04 59 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
     "packages/native-core",
     "packages/native-core-macro",
     "packages/rsx-rosetta",
+    "packages/signals",
     "docs/guide",
 ]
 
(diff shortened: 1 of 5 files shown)


jkelleyrtp:
More changes incoming, but merging it to have it experimentally available.

Issue #619, opened 2022-11-13
Value not updating with UseState ==


ealmloff:
**Problem**

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

fn app(cx: Scope) -> Element {
    let mut count = use_state(&cx, || 0);
    count += 1;
    if count == 0 {
        panic!("count should be 1");
    }

    cx.render(rsx! {
        div{
            "{count}"
        }
    })
}
```

**Expected behavior**

The count should be 1 and the program should not panic.

**Environment:**
 - Dioxus version: tested on `master` and `0.2.4`
 - Rust version: `1.65.0`
 - OS info: Windows
 - App platform: All

Demonthos:
I've been thinking about this for a while. I don't think there is any way to resolve this problem while keeping the Deref behavior and not panicking if the value is currently mutably borrowed on PartialEq, Add, etc implementations which is the main difference between UseState and UseRef...

We either need to give up that behavior and unify UseState and UseRef or keep the behavior and leave odd updating behavior as a footgun.

If we keep the behavior we need to update the documentation to state that the value is only valid before writes instead of only warning it is invalid in futures.

PR #704, opened 2023-01-02
Use invariant lifetime and immutable hooks to allow spawning futures without "to_owned"

jkelleyrtp:
This is an attempt at a pretty major change to Dioxus, namely `use_hook`, borrowing some ideas from Sycamore to make our state management simpler in async contexts.

Thesis
---
Generally, I'm not convinced that `use_state` is a successful hook right now. As per #619 , using use_state in a future can be very confusing and generally lead to wrong results. This is because the current value is accessible through the hook, and needs to stay that way because it's tied to the lifetime of that render. This was the original innovation that enabled hooks to borrow into callbacks, so I'm not necessarily disappointed that it's bad that it works that way.

The current issue we have is that every hook needs to behave the same in async contexts as it does in sync contexts. This requires cloning the hook into an async context, leading to a lot of extra code on the hook side and lots of consideration in how a hook she be "to_owned!" into an async environment. This is also super noisy and gets in the way of writing apps faster.

I'm also not convinced that hooks returning `&mut T` is *that* worth the headache they induce. It's great that it's possible, but I don't think any of the hooks we use right now particularly rely on the fact that they can get a mutable reference. Only `use_state` deeply relies on this, and we've mentioned it's already flawed due to the problems related to this "feature".

---

So, my proposal:  `use_hook` only returns immutable references. This is still tied to bump allocation like before and these values can be shared into handlers, like before. But now, we gain a few things:

- Futures can be spawned inline without calling to_owned on every hook
- Hooks don't need to implement to_owned
- `on` handlers can be async

---

Addressing the holes this leaves:

- Spawning is managed by an invariant lifetime.
- use_state is replaced by a signal-like API 
- hooks like use_route that provide an &T to a value now use a signal-like API 

---

An Example.

What a component with some async looked like before:

```rust
    let emails_sent = use_ref(cx, || vec![]);
    let status = use_state(cx, || Drafting);
    let delay = use_state(cx, || false);

    let tx = use_coroutine(cx, |mut rx| {
        let emails_sent = emails_sent.clone();
        let status = status.clone();
        let delay = delay.clone();

        async move {
            while let Some(message) = rx.next().await {
               status.set(Sending);
               sleep(Duration::from_millis(*delay.get_rc())).await;
                emails_sent.write().push(message);
               status.set(Sent);
            }
        }
    });

```

What it looks like now:

```rust
    let emails_sent = use_state(cx, || vec![]);
    let status = use_state(cx, || Drafting);
    let delay = use_state(cx, || false);

    let tx = use_coroutine(cx, |mut rx| async move {
        while let Some(message) = rx.next().await {
           status.set(Sending);
           sleep(Duration::from_millis(delay()).await;
            emails_sent.write().push(message);
           status.set(Sent);
        }
    });
```

### DIFF
--- a/examples/readme.rs
+++ b/examples/readme.rs
@@ -2,18 +2,125 @@
 //!
 //! The example from the README.md.
 
-use dioxus::prelude::*;
+use std::{any::Any, cell::RefCell, marker::PhantomData, pin::Pin};
+
+use dioxus::prelude::{use_state, Element, Scoped};
+// use dioxus::prelude::*;
+use futures_util::{future::abortable, Future};
 
 fn main() {
-    dioxus_desktop::launch(app);
+    // dioxus_desktop::launch(app);
 }
 
-fn app(cx: Scope) -> Element {
-    let mut count = use_state(cx, || 0);
+fn app(cx: Scope) {
+    // let name = use_state(cx, || "world".to_string());
 
-    cx.render(rsx! {
-        h1 { "High-Five counter: {count}" }
-        button { onclick: move |_| count += 1, "Up high!" }
-        button { onclick: move |_| count -= 1, "Down low!" }
-    })
+    let name = cx.use_hook(|| "asdasd".to_string());
+
+    cx.spawn(async {
+        println!("Hello, world! {name}");
+    });
+
+    use_future(cx, async move {
+        println!("Hello, world! {name}");
+    });
+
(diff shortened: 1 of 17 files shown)


jkelleyrtp:
I don't think this is compatible with the prop lifetime since you'd be carrying the lifetime of props into the future.

PR #707, opened 2023-01-03
Fix the namespace and volatility of dynamic attributes

Demonthos:
The current implementation defaults to None for the namespace and false for volatile instead of using the values defined in the html crate.
This breaks the following example:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        h1 {
            color: "{\"red\"}",
            "This should be red"
        }
    })
}
```

This PR fixes that issue.


jkelleyrtp:
Good catch, thank you!

PR #708, opened 2023-01-03
Support optional attributes

### LINKED ISSUE #530: Support optionally adding an attribute

Demonthos:
This PR adds support for optionally adding attributes with Option<AttributeValue>. This makes code to toggle attributes much simpler and more efficient.

closes #530

### DIFF
--- a/packages/core/src/nodes.rs
+++ b/packages/core/src/nodes.rs
@@ -5,7 +5,7 @@ use bumpalo::boxed::Box as BumpBox;
 use bumpalo::Bump;
 use std::{
     any::{Any, TypeId},
-    cell::{self, Cell, RefCell, UnsafeCell},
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Nice!!

PR #709, opened 2023-01-03
Fix potential typo: Strign -> String

MauriceKayser:
Since it compiles without the fix, I am not sure if this is intentional or a typo.


Demonthos (approved):
Thanks! Definitely a typo, the html crate has some unused types for each attribute which could be used in the future to improve type safety or check the attributes.

PR #712, opened 2023-01-04
Easy peasy custom elements using trait namespaces

jkelleyrtp:
cc @olanod @Demonthos , curious to get thoughts here

---

This makes it *really* easy to extend the default element set using traits.

To make this whole process easier, this PR exposes a new crate with that provides a macro for declaring a new namespace of elements.

```rust
custom_elements! {
    // The name of the namespace
    HtmlElements;

    // The name of the generated element in case you need to extend it manually
    RawElement;

    // An element
    base None {
        href: Uri DEFAULT,
        target: Target DEFAULT,
    };

    // A list of attributes that apply to every element in this namespace, applied via traits
    [
        /// Prevent the default action for this element.
        ///
        /// For more information, see the MDN docs:
        /// <https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault>
        prevent_default: "dioxus-prevent-default";
    ]
}
```

For instance, all the html elements are now accessible behind the `HtmlElements` trait implemented for `DioxusElements`. 

To implement your own namespace, you'll do something like this

```rust
/// A generic type that contains metadata for each element
struct HtmlElement<T = ()> {
   name: &'static str,
   namespace: Option<&'static str>,
   _p: PhantomData<T>
}

/// A blanket impl for all elements so they all get a common set of attributes
/// These impls get merged during optimization since every element is a ZST.
impl<T> HtmlElement<T> {
    /// An attribute available for every element in this namespace
    pub const fn class(&self) -> AttributeDescription {
        AttributeDescription {
            name: "class",
            namespace: None,
        }
    }    
}

/// A marker ZST that extends just this element with new attributes
struct link;
impl HtmlElement<link> {
    /// An attribute available for just the link element in this namespace
    pub const fn href(&self) -> AttributeDescription {
        AttributeDescription {
            name: "href",
            namespace: None,
        }
    }    
}

/// A trait that we'll use to extend the ScopeState namespace with
pub trait HtmlElements {
    const div: HtmlElement = HtmlElement::new("div", None);
    const link: HtmlElement<link> = HtmlElement::new("link", None);
}

/// Actually extending the scopestate namespace
impl HtmlElements for DioxusElements {}
```

Now, when the rsx macro operates, it generates code roughly similar to what it did before, but using this new knowledge of traits instead of the module namespace.

```rust
TemplateNode::Element {
    tag: DioxusElements::link.tag,
    namespace: DioxusElements::link.namespace,
    attrs: &[
        TemplateAttribute::Static {
            name: DioxusElements::link.class().name,
            namespace: DioxusElements::link.class().namespace,
            value: "asd",
        }
    ],
    children: &[],
};
```


Demonthos:
This seems like a good approach. It avoids the issues cluttering the global import, and allows should allow combining different namespaces easily!

PR #715, opened 2023-01-04
don't remove real dom nodes when the element id changes

Demonthos:
The current implementation tries to remove nodes when the virtual dom recycles the id. These nodes should be removed with remove mutations so this behavior is not necessary, and it causes issues when the same id is set twice.


jkelleyrtp (approved):

jkelleyrtp:
Yes, the virtualdom recycles automatically

PR #721, opened 2023-01-05
Add fuzzing test to dioxus-core

Demonthos:
This PR adds a fuzzing test to dioxus-core that creates and then diffs random nodes and templates. It is notably missing code to fuzzes keyed diffing. Because this generates random nodes it will hit code paths that are more uncommon. The manual tests we already have can better test diffing nodes that are more similar.

### DIFF
--- a/packages/core/tests/fuzzing.rs
+++ b/packages/core/tests/fuzzing.rs
@@ -32,6 +32,7 @@ fn create_random_attribute(attr_idx: &mut usize) -> TemplateAttribute<'static> {
 }
 
 fn create_random_template_node(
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
It seems like fuzzing found a bug in core. Are you interested in digging into it, or would you like me to?

Demonthos:
I can dig into it

jkelleyrtp:
Do you mind resolving the conflicts? Seems like this now passes.

jkelleyrtp (approved):

jkelleyrtp:
Yayyyyy

PR #727, opened 2023-01-07
feat: add `routable` trait to the router to enable based-match routing

jkelleyrtp:


Demonthos:
This work is being continued in #1020

PR #745, opened 2023-01-10
Make core pass clippy and remove some unnecessary unsafe

Demonthos:
Fixes clippy on master

PR #746, opened 2023-01-10
UnsafeCell<Bump> to UnsafeCellWith<Bump>

FrankReh:
Original intent was to satisfy clippy as this replaces code that used &self to return a &mut Bump with a `with_mut` method.

Idea comes from Tokio. Wrap the UnsafeCell<T> in a new type that provides `with` and a `with_mut` methods. The scope of the unsafe code might be more clear at the call site.

The dereference of the pointer, the *const T, or the *mut T, still has to be wrapped in `unsafe` but it may make it more clear to the caller what safety concerns they are taking into their own hands: whether it is just that they can guarantee exclusive access to the Bump, or whether they also should be guaranteeing the lifetime of the allocation won't exceed the Bump being reset.

### DIFF
--- a/packages/core/src/bump_frame.rs
+++ b/packages/core/src/bump_frame.rs
@@ -3,7 +3,7 @@ use bumpalo::Bump;
 use std::cell::{Cell, UnsafeCell};
 
 pub(crate) struct BumpFrame {
-    pub bump: UnsafeCell<Bump>,
+    pub bump: UnsafeCellWith<Bump>,
     pub node: Cell<*const RenderReturn<'static>>,
 }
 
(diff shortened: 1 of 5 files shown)


FrankReh:
Just a draft while I have so many comment TODOs in the changed code.

This doesn't actually fix anything, except for a clippy warning and a clippy error.

The code is meant to do what it did before, but with a few `unsafe` blocks made explicit.

And before I got any deeper down this rabbit hole, wanted to show the repo maintainers. If you're not interested, not problem, you or I can just close this PR.

FrankReh:
Just saw #745. So maybe changes like these are even less important. Maybe another PR is warranted to address the comments about safety contracts where undocumented unsafe blocks exist.

jkelleyrtp:
Thanks! I just merged a fix to pass clippy, but we need to rework some of this since it's throwing a false alert in MIRI.

PR #751, opened 2023-01-11
Create with_event_handler to expose window event on desktop

Demonthos:
Exposes a callback for the web view renderer to allow users to listen for web view and window events.

### DIFF
--- a/packages/desktop/src/cfg.rs
+++ b/packages/desktop/src/cfg.rs
@@ -1,5 +1,7 @@
 use std::path::PathBuf;
 
+use wry::application::event::Event;
+use wry::application::event_loop::EventLoopWindowTarget;
 use wry::application::window::Icon;
 use wry::{
     application::window::{Window, WindowBuilder},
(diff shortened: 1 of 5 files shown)


packages/desktop/src/cfg.rs
```diff
@@ -77,14 +84,15 @@ impl Config {
...
-    //     handler: impl Fn(&mut EventLoop<()>, &mut WebView) + 'static,
-    // ) -> Self {
-    //     self.event_handler = Some(Box::new(handler));
-    //     self
-    // }
+    /// Set a custom wry event handler. This can be used to listen to window and webview events
+    /// This only has an effect on the main window
+    pub fn with_event_handler(
+        mut self,
+        handler: impl Fn(&Event<UserWindowEvent>, &EventLoopWindowTarget<UserWindowEvent>) + 'static,
```
jkelleyrtp, line 91: This can/should be FnMut.

jkelleyrtp:
I'm not 100% convinced that this is the right answer, since it requires some state being bridge from launch into the state while the app is running. You can't really add handlers at runtime.

The **right** answer is a hook or a "listener" component that gets triggered for every event that can be registered/unregistered at runtime.

Demonthos:
Moving this into a hook to dynamically add and remove hooks makes sense. I think a method on the window would work here. That could be accessed with the use_window hook.

We should also change the drag handler to behave the same way. This would be a breaking change. I can leave the old approach in for now.

jkelleyrtp (approved):

PR #754, opened 2023-01-11
Implement hot reloading for Desktop, TUI, and Liveview

Demonthos:
- [x] TUI
- [x] Desktop
- [x] LiveView
~- SSR-~ (not possible with the current SSR API that takes a immutable reference to the virtual dom, but this does expose hooks that should make it easy for users to implement on there own project)
- [x] Allow configuring the paths the file listener uses
- [x] Update guide
- [x] Allow configuring the hot reloading context for different namespaces

This PR creates hot reloading handler for each platform and enables hot reloading without the CLI by using a hot reloading macro. Each hot reloading handler uses a socket to communicate with the file listener, so that external tools like the CLI can hook into the existing hot reloading handler.


jkelleyrtp:
Looks great. The use of the macro is to get access to the env var to line up the watch directory, right?

Demonthos:
> Looks great. The use of the macro is to get access to the env var to line up the watch directory, right?

Yes

> I'm a tiny bit worried that it will be confusing that the project doesn't hot-reload automatically when logic changes, but I'm not sure if we can get that kind of tooling wired up through a macro.

I have the same concern. I will make it more clear that the user needs to manually rebuild the project to get updates when in the message when the hot reloading fails.

We should be able to do anything that the CLI does with the macro approach. It would require more logic to check if the dev server was started in a previous call and prevent the main thread from closing after shutting down the outdated application, but it should be possible.

packages/hot-reload/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Contributing
+
+- Report issues on our [issue tracker](https://github.com/dioxuslabs/dioxus/issues).
+- Join the discord and ask questions!
+
+## License
+This project is licensed under the [MIT license].
+
+[mit license]: https://github.com/DioxusLabs/dioxus/blob/master/LICENSE-MIT
```
d4h0, line 149: I'm not 100% sure, but shouldn't this be `[MIT license]: https[...]`?

d4h0:
@Demonthos: This looks great!

I have a few questions:

> The interpreter can only use expressions that existed on the last full recompile.

`full recompile` includes `incremental recompilation`, right?

I'm asking, because I've noticed some unexpected behavior when I experimented with variable interpolation in RSX templates in my Wasm app (it seemed a bit buggy, but I might have done something wrong. I'm planning to experiment a bit more, and then maybe open an issue with my findings).

For example `let x = 1; rsx!("foo {x}")` => `let x = "bar"; rsx!("foo {x}")` lead to issues, if I remember correctly. Sometimes `hot-reload` stopped working at all – I believe `let x = 1; rsx!("foo {x}")` => `let y = 1; rsx!("foo {y}")` was such a case – and I had to restart the CLI.

As mentioned, I'll experiment a bit more, and then possibly open an issue.

---

Right now, this isn't compatible with tools like `cargo watch`.

Is my assumption correct, that `dioxus-hot-reload` is able to recognize changes outside RSX macros, that require full recompilation?

(After looking at the changes in this PR a bit more: The answer appears to be 'Yes')

If so, could there be an option to create/change a specific file, which could be used to trigger `cargo watch` to recompile the project, if `dioxus-hot-reload` can't `hot-reload` the changes?

For example, `dioxus-hot-reload` could create/change `./.recompile`. Then we could start our programs via `cargo watch --exec run --watch './.recompile'`, which would instruct `cargo watch` to only watch `./.recompile` for changes.

--- 

With the above, TUI apps could be developed with full `hot-relaod` support by using `dioxus-hot-reload` and then using `cargo watch --exec run --watch './.recompile'`.

---

I've added support for automatic reconnects to the WebSocket after disconnects to my fork of `dioxus-liveview` (I'll open a PR later to evaluate if this can be added to Dioxus). My PR plus `dioxus-hot-reload` and `cargo watch` would also enable full `hot-reload` for `dioxus-liveview` (my PR and `cargo watch` alone would already enable `hot-reload`. But with `dioxus-hot-reload` this setup would also support the lightning fast `hot-reload` that is available for Wasm apps).

(While my PR can be used to get `hot-reloading` working, native support within the CLI would be better, of course. However, `WebSocket reconnecting` should anyway be added to Dioxus. I think this is important for production use, and while developing Dioxus apps).

---

> By default the dev server watches on the `src` and `examples` folders in the root crate directory.

Other paths (e.g., `Cargo.toml` and local dependencies) probably also should be watched, to avoid unexpected issues. It's probably a good idea to do what `cargo watch` does by default.

Demonthos:
Thanks for the feedback!

> Right now, this isn't compatible with tools like `cargo watch`.
> 
> Is my assumption correct, that `dioxus-hot-reload` is able to recognize changes outside RSX macros, that require full recompilation?
> 
> (After looking at the changes in this PR a bit more: The answer appears to be 'Yes')
> 
> If so, could there be an option to create/change a specific file, which could be used to trigger `cargo watch` to recompile the project, if `dioxus-hot-reload` can't `hot-reload` the changes?
> 
> For example, `dioxus-hot-reload` could create/change `./.recompile`. Then we could start our programs via `cargo watch --exec run --watch './.recompile'`, which would instruct `cargo watch` to only watch `./.recompile` for changes.
> 
> With the above, TUI apps could be developed with full `hot-relaod` support by using `dioxus-hot-reload` and then using `cargo watch --exec run --watch './.recompile'`.
> 
> I've added support for automatic reconnects to the WebSocket after disconnects to my fork of `dioxus-liveview` (I'll open a PR later to evaluate if this can be added to Dioxus). My PR plus `dioxus-hot-reload` and `cargo watch` would also enable full `hot-reload` for `dioxus-liveview` (my PR and `cargo watch` alone would already enable `hot-reload`. But with `dioxus-hot-reload` this setup would also support the lightning fast `hot-reload` that is available for Wasm apps).
> 
> (While my PR can be used to get `hot-reloading` working, native support within the CLI would be better, of course. However, `WebSocket reconnecting` should anyway be added to Dioxus. I think this is important for production use, and while developing Dioxus apps).

I need to go back and update the docs for the latest changes but I just pushed some changes to allow configuring a command to run when the hot reloading fails. This should be more flexible than creating a empty file. You could create a build command or create a directory to trigger cargo watch if you prefer that method. With this combined with your PR for reconnecting web sockets should provide a full CLI-like dev experience.

> > By default the dev server watches on the `src` and `examples` folders in the root crate directory.
> 
> Other paths (e.g., `Cargo.toml` and local dependencies) probably also should be watched, to avoid unexpected issues. It's probably a good idea to do what `cargo watch` does by default.

I changed this to more broad. It now listens for any file changes from the root directory except the target directory and any files in the .gitignore.

Demonthos:
Docs are updated. Here is what the new macro looks like:


https://user-images.githubusercontent.com/66571940/212437899-9aac647a-ba82-4013-871d-5a541196f166.mp4

d4h0:
> I just pushed some changes to allow configuring a command to run when the hot reloading fails. This should be more flexible than creating a empty file.

Good idea, that makes much more sense. If `dioxus-hot-reload` handles changes like `cargo`, then `cargo watch` would not be required at all, which is great.

I'm not 100% sure if this makes sense, but maybe the option to execute a Rust expression or call a Rust function would be useful, in some cases.

Similarly, in some cases it might be useful to execute a command or Rust code after a hot-reload.

However, I can't really come up with a real use-case for those suggestions (executing Rust code could save having to execute a separate Bash script, if some more complex action is required).

> I changed this to more broad. It now listens for any file changes from the root directory except the target directory and any files in the .gitignore.

Thanks for considering my suggestion!

I think, watching a too broad range of files wouldn't be good either, however.

For example, I wouldn't want `dioxus-hot-reload` to be triggered when I edit a README file, or files in a potential `wiki` directory, etc.

Ideally, `dioxus-hot-reload` would use the same heuristic that `cargo` itself uses (plus the option to watch additional files, that are specific to the project – e.g., CSS files, and so on).

I'm not 100% sure how exactly that heuristic works, but I think `cargo` considers at least the following:

- Files in the `src` directory
- `Cargo.toml`
- `Cargo.lock` (most likely)
- The above for all local dependencies (dependencies that have a `path` field, probably also local git repos)

In case you are not aware of these, there are a few crates that can be used to get the required metadata:

- [cargo_metadata](https://crates.io/crates/cargo_metadata/)
- [krates](https://crates.io/crates/krates)

There might be subtle, hard to debug bugs, if an incompatible heuristic is used.

> Here is what the new macro looks like working:

That looks great!

Demonthos:
> Good idea, that makes much more sense. If `dioxus-hot-reload` handles changes like `cargo`, then `cargo watch` would not be required at all, which is great.
> 
> I'm not 100% sure if this makes sense, but maybe the option to execute a Rust expression or call a Rust function would be useful, in some cases.
> 
> Similarly, in some cases it might be useful to execute a command or Rust code after a hot-reload.
> 
> However, I can't really come up with a real use-case for those suggestions (executing Rust code could save having to execute a separate Bash script, if some more complex action is required).

That was one of the approaches I looked at, unfortunately there doesn't seem to be a good cross platform way to spawn a detached task outside of spawning new commands. Which means what ever code is run in that closure would need to finish running before any rebuild was triggered so something like using cargo as a library would not work. Ultimately, without a use case that would require this I don't think it is worth complicating the API.

> Thanks for considering my suggestion!
> 
> I think, watching a too broad range of files wouldn't be good either, however.
> 
> For example, I wouldn't want `dioxus-hot-reload` to be triggered when I edit a README file, or files in a potential `wiki` directory, etc.
> 
> Ideally, `dioxus-hot-reload` would use the same heuristic that `cargo` itself uses (plus the option to watch additional files, that are specific to the project – e.g., CSS files, and so on).
> 
> I'm not 100% sure how exactly that heuristic works, but I think `cargo` considers at least the following:
> 
> * Files in the `src` directory
> * `Cargo.toml`
> * `Cargo.lock` (most likely)
> * The above for all local dependencies (dependencies that have a `path` field, probably also local git repos)
> 
> In case you are not aware of these, there are a few crates that can be used to get the required metadata:
> 
> * [cargo_metadata](https://crates.io/crates/cargo_metadata/)
> * [krates](https://crates.io/crates/krates)
> 
> There might be subtle, hard to debug bugs, if an incompatible heuristic is used.

I should have been more clear in my comment. It filters out all changes with files that do not match one of these extensions: `rs`, `toml`, `css`, `html`, and `js`. This seems to be similar to the approach cargo watch uses. This could result in some unwanted recompiles for files that are never actually in the binary, but I think it is fairly reasonable approach.

Tracking local dependencies could be a great feature to add in the future, but I don't think it is necessary for the initial implementation. Not tracking local dependencies shouldn't result in any unexpected behavior.

There are also some concerns that could become an issue as this crate grows:

- The server is running in debug mode because the crate it is compiled in is running in debug mode, so performance could be an issue.

- I believe Cargo watch is able to store hashes of the files instead of the actual file itself because it does not care about how the contents of the file change, just if they do. This implementation needs to store the files in memory to be able to diff the new and previous versions which is significantly more expensive. Specifically when the hot reloading initially starts and the application launches it needs to read all .rs files within the main directory into memory

d4h0:
> Ultimately, without a use case that would require this I don't think it is worth complicating the API.

Yes, I think this is reasonable.

> Which means what ever code is run in that closure would need to finish running before any rebuild was triggered so something like using cargo as a library would not work.

Just to clarify, [cargo_metadata](https://crates.io/crates/cargo_metadata/) and [krates](https://crates.io/crates/krates) use the output of `cargo metadata` (i.e., a build in sub-command of Cargo's CLI, which outputs JSON), and they don't depend on the `cargo` crate. 

> Source files could really be anywhere with macros (include_bytes, etc), so there isn't really a way we can capture all of the files.

You are right. Without interpreting the source code, there always will be possible corner cases.

> Tracking local dependencies could be a great feature to add in the future, but I don't think it is necessary for the initial implementation.

That is reasonable.

> Not tracking local dependencies shouldn't result in any unexpected behavior.

Wouldn't the following lead to unexpected behavior:

`crate_1`:

```rust
pub const FOO: i32 = 1;
```

`crate_2`:

```rust
rsx!{ "{crate_1::FOO}" }
```

If `FOO` is changed to `2`, `dioxus-hot-reload` should recognize that a full re-build is required.

However, it would be reasonable to just mentioning in the docs that dependencies are not tracked.

---

The biggest disadvantage of not doing exactly what `cargo` or `cargo watch` does is, that `dioxus-hot-reload` can't be used as a (perfect) replacement of `cargo watch`. And I have no idea, how to archive that (any other way).

Even if `dioxus-hot-reload` creates a file if a complete rebuild is needed (as described above), and I run `cargo watch --exec run --watch './.recompile'`, the `--watch` option results in `cargo watch` not watching local dependencies anymore (which I just realized).

---

> The server is running in debug mode because the crate it is compiled in is running in debug mode, so performance could be an issue.

I had another look at the source code of this PR, and it seems the main reason why this is a macro, and not a sub-command of the CLI is to get `CARGO_MANIFEST_DIR` – is that correct?

If so: You can use the above-mentioned `cargo_metadata` crate to get that data (see [here](https://docs.rs/cargo_metadata/latest/cargo_metadata/?search=manifest)). `cargo_metadata` is intended for exactly this use-case.

Then the CLI could have a sub-command so that users would just execute something like `dioxus run --hot-reload`, without having to add a macro and a Cargo feature to activate/deactivate `hot-reloading`.

This would have the added benefit, that the CLI is compiled in release mode, and that `dioxus-hot-reload` (and its dependencies) wouldn't slow down compilation of the users' application.

d4h0:
Also, with a `dioxus run --hot-reload` (or `dioxus watch`) sub-command it probably would be not unreasonable to default to recompiling the application when a full re-build is required. With a macro that would seem a bit odd.

Demonthos:
> > Not tracking local dependencies shouldn't result in any unexpected behavior.
> 
> Wouldn't the following lead to unexpected behavior:
> 
> `crate_1`:
> 
> ```rust
> pub const FOO: i32 = 1;
> ```
> 
> `crate_2`:
> 
> ```rust
> rsx!{ "{crate_1::FOO}" }
> ```
> 
> If `FOO` is changed to `2`, `dioxus-hot-reload` should recognize that a full re-build is required.
> 
> However, it would be reasonable to just mentioning in the docs that dependencies are not tracked.

> The biggest disadvantage of not doing exactly what `cargo` or `cargo watch` does is, that `dioxus-hot-reload` can't be used as a (perfect) replacement of `cargo watch`. And I have no idea, how to archive that (any other way).
> 
> Even if `dioxus-hot-reload` creates a file if a complete rebuild is needed (as described above), and I run `cargo watch --exec run --watch './.recompile'`, the `--watch` option results in `cargo watch` not watching local dependencies anymore (which I just realized).

Personally I wouldn't expect this. This is the same behavior that the Dioxus CLI and Trunk have. This would be a great feature to add, but I don't think it is expected. In the same way if I used C bindings with cargo watch I wouldn't expect it to notice changes to the C bindings. You need to draw the line somewhere on what it watches or doesn't, different tools choose different points.

If you do want to track local dependencies, you could always add those paths to the watch paths explicitly. Or if this feature is important to you, you could add it in a follow up PR. I just don't personally have a use case for it, but I would be happy to help if you want help getting started! I believe there is a way to ignore files in cargo watch, could you ignore all of the files in the crate but still watch the dependencies?

Demonthos:
> > The server is running in debug mode because the crate it is compiled in is running in debug mode, so performance could be an issue.
> 
> I had another look at the source code of this PR, and it seems the main reason why this is a macro, and not a sub-command of the CLI is to get `CARGO_MANIFEST_DIR` – is that correct?

The main reason for using a macro instead of integrating with the CLI is making hot reloading work without having to install the CLI.

> If so: You can use the above-mentioned `cargo_metadata` crate to get that data (see [here](https://docs.rs/cargo_metadata/latest/cargo_metadata/?search=manifest)). `cargo_metadata` is intended for exactly this use-case.
> 
> Then the CLI could have a sub-command so that users would just execute something like `dioxus run --hot-reload`, without having to add a macro and a Cargo feature to activate/deactivate `hot-reloading`.
> 
> This would have the added benefit, that the CLI is compiled in release mode, and that `dioxus-hot-reload` (and its dependencies) wouldn't slow down compilation of the users' application.
> 
> Also, with a dioxus run --hot-reload (or dioxus watch) sub-command it probably would be not unreasonable to default to recompiling the application when a full re-build is required. With a macro that would seem a bit odd.

Each approach has tradeoffs:

Advantages for the CLI:
- Better performance and faster compile times
- Unification between web and native tools
- Deduplication of commands: In order to get the full cli-like behavior with the macro approach you need to call the initial build command and add that command to the configuration of the macro.
- You can track for file changes during recompiles: In the macro system if you trigger a recompile, and change a file during the recompile it will not detect the change because the watcher has not started yet.
- External configuration files can be easily added to a .gitignore file to be ignored

Advantages for the macro approach:
- You just need cargo run to get started / You don't need to install any external tools
- Configuration is just rust code
  - Autocomplete and inline documentation for configuration
- It is more clear how different namespaces could work with this approach. For libraries that use a non-html namespace with Dioxus (eg. Freya) there needs to be some way to configure a mapping between rust code and the resulting attribute/element names. With the macro approach these libraries can re-export the macro with a custom namespace default. With the CLI approach it is unclear how this would work. One approach is reading the configuration from a file, but that would require every user of Freya to download that file.


I'm open to either approach. Most of the hot reloading crate could be re-used as a library in the CLI.

d4h0:
Alright, that all makes sense. Thanks, for answering my questions and responding to my suggestions.

> Each approach has tradeoffs:
> Advantages for the macro approach:

Ah, I see. 

> I'm open to either approach. Most of the hot reloading crate could be re-used as a library in the CLI.

I'm interested in going through the hot-reloading code to see how it works, at some point. Maybe I'll then try to build a CLI sub-command with it (if nobody else has built it yet). It seems, both options would be valuable.

jkelleyrtp (approved):
Looks great! We might want to play with the API a tad - maybe the configuration a part of config or something so the hot reloading context is tied to a specific app instance. But for now, looks great.

Issue #736, opened 2023-01-09
dioxus_liveview::interpreter_glue should not wrap javascript in <script> tag


DusterTheFirst:
## Specific Demand

`dioxus_liveview::interpreter_glue` will always return a string of javascript wrapped in `<script>` tags. This forces using the glue inline with the HTML, and not as a separate resource. This can cause problems when serving a website with a `Content-Security-Policy` as it is unsafe and unwise to allow un-hashed/un-nonced inline scripts in your page. By not letting the user create the script tags, these attributes are very hard if not impossible to add to the script tags given. Another use case would be allowing the interpreter glue to be served as a separate asset, for example `https://example.com/glue.js`.

## Implement Suggestion

Remove the `<script>` tags from the content of the string returned by `dioxus_liveview::interpreter_glue`

PR #762, opened 2023-01-13
Add WebSocket reconnect functionality to `dioxus-liveview`

d4h0:
This adds functionality to `dioxus-liveview` needed for automatically reconnecting when the WebSocket was closed for any reason.

IMO, this is both important for production and for developer experience.

This PR in combination with `cargo watch` also enables an easy way to get `hot-reloading`/`live-reloading` working – for example, via `cargo watch --exec run --delay 0.1 --clear`.

Getting `live-reloading` working was my initial goal, but it turned out that automatic WebSocket reconnecting was the only missing part.

When `cargo watch` detects a change to the application source code, recompiles it, and then restarts the server, the WebSocket connection is closed. After the server successfully is restarted, the WebSocket client (which polls the server) reconnects again, which will trigger a full re-render of the app.

Please let me know if my implementation is appropriate, and you are interested in merging it into `dioxus-liveview`. 

There are still a few things to do, but I first wanted to get some feedback on this PR.

There are also a few decisions to make, where I was not sure what would be ideal.

Please see the comments marked with `TODO`.


d4h0:
It probably makes sense to move some comments, that are not specific to certain lines of code, here, so they are easier to discuss. I therefore removed those comments, and added them below (with some more comments/ideas).

---

Ideally, there would be a way to set up an action that is sent from the server to the client after a connection is established, which is executed on disconnects (e.g., to display a message that says that the server can't be reached). Is there already such functionality? If not (and it would be difficult to implement), this can be added via a different PR (probably by someone else). I think, this could be as simple as "On disconnect, set element $selector to $rsx"

---

Similarly important would be, to signal to the app itself that the client disconnected (i.e., to the programmer of the app). Otherwise, it's not possible to save the current state of the user.

Ideally, it would somehow be possible to restore the state before the disconnect, when a client reconnects. At the moment, the app is just re-rendered in the browser, which leads to bad user experience (e.g., for end-users with spotty internet connections). 

What would be a good way to restore state after reconnects? Could the reconnecting client send something to the server that would result in the restoration of the state? This seems slightly similar to rehydration.

---

It seems, the last two sections are also relevant to other renderers.

For example, let's say, the backend of an `dioxus-desktop` app panics, and the connection to the WebView is severed – what would happen?

Can Dioxus run in a thread that's not the main thread, which would mean that Dioxus can crash, without the program being closed?

I imagine, the UI would just freeze, then.

With a `disconnet event handler` (provided at startup, either by the user, or a default one), the UI/interpreter at least could display an error.

Similarly, if there is a bug in the interpreter that powers `dioxus-desktop`, the UI could crash/freeze (and disconnect from the backend), in which case the backend should be able to save the current state of the app.

Therefore, it would make sense to implement such error handling functionality globally for all renderers.

Please let me know if that makes sense. I will then open a separate issue with the above information.

---

From the source code:

> Browsers log un-catchable errors to the developer console, when network requests fail. Those errors will potentially be logged many times until we successfully reconnect. Unfortunately, there doesn't seem to be a good way to avoid this. To not add to the noise with our own log messages, we'll not log anything while reconnecting.

I've read, that WebWorker network errors will not be logged to the developer console. So moving parts of this script to a WebWorker might be a solution. 

However, that seems to add too much complexity, to be a reasonable option, IMO. With incremental re-compilation and the retry delay settings (below the above comment) the situation doesn't seem to be too bad.

Should I still try to create a WebWorker to try to suppress those errors?

Currently, the following is logged, with incremental recompilation (I've added support for logging, in Debug mode):

```
[Dioxus] WebSocket closed – code: [1006], reason: []
[Dioxus] Attempting to reconnect
[Dioxus] Note: WebSocket errors will be logged below, until we are able to reconnect:
> WebSocket connection to 'ws://127.0.0.1:3030/ws' failed:
> WebSocket connection to 'ws://127.0.0.1:3030/ws' failed:
> WebSocket connection to 'ws://127.0.0.1:3030/ws' failed:
> WebSocket connection to 'ws://127.0.0.1:3030/ws' failed:
> WebSocket connection to 'ws://127.0.0.1:3030/ws' failed:
[Dioxus] Connected to WebSocket
```

(The `WebSocket connection to` lines are displayed red (i.e., as error). That's the part that can't be suppressed)

...which doesn't seem too bad. But if the client can't reconnect for a few minutes, there would be hundreds of errors.

---

Should 'WebSocket auto-reconnect' be optional via a Cargo feature flag? It does seem to be very useful in both, production and while developing. However, it is a feature that might send many requests to the server, which might be unexpected. I think, having it always active and configurable (at runtime), and documenting the behavior would be the way to go.

jkelleyrtp:
Thanks for the PR! This is definitely something that's missing, and hasn't been something we approached due to all the complexities you've discovered.
-----
With complex enough JS on the client, I don't think you need a web worker. Can't we just handle the disconnect and try to loop re-connect?

IE 

```
loop {
    try {
        connect().await;
        liveview_loop().await;
    } catch (err) {
       // connection lost, wait some time before trying again
    }
}
```

I imagine the same thing could be added to all the interpreters instead of using the global callback thing currently in place.

----

That being said, it'd be very nice to have a handler on the server side that, like you said, could wait for a connection to be disconnected and give it maybe 30 seconds before serializing the state or just killing the app. 

It would also mean we need to somehow indicate to the client that its state on the server was killed and it should just try to reload or wipe the page since the UI is likely out of sync.

d4h0:
Thanks for evaluating my PR, @jkelleyrtp.

My description above probably wasn't very good, and I'd like to clarify a few things:

> With complex enough JS on the client, I don't think you need a web worker.

So this PR works already perfectly fine (for the `hot-reload` use-case, some stuff is missing for production).

You can give it a try by following the following steps:

1. `git clone https://github.com/d4h0/dioxus`
2. `cd dioxus/packages/liveview/`
3. `git switch liveview-ws-reconnect`
4. `cargo run --example axum --features axum`
5. Go to http://127.0.0.1:3030, and then open the DevTools Console

If you now close the server, you will see that the client tries to reconnect, which should work when you start the server again.

As you can see, there are some errors logged to the console (i.e., `WebSocket connection to 'ws://127.0.0.1:3030/ws' failed`).

These errors are specific to the DevTools Console (i.e., JavaScript has no access to them, and can not suppress them).

To demonstrate that, please run the following code in the DevTools Console:

```javascript
await fetch("http://wecewcfwecfewfc.com").catch(e => console.log("catched!"))
```

As you can see, we catch the error (`catched!` is logged), but there is still an error logged: `GET http://wecewcfwecfewfc.com/ net::ERR_NAME_NOT_RESOLVED`.

As mentioned, such network errors can't be suppressed from JavaScript, which isn't ideal, because we have to poll the server many times until we potentially reconnect.

This might confuse or annoy users of Dioxus, but there isn't anything we can do about it.

The only potential solution I've seen, is to do the polling in a WebWorker. According to the comment I saw, network errors of WebWorkers are not displayed in the DevTools Console (but I haven't tried that yet).

I think, the added complexity of a WebWorker, just to avoid those errors, isn't worth it.

But if you think it is, I'll give it a try.

> Can't we just handle the disconnect and try to loop re-connect?

Yes, that's exactly what the JavaScript client does :)

Unfortunately, WebSockets are callback-based, not Promise-based, so `async/await` can't be used.

> I imagine the same thing could be added to all the interpreters instead of using the global callback thing currently in place.

Can you explain that a bit more? As far as I know, only `dioxus-liveview` uses WebSockets (`dioxus-desktop` seems to use the `interprocess`, which makes much more sense than WebSocket, for desktop apps).

At this point, I have only read the guide, created a simple `hello-world` app, and looked at the source code of `dioxus-liveview`, so I might miss something.

> That being said, it'd be very nice to have a handler on the server side that, like you said, could wait for a connection to be disconnected and give it maybe 30 seconds before serializing the state or just killing the app.

As mentioned above, I haven't really used Dioxus, so the following might not make much sense.

I think, we should just signal the disconnect to the app, and let the programmer handle that situation.

Something like this:

```rust
use dioxus::prelude::*;

pub fn run(cx: Scope) -> Element {
    // Setup an action that will be executed by the client on disconnects:

    // (This would be sent to the client when the connection is established)

    set_client_disconnect_action(&cx, "#css-selector", || rsx! {
        "Server can't be reachead! Trying to reconnect..."
    });

    // ^ On disconnects the client will replace `innerHtml` of `#css-selector`
    // with the HTML generated by `rsx!`

    // Maybe we also should support custom JavaScript:

    let js = DangerousJs("alert('Server can't be reachead! Trying to reconnect...')");

    // ^ `DangerousJs` makes it clear that care is needed to prevent
    // Cross-Site-Scripting vulnerabilities, etc.

    set_client_disconnect_js_action(&cx, js);

    // Handle disconnects and reconnects on the server:
    
    let user_id = get_user_id();
    let state = use_state(&cx, move || get_user_state(user_id));

    on_disconnect(&cx, move || save_user_state(user_id, state));
    on_reconnect(&cx, move || state.modify(|state| *state = get_user_state(user_id));

    // To enable handling disconnects after a delay, the following API could be offered:

    let delay = std::time::Duration::from_secs(30);

    on_disconnect_with_delay(&cx, delay, move || todo!());

    // `delay` could be `Into<DisconnectDelay>`, which would be implemented for
    // `Duration` and `i32` (which would be converted into a `Duration` value,
    // e.g. via `Duration::from_secs`), so the following would be possible:

    on_disconnect_with_delay(&cx, 30, move || todo!());

    cx.render(rsx! {
        div {
            "Hello, world!"
        }
    })
}
```

> It would also mean we need to somehow indicate to the client that its state on the server was killed and it should just try to reload or wipe the page since the UI is likely out of sync.

So right now, the app is re-rendered on every reconnect (`innerHtml` of `root` is set to ""). And one of my goals is to properly restore the server state.

This probably will involve that clients get an ID (e.g., a UUID), that is used to communicate between client and server (so that a client can be linked to a session).

With such an ID it should be easy to tell the client if its server state was deleted.

However, I think that logic should ideally be handled on the server. The client sends its session ID to the server, and the server sends the appropriate edits (so that we don't need to set `innerHTML` of `root` to "" on the client).

jkelleyrtp (approved):

d4h0:
@jkelleyrtp: I realized that setting `root.innerHTML` to `""` does not work (this seems to crash the `Interpreter` if some actual changes happen). I'm now trying to figure out how to properly manage the state across reconnects.

Also: I will add a builder to configure the generation of the interpreter glue. This makes it possible to configure how/if reconnects happen, and can be used to support features like https://github.com/DioxusLabs/dioxus/issues/736.

After that, or if I hit some problem I don't know how to overcome, I'll re-summarize the situation here (but feel free to add suggestions, if you have some).

d4h0:
Quick update: I'm still working on this, and went through a couple iterations already. Some unforeseen complexities gave me a bit of a headache, but my last version seems to work well. I have an idea to make things more flexible, and will explore that route next week. I guess, after that I'll have something to show.

d4h0:
@jkelleyrtp: The code is now ready to be reviewed.

In this comment I'll give you a high-level overview of the current state of the PR.

Before I start, I'd like to add a few warnings:

1. I discovered a few issues with the previous code which I've fixed (e.g. there was no proper error handling, WebSocket messages where just converted to strings, which resulted in close messages being ignored, and silently ignored messages in general). At some point I started to consider this PR as a complete rewrite of the previous code. There are not many lines left that existed before this PR.

2. I basically never have worked with other people on software projects before (besides a few small PRs), which means A) I wasn't forced to learn Git (i.e. I have almost no Git skills), and B) I might have done things that are inappropriate. Please let me know if there are any issues in this regard.

3. English isn't my native language, so the names I've chosen might not be good. Feel free to tell me if there is anything that could be named better (or just rename it). I've used `cargo spellcheck`, so hopefully there are not too many spelling errors.

With that being out of the way, I discovered two issues in regard to letting clients automatically reconnect:

**Issue 1:** In a distributed system (i.e. more than one server), clients might reconnect to a different server.

Because of that, I quickly gave up the idea to automagically store and restore the state in Dioxus. Instead, I've implemented simple tools that users can use to store/restore state:

1. A `use_disconnect_handler` hook, that provides a callback to Dioxus, which is called after a client disconnects (and can be used to save the state). Every component can have one or more handlers.

2. A `ClientStatus` enum is provided via `get_client_status`, which can be used to detect if a client is connecting or reconnecting (e.g. to restore state when a client reconnects).

3. A `use_disconnect_client_actions` hook, which can be used to define actions that will be executed on the client, after the connection is lost. This can be used, for example, to display a warning message, or to disable form element to prevent data loss.

With those tools, users are able to implement proper session management (e.g. via Redis or a database, in a distributed system).

A previous iteration also implemented this session management itself (i.e. it sent a session ID to the client, which it had to provide to the server when reconnecting).

However, it turned out, that it is better if the user handles the session management. This is both more flexible and more secure (because the user can use secure, HTTP-only, encrypted cookies to assign session IDs to clients. What I had implemented stored the session ID via JavaScript, in memory).

**Issue 2:** Saving state if the server crashes or restarts.

Basically, this is pretty hard to archive. At the moment, the best way to archive this, is to just save important state (e.g. form data) after every change. There is a comment somewhere in the code where I discuss this in more detail, and what potential options I see.

Besides the above tools, I've added/changed the following things:

1. I've added a proper error type (previously, the error type had only one `SendingFailed` variant, without any information, that was returned for every failure)

2. Adapters now need to produce/consume a `WebSocketMsg` enum (instead of producing/consuming strings). I've done this to properly handle WebSocket close messages (previously, close message were converted to an empty string, which wasn't handled, which both isn't great)

3. `main.js` was rewritten to enable automatically reconnecting, logging, and client-side actions

4. Besides `interpreter_glue` there now also is `InterpreterGlueBuilder`, which can be used to configure the generated JavaScript code

5. I've added a comprehensive example that demonstrates session management (`sessions.rs`)

**What is missing?**

There are a few things that I'd like to implement only after I got some feedback from you:

1. Updating/adding/finishing the documentation
2. I've only ported the Axum adapter (but I looked at the docs of Warp and Salvo, and porting those adapters shouldn't be a problem)
3. Updating the examples (there might be some minor things to do)
4. A few minor things marked by `TODO` comments (however, I'm not planning to address all TODO items with this PR. This PR is probably already way too big)

There are also comments marked with `XXX`, which directly address you, and where I'd like to get feedback from you.

Besides general feedback regarding the approach I've chosen, some of those XXX comments (or rather the issues they talk about) are important for finishing this PR. I've tried to decide as much as I could, but a few times I wasn't sure what the best approach would be.

**Recommended review order**

I thought, maybe it would be useful to give you a plan to review this PR. I'd do it in this order:

1. `examples/session.rs`: This will give you another high-level overview of what is going on
2. `lib.rs`: Mostly, the error type and the new glue code builder
3. `adapters/mod.rs`: The updated adapter traits, and the new `WebSocketMsg` type
4. `adapter/axum_adapter.rs`
5. `hooks.rs`: The functions here are all pretty simple, but there are some important comments
6. `pool.rs`
7. `main.js`

As I wrote above, there are not many lines left from the previous version of the code. So I don't think it would make sense to view Git diffs. I'd just clone my repo and review it in an IDE with Rust-Analyzer.

d4h0:
There seem to be some merge conflicts. I'll try to resolve them next Tuesday.

d4h0:
Alright, I've fixed the merge conflicts. This was the first time that I have resolved conflicts, so I hope this was done correctly.

d4h0:
@Demonthos: There were two issues with the `hot-reload` feature you have added to `liveview` (which I have just merged into my branch):

1. The `hot-reload` feature was added to the default features, which resulted in 100% of the CPU being used. I guess, because the `hot-reload` macro isn't executed anywhere in the code I use for testing (so the `hot-reload` server channel always returns `None`). 

3. Compilation failed if the default features were disabled because some code that depends on `dioxus_hot_reload` wasn't guarded by the `hot-reload` feature flag.

I have removed `hot-reload` from the default features, and I've put all code that depends on `dioxus_hot_reload` behind the feature flag (see my last commit above).

When I merged the `hot-reload` changes I was already planning to propose to not add `hot-reload` to the default features. I think this incident is a good example for what could happen (it also doesn't feel right to have development features in the list of default features).

You might want to check the other places, where you have added the `hot-reload` functionality, for the above issues.

Ideally, the `hot-reload` functionality would check if the server is running, and panic if it does not.

Demonthos:
> @Demonthos: There were two issues with the `hot-reload` feature you have added to `liveview` (which I have just merged into my branch):
> 
> 1. The `hot-reload` feature was added to the default features, which resulted in 100% of the CPU being used. I guess, because the `hot-reload` macro isn't executed anywhere in the code I use for testing (so the `hot-reload` server channel always returns `None`).

> 2. Compilation failed if the default features were disabled because some code that depends on `dioxus_hot_reload` wasn't guarded by the `hot-reload` feature flag.
> 
> You might want to check the other places, where you have added the `hot-reload` functionality, for the above issues.

Thanks for letting me know about these issues. I think both of these issues should be fixed in https://github.com/DioxusLabs/dioxus/pull/818. I moved the match into the future select so it will only resolve the future and update the vdom if the channel is connected which should resolve the performance issues.

> I have removed `hot-reload` from the default features, and I've put all code that depends on `dioxus_hot_reload` behind the feature flag (see my last commit above).
> 
> When I merged the `hot-reload` changes I was already planning to propose to not add `hot-reload` to the default features. I think this incident is a good example for what could happen (it also doesn't feel right to have development features in the list of default features).

Here is my reasoning behind enabling the hot-reloading features by default:
- It makes it easier to get started with hot reloading: You just need to call the macro without changing the Cargo.toml
- It is also only enabled in debug builds. For me this is similar to having more extensive checking in debug mode: It makes the developer experience better by default, and you don't need to pay the runtime costs in release builds.
- I think for most cases having hot reloading enabled in debug mode is beneficial. The feature is pretty new, so there could be bugs like this, but users can always disable the default features and enable them manually if there is an issue with the feature. We could disable it as a default feature until it is more stable, but in the long term I think it should be enabled by default.

It could also be useful to get some broader community feedback on how hot reloading should work through a poll:
- Should the configuration be in rust code or in a .toml file?
- Should we use the macro approach for hot reloading, the dioxus cli, or both?
- Should hot reloading be enabled by default?
- Other issues, or suggestions for hot reloading?

> Ideally, the `hot-reload` functionality would check if the server is running, and panic if it does not.

I don't think the code should panic if hot reloading fails to connect. If it does not connect, it is not an issue with the users code, but with the library. We should provide some feedback to the user, but I don't think panicking is the right solution here. Currently, hot reloading logs when it successfully connects, we could add a different log if it fails to connect after a certain amount of time.

d4h0:
> I think both of these issues should be fixed in https://github.com/DioxusLabs/dioxus/pull/818.

Yes, that should fix it.

> Here is my reasoning behind enabling the hot-reloading features by default:

Alright, I'm not completely against it. It just feels a bit wrong to add a developer feature to the default features. But maybe it's a worthy compromise.

---

This is probably silly, but just in case it isn't, because `liveview` is a server and performance is important – do you think the following would make any significant difference?

```rust
    // #1: Moving `Pending` out of the loop:
    #[cfg(not(all(feature = "hot-reload", debug_assertions)))]
    let hot_reload_wait: std::future::Pending<Option<()>> = std::future::pending();

    loop {
        #[cfg(all(feature = "hot-reload", debug_assertions))]
        let hot_reload_wait = hot_reload_rx.recv();

        tokio::select! {
            // ..

            // #2: Using a condition that is always `false`, if `hot-reload` is disabled:
            Some(msg) = hot_reload_wait, if cfg!(all(feature = "hot-reload", debug_assertions)) => { .. }
        }
    }
```

Demonthos:
> This is probably silly, but just in case it isn't, because `liveview` is a server and performance is important – do you think the following would make any significant difference?
> 
> ```rust
>     // #1: Moving `Pending` out of the loop:
>     #[cfg(not(all(feature = "hot-reload", debug_assertions)))]
>     let hot_reload_wait: std::future::Pending<Option<()>> = std::future::pending();
> 
>     loop {
>         #[cfg(all(feature = "hot-reload", debug_assertions))]
>         let hot_reload_wait = hot_reload_rx.recv();
> 
>         tokio::select! {
>             // ..
> 
>             // #2: Using a condition that is always `false`, if `hot-reload` is disabled:
>             Some(msg) = hot_reload_wait, if cfg!(all(feature = "hot-reload", debug_assertions)) => { .. }
>         }
>     }
> ```

I would imagine it does not make much of a difference, but if you are still seeing performance issues in that area, you could try it. It could be more clear what the code was doing if we make two entirely separate switch statements depending on the features enabled, and the release build would not be polling any extra futures.

d4h0:
Alright. If I ever build something with Dioxus-Liveview that gets significant traffic, I'll probably anyway do some general profiling and benchmarking. Better to wait until then, before making the code more complex.

d4h0:
@jkelleyrtp: Do you have any idea when you will be able to have a look at this? See [my previous comment](https://github.com/DioxusLabs/dioxus/pull/762#issuecomment-1418247001), in case you missed it.

If there is an issue with what I came up with – even if it is fundamental – just let me know.

d4h0:
Due to the lack of feedback, and me therefore losing interest in working on this, I'll close this PR now. If anybody wants to continue, they can fork [my branch](https://github.com/d4h0/dioxus/tree/liveview-ws-reconnect) (which I will not delete, for now).

PR #786, opened 2023-01-20
Clean up removed examples from examples/README.md

Firfi:
An example was removed in https://github.com/DioxusLabs/dioxus/commit/c4a18bc24d8722dda6c7c432a18f11671046718a but is stuck in README still.

Additionally, all tui_* examples were removed at some point too. Synched it up with README.


jkelleyrtp:
Thank you!

PR #787, opened 2023-01-20
Stable Docs Workflow

Demonthos:
Creates a manually triggered workflow to publish the docs on the stable directory instead of nightly


jkelleyrtp (approved):

jkelleyrtp:
Thanks! Didn't realize we could have empty workflow triggers. We might want to tweak the docsite to always point to a specific version instead of a "stable" link so we don't always break permalinks.

PR #789, opened 2023-01-20
Fix replacing a dynamic text node in native core

Demonthos:
This fixes dynamic text nodes not adding their ids to the element id -> node id mapping causing replacing dynamic text nodes to fail

This fixes a bug encountered in #788


jkelleyrtp:
Thanks!

PR #794, opened 2023-01-22
do not box scope states

Demonthos:
Scope states are currently individually boxed which causes each scope state to create a new memory allocation when created and deallocation when removed. If this box is removed, the Vec backing the slab will often have memory pre-allocated that can be used instead of calling the global allocator. 

This is especially useful any time scopes are created and then destroyed. For example this benchmark is ~15% without boxing scope states:
```rust
use criterion::{criterion_group, criterion_main, Criterion};
use dioxus::prelude::*;
use rand::prelude::*;

criterion_group!(mbenches, create_rows);
criterion_main!(mbenches);

fn create_rows(c: &mut Criterion) {
        let num = if cx.generation() % 2 == 0 { 10_000 } else { 0 };
        render!(
            table {
                tbody {
                    (0..num).map(|f| {
                        let label = Label(["hello world", "this is a test", "testing"]);
                        rsx!( Row { row_id: f, label: label } )
                    })
                }
            }
        )
    }

    c.bench_function("create rows", |b| {
        let mut dom = VirtualDom::new(app);
        let _ = dom.rebuild();

        b.iter(|| {
            // let g = dom.rebuild();
            let g = dom.mark_dirty(ScopeId(0));
            dom.render_immediate();
            // assert!(g.edits.len() > 1);
        })
    });
}

#[derive(PartialEq, Props)]
struct RowProps {
    row_id: usize,
    label: Label,
}
fn Row(cx: Scope<RowProps>) -> Element {
    let [adj, col, noun] = cx.props.label.0;
    cx.render(rsx! {
        tr {
            td { class:"col-md-1", "{cx.props.row_id}" }
            td { class:"col-md-1", onclick: move |_| { /* run onselect */ },
                a { class: "lbl", "{adj}" "{col}" "{noun}" }
            }
            td { class: "col-md-1",
                a { class: "remove", onclick: move |_| {/* remove */},
                    span { class: "glyphicon glyphicon-remove remove", aria_hidden: "true" }
                }
            }
            td { class: "col-md-6" }
        }
    })
}

#[derive(PartialEq)]
struct Label([&'static str; 3]);
```

### DIFF
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -37,7 +37,7 @@ log = "0.4.17"
 # Serialize the Edits for use in Webview/Liveview instances
 serde = { version = "1", features = ["derive"], optional = true }
 
-bumpslab = { git = "https://github.com/demonthos/bumpslab", branch = "reduce-unsafe"}
+bumpslab = { git = "https://github.com/jkelleyrtp/bumpslab" }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -37,7 +37,7 @@ log = "0.4.17"
 # Serialize the Edits for use in Webview/Liveview instances
 serde = { version = "1", features = ["derive"], optional = true }
 
-bumpslab = { git = "https://github.com/jkelleyrtp/bumpslab" }
+bumpslab = { git = "https://github.com/demonthos/bumpslab", branch = "reduce-unsafe"}
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -37,7 +37,7 @@ log = "0.4.17"
 # Serialize the Edits for use in Webview/Liveview instances
 serde = { version = "1", features = ["derive"], optional = true }
 
-bumpslab = "0.1.0"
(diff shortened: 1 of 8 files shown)


jkelleyrtp:
If the slab gets resized then all the pointers become invalid. I made a crate for this called bump slab which I think we're already using that provides a bump-backed storage solution that recycles and won't invalidate pointers. We should use that instead.

jkelleyrtp:
By "already using" I meant in our cargo.toml - please use it here instead of boxing! It's not being used in scope allocation, and I think you'll find similar performance wins.

Demonthos:
The implementation is a bit ugly because we are locked into usize ids because the usize in ScopeId is a public API, so changing it would be a breaking change

Demonthos:
https://github.com/jkelleyrtp/bumpslab/pull/1 fixes the miri test, but the fuzzing test still fails... I'm trying to run it though miri now, but because of how slow miri is it will take a while.

It also looks like bump slab never drops values that are not removed manually so creating a bump slab, adding values, and dropping it will leak memory.

jkelleyrtp:
Yes, bumpslab expects values to be reclaimed using the reclaim method.

Demonthos:
I think all the safely issues with BumpSlab are fixed now, and the performance improvements is about the same with this approach

jkelleyrtp:
Great! I will release a new version of bumpslab once this is merged and then we can bump forward the released Dioxus minor version.

PR #797, opened 2023-01-26
Fix drag event serialization in JS interpreter

### LINKED ISSUE #796: Crash with drag events in Desktop App
I have a simple component that acts upon drag events like so:

``` rs
#[inline_props]
fn ResizeComponent(cx: Scope, size: UseState<Option<f64>>) -> Element {
    cx.render(rsx!(div {
        class: "resize",
        draggable: true,
        ondrag: move |event| {
            let value = event.data.mouse.page_coordinates().x;
            if value < 0.0 || value > 0.0 {
                size.set(Some(value))
            }
        },
        ondragstart: move |event| { size.set(Some(event.data.mouse.page_coordinates().x)) }
    }))
}
```

When I run my project and start to drag then this happens (excerpt):

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MissingField("mouse")', /Users/terhechte/.cargo/git/checkouts/dioxus-1e619ce595d3799d/1b597f4/packages/html/src/transit.rs:36:41
stack backtrace:
  16:        0x10561c740 - core::result::Result<T,E>::unwrap::hd33bb8ea583e058b
                               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1113:23
  17:        0x1055fb90c - <dioxus_html::transit::HtmlEvent as serde::de::Deserialize>::deserialize::h6d10adcf429e3294

```

This works fine when using `Dioxus-web`.

**Environment:**
 - Dioxus version: master
 - OS info: macOS Ventura
 - App platform: desktop

Demonthos:
Adds the mouse data under the mouse field instead of serializing the mouse data directly for drag events
fixes #796

### DIFF
--- a/packages/interpreter/src/interpreter.js
+++ b/packages/interpreter/src/interpreter.js
@@ -445,6 +445,41 @@ class Interpreter {
   }
 }
 
+function get_mouse_data(event) {
+  const {
+    altKey,
+    button,
+    buttons,
+    clientX,
+    clientY,
+    ctrlKey,
+    metaKey,
+    offsetX,
+    offsetY,
+    pageX,
+    pageY,
+    screenX,
+    screenY,
+    shiftKey,
+  } = event;
+  return {
+    alt_key: altKey,
+    button: button,
+    buttons: buttons,
+    client_x: clientX,
+    client_y: clientY,
+    ctrl_key: ctrlKey,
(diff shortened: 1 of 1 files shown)


terhechte:
Thanks for the quick fix! 👍

PR #802, opened 2023-01-27
Fix release build

kurotych:
Compiling dioxus app doesn't work in release mode 

image


Demonthos (approved):
cargo clippy seems to failing for an reason unrelated to this PR

PR #803, opened 2023-01-28
Fix Clippy warnings in core

Demonthos:
The Clippy version of CI seems to have updated to warn about non-inlined arguments in format_arguments


jkelleyrtp:
This is a new lint in the most recent version of Rust.

PR #811, opened 2023-01-31
Implement Hydration

### LINKED ISSUE #809: Web Hydration Fails
**Problem**

The web hydration example results in the `web` crate results in a second version of the page being created instead of hydrating the existing page.

**Steps To Reproduce**

Run the [web hydration](https://github.com/DioxusLabs/dioxus/blob/master/packages/web/examples/hydrate.rs) example.

**Expected behavior**

One version of the page is rendered, then hydrated.

**Environment:**
 - Dioxus version: [`master`]
 - Rust version: [`nightly`]
 - OS info: [Windows]
 - App platform: [`web`]

**Questionnaire**

Demonthos:
Hydration functionality was removed when templates were added. This re-implements hydration.

closes #809


jkelleyrtp:
Sweeet!

PR #816, opened 2023-02-06
Async event handlers

jkelleyrtp:
Allows async event handlers for the HTML crate by adding support to Dioxus Core.

This is useful with crates like dioxus-signal which give a static handle to some state.


Demonthos (approved):
I'm suprized that we can make this change without breaking changes, but it seems to work with the old and new event handlers. This should make data loading triggered by events much nicer!

PR #824, opened 2023-02-09
Fix keyed diffing with fragment node children

### LINKED ISSUE #821: Reordering Lists does not work with multiple sub elements
**Problem**

When reordering list elements with for example multiple `tr` rows the ordering does not work correctly.

**Steps To Reproduce**

The example below reprocudes the issue (and the correct issue if the second `tr` is removed)

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

    let mut count = use_state(&cx, || 0);

    cx.render(rsx!(ListComponent {}))
}

#[inline_props]
pub fn ListElement(cx: Scope<'_>, number: i64) -> Element<'_> {
    cx.render(rsx! {
        tr {
            td { "{number}" }
        },
        // without this `tr` below the reordering works fine!
        tr { 
            td { "Details to {number}"}
        }
    })
}

pub fn ListComponent(cx: Scope<'_>) -> Element<'_> {
    let list_state = use_state(&cx, || vec![1, 2, 3, 4, 5]);

    cx.render(rsx! {
        button { onclick: move |_| { list_state.modify(|e| { let mut e = e.clone(); e.reverse(); e }) }, "Reverse" }
        div { list_state.current().iter().map(|e| rsx! { "{e}" }) }
        table {
            tbody {
                list_state.current().iter().map(|e| {
                    rsx! { ListElement {
                        key: "{e.to_string()}",
                        number: *e,
                    }}
                })
            }
        }
    })
}
```

**Expected behavior**

The list should be reordered correctly even if multiple sub elements are inside the ordered elements

With the second `tr` inside the ListElement component the output changes to
```
54321
3
Details to 3
2
Details to 2
1
Details to 1
4
Details to 4
5
Details to 5
```
without it the reordering works fine

```
54321
5
4
3
2
1
```

**Environment:**
 - Dioxus version: `master`
 - Rust version: 1.66
 - OS info: Windows
 - App platform: `web`

**Questionnaire**

Demonthos:
fixes #821
This changes the push all real nodes from counting the nodes created to summing them to get the true total

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -717,7 +717,7 @@ impl<'b> VirtualDom {
                     Fragment(nodes) => nodes
                         .iter()
                         .map(|node| self.push_all_real_nodes(node))
-                        .count(),
+                        .sum(),
 
                     Component(comp) => {
                         let scope = comp.scope.get().unwrap();
@@ -729,7 +729,7 @@ impl<'b> VirtualDom {
                     }
                 }
             })
-            .count()
+            .sum()
     }
 
     fn create_children(&mut self, nodes: impl IntoIterator<Item = &'b VNode<'b>>) -> usize {


jkelleyrtp:
Good catch!

PR #835, opened 2023-02-20
Fix anchor link for block-level elements.

marcerhans:
Block-level elements/tags does not open the systems browser for external links.

An edited version of the "link" example demonstrates this.

```Rust
p {
    a { href: "http://dioxuslabs.com/", div { "Default link - links outside of your app" } }
}
```
The code above would try to open the link in the application itself. External links would only work with *direct* anchor element links (just text).

This fix searches for the closest anchor parent element (or self) in the DOM and fetches that.


marcerhans:
It appears my "fix" breaks the router Link types, I will just report it as a bug.

PR #840, opened 2023-02-22
PartialEq for UseState/UseCoroutine

mirkootter:
Add `PartialEq`-implementation for `dioxus_hooks::Coroutine` and `dioxus_hooks::UseState`.

Both are good candidates to pass as a prop. `UseState` already implements `PartialEq`, but I removed the trait bound on T.


jkelleyrtp:
Looks good! Thank you!

PR #841, opened 2023-02-23
0.2 -> 0.3 migration changes

flisky:
Here're some changes which is required for us to upgrade. 

And I'm glad to help if you have any better idea (API change, method deprecated or even removed).

### DIFF
--- a/packages/fermi/src/root.rs
+++ b/packages/fermi/src/root.rs
@@ -13,8 +13,8 @@ use crate::Readable;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct AtomId {
-    pub(crate) ptr: *const (),
-    pub(crate) type_id: TypeId,
+    pub ptr: *const (),
+    pub type_id: TypeId,
 }
 
 pub struct AtomRoot {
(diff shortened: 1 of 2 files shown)


flisky:
I figure out why `task` is marked `todo()`, because no way to resetting when future is resolved.

Here's a not-so-effecient solution at https://github.com/flisky/dioxus/commit/271dcbbc931b9e200fd01029343e14ca630302b8:

* add a sequence id to task id
* attach sequence id to LocalTask
* take a peek on scheduler tasks when call `UseFuture::task`.

I can turn it into a PR if you're interested.

flisky:
Sorry to bother again, but I think there's some race condition on the `use_future` hook. @jkelleyrtp @Demonthos 

We have two components, `Parent` and `Child`, both have its own `use_future` with a same `REFRESH` fermi. However we find that the parent `use_future` has almost 50% chance not to refresh after modify `REFRESH` (I can comfirm the previous future is already resolved).

I suspect something wrong with https://github.com/DioxusLabs/dioxus/blob/fdca2abf05a076e0268b070f92ed6b90aab8dba9/packages/hooks/src/usefuture.rs#L57-L60

I fixed it by https://github.com/flisky/dioxus/commit/d6a59feec921e8e18fbcaf9dc1783540a788839a, only remove the exactly current task.

Hope it help.

Issue #75, opened 2022-01-06
Disable reload context menu


UARTman:
Running a basic "Hello World" app in Linux I notices a bunch of strange stuff going on:

- Multiple-second load times for the content after the window is initialized (might be inherent to platform, not sure if can be fixed)
- Weird menus: App has two menus in the menubar: "First menu" and "Second menu". First menu gives a number of options like "About App", "Quit", etc., but they do nothing. Second menu creates a strange blank menu that has low width and big height.
- Right-clicking in the window shows a menu with greyed out "Forward", "Stop", and "Backward", as well as working "Reload" and "Inspect Element". Hitting reload breaks the app.

The menus might be intended, but reload breaking stuff is clearly not.

jkelleyrtp:
Thanks for the report!

- I've identified why 1) is happening - GTK/Windows don't receive a MainEventsCleared on startup.
- Menus are a result of poor config merging, this is sorted out in the master branch
- The context menu on right-click is a result of the platform/Tao (not inserted by us) and should be disabled in release mode of your app. I know the inspect menu goes away in release menu. Forward/back is the result of poor config merging.

The one thing I don't exactly know how to fix is reloading the page, so I think we need to override that manually.

<img width="823" alt="Screen Shot 2022-01-07 at 1 16 03 AM" src="https://user-images.githubusercontent.com/10237910/148500645-4dc7840e-64d5-4340-8477-98f8044c6c22.png">

jkelleyrtp:
> Running a basic "Hello World" app in Linux I notices a bunch of strange stuff going on:
> 
> * Multiple-second load times for the content after the window is initialized (might be inherent to platform, not sure if can be fixed)
> * Weird menus: App has two menus in the menubar: "First menu" and "Second menu". First menu gives a number of options like "About App", "Quit", etc., but they do nothing. Second menu creates a strange blank menu that has low width and big height.
> * Right-clicking in the window shows a menu with greyed out "Forward", "Stop", and "Backward", as well as working "Reload" and "Inspect Element". Hitting reload breaks the app.
> 
> The menus might be intended, but reload breaking stuff is clearly not.

Can you check is #84 works for you?

```
git checkout https://github.com/DioxusLabs/dioxus.git
git pull
git checkout jk/windows-lag
cargo run --example readme --features "desktop"
```

It should load immediately now.

UARTman:
> Can you check is #84 works for you with respect to startup lag?

[...]

PR #856, opened 2023-03-01
Create global shortcut handler for Dioxus desktop

### LINKED ISSUE #855: Expose `GlobalHotKeyManager`
## Specific Demand

I'd like to be able to use [global hotkeys](https://github.com/tauri-apps/global-hotkey) from Dioxus.

My understanding is that this should be fairly trivial since the functionality is exposed in Tao.

Thanks.

Demonthos:
closes #855 

The API here is similar to what was outlined in #751 for custom handlers.

There are methods to add and remove shortcuts on the desktop context and a use_global_shortcut hook that will create a shortcut that will be removed when the hook is dropped.

The same method of creating shortcuts is also used to override CTRL+R to fix ctrl+r breaking all apps as a follow up to #75


jkelleyrtp:
Great PR. Code looks clean and exactly what I was expecting the API to look like. A+ work.

PR #873, opened 2023-03-10
Update Router book to reflect crate separation and need for `web` feature.

askreet:
This tripped me up while setting up my first Dioxus app for deployment in a browser. I believe this changed from 0.2 -> 0.3.


Demonthos (approved):
Thank you!

PR #875, opened 2023-03-11
Accept Arbitrary Attribute Values for Loop

stephenandary:
WIP - Draft PR to track progress and information

Below is direction from @Demonthos:

1) Change the key type in the node: https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L51
2) Make arbitrary attribute Hash here: https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L558
3) Change the parsing here https://github.com/DioxusLabs/dioxus/blob/6a16df835d232a5565c10f35a9319e8df7907d2d/packages/rsx/src/element.rs#L16 and https://github.com/DioxusLabs/dioxus/blob/6a16df835d232a5565c10f35a9319e8df7907d2d/packages/rsx/src/component.rs#L59 

Design Considerations: 

1) arbitrary values should probably not need to implement hash. They are used to allow custom elements to accept other types. We can just panic instead.
2) f64 does not implement hash. We can either panic or create a custom implementation

The main use case I see for this is raw indexes for arrays as keys
It might be better to create a different type that does not allow custom values and does not include f64. We should be able to reuse most of the code for arbitrary attribute value

The code for arbitrary attributes in core and rsx could be useful here. The main thing they do is create a trait that can convert from a normal attribute to a arbitrary value here: https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L839
Then create attributes here in the macro: https://github.com/DioxusLabs/dioxus/blob/master/packages/rsx/src/element.rs#L248, which accepts anything that can implements that trait here https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/scopes.rs#L517


packages/core/src/nodes.rs
```diff
@@ -48,7 +48,10 @@ pub struct VNode<'a> {
     /// The key given to the root of this template.
     ///
     /// In fragments, this is the key of the first child. In other cases, it is the key of the root.
-    pub key: Option<&'a str>,
+    ////////////////////////////////////////////////////////////////////////////
+    //////////////////// Change the key type in the node ///////////////////////
+    ////////////////////////////////////////////////////////////////////////////
+    pub key: Option<&'a KeyValue<'a>>,
```
Demonthos, line 54: This could be changed to not borrow KeyValue. The old version borrowed the string because it does not have know size, but KeyValue is Copy and Sized so it is no longer needed

packages/core/src/nodes.rs
```diff
@@ -894,3 +897,70 @@ impl<'a, T: IntoAttributeValue<'a>> IntoAttributeValue<'a> for Option<T> {
...
+    Text(&'a str),
+
+    /// Signed integer key value
+    Int(i64),
+}
+
+// need to understand if these traits are necessary
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serialize", serde(untagged))]
+pub enum BorrowedKeyValue<'a> {
```
Demonthos, line 913: This was necessary to support Arbitrary Any values in Arbitrary Attribute but for Keys Arbitrary values make less sense so we don't need a borrowed variant

stephenandary:
I can deal with the hash trait issues tomorrow. You mentioned this in directions, I will take a gander in the morning.

stephenandary:
I am away for weekend, but I am bringing laptop. I will try to tackle this when I have down time (hopefully today).

stephenandary:
@Demonthos sorry for the delay, I didn't realize I could just run Cargo Build at the root of the core directory. This is my first Open Source Rust PR. I was just using rust analyzer until I relaized that.

stephenandary:
Closing to start over

PR #877, opened 2023-03-12
Don't accept raw attributes in components without trailing commas

### LINKED ISSUE #876: Component attribute formatting fails
**Problem**

Components with attributes without trailing commas followed by another attribute do not trigger the automatic formatting feature.

**Steps To Reproduce**

- Run
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    let idx = 0;
    cx.render(rsx! {
        Comp {
            id: "myid-{idx}" // note no trailing comma
            num: 10
        }
    })
}

#[inline_props]
fn Comp<'a>(cx: Scope<'a>, id: &'a str, num: u32) -> Element<'a> {
    cx.render(rsx! {
        div {
            id: *id
        }
    })
}
```
- Inspect the element and see that the id is "myid-{idx}" instead of "myid-0"

**Expected behavior**

- The element should have an id of "myid-0"

**Screenshots**

**Environment:**
 - Dioxus version: `0.3`
 - Rust version: `v1.67.1`
 - OS info: `All`
 - App platform: `All`

**Questionnaire**

Demonthos:
fixes #876 

It appears that commas are already optional for non-string attribute values in components. This PR makes it optional for all component attributes

### DIFF
--- a/packages/router/tests/web_router.rs
+++ b/packages/router/tests/web_router.rs
@@ -22,7 +22,7 @@ fn simple_test() {
                 onchange: move |router: RouterContext| log::trace!("route changed to {:?}", router.current_location()),
                 active_class: "is-active",
                 Route { to: "/", Home {} }
-                Route { to: "blog"
+                Route { to: "blog",
                     Route { to: "/", BlogList {} }
                     Route { to: ":id", BlogPost {} }
                 }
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
Correct me if I'm wrong, but I think the right fix here is to always enforce the trailing comma, at least to match the behavior of elements right?

Demonthos:
> Correct me if I'm wrong, but I think the right fix here is to always enforce the trailing comma, at least to match the behavior of elements right?

I initially removed the restriction instead because it does not break existing code, but I'm also happy with that option as long as it is consistent.

jkelleyrtp:
Thanks @Demonthos. Rolling this into #1131 tracking

PR #886, opened 2023-03-15
Added mention of StreamExt trait extension.

marcerhans:
Noted that there was no mention of the trait extension (StreamExt) in the docs when trying to send messages via a UnboundedReceiver. For me it was quite confusing where the ".next()" method came from, but figured it out by looking at the test cases. Added a short note/mention of it in the docs.


Demonthos (approved):
Looks good. It might makes sense to add this to the prelude, or exports in the future, so that you don't need to add a separate dependency, but for now this is good. Thank you!

PR #894, opened 2023-03-19
Create onmounted Event

### LINKED ISSUE #423: Element Focus Behavior
## Element Focus Behavior

I want element like `input` can be focused, but native `autofocus` attribute has no effect in `Dioxus` web (Firefox) at the following example:
```rust
use dioxus::prelude::*;
    cx.render(rsx!( div {
        input {
            name: "input_test",
            r#type: "text",
            autofocus: "",
        }
    }))
}

fn main() {
    dioxus::web::launch(app);
}
```

the reason is the `autofocus` attribute has effect only when page loaded but not element dynamic created. 

## Implement Suggestion

`React.js` have two solutions
* `autoFocus` attribute like this
```js
const MyComponent = () => (
  <label>
    Title
    <input type="text" defaultValue="Untitled" autoFocus />
  </label>
);
```
* DOM API `focus()`
```js
const input = document.querySelector("input");
input.focus();
input.select();
```

To support DOM API call, `Dioxus` may need implement some special event e.g.
```
input {
    onmounted: |event| {
            let element = event.target();
            element.focus();
            // or just element id
            // element = document.get_element_by_id(event.element_id);
    },
}
```
so we can get element object. 

I'm not sure whether the feature could be implemented in `dixous-core` crate or just Web DOM based platform.

### LINKED ISSUE #517: Support something like React's componentDidMount and related hooks
## Specific Demand

I'm trying to implement infinite scroll for a page. As part of that, I want to trigger loading more elements when the bottom X elements are visible. On the web, this can done by using an [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver). However, the `IntersectionObserver.observe` method takes elements, not ids. That means I cannot create an observer until after the component is in the DOM.

### LINKED ISSUE #631: Node Refs

### LINKED ISSUE #804: use_eval with scrolling after component is rendered
## Specific Demand

I want to execute eval each time after the component is rendered (I want to make scroll to a particular object) 
In the code below scrolling doesn't work. I think because eval executes before HTML is rendered. 

Looks like it's not implemented https://github.com/DioxusLabs/dioxus/issues/631 https://github.com/DioxusLabs/dioxus/issues/517 but maybe you can suggest me any better workaround that I have? 
Because I've found nothing better than just sleeping in the "Component" function  

```rust
    let eval = dioxus_desktop::use_eval(cx);
    let res = cx.render(rsx! {
        div {
            dangerous_inner_html: "{html}"
        }
    });
    eval(JsScroll.to_string()); // document.querySelector(...).scrollIntoView({behavior: 'smooth'})
    res
```

### Workaround 1
```rust
    static JsScroll: &str = r#"
      setTimeout(function(){
      document.querySelector(`[data-sourcepos='138:1-138:14']`).scrollIntoView({behavior: 'smooth'})
      }, 100);
    "#;

    let eval = dioxus_desktop::use_eval(cx);
    let res = cx.render(rsx! {
        div {
            dangerous_inner_html: "{html}"
        }
    });
    eval(JsScroll.to_string());
    res
```
 
### Workaround 2 
I use `use_effect` where I "update" Component with the same value, it triggers "Component function" again and scrolling (eval call) works a second time as expected  

```rust
static MARKDOWN_CONTENT: Atom<String> = |_| "".to_string();
static JsScroll2: &str = r#"
document.querySelector(`[data-sourcepos='138:1-138:14']`).scrollIntoView({behavior: 'smooth'})
"#;
#[inline_props]
pub fn Markdown(cx: Scope<'a>) -> Element {
    let con = use_read(cx, MARKDOWN_CONTENT);
    let setContent = use_set(cx, MARKDOWN_CONTENT).clone();

    use_effect(cx, con, |con| async move {
        setContent(con);
    });

    let parser = &mut markdown_it::MarkdownIt::new();

    let html = parser.parse(con).render();
    let eval = dioxus_desktop::use_eval(cx);
    let res = cx.render(rsx! {
        div {
            dangerous_inner_html: "{html}"
        }
    });
    eval(JsScroll2.to_string());
    res
}
```

Demonthos:
resolves #631
resolves #517
resolves #804
resolves #423 

Implements a new onmounted event with a trait object in the HTML crate. (More API discussion in #631)

The element allows:
- The raw element type of the platform
- The position of the element
- Scrolling to the element
- Focusing the element
Each of these APIs are optional for the underlying renderer to implement.

Todo:
- [x] Create new event in html crate
- [x] Platform specific implementations
  - [x] Web
  - [x] Desktop
  - [x] Liveview
  - [x] TUI
- [x] Add example of onmounted event


yusefnapora:
This looks so cool! I've been writing components that render svg shapes and have been using polling hacks to query JS for the the rendered size of the SVG element so I can calculate scaling factors, etc. Can't wait to toss that code out 😄 This seems like it will open up quite a few new possibilities.

jkelleyrtp:
Awesome! Thank you! @Demonthos

zoechi:
A bit late, but should the event name not be `onmount`?
All event names in the browser are in this form `onmousemove` `onmouseenter`, `onkeypress`, ...
I'm not aware of one event name in past tense.

zoechi:
> Hooks are dropped when the component is dropped. 

Right, I remember seeing a comment in Discord where you mentioned that :facepalm: 

>  I can't think of a use case that would use the element.

Right, the information I would be interested in is available in the components state anyway.

PR #895, opened 2023-03-20
Allow raw elements if they include a dash

### LINKED ISSUE #893: Support custom elements
## Specific Demand

Custom elements are part of [Web Component](https://developer.mozilla.org/en-US/docs/Web/Web_Components) spec. They are used by popular libraries. For example, [Material Web](https://github.com/material-components/material-web/) to provide framework agnostic implementations of their design.

## Current Behavior 

Not only are custom elements not supported, trying to use them results in a very unhelpful error message

```rust
// src/main.rs
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    // launch the web app
    dioxus_web::launch(App);
}

    cx.render(rsx! {
        div {
            "Hello, world!"
            mat-button {
                "Click me"
            }   
        }
    })
}
```

```
>> cargo build --target wasm32-unknown-unknown
   Compiling dioxus-demo v0.1.0 (/home/hamza/code/dioxus-demo)
error: expected identifier or integer
  --> src/main.rs:14:17
   |
14 |                 "Click me"
   |                 ^^^^^^^^^^

error: could not compile `dioxus-demo` due to previous error
```

Demonthos:
closes #893 

A different idea for quicker custom elements. Allows inline custom elements. This does not allow the typesafety of normal elements. It is very similar to raw attribute names surrounded by quotes. This is useful for one use components, for libraries and components that are used many times #539 would be a better solution

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        web-component {
            "my-prop": "5%",
        }
    })
}
```

Why not use quotes like custom attributes use:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        "web-component" {
            "my-prop": "5%",
        }
    })
}
```
Currently commas are optional between custom elements and raw expression, so this is currently accepted as the text node "web-component" followed by a raw expression with an error. This is especially problematic when there are no attributes in the raw component:

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        "web-component" {
        }
    })
}
```
Compiles without errors in both the latest release, and the hypothetical version that accepts raw elements surrounded by quotes, but has a completely different meaning.

### DIFF
--- a/examples/web_component.rs
+++ b/examples/web_component.rs
@@ -1,0 +1,15 @@
+
+
+use dioxus::prelude::*;
+
+fn main() {
+    dioxus_desktop::launch(app);
+}
+
+fn app(cx: Scope) -> Element {
+    cx.render(rsx! {
+        web-component {
+            "my-prop": "5%",
+        }
+    })
+}
--- a/examples/web_component.rs
+++ b/examples/web_component.rs
@@ -1,5 +1,3 @@
-
-
 use dioxus::prelude::*;
 
 fn main() {
--- a/packages/rsx-rosetta/src/lib.rs
+++ b/packages/rsx-rosetta/src/lib.rs
@@ -1,6 +1,6 @@
 use convert_case::{Case, Casing};
 use dioxus_rsx::{
-    BodyNode, CallBody, Component, Element, ElementAttr, ElementAttrNamed, IfmtInput,
+    BodyNode, CallBody, Component, Element, ElementAttr, ElementAttrNamed, ElementName, IfmtInput,
 };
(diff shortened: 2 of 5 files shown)


jkelleyrtp:
Thought about this for a while and I think this is good. Makes it really easy to wrap webcomponents. I'm a little cautious since this conflicts with expression parsing, but it's very unlikely anyone will try to do inline math.

PR #898, opened 2023-03-20
Make web and desktop use_eval compatible

Demonthos:
Makes the web and desktop versions of use_eval both only accept Strings and return a Rc<dyn Fn(String) -> EvalReturn>. Currently the web version of use_eval only returns a borrowed version of the closure which makes it difficult/impossible to use in futures


jkelleyrtp:
Thanks! Rolling this into #1131

PR #907, opened 2023-03-23
Allow paths in format expressions

### LINKED ISSUE #901: `rsx!{ "{foo::BAR}" }` fails to compile: "expected `}` in format string"
**Problem**

`rsx!` doesn't seem to support paths: `rsx!{ "{foo::BAR}" }` 

**Steps To Reproduce**

Add the following to a Dioxus app:

```rust
mod foo {
    const BAR: &str = "baz";
}

rsx! { "{foo::BAR}" };

```

**Expected behavior**

The code compiles.

Instead, fails with:

```
error: invalid format string: expected `'}'`, found `':'`
  --> src/main.rs:82:5
   |
20 |     rsx! { "{foo::BAR}" };
   |     ^^^^^^^^^^^^^^^^^^^^^ expected `}` in format string
   |
   = note: if you intended to print `{`, you can escape it using `{{`
```

**Environment:**
 - Dioxus version: [My fork based on master](https://github.com/DioxusLabs/dioxus/pull/762) (updates after Feb 14 are missing)
 - Rust version: 1.68.0
 - OS info: Linux
 - App platform: LiveView

**Questionnaire**

Demonthos:
Makes inline formatting slightly more flexable by allowing paths.

closes #901

### DIFF
--- a/packages/rsx/src/ifmt.rs
+++ b/packages/rsx/src/ifmt.rs
@@ -59,6 +59,11 @@ impl FromStr for IfmtInput {
                 let mut current_captured = String::new();
                 while let Some(c) = chars.next() {
                     if c == ':' {
+                        // two :s in a row is a path, not a format arg
+                        if chars.next_if(|c| *c == ':').is_some() {
+                            current_captured.push_str("::");
+                            continue;
+                        }
                         let mut current_format_args = String::new();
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Hmmmm. Hopefully this will be forward compatible.

PR #730, opened 2023-01-08
Simplify native core state with ECS and make native-core work with other frameworks

Demonthos:
This PR reworks the state system of native-core centered around an ECS model.

It also resurrects the Rink crate as the core framework agnostic crate backing Dioxus TUI

Goals:
- Make resolving state parallel
- Unify child/parent/node dependent state
- Allow more complex interactions between states (make https://github.com/DioxusLabs/blitz/pull/19 possible)
- Provide a nicer API for libraries outside of Dioxus to use renderers built with native-core

Todo:
- [x] Make resolving state parallel
- [x] Unify child/parent/node dependent state
- [x] Create node-mutation listener API for extensions that track specific nodes (to make focus system and widgets available without Dioxus)
- [x] Update Docs 
  - [x] Add examples for non-Dioxus frameworks to TUI
  - [x] Update native-core crate docs
  - [x] Update custom renderer guide docs
- [x] Update TUI renderer
- [x] Update Blitz renderer (https://github.com/DioxusLabs/blitz/pull/22)
- [x] Make Dioxus dependency optional
~~- [ ] Resolve issues around optional dependencies ([discussion](https://discord.com/channels/899851952891002890/954257659597553664/1074100965919182930))~~ The API should allow for this as a non-breaking performance improvement in the future

PR #913, opened 2023-03-28
Avoid duplicating dirty nodes in native core

Demonthos:
Fixes duplicate nodes in the dirty nodes struct causing passes to rerun more than once per node. This fixes CI in main branch introduced in #730.


Demonthos:
I'm not sure what is causing CI to fail here. The tests pass on my local msvc-windows machine. This could be a caching issue?

PR #915, opened 2023-03-28
Don't add no op mutations when diffing lists

Demonthos:
When diffing lists, the Virtual Dom will sometimes insert mutations that do nothing like appending no children or inserting no nodes before or after a node. This checks for those cases and only inserts mutations when needed


jkelleyrtp:
Looks like the behavior of some tests changed.

PR #926, opened 2023-04-02
Prevent default redirect on web forms

### LINKED ISSUE #924: onsubmit example doesn't behave as expected.
**Problem**

When I use the code in https://dioxuslabs.com/docs/0.3/guide/en/interactivity/user_input.html#uncontrolled-inputs, I do not see the expected output.

**Steps To Reproduce**

```
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

    cx.render(rsx!(
        body {
            form {
                onsubmit: move |event| {
                    println!("Submitted! {event:?}")
                },
                input { name: "name", },
                input { name: "age", },
                input { name: "date", },
                input { r#type: "submit", },
            }

        }
    ))
}
```

`dioxus serve`

Press "Submit Query".

**Expected behavior**

`Submitted! UiEvent { data: FormData { value: "", values: {"age": "very old", "date": "1966", "name": "Fred"} } }`

**Observed behavior**

Immediate HTTP GET request.  I don't see any evidence of the onsubmit action.

**Environment:**
 - Dioxus version: v0.3.0
 - Rust version: 1.68.1
 - OS info: Linux
 - App platform: web

**Questionnaire**

Demonthos:
Prevent the default redirect behavior of forms in the web renderer to match the desktop renderer, and documentation (the [guide](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/user_input.html#uncontrolled-inputs) and [examples](https://github.com/DioxusLabs/dioxus/blob/adade2d5c54c7c44c3ab7ad2632b0ed678a7dbd6/examples/form.rs#L4))

Fixes #924


kylerlaird:
Does this allow the option of standard form submission or not (by returning false in JavaScript)?
https://www.tutorialspoint.com/How-to-stop-form-submission-using-JavaScript

Demonthos:
> Does this allow the option of standard form submission or not (by returning false in JavaScript)? https://www.tutorialspoint.com/How-to-stop-form-submission-using-JavaScript

Adding prevent default will now make the form submit. In the original example, if you wanted the form to redirect you could use:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

    cx.render(rsx!(
        body {
            form {
                prevent_default: "onsubmit",
                onsubmit: move |event| {
                    println!("Submitted! {event:?}")
                },
                input { name: "name", },
                input { name: "age", },
                input { name: "date", },
                input { r#type: "submit", },
            }

        }
    ))
}
```

Demonthos:
> Adding "prevent_default" will now make the form submit (which is the default behavior)? That seems backward to me.

I agree making the default different from JavaScript could be unexpected, but it could be necessary in order to make rendering consistent between different platforms.

Dioxus supports two main renderers right now: Web and Desktop. For the web renderer, making redirecting the default behavior might make sense, but for the desktop renderer redirecting out of the application into the browser does not make sense. In other renderers, like the [Blitz](https://github.com/DioxusLabs/blitz) and [TUI](https://github.com/DioxusLabs/dioxus/blob/master/packages/dioxus-tui/README.md) renderers, there is no concept of redirecting at all so the default form behavior makes even less sense.

> Is the onsubmit handler able to specify whether or not the form submits normally after the handler completes (such as returning false in JavaScript)?

There is currently no way to conditionally prevent default on events. If we added this feature it would still leave issues with the default form behavior on web breaking the desktop renderer.

I'm open to other ideas, but I think a solution that is cross platform by default should be preferred.

jkelleyrtp:
I'd rather not override default behavior for web. I think it makes sense to do it for desktop/mobile since the experience would be completely jarring, but for web we want to match whatever the standard is. What would make more sense is to automatically prevent default if the event handler is present since... why would you want to handle the event if the page is redirected?

PR #928, opened 2023-04-02
Add more features to todomvc

pickfire:
- toggle-all button
- double click modify todo item
- filter state show mouse pointer
- individual todo item remove button
- correct active item count

Screenshot_20230402_102942


Demonthos (approved):
Looks good. Thank you! We can further improve this example once #894 lands by auto focusing the input after receiving a double click event

PR #931, opened 2023-04-02
doc: fermi: add counter example to readme

### LINKED ISSUE #660: Website globalstate.rs fermi example use_state instead of use_atom
**Problem**

On https://dioxuslabs.com/, there are the codesnippets early on the page, one of them is `globalstate.rs` with the content

```
//! Use Fermi to easily manage global state with a simple Atom-based API

// Define an atom of state
static COUNT: Atom<i32> = |_| 0;

// Read it anywhere
fn Demo(cx: Scope) -> Element {
    let count = use_read(&cx, || COUNT);

    render!("Count: {count}")
}

// Or write to it from anywhere
fn Increment(cx: Scope) -> Element {
    let mut count = use_atom(&cx, COUNT);

    render!(button { onclick: move |_| count += 1, "Increment" } )
}

```

The line `let mut count = use_atom(&cx, COUNT);` has to be changed to either `use_set` or `use_atom_state` (I believe) to compile. It's also this way on the [GitHub Readme of fermi](https://github.com/DioxusLabs/dioxus/tree/master/packages/fermi).

**Questionnaire**

## Edit

I just noticed it's also documented with `use_atom` [here](https://dioxuslabs.com/guide/state/fermi.html). Also, documentation is not consistent about how to use it. on crates.io, the README and the guide all have different instrucitons:

[crates.io:](https://crates.io/crates/fermi)
```
[depdencies]
fermi = { git = "https://github.com/dioxuslabs/fermi" }
```

[README](https://github.com/DioxusLabs/dioxus/tree/master/packages/fermi)
```
[dependencies]
fermi = { git = "https://github.com/dioxuslabs/dioxus" }
```

[Guide](https://dioxuslabs.com/guide/state/fermi.html)
```
[dependencies]
dioxus = { version = "0.2", features = ["desktop", "fermi"] }
```

arqalite:
Partially closes #660, the other half is in DioxusLabs/docsite#59.

The Fermi snippet on the website shows an increment/counter example that doesn't work. The reporter also mentioned the Fermi readme had the same issue, but it's been fixed with the 0.3 release I think.

However the examples in the readme only show reading, and overwriting the value, but a newcomer wouldn't be able to figure out the common scenario of updating a value based on itself (i.e. counters). I added this example:

> If needed, we can update the atom's value, based on itself:

```rust, ignore
static COUNT: Atom<i32> = |_| 0;

fn Counter(cx: Scope) -> Element {
    let mut count = use_atom_state(cx, COUNT);

    cx.render(rsx!{
        p {
          "{count}"
        }
        button {
            onclick: move |_| count += 1,
            "Increment counter"
        }
    })
}
```

### DIFF
--- a/packages/fermi/README.md
+++ b/packages/fermi/README.md
@@ -44,7 +44,7 @@ From anywhere in our app, we can read the value of our atom:
 
 ```rust, ignores
 fn NameCard(cx: Scope) -> Element {
-    let name = use_atom_state(cx, NAME);
+    let name = use_read(cx, NAME);
     cx.render(rsx!{ h1 { "Hello, {name}"} })
 }
 ```
--- a/packages/fermi/README.md
+++ b/packages/fermi/README.md
@@ -44,7 +44,7 @@ From anywhere in our app, we can read the value of our atom:
 
 ```rust, ignores
 fn NameCard(cx: Scope) -> Element {
-    let name = use_read(cx, NAME);
+    let name = use_atom_state(cx, NAME);
     cx.render(rsx!{ h1 { "Hello, {name}"} })
 }
 ```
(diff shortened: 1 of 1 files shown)


Demonthos (approved):
Thank You! This is a very common situation and a great addition to the readme. We should probably also add something about AtomRef for non-copy types and document why use_set should be preferred over use_atom_state to the readme. use_set does not subscribe the component to update when the atom changes, but use_atom_state does (because you can not read the value from set, but can from use_atom_state)

PR #935, opened 2023-04-05
Hotfix wry segfaulting by commenting out webcontext

jkelleyrtp:
Seems to be an issue in wry where passing the data directory pointer doesn't follow proper lifetime rules.

This just comments out the webcontext code. We should find a long term solution - like leaking the webcontext.


Demonthos (approved):
This breaks bundling on windows, but if it fixes the seg fault that is fine for now

PR #936, opened 2023-04-06
Add web context back and ensure it isn't dropped prematurely.

ProfXwing:
When wry is building an innerwebview, it will either push the custom protocols to its own struct field, or it will add it to the web context. The problem comes when web context is dropped prematurely. Then, the obj c trickery wry uses doesn't know how to handle the protocols and segfaults. 

Wry is actually *supposed* to have some logic (in the attached image) to not handle the protocol if web context is dropped, but it seems like they haven't quite handled it correctly. Maybe they should unset the ivar when a WebContext is dropped? I'm not an expert on unsafe or objc so idk lol.

This pr just stores the web context in a WebViewHandler, but there's probably a better way to do it.

Also, I went ahead and updated wry to the latest stable release to make sure nothing else fishy is going on.

### DIFF
--- a/packages/desktop/Cargo.toml
+++ b/packages/desktop/Cargo.toml
@@ -21,7 +21,7 @@ serde = "1.0.136"
 serde_json = "1.0.79"
 thiserror = "1.0.30"
 log = "0.4.14"
-wry = { version = "0.23.4" }
+wry = { version = "0.27.2" }
 futures-channel = "0.3.21"
 tokio = { version = "1.16.1", features = [
     "sync",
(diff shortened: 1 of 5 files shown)


jkelleyrtp:
Can we Box and Pin the webcontext so 1) the pointer is stable in the heap and 2) the webcontext can't be moved?

jkelleyrtp:
Also, it looks like the bump in wry changes which packages need to be pre-installed.

jkelleyrtp:
Thanks!

PR #939, opened 2023-04-06
Disable browser shortcut keys on windows

### LINKED ISSUE #937: Reloading page breaks desktop app
**Problem**

Pressing `F5` or `Ctrl+r` reloads the app's page and the HTML content doesn't seem to be re-rendered so there is only a blank white page.

I noticed that there was a [previous issue](https://github.com/DioxusLabs/dioxus/issues/75) that disabled the reload context menu when the app was built in release mode.

**Steps To Reproduce**

- Create a simple desktop app as specified by the [guide](https://dioxuslabs.com/docs/0.3/guide/en/getting_started/desktop.html)
- Build it in release mode or just `cargo run` works as well.
- Press `F5` while the opened app is in focus.

**Expected behavior**

The app should remain functional, either not reloading or even better actually re-rendering the content.

**Environment:**
 - Dioxus version: v0.3.2
 - Rust version: 1.68.2
 - OS info: Windows 10
 - App platform: `desktop`

**Questionnaire**

Demonthos:
This enables a webview setting in the newest version of wry that disables browser specific shortcuts. This setting only works for windows shortcuts like CTRL+R and F5 that break Dioxus applications.

fixes #937

### DIFF
--- a/examples/counter.rs
+++ b/examples/counter.rs
@@ -1,0 +1,36 @@
+//! Comparison example with leptos' counter example
+//! https://github.com/leptos-rs/leptos/blob/main/examples/counters/src/lib.rs
+
+use dioxus::prelude::*;
+
+fn main() {
+    dioxus_desktop::launch(app);
+}
+
+fn app(cx: Scope) -> Element {
+    let counters = use_state(cx, || vec![0, 0, 0]);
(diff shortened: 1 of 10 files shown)


jkelleyrtp:
Thank you!

PR #940, opened 2023-04-06
Make TUI widgets cross framework

Demonthos:
This changes widgets from Dioxus specific Components to framework agnostic elements. Once this work is finished most of these widgets can be reused in [blitz](https://github.com/DioxusLabs/blitz) with some minor modifications.

In order to allow this we need to roughly implement the [shadow dom](https://javascript.info/shadow-dom) that allows browsers to render widget elements. This PR will expose a API similar to the web components to create widgets.

The implementation of a shadow DOM in native-core:
- Elements have a optional shadow root property that points to a shadow DOM root (in the same world). The shadow root can then have a single element where children in the light DOM are inserted
- Some states will need to traverse through the shadow DOMs like the Layout state, but other need to skip over the shadow DOMs like the style states. We can have a `const ISOLATE_SHADOW_DOM: bool` property in the state struct to control this.


Demonthos:
Input widgets are now working without Dioxus and subtrees are isolated, but there are some caveats with the current implementation of widgets. Currently the widgets do modify the light DOM node to mimic the attributes browsers apply to the input element in the [default stylesheet](https://trac.webkit.org/browser/trunk/Source/WebCore/css/html.css). This could be unexpected when other renderers try to read those attributes. In the future we need to implement selectors and a default style sheet, then we can remove the widget rendering side effects.

These widgets do allow modifying arbitrary attributes which makes them more flexible than the old widget system.

jkelleyrtp:
Do you think we can get the merge conflicts worked out here?

PR #945, opened 2023-04-08
make new_window return desktop context instead of webview

ProfXwing:
It seems like DesktopContext only contains fields related to the new window created anyway. This makes it easier to reference a new window from the main one. It's especially useful for creating events in the new window using a closure in the main window! 

This just makes create_new_window return a tuple with the WebViewHandler and DesktopContext. Perhaps DesktopContext should hold the WebViewHandler, or another struct should be added?


packages/desktop/src/desktop_context.rs
```diff
@@ -133,11 +132,9 @@ impl DesktopContext {
             .send_event(UserWindowEvent(EventData::Poll, id))
             .unwrap();
 
-        let webview = window.webview.clone();
-
         self.pending_windows.borrow_mut().push(window);
 
-        Rc::downgrade(&webview)
+        desktop_context
```
Demonthos, line 137: If we return a strong reference to the Rc here it prevents the child window from closing before the parent window closes.

Try running this example and closing the popup:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    let window = dioxus_desktop::use_window(cx);
    let windows = use_ref(cx, || vec![window.clone()]);

    cx.render(rsx! {
        div {
            button {
                onclick: move |_| {
                    let dom = VirtualDom::new(popup);
                    windows.write().push(window.new_window(dom, Default::default()));
                },
                "New Window"
            }
        }
    })
}

fn popup(cx: Scope) -> Element {
    cx.render(rsx! {
        div { "This is a popup!" }
    })
}
```

We could remove the interior `Rc`s in DesktopContext and wrap the entire context in an Rc instead. We could then return a Weak<DesktopContext> when creating a window and provide Rc<DesktopContext> in the context.

Demonthos (changes requested):

Demonthos:
For context, this was discussed on discord to make registering a wry event handler or keyboard shortcuts that are scoped to a specific window for a child window from a parent window much easier. For example if the main window had a shortcuts settings menu and a popup, the main window needs to control the popup window's shortcuts

packages/desktop/src/desktop_context.rs
ProfXwing: Is this good? I wasn't quite sure what interior `Rc`s you were referring to, but DesktopContext should always be wrapped in an `Rc` now. I'll note that your example will still break if you call `.upgrade` on the `Weak<DesktopContext>`, just as it did with the `Weak<Webview>`. Hopefully it's more obvious to the user that you shouldn't keep that reference now though!

Demonthos: Much better, I should have been more clear in my original comment. By the interior `Rc`s I meant the `Rc<WebView>` inside of the desktop context. That used to be necessary to make the desktop context `Clone`, but now that the whole context is wrapped in an `Rc`, we don't need it to implement `Clone`

We could also rename desktop context to desktop service and make desktop context a type definition for Rc<desktopService> so that this change is less breaking. This is similar to what [RouterContext](https://docs.rs/dioxus-router/latest/dioxus_router/type.RouterContext.html) does.

ProfXwing: Gotcha. I removed the `Rc` around webview. It's worth noting that the Rc<Webview> was owned by both the DesktopContext and WebviewHandler. My solution was to change the WebviewHandler to hold a DesktopContext. Another option I saw was to change WebviewHandler to hold a reference to the webview, but that would involve a lot of messy lifetimes. Let me know if there's a better way to handle that. 

I did go ahead and rename DesktopContext to DesktopService and add a type for DesktopContext to avoid verbose Rc<_> types.

Demonthos (approved):
Much better, Thank You! The window API will change a lot when subtrees are introduced but until then this should be slightly more flexible.

(Waiting for a second review before this is merged because this is a breaking change)

jkelleyrtp:
Sorry for the delay! We haven't quite nailed down how to get breaking changes in aside bug fixes. This is mergable now that we're finally rolling onto 0.4-pre.

Thanks so much!

PR #946, opened 2023-04-08
Fix event bubbling

### LINKED ISSUE #867: Event propagation fails
**Problem**

Children with onpointerdown or onclick events block propagation of pointer events to their parent.

**Steps To Reproduce**

1) make a parent div with two child divs.
2) assign `onpointermove` and `onpointerdown` events to the parent that log when triggered.
3) assign an `onpointerdown` or `onclick` event to only one of the children that logs when triggered.
4) observe that mouse interaction with the event-free child triggers the parent's event logs as expected.
4) observe that when the mouse is moved across the child with an `onpointerdown` event, the parent's `onpointermove` event is not logged.
5) observe that when the mouse is clicked on the child with the `onpointerdown` event, the parent's `onpointerdown` event is not logged.

**Expected behavior**

moving the mouse across either child div should trigger the parent's `onpointermove` event log. Would be helpful in making draggable divs.

**Environment:**
 - Dioxus version: 3.2
 - Rust version: 1.67.1
 - OS info: WSL2 Ubuntu, running the app with Trunk.
 - App platform: web - 0.3.1

**Questionnaire**

Main.rs
```
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    wasm_logger::init(wasm_logger::Config::new(log::Level::Info));
    dioxus_web::launch(app);
}

    cx.render(rsx!{
        div {
            style: "width: 300px; height: 300px; background-color: blue;",
            onpointermove: move |_| log::info!("pointer moving on container"),
            onpointerdown: move |_| log::info!("pointer down on container"),
            div {
                style: "margin: 10px; width: 100px; height: 100px; background-color: yellow;",
                "no onpointerdown or click event",
            },
            div {
                style: "margin: 10px; width: 100px; height: 100px; background-color: red;",
                onpointerdown: move |_| log::info!("pointer down on CHILD"),
                "onpointerdown event"
            },
        },
        p {
            "Children with onpointerdown or onclick events specified do not allow pointer events to propagate to their parent"
        },
        p {
            "view console for details"
        }
    })
}
```
index.html
```
<!DOCTYPE html>
<html lang="en">
    <head> 
        <meta charset="UTF-8">
        <meta http-equiv="X-UA-Compatible" content="IE=edge">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        
        <title>Pointer Event Bug</title>
    </head>
    <body>
      <div id = "main">
       
      </div>
    </body>
</html>
```
Cargo.toml
```
[dependencies]
dioxus = "0.3.2"
dioxus-web = "0.3.1"

log = "0.4"
wasm-logger = "0.2.0"
```

Demonthos:
Fixes event propagation between components.

1) Changes parent element id to be wrapped in a cell to make it mutable (breaking)
2) We only track nodes that have a id that is mounted in the renderer, but when bubbling events we need to know where to resume bubbling in the parent component. In order to support that we need to keep track of some dynamic nodes that are never mounted like fragment and component (breaking). This PR adds a mounted id to those dynamic nodes, but does set the id in the mutations. Alternatively we could use a separate slab and id system for ids that are only used within the Virtual DOM.

Closes #867


jkelleyrtp (approved):

jkelleyrtp:
Tests still aren't passing :(

Demonthos:
Closing this in favor of #1225

PR #949, opened 2023-04-09
Add timeout count example

pickfire:
https://jakelazaroff.com/words/were-react-hooks-a-mistake/ Although dioxus have its own foot guns like infinite loop but I noticed some parts feels more explicit while some parts are more magical.


Demonthos:
I'm not sure this is distinct enough from the [button example](https://github.com/DioxusLabs/dioxus/blob/9571adea3068bee5f62e2229004cb873e8dcabaa/examples/button.rs) (with futures triggered by a button) and the [readme example](https://github.com/DioxusLabs/dioxus/blob/9571adea3068bee5f62e2229004cb873e8dcabaa/examples/readme.rs) (with a button that increments a count) to warrant a separate example.

It could see it being beneficial to add a note in the guide about Gloo for more ergonomic JavaScript API bindings and [async-std](https://docs.rs/async-std/latest/async_std/) as an alternative to Tokio on WASM.

pickfire:
> I'm not sure this is distinct enough from the [button example](https://github.com/DioxusLabs/dioxus/blob/9571adea3068bee5f62e2229004cb873e8dcabaa/examples/button.rs) (with futures triggered by a button)

It's not just about triggering a future, it also needs to print the value from the counter which is incremented from the button, inspired by https://jakelazaroff.com/words/were-react-hooks-a-mistake/, I saw on dioxus blog that dioxus also takes in react issues, but I am surprised I don't get the same foot gun here.

jkelleyrtp:
Thanks for the example!

PR #954, opened 2023-04-11
Fix Linux Wry dependencies

Demonthos:
When wry was updated the required version of libwebkit2gtk was updated. This bumps the version in CI and docs to hopefully fix the Linux CI


jkelleyrtp:
Thank you!

PR #960, opened 2023-04-15
fix: Update logos and custom assets example

marc2332:
- Updated the default app window icon of `dioxus-desktop`
- Updated the  custom assets example dioxus logo
- Removed old code from the custom assets example

image


Demonthos (approved):
Thank You

Issue #595, opened 2022-10-17
idiomatic way of uploading files


neevek:
Hi,
I am using Dioxus to write a simple file server, which is designed to support uploading multiple files at once, I tried to handle the upload action with a normal `onsubmit` event handler, when I select multiple files, `event.values` contains only once file with a weird file path, which looks like a file path on Windows, but I am testing on a Mac:
```
{"file": "C:\\fakepath\\testfile.txt"}
```

I have the following code for the form:
```rust
form {
    prevent_default: "onsubmit",
    onsubmit: handle_upload_files,
    method: "post",
    enctype: "multipart/form-data",
    input {
        r#type: "file",
        name: "file",
        multiple: "true",
    }
    input {
        r#type: "submit",
        value: "Upload"
    }
}
```

And how do I upload files with that file path?

santokalayil:
Same issue here.. Help us to find a solution.

Issue #426, opened 2022-05-26
Add support for native system dialog


st-little:
## Specific Demand

Add support for native system dialog such as [electron dialog](https://www.electronjs.org/docs/latest/api/dialog)

rMazeiks:
Agreed, this would be a nice feature!

In the meantime, consider workarounds:

- Using a library for these dialogs. [native-dialog](https://lib.rs/crates/native-dialog) sounds like what you're looking for
- For files, `<input type="file">` might do what you need

mrxiaozhuox:
> I recommend using NFD directly in handlers - we will be integrating NFD into file input elements in a future release.

use `NFD` replace `<input type="file">` or just add a file-dialog component ?

jkelleyrtp:
> > I recommend using NFD directly in handlers - we will be integrating NFD into file input elements in a future release.
> 
> use `NFD` replace `<input type="file">` or just add a file-dialog component ?

Our `input` element used in desktop will need to support onlick -> NFD (which it doesn't) and register a background file-handler with wry for drag-and-drop onto it.

jkelleyrtp:
In #692 , we finally run the VirtualDom on the main thread, unlocking this capability.

jkelleyrtp:
This has been somewhat added, but we recommend deferring to RFD directly.

PR #965, opened 2023-04-18
Implement file engine on each platform

Demonthos:
The form event has a file engine property, but it is unpopulated on all platforms. This PR implements the file engine for each platform.

closes https://github.com/DioxusLabs/dioxus/issues/595
related to https://github.com/DioxusLabs/dioxus/issues/426
stacked on #938 

Todo:
- [x] Web implementation
- [x] Desktop implementation (with RFD)
- [x] Liveview implementation
- [x] Native implementation (This PR does not create file input widgets for TUI/Blitz, but the file engine can be used for those platforms once the widget is implemented)


Demonthos:
#894 introduces some utilities for queries on desktop and liveview that will be useful for the implementation on those platforms. Pausing work on this until that PR is merged or we pull out the utilities from that PR

Demonthos:
File events are working on all platoforms with the following caviats:
1) Liveview file uploads serialize the binary data into text. This would be much more efficient with a binary encoding, but would require a breaking change to the adapters, so it is not implemented in this PR
2) Desktop file uploads use the [RFD](https://docs.rs/rfd/latest/rfd/) library to get the real path of files uploaded. This works great for input elements, but there is no way to set the file in Javascript to make form submissions with file uploads work on desktop. We could use the same mechanism we use for Liveview for file uploads, but we would not be able to get the full system path of the uploaded files.

jkelleyrtp:
Thank you <3

PR #973, opened 2023-04-23
added select element example

Icekey:
Added a example for the select element. 
1. The "multiple" property seems bugged. The FormEvent returns "FormEvent { value: "Option 1", values: {} }" even when multiple options are selected.
2. The "selected" property does nothing and i could not find a way to preselect one specific option.


Demonthos:
1. See https://github.com/DioxusLabs/dioxus/pull/938
2. See #974

Icekey:
Did the syntax of "onclick" change? I tried to update my example but the "onclick" event is never executed.

jkelleyrtp:
Thank you!!

Icekey:
I think the multi selection of options is still not working right. When you have multiple options selected the emitted FormEvent only contains the first selected option in the "value" field. I would have expected that the "values" field gets filled in that case. Maybe it would be better if FormData got changed to an enum so that single and multi value events have a dedicated type Instead of just leaving the unrelated field empty.

PR #976, opened 2023-04-23
Fix links in Liveview 

Demonthos:
In the desktop renderer, click events on `<a>` links are intercepted and forwarded as an event to the desktop renderer so that they can be opened in a new window of the system browser. When the LiveView renderer was created this behavior was copied.

This PR makes links open normally with the LiveView renderer instead of redirecting the event to the server.


jkelleyrtp:
Sorry this has been sitting here so long - can we fix the merge conflict?

PR #981, opened 2023-04-24
Update crossterm version

Demonthos:
Bumps the crossterm version which has new support for more terminals and key codes.


jkelleyrtp:
Thanks! Rolling this into #1131

PR #983, opened 2023-04-24
fix: change crossterm poll timeout to 10ms from 1s

### LINKED ISSUE #978: `crossterm::event::poll()` causing delayed key inputs in Dioxus TUI
**Problem**

When using text input widgets, key events register with a noticeable delay. Moving the mouse causes buffered / backlogged key events to flush, but typing without mouse movement is extremely slow.

- Clone the dioxus repo
- Run the `widgets` example in the `dioxus_tui` package 
- Enter the text or number input widget
- Type

**Expected behavior**

Typed characters should register immediately, without noticeable delay.

**Environment:**
 - Dioxus version: master (459d8d6)
 - Rust version: 1.69 stable
 - OS info: Linux
 - App platform: TUI

Mix of these two:
- [X] I would like to fix and I have a solution (Idea, but not solution)

I decreased the poll timeout from 1000ms to 10ms on line 101 of packages/rink/lib.rs for testing, and the key delays were significantly improved. There wasn't a noticable increase in CPU usage (in htop, at least), but I'm assuming the 1000ms timeout was implemented for performance-related reasons, so I'm not sure whether or not changing that would have significant drawbacks.

Happy to contribute a PR if wanted :)

willothy:
Problem: Key events were delayed signifantly in TUI widgets.
Solution: Decrease crossterm event poll timeout from 1000ms to 10ms.

Fixes #978

### DIFF
--- a/packages/rink/src/lib.rs
+++ b/packages/rink/src/lib.rs
@@ -98,7 +98,8 @@ pub fn render<R: Driver>(
     let event_tx_clone = event_tx.clone();
     if !cfg.headless {
         std::thread::spawn(move || {
-            let tick_rate = Duration::from_millis(1000);
+            // Timeout after 10ms when waiting for events
+            let tick_rate = Duration::from_millis(10);
             loop {
                 if crossterm::event::poll(tick_rate).unwrap() {
                     let evt = crossterm::event::read().unwrap();


Demonthos (approved):
Thank You! (CI error is unrelated to the changes in this PR)

PR #984, opened 2023-04-24
Fix HTML to RSX conversions example

### LINKED ISSUE #943: `dioxus translate` panics regardless of input.
**Problem**

running `dioxus translate` panics with the following error:

```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/dioxus-autofmt-0.3.0/src/element.rs:279:25
```

**Steps To Reproduce**

- run `dioxus translate` on an html file or selection (in my test case: `dioxus translate -f src/test.html`)

**Expected behavior**

There should be output for a converted RSX translation.

**Screenshots**

Screenshot from 2023-04-07 10-36-41

**Environment:**
 - Dioxus version: 0.3.0
 - Rust version: 1.68.2
 - OS info: Ubuntu 22.04 (Pop_OS! 22.04)
 - App platform: web

**Questionnaire**

Demonthos:
The autoformat package assumes that the source is non-empty, but `write_block_out` does not supply a source. This fixes the code to allow an empty source string. It also adds some tests to check the behavior of rsx-rosetta

fixes #943

### DIFF
--- a/packages/autofmt/src/element.rs
+++ b/packages/autofmt/src/element.rs
@@ -276,13 +276,12 @@ impl Writer<'_> {
         let start = location.start();
         let line_start = start.line - 1;
 
-        let this_line = self.src[line_start];
-
-        let beginning = if this_line.len() > start.column {
-            this_line[..start.column].trim()
-        } else {
-            ""
-        };
+        let beginning = self
+            .src
+            .get(line_start)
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
Thank you! <3 <3

PR #987, opened 2023-04-25
Allow multiple prevent default for Web, Desktop and LiveView

### LINKED ISSUE #535: Issues with multiple prevent_default attributes
**Problem**

As per the docs/examples it seems to be possible to prevent multiple defaults in the same element, but it is only working for the last one defined.

**Steps To Reproduce**

- Run and open in the browser the following app (see below)
- Drag and drop a file

```rust
use dioxus::prelude::*;

fn main() {
  tracing_wasm::set_as_global_default();
  dioxus_web::launch(app);
}

  cx.render(rsx!(
    div {
      ondrop: |_| tracing::info!("On drop"),
      ondragover: |_| tracing::info!("On drag over"),
      prevent_default: "ondrop",
      prevent_default: "ondragover",
      style: "border: 1px solid black; width: 300px; height: 200px;",
      "Drag and drop a file"
    }
  ))
}
```

**Expected behavior**

It should prevent the browser from opening the file and just show the logs, but it is instead opening the file, tested in Chrome and Safari.

**Environment:**
 - Dioxus version: `master` -> 8347de59565d39a275959accd441c27bc460c8d1
 - Rust version: 1.63.0
 - OS info: MacOS
 - App platform: `web`

**Questionnaire**

attilio-oliva:
At the moment, the DOM interpreter expects only one event to prevent the default behavior. This pr allows defining multiple events separated by spaces and prevent all of them.
This is the most straightforward way to do it, but maybe the code should be reviewed to be simplified.
Fixes  #535

### DIFF
--- a/docs/guide/examples/event_prevent_default.rs
+++ b/docs/guide/examples/event_prevent_default.rs
@@ -10,8 +10,7 @@ fn App(cx: Scope) -> Element {
     // ANCHOR: prevent_default
 cx.render(rsx! {
     input {
-        prevent_default: "oninput",
-        prevent_default: "onclick",
+        prevent_default: "oninput onclick",
     }
 })
     // ANCHOR_END: prevent_default
--- a/docs/guide/src/en/interactivity/event_handlers.md
+++ b/docs/guide/src/en/interactivity/event_handlers.md
@@ -41,7 +41,7 @@ If you want to prevent this behavior, you can call `stop_propogation()` on the e
 
 Some events have a default behavior. For keyboard events, this might be entering the typed character. For mouse events, this might be selecting some text.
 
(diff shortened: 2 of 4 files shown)


Demonthos:
This fixes web events, but doesn't effect other platforms. LiveView and desktop events are handled [here](https://github.com/DioxusLabs/dioxus/blob/master/packages/interpreter/src/interpreter.js#L354). We also need to update the documentation [here](https://github.com/DioxusLabs/dioxus/blob/master/docs/guide/src/en/interactivity/event_handlers.md#prevent-default) to mention separators instead of multiple values

Demonthos (approved):
Thank You

PR #990, opened 2023-04-26
make files optional when deserializing form data

Demonthos:
Fixes form events on desktop broken in #965. (I only noticed this issue when updating other PRs with the changes)

Form events on desktop never contain the file field which causes crashes when deserializing input events. This PR makes the field optional when deserializing which fixes the issue

### DIFF
--- a/packages/html/src/events/form.rs
+++ b/packages/html/src/events/form.rs
@@ -14,7 +14,11 @@ pub struct FormData {
 
     #[cfg_attr(
         feature = "serialize",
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I would really like to get some auto testing setup for all of our platforms at some point. We keep having these little regressions that can't really be caught with our current setup.

Demonthos:
> I would really like to get some auto testing setup for all of our platforms at some point. We keep having these little regressions that can't really be caught with our current setup.

Working on some automated browser tests in #995

Issue #986, opened 2023-04-25
DragEvent should have data about the dragged files, just like FormEvent


attilio-oliva:
FormEvent have API to manage drag and drop files, but DragEvent do not. This prevents from implementing a common drag and drop area like this:
```
div { 
            prevent_default: "ondrop ondragover",
             ondrop: move |event: DragEvent| {
                   event.stop_propagation();
                   // This is how you would do to get all the file names using FormEvent
                   // Maybe not the best naming choice, I would expect something like `files.names()`
                   log!(format!("{:?}", event.files.files());
              },
              ondragover: move |event: DragEvent| {
                    event.stop_propagation();
                }
}
```

Demonthos:
This is blocked on #965. Once that is finished, this should be a simple (but breaking) change

attilio-oliva:
Any update on this issue?
Everything should be in place thanks to the pr #992, but it is still not merged. Is there a specific reason about this choice?

PR #992, opened 2023-04-27
Add file property to drag data

ealmloff:
Closes https://github.com/DioxusLabs/dioxus/issues/986

Adds a files property to the drag event that contains the file engine from FormData.

> Note: The example will not work until #987 is merged


Demonthos:
Note: Both the web and desktop implementation read the file in javascript (unlike the desktop file picker which reads the file in rust)

PR #1020, opened 2023-05-12
Enum Based Router

Demonthos:
closes #509 

This document is a summery of some API ideas suggested in the Dioxus discord and built on top of the fantastic work @TeFiLeDo has done in #637. The functionality of that PR is a massive improvement from the current router, but the API does become more complicated. This PR is an attempt to unify some of the concepts in that PR with the goal of simplifying the API.

Goals:
- Simple API
- Internal links should be type safe and infallible
- Allow layout components to be [run in parallel](https://remix.run/blog/remixing-react-router) (More of an architecture consideration, this will not be implemented in this PR)

# How Routing Works

Routes and Redirects are defined in an enum. Routing priority goes to static segments first, and then follows the order the enum routes were defined

```rust
#[derive(Clone, Debug, PartialEq, Routable)]
enum Route {
    // Each Variant is a route with a linked component, dynamic segments are defined with the syntax: `:name` and the type is inferred from the field type. The type must implement FromStr
    #[route("/:dynamic" Component1)]
    Route1 { dynamic: String },
    // You can nest routes which makes all routes in the block relative to a parent route. Nested routes are flattened into the parent enum
    // Nest accepts a optional layout component. The layout component that wraps all children and renders them where the Outlet component is found. It can accept parameters from the nested route, just like a normal route
    #[nest("/hello_world" layout = Component)]
        // If the component is not specified, the component is assumed to be at the path of the route (in this case /pages/hello_world.rs or /pages/hello_world/index.rs)
        #[route("/")]
        // You can opt out of a parent Layout
        #[!layout(Component)]
        Route2 {},
    #[end_nest]
    
    // Queries are defined with the syntax: `?:name` and the type is inferred from the field type. The type must implement From<&str> (not FromStr because the query parsing must be infallible). The query part of the url is not included in the route path for file based routing. (in this case /pages/takes_query.rs or /pages/takes_query/index.rs)
    #[route("/takes_query?:dynamic")]
    Route3 { dynamic: u32 },
    // Redirects are defined with the redirect attribute
    #[redirect("/old_hello_world/:dynamic")]
    #[route("/hello_world/:dynamic")]
    Route4 { dynamic: u32 },
    // members that can be parsed from all trailing segments are defined with the syntax: `:...name` and the type is inferred from the field type. The type must implement FromSegments.
    // Because this route is defined after Route3, it will only be matched if Route3 does not match and it will act as a fallback
    #[route("/:...segments")]
    Route5 { segments: Vec<String> },
}
```

## Changes from the Previous API
`named` routes: Named routes are replaced by enum Variants, Internal routing is now type safe and infallible by default
`fixed`, `parameter`, and `matching` routes are all combined into a macro. This PR uses the FromStr and FromRoute traits to fallibly parse routes. If the parsing fails the next route is attempted. This also allows us to build up clear error messages about *why* a route did not match:
Going to the route: `1234/-2` with the router:
```rust
#[derive(Clone, Debug, PartialEq, Routable)]
enum Route {
    #[route("/:dynamic")]
    Route1 { dynamic: String },
    #[route("/hello_world")]
    Route2 {},
    #[route("/hello_world/:dynamic")]
    Route3 { dynamic: u32 },
    #[route("/:number1/:number2")]
    Route4 { number1: u32, number2: u32 },
    #[route("/")]
    Route5 {},
}
```
would generate this error message
```
Route did not match:
Attempted Matches:
1) Route 'Route2' ('/hello_world') did not match:
Static segment 'hello_world' did not match
2) Route 'Route5' ('/') did not match:
Static segment '' did not match
3) Route 'Route1' ('/(dynamic)') did not match:
Found additional trailing segments: -2
4) Route 'Route4' ('/(number1)/(number2)') did not match:
Dynamic segment '(number2:u32)' did not match: invalid digit found in string
```

# Links

The Link Component can remain largely unchanged from #637 with the exception that the link passed in can either be the enum, or a &str. If the link is an enum, the link is guarantied to be valid.

# Open Questions

- Working with other frameworks?

The conversion of enum routes to and from a route should be possible to make cross framework, but it is less clear how the rendering of the router can be made platform agnostic because each route takes different parameters unlike #637


TeFiLeDo:
Sorry it took me so long, I finally had a chance to look at this.

I don't have much experience when it comes to writing or reading macros, so please correct me if any of my observations are wrong.

---

I really, really like the idea of internal links being type safe and infallible. I've tried and failed to create a similar system myself and failed, so it's very cool to see how you have accomplished this.

That being said, I have a few observations:

1. As far as I can tell you don't represent the routes as an actual tree, instead flattening them into a list of enum variants. This seems to cause more code duplication. For example would `/:name/profile` and `/:name/feed` (with a nesting block) both need to define the `name` parameter? This might get especially confusing if the types don't match, therefore causing one route to accept different names than the other. Maybe your macro could look out for that.
   
2. It might be a good idea to add a special case for `/`, or maybe a "keyword". I personally find it confusing to have the most basic route at the end, and such a feature would make that unnecessary. This might be confirmation bias, but the existence of the [`exact` feature of React Router 5](https://v5.reactrouter.com/web/api/Route/exact-bool) and the fact that the [Vue Router tutorial has `/` first](https://router.vuejs.org/guide/#javascript) seem to suggest that other people have similar feelings about this.
   
3. I'm confused by `Route5` in your first example. Wouldn't `number2` need to be a `Vec`, or does it only get the first or last segment?
   
4. One idea I had during my aforementioned attempt was to change the history providers. Instead of storing the URL, they could store the actual routing state. That way it could contain additional data not part of the URL. 
   
   I know that one JS router (I don't remember which one) has a feature that allows defining a route containing `../`. The advantage of that is that when sharing the URL the recipient gets a different page, e.g. only the specific content instead of a popup above a feed.
   
   Such a feature could be implemented if the history provider stores the state itself.
   
5. It seems that only the innermost routing component gets compile checked access to the URL parameters. However, it might be useful to access them from a layout component. For example, `/:name/profile` and `/:name/feed` might both share a header with `:name`s name and profile picture. If access to the parameters is only possible from within the `Profile` and `Feed` components, the nesting feature becomes basically useless in such a scenario.

Demonthos:
> Sorry it took me so long, I finally had a chance to look at this.
> 
> I don't have much experience when it comes to writing or reading macros, so please correct me if any of my observations are wrong.

Thank you for taking the time to look at this! I really appreciate your feedback

> I really, really like the idea of internal links being type safe and infallible. I've tried and failed to create a similar system myself and failed, so it's very cool to see how you have accomplished this.
> 
> That being said, I have a few observations:
> 
> 1 & 5:
> As far as I can tell you don't represent the routes as an actual tree, instead flattening them into a list of enum variants. This seems to cause more code duplication. For example would `/:name/profile` and `/:name/feed` (with a nesting block) both need to define the `name` parameter? This might get especially confusing if the types don't match, therefore causing one route to accept different names than the other. Maybe your macro could look out for that.
> It seems that only the innermost routing component gets compile checked access to the URL parameters. However, it might be useful to access them from a layout component. For example, `/:name/profile` and `/:name/feed` might both share a header with `:name`s name and profile picture. If access to the parameters is only possible from within the `Profile` and `Feed` components, the nesting feature becomes basically useless in such a scenario.

I agree we could improve this. What if we automatically add dynamic fields in a nest to all children components and routes like this:

```rust
#[derive(Clone, Debug, PartialEq, Routable)]
enum Route {
    // Nests with parameters require a {(param_name: Type,)*} field that defines the types that all child routes and Components (both layout and non-layout) will receive
    #[nest("user/:user_id" {user_id: String})]
        // Everything inside the nest has the added parameter `user_id: String`
        // UserFrame is a layout component that will receive the `user_id: String` parameter
        #[layout(UserFrame)]
            // Route1 is a non-layout component that will receive the `user_id: String` and `dynamic: String` parameters
            #[route("/:dynamic")]
            Route1 {
                // Automatically adds a `user_id: String` field
                dynamic: String
            },
        #[end_layout]
        // Route2 is a non-layout component that will receive the `user_id: String` parameter
        #[route("/hello_world")]
        Route2 {
            // Automatically adds a `user_id: String` field
        },
    #[end_nest]
}
```

> 2. It might be a good idea to add a special case for `/`, or maybe a "keyword". I personally find it confusing to have the most basic route at the end, and such a feature would make that unnecessary. This might be confirmation bias, but the existence of the [`exact` feature of React Router 5](https://v5.reactrouter.com/web/api/Route/exact-bool) and the fact that the [Vue Router tutorial has `/` first](https://router.vuejs.org/guide/#javascript) seem to suggest that other people have similar feelings about this.

There are two layer of ordering:
1) More specific routes match first (first static routes, then single dynamic routes, then match all routes)
2) When there is a tie, the order in the enum is followed

Because catch all routes are less specific, you can have a root catch all route first in your code but it will only match when all non-catch all routes are not matched.

Eg:
```rust
use dioxus::prelude::*;
use dioxus_router_core::{history::MemoryHistory, *};
use dioxus_router_macro::*;
use router::Routable;
use std::str::FromStr;

fn root(cx: Scope) -> Element {
    let current_route = use_ref(cx, String::new);
    let parsed = Route::from_str(&current_route.read());
    let url = parsed.as_ref().map(|route| route.to_string());

    render! {
        input {
            oninput: |evt| {
                *current_route.write() = evt.value.clone()
            },
            value: "{current_route.read()}"
        }

        Router::<Route, MemoryHistory> {
            current_route: current_route.read().clone(),
        }
        pre {
            "rust:\n{parsed:#?}\n"
            "url:\n{url:#?}"
        }
    }
}

fn main() {
    dioxus_desktop::launch(root);
}

#[inline_props]
fn Route1(cx: Scope, segments: Vec<String>) -> Element {
    render! {
        pre {
            "Route1{{\n\tsegments:{segments:?}\n}}"
        }
    }
}

#[inline_props]
fn Route2(cx: Scope, dynamic: usize) -> Element {
    render! {
        pre {
            "Route2{{\n\tdynamic:{dynamic}\n}}"
        }
    }
}

#[inline_props]
fn Route3(cx: Scope, dynamic: String) -> Element {
    render! {
        pre {
            "Route3{{\n\tdynamic:{dynamic}\n}}"
        }
    }
}

#[rustfmt::skip]
#[routable]
#[derive(Clone, Debug, PartialEq)]
enum Route {
    #[route("/:...segments" Route1)]
    Route1 { segments: Vec<String> },
    #[route("/:dynamic" Route2)]
    Route2 { dynamic: usize },
    #[route("/:dynamic" Route3)]
    Route3 { dynamic: String },
}
```

One of the more specific routes matched (because this is a tie, it tries to match in the order of the enum)
image

Fallback to the second `/:dynamic` route
image

When both of those fail, try to match the catch all route
image

> 3. I'm confused by `Route5` in your first example. Wouldn't `number2` need to be a `Vec`, or does it only get the first or last segment?

Yes, that example should be fixed now

> 4. One idea I had during my aforementioned attempt was to change the history providers. Instead of storing the URL, they could store the actual routing state. That way it could contain additional data not part of the URL.
>    I know that one JS router (I don't remember which one) has a feature that allows defining a route containing `../`. The advantage of that is that when sharing the URL the recipient gets a different page, e.g. only the specific content instead of a popup above a feed.
>    Such a feature could be implemented if the history provider stores the state itself.

Right now I'm thinking of Routable meaning three things:
1) Derserializable from a route (to make the initial route parsable)
2) Serializable to a route (to change the url in the web version)
3) Renderable to a component (to actually display the content of the route)

I have been thinking about this as well. It would reduce the amount of parsing we would need to do and make the in memory router much simpler. My main concern is this would make internal and external navigations different.

If some of the router state is not stored within the URL, there could be a difference between navigating to a page outside of the SPA and then clicking the back button and navigating to a page within the SPA. If you navigate with the browser back button from an external page, you would need to be provided some default parameters for fields that are not encoded in the URL. If you navigate from a internal page, these fields would be filled in by the Dioxus router.

TeFiLeDo:
> I agree we could improve this. What if we automatically add dynamic fields in a nest to all children components and routes like this:
>
>     lots of code

I like that. Couple of questions:

1. I believe `#[nest]` can be nested. Will inner layouts get their parents parameters?
2. Can a `#[nest]` define a query parameter?

> There are two layer of ordering:
>
> 1. More specific routes match first (first static routes, then single dynamic routes, then catch all routes)
> 2. When there is a tie, the order in the enum is followed
>
> Because catch all routes are less specific, you can have a root catch all route first in your code but it will only match when all non-catch all routes are not matched.

Ah ok, thanks for explaining.

> My main concern is this would make internal and external navigations different.

In the browser you cannot get around doing that.

1. You cannot use the history API for external navigation. All you get is an exception.
2. If you don't use the history API for internal navigation you have to deal with all the problems it was created to solve. Among others things that includes the back button not working properly and a lot of unnecessary full page loads.

> If some of the router state is not stored within the URL, there could be a difference between navigating to a page outside of the SPA and then clicking the back button and navigating to a page within the SPA and click ing the back button. If you navigate with the browser back button from an external page, you would need to be provided some default parameters for fields that are not encoded in the URL. If you navigate from a internal page, these fields would be filled in by the Dioxus router.

The history API allows you to store arbitrary state. My web history providers use that for scroll restoration. It could also be used to store the actual state. In fact, the function to change the current location is called [`pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) and the `url` parameter is optional.

I can take care of that part, if you'd like me to.

Demonthos:
> I like that. Couple of questions:
> 
> 1. I believe `#[nest]` can be nested. Will inner layouts get their parents parameters?

Yes

> 2. Can a `#[nest]` define a query parameter?

We could treat query parameters similar to the main route:
```rust
#[nest("/?:query_param")]
```
Where query_param implements Default, Display and From<&str>
It could be encoded in the url as: `query_param=value`

We could allow in the route only one catch all query:
```rust
#[nest("/?:...query_params")]
```
Where query_params implements Default, Into<impl Iterator<(String, String)> and From<HashMap<String, String>>
It could be encoded in the url as: `key=value&key2=value2&...`
It would need to not output any keys that match the keys in the other query parameters

> The history API takes care of that. It allows you to store arbitrary state. My web history providers use that for scroll restoration. It could also be used to store the actual state. In fact, the function to change the current location is called [`pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) and the `url` parameter is optional.
> 
> I can take care of that part, if you'd like me to.

That makes sense. If you can store arbitrary data in the browser's history API, adding extra states that are not included in the patterns in nests and routes can be allowed as long as they implement Default

Demonthos:
I'm not sure that the macro should change the definition of the enum. It could make it difficult to see what the fields of each variant is.

Instead of declaring the type being declared with the nest like this:

```rust
#[derive(Clone, Debug, PartialEq)]
#[routable]
enum Route {
    // Nests with parameters require a {(param_name: Type,)*} field that defines the types that all child routes and Components (both layout and non-layout) will receive
    #[nest("user/:user_id" {user_id: String})]
        // Everything inside the nest has the added parameter `user_id: String`
        // UserFrame is a layout component that will receive the `user_id: String` parameter
        #[layout(UserFrame)]
            // Route1 is a non-layout component that will receive the `user_id: String` and `dynamic: String` parameters
            #[route("/:dynamic")]
            Route1 {
                // Automatically adds a `user_id: String` field
                dynamic: String
            },
        #[end_layout]
        // Route2 is a non-layout component that will receive the `user_id: String` parameter
        #[route("/hello_world")]
        Route2 {
            // Automatically adds a `user_id: String` field
        },
    #[end_nest]
}
```

We could require all of the children of a dynamic nest to share the same type for the parameter:

```rust
#[derive(Clone, Debug, PartialEq)]
#[routable]
enum Route {
    // Nests with parameters have types taken from child routes
    #[nest("user/:user_id")]
        // Everything inside the nest has the added parameter `user_id: String`
        // UserFrame is a layout component that will receive the `user_id: String` parameter
        #[layout(UserFrame)]
            // Route1 is a non-layout component that will receive the `user_id: String` and `dynamic: String` parameters
            #[route("/:dynamic")]
            Route1 {
                // The type is taken from the first instance of the dynamic parameter
                user_id: String,
                dynamic: String
            },
        #[end_layout]
        // Route2 is a non-layout component that will receive the `user_id: String` parameter
        #[route("/hello_world")]
        Route2 {
            user_id: String,
        },
    #[end_nest]
}
```

docs/router/src/reference/layouts.md
```diff
@@ -0,0 +1,19 @@
+# Layouts
+
+Layouts allow you to 
+
+[`Outlet`]s tell the router where to render content in layouts. In the following example,
```
stephenandary, line 5: The initial Layouts sentence is incomplete, and - really pedantic, and I apologize ahead of time lol - when reading this it may be nicer to see: 

`[Outlet] tells the router ...`

Looks like this right now: 
image

docs/router/src/example/navigation-targets.md
```diff
@@ -0,0 +1,27 @@
...
+
+## What is a navigation target?
+
+A [`NavigationTarget`] is similar to the `href` of an HTML anchor element. It
+tells the router where to navigate to. The Dioxus Router knows two kinds of
+navigation targets:
+
+- [`Internal`]: We used internal links in the previous chapter. It's a link to a page within our
+  app represented as a Route enum.
+- [`External`]: This works exactly like an HTML anchors' `href`. Don't use this for in-app
```
stephenandary, line 14: :heart:

docs/router/src/reference/layouts.md
Demonthos: Oops, should be fixed now

docs/router/examples/full_example.rs
```diff
@@ -0,0 +1,112 @@
+#![allow(non_snake_case)]
```
stephenandary, line 1: Something I have noticed which may be my error (digging into now), is when refactoring out components with inline props, one may be required to import the components props as well as the component. 

If I am not making a mistake, is it possible to not require this as it can be slighly confusing if the code does not exist explicitly. 

Current File Structure: 
```
src
  |_ main.rs
  |_ components/
        |_ navbar.rs
              |_ NavBar
```
Include NavBar in another file (e.g. main.rs here)
`use crate::components::navbar::NavBar;`

navbar.rs (just refactoring fullstack example for now to integrate some auth flow)
```rust
#[inline_props]
pub fn NavBar(cx: Scope) -> Element {
    render! {
        nav {
            ul {
                li { Link { target: Route::Home {}, "Home" } }
                li { Link { target: Route::BlogList {}, "Blog" } }
            }
        }
        Outlet {}
    }
}
```
Analyzer Error: 
```
cannot find struct, variant or union type `NavBarProps` in this scope
not found in this scoperustcClick for full compiler diagnostic
router.rs(2, 1): consider importing this struct: `use crate::components::navbar::NavBarProps;
`
```
This is easy enough to get around, but is it something to consider abstracting?

stephenandary: Moving a few things may end up looking like this from current example: 

```rust
use crate::components::{
    blog::{Blog, BlogList, BlogListProps, BlogPostProps, BlogProps},
    home::{Home, HomeProps},
    navbar::{NavBar, NavBarProps},
    page_not_found::{PageNotFound, PageNotFoundProps},
};
```

Demonthos: I agree if we could avoid the import it would be easier, but I don't know if it is possible. I don't see a way to get an associated type from a function. [function item types](https://doc.rust-lang.org/reference/types/function-item.html) are unnamable, and I don't think we can get the function pointer type in a way that allows us to access associated types

My attempt:
```rust

fn test(){
    <Wrapper::<app2, _> as HasProps>::Props::IS_STATIC;
}

trait HasProps{
    type Props;

    fn takes_self(&self){}
}

struct Wrapper<F: Fn(Scope<T>) -> Element, T>{
    phantom:PhantomData<(T, F)>
}

impl<T, F: Fn(Scope<T>) -> Element> HasProps for Wrapper<F, T> {
    type Props = T;
}

#[inline_props]
pub fn app2(cx: Scope) -> Element {
    todo!()
}
```

stephenandary: Would a good option be just to document this then somehow? It is really not a lot of work and the analyzer covers bases well.  May be nice to say "when using inline props, you must import props with component in files" somewhere.

jkelleyrtp: In components we can get the props of a component by using a const fn. It's called `fc_to_builder`.

I imagine we could have a similar utility function?

```
/// This utility function launches the builder method so rsx! and html! macros can use the typed-builder pattern
/// to initialize a component's props.
pub fn fc_to_builder<'a, A, T: Properties + 'a>(_: fn(Scope<'a, T>) -> A) -> T::Builder {
    T::builder()
}
```

jkelleyrtp:
Just some thoughts...

-  Is there a way we can do nesting without requiring the multi-block syntax? It's rather unusual syntax - I haven't seen it be popular anywhere else in the Rust ecosystem. Clap has some inspiration to draw from here since they do nests by deferring into a second enum. I know that prevents us from being type-safe on "upwards" segments, but those *are* atypical anyways. Maybe even a flatten variant of nesting where you name the nest via an additional attribute.
- Similar thing for the layout syntax. 
- I'm not a huge fan of having to turn off Rustfmt. It gives some code smell.
- The catchall syntax is also unusual but not hard to learn.
- Automatically generating the use_navigator is.... weird. Could we attach the navigator hook onto the Routes object via an impl so people can opt into pub-using in their app. Something like `Routes::use_navgiator`?

From a code perspective though, the new Router is miles better. I really like Outlets, Layouts, the enum, and using the namespace. It's all very clean.

stephenandary:
Is the multi-block syntax and catchall changing before release?

> Just some thoughts...
> 
> * Is there a way we can do nesting without requiring the multi-block syntax? It's rather unusual syntax - I haven't seen it be popular anywhere else in the Rust ecosystem. Clap has some inspiration to draw from here since they do nests by deferring into a second enum. I know that prevents us from being type-safe on "upwards" segments, but those _are_ atypical anyways. Maybe even a flatten variant of nesting where you name the nest via an additional attribute.
> * Similar thing for the layout syntax.
> * I'm not a huge fan of having to turn off Rustfmt. It gives some code smell.
> * The catchall syntax is also unusual but not hard to learn.
> * Automatically generating the use_navigator is.... weird. Could we attach the navigator hook onto the Routes object via an impl so people can opt into pub-using in their app. Something like `Routes::use_navgiator`?
> * The PR is cluttered thanks to all the Cargo.toml changes with workspace deps. We should clean that up first.
> 
> From a code perspective though, the new Router is miles better. I really like Outlets, Layouts, the enum, and using the namespace. It's all very clean.

What is plan regarding these comments?

stephenandary:
> > * Is there a way we can do nesting without requiring the multi-block syntax? It's rather unusual syntax - I haven't seen it be popular anywhere else in the Rust ecosystem. Clap has some inspiration to draw from here since they do nests by deferring into a second enum. I know that prevents us from being type-safe on "upwards" segments, but those _are_ atypical anyways. Maybe even a flatten variant of nesting where you name the nest via an additional attribute.
> > * Similar thing for the layout syntax.
> > * I'm not a huge fan of having to turn off Rustfmt. It gives some code smell.
> 
> I agree, the macro is ugly, but there are some constraints here:
> 
> * In the future, if we enable layout based data fetching, we need all routes to be in the same enum so that a child layout can fire a request based on a parent layout's url parameter.
>   For example a layout attached to the `:comment` in `/:user/comment/:comment` may need to read the user id to fetch the comment contents from the database
> * We need all routes to be in the same enum to allow children to opt-out of parent layouts
> * All routes being in the same enum makes linking to routes easier than nested routes in terms of how much you are typing:
> 
> ```rust
> Route::User(UserRoute::Comment(1234))
> ```
> 
> is more verbose than
> 
> ```rust
> Route::Comment(1234)
> ```
> 
> That said, I'm also not sold on the syntax and I'm open to other ideas. I would like to avoid an rsx-like macro because it cannot act like rsx with if statements loops, etc.
> 
> One other idea we discussed that would make the macro flat (but more verbose) is only applying layouts to one route and not allowing nests:
> 
> ```
> #[derive(Clone, Debug, PartialEq, Routable)]
> enum Route {
>     #[route("/:dynamic" Component1)]
>     Route1 { dynamic: String },
>     // layouts only apply to one route, which makes them... less useful
>     #[layout("/hello_world", Component)]
>     #[route("/")]
>     Route2 {},
>     // nests do not exist, all routes must be from the root
>     #[route("/hello/world")]
>     Route2 {},
> }
> ```
> 
> > * Automatically generating the use_navigator is.... weird. Could we attach the navigator hook onto the Routes object via an impl so people can opt into pub-using in their app. Something like `Routes::use_navgiator`?
> 
> Those hooks/components break when you use the wrong Route. If we expose the same method on all of them, when there are multiple routers involved it can be easy to mix them up (like in the new docsite). We should make a way to opt out of auto generating them.

Is Verbosity a bad thing here?  

From your example: 

(Thinking UserRoute is the Nest here)
`Route::User(UserRoute::Comment(1234))`
is more verbose than
`Route::Comment(1234)`

What if a user has a route for comments other than through User path there? 

e.g. having both `Route::User(UserRoute::Comment(1234))` and `Route::Other(OtherRoute::Comment(1234))`?

Demonthos:
> Is Verbosity a bad thing here?

> From your example:

> (Thinking UserRoute is the Nest here)
> Route::User(UserRoute::Comment(1234))
> is more verbose than
> Route::Comment(1234)
> 
> What if a user has a route for comments other than through User path there?
> 
> e.g. having both Route::User(UserRoute::Comment(1234)) and Route::Other(OtherRoute::Comment(1234))?
> 
> Do they both become Route::Comment(1234)?

You can choose the name when you define the router. I would use `Route::UserComment`, `Route::OtherComment`. I'm worried that making the Route verbose makes it too difficult to use in comparison to string routes.

This can be more of an issue when dealing with deeply nested routes:
String based:
```
"https://dioxuslabs.com/docs/nightly/guide/en/fullstack/server_functions.html"
```
Nested enums:
```
Route::Docs(DocsRoute::Nightly(NightlyRoute::Guide(GuideRoute::en(EnRoute::Fullstack(FullstackRoute::ServerFunctions)))))
```
Flattened enum:
```
Route::NightlyEnGuideServerFunctions {}
```

stephenandary:
Thank you @Demonthos, that makes sense. I noticed that when I reread your initial response above where you used `:comment` to name it. 

I find myself wondering if a macro can be used to allow a user to write nested enum paths like a flattened enum path, but I think that is not worth thinking about since you mentioned the flattened enum was important for 2 or 3 other purposes also.

Demonthos:
> I find myself wondering if a macro can be used to allow a user to write nested enum paths like a flattened enum path, but I think that is not worth thinking about since you mentioned the flattened enum was important for 2 or 3 other purposes also.

The current macro does secretly support nested routers. I just think it should not be the default option:
https://github.com/Demonthos/docsite/blob/ea/docs-new-router/src/lib.rs#L207

attilio-oliva:
Good job to everyone, involved on this interesting feature.
I wanted to share some thoughts as a third party average user: the current proposed API for nested routes is a bit awkward and not so idiomatic. It is simple once you get used to it but it would require an explicit new formatting rule, and would be initially strange to rust programmers.
Maybe a more intuitive and readable way would be something like this:

```rust
#[derive(Clone, Debug, PartialEq)]
#[routable]
enum Route {
   #[route("/normal_route")]
    Normal {},

   #[nest(UserRoute)]
   #[route("user/:user_id")]
    User {}
}

#[derive(Clone, Debug, PartialEq)]
#[routable]
#[nestable("user_id")]  //explicit parameters that have to be defined in the parent route and can be used by inner routes
enum UserRoute {
  #[route("/:dynamic")]
  #[layout(UserFrame)]
   Route1 {
      user_id: String,
      dynamic: String
   },

   #[route("/hello_world")]
    Route2 {
       user_id: String,
    },
}
``` 

If it is desired to have a big single route definition, it could be at least translated to something like the following. Doing so may be a bit less awkward, instead of having a macro to end a nested route definition.

```rust
#[derive(Clone, Debug, PartialEq)]
#[routable]
enum Route {
   #[route("/normal_route")]
    Normal {},

   #[nest]  //if the previous method can be used as well, keep using this macro like this
   #[route("user/:user_id")]
    User {
      #[route("/:dynamic")]
      #[layout(UserFrame)]
       Route1 {
          user_id: String,
          dynamic: String
       },

        #[route("/hello_world")]
        Route2 {
            user_id: String,
        },
     }
  
}

``` 

For the problem related to the resulting verbose enums, as already proposed, flattening would be the solution

jkelleyrtp:
> Good job to everyone, involved on this interesting feature. I wanted to share some thoughts as a third party average user: the current proposed API for nested routes is a bit awkward and not so idiomatic. It is simple once you get used to it but it would require an explicit new formatting rule, and would be initially strange to rust programmers. Maybe a more intuitive and readable way would be something like this:
> 
> ```rust
> #[derive(Clone, Debug, PartialEq)]
> #[routable]
> enum Route {
>    #[route("/normal_route")]
>     Normal {},
> 
>    #[nest(UserRoute)]
>    #[route("user/:user_id")]
>     User {}
> }
> 
> #[derive(Clone, Debug, PartialEq)]
> #[routable]
> #[nestable("user_id")]  //explicit parameters that have to be defined in the parent route and can be used by inner routes
> enum UserRoute {
>   #[route("/:dynamic")]
>   #[layout(UserFrame)]
>    Route1 {
>       user_id: String,
>       dynamic: String
>    },
> 
>    #[route("/hello_world")]
>     Route2 {
>        user_id: String,
>     },
> }
> ```
> 
> If it is desired to have a big single route definition, it could be at least translated to something like the following. Doing so may be a bit less awkward, instead of having a macro to end a nested route definition.
> 
> ```rust
> #[derive(Clone, Debug, PartialEq)]
> #[routable]
> enum Route {
>    #[route("/normal_route")]
>     Normal {},
> 
>    #[nest]  //if the previous method can be used as well, keep using this macro like this
>    #[route("user/:user_id")]
>     User {
>       #[route("/:dynamic")]
>       #[layout(UserFrame)]
>        Route1 {
>           user_id: String,
>           dynamic: String
>        },
> 
>         #[route("/hello_world")]
>         Route2 {
>             user_id: String,
>         },
>      }
>   
> }
> ```
> 
> For the problem related to the resulting verbose enums, as already proposed, flattening would be the solution

I really like the solution you proposed in the first code block. Being able to "forward" the arguments via the routable macro makes a lot of sense.

Also, generally for feedback here, I would prefer to have a single attribute with multiple parameters rather than multiple stacked attributes. So instead of 
```rust
      #[route("/:dynamic")]
      #[layout(UserFrame)]
       Route1 {
          user_id: String,
          dynamic: String
       },
```

It would be 
```rust
      #[route("/:dynamic", layout = UserFrame)]
       Route1 {
          user_id: String,
          dynamic: String
       },
```

docs/router/examples/full_example.rs
```diff
@@ -0,0 +1,112 @@
...
+// ANCHOR: router
+#[derive(Routable, Clone)]
+#[rustfmt::skip]
+enum Route {
+    #[layout(NavBar)]
+        #[route("/")]
+        Home {},
+        #[nest("/blog")]
+            #[layout(Blog)]
+                #[route("/")]
```
stephenandary, line 22: should attribute here be `"#[route("/blog")]`?

Demonthos: The route attribute is relative to the nests it is in, because this route is under a `nest("/blog")`, it will end up being rendered at the route "/blog/"

stephenandary: Why do you need to do `/blog/:name"` below then instead of `\/:name`?

Demonthos: I'm not sure why I did that... I will fix that in the docs rewrite branch

PR #1023, opened 2023-05-13
Mention futures-util in use_coroutine hook

DogeDark:
Adds a mention to ``futures-util`` in order to use ``rx.next().await``.


jkelleyrtp:
Thank you!!

PR #1024, opened 2023-05-14
fix router book link

onweru:
Currently the link to the [router book](https://dioxuslabs.com/docs/0.3/router/) is broken on the [routing](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/router.html#more-reading) section in the docs. The same link can be accessed from the top navigation, but it's better if this one works too.


Demonthos (approved):
Thank you!

PR #1036, opened 2023-05-20
feat: Move layout_attributes behind a feature in native-core

marc2332:
### Changes in `native-core`
- Moved  the `layout_attributes` module, taffy and lighningss behind the `layout-attributes` feature
- Removed unused dependencies (slab, crossbeam and hashbrown)


Demonthos (approved):
We might want to think about if pulling the features out into separate crates makes sense at some point. I believe the taffy attribute parsing code is being unstreamed, so that integration may become unnecessary soon. If other frameworks adopt native-core, it probably makes sense to pull the Dioxus integration into a separate crate.

But for now, this looks good. Thank you!

(This is a breaking change; waiting for a second review before merging)

jkelleyrtp:
Thanks @marc2332, we're trying to get you a published build ASAP for Freya :)

PR #1044, opened 2023-05-24
Clarify hooks rules to mention early returns

elliotwaite:
Let me know if you think this should be reworded in any way, or perhaps the additional text could be mentioned below the list of rules if you think that would look better. Or if you don't think this clarification is needed, feel free to close this PR without merging it.


jkelleyrtp:
This is good - we might move the warning below the bullet points but this is good for now.

elliotwaite:
Okay, sounds good.

PR #1056, opened 2023-05-30
Create use_eval for liveview

Demonthos:
Creates a use_eval hook (functioning the same as the desktop use_eval hook) for the liveview renderer to allow liveview programs to interact with browser APIs

### DIFF
--- a/packages/desktop/src/eval.rs
+++ b/packages/desktop/src/eval.rs
@@ -31,6 +31,19 @@ impl IntoFuture for EvalResult {
 }
 
 /// Get a closure that executes any JavaScript in the WebView context.
+///
+/// ```rust
+/// fn app(cx: Scope) -> Element {
+///     let eval = use_eval(cx);
+///     let result = eval(r#"
+///         // You can return values from javascript
+///         return document.getElementById("test-id").getAttribute("class");
+///     "#);
+///     cx.spawn(async move {
+///         println!("{:?}", result.await);
+///     });
+/// }
+/// ```
 pub fn use_eval(cx: &ScopeState) -> &Rc<dyn Fn(String) -> EvalResult> {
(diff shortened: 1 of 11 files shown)


jonay2000:
This feature is hard to use, since the error use_eval returns (`QueryError`) is public in a private module (`query`) and can therefore not be imported. Is `QueryErrror` the right error for this?

jonay2000:
I figured out why the future never resolves, the resoponse to the query contains no "data" field, which means serde cannot deserialize the response. Unfortunately, the error is silent.

jonay2000:
I think this could be filled by adding 
```js
|| null
```

to this line https://github.com/DioxusLabs/dioxus/blob/b264211cc2685232426317d3055046838047d3db/packages/liveview/src/query.rs#L49

jonay2000:
As well as that, I recommend logging errors when deserialization fails here
https://github.com/DioxusLabs/dioxus/blob/b264211cc2685232426317d3055046838047d3db/packages/liveview/src/pool.rs#L163

jonay2000 (changes requested):
see previous comments

Demonthos:
Thanks for letting me know about those issues. They should be fixed now, if it fails to deserialize the return value, it will now default to Null.

You don't need to await a use_eval hook that returns nothing unless you need to know when it finishes. The code will run in the browser even if you do not await the future.

jonay2000:
awesome!

I know I don't need to await it, but for my logic it makes a lot of sense to do so anyway.

jonay2000:
and I see QueryError is public now too :)

jonay2000 (approved):

jonay2000:
One more thing, maybe put in the documentation that to return a value you need to explicitly use `return`.

PR #1057, opened 2023-05-30
Add richer API for describing global shortcut accelerators.

ndarilek:


jkelleyrtp:
I like this! Great change, thanks @ndarilek . It's breaking so it'll have to be released in 0.4-pre.0 before making it into stable.

jkelleyrtp (approved):

PR #1063, opened 2023-06-02
Add initial_value attribute & fix static special attribute handling

### LINKED ISSUE #1061: Add InitialValue Attribute
## Specific Demand

There is currently no clear way to set a initial value for an uncontrolled input. Controlled inputs can have an impact on performance especially in situations where the latency is high (like in the liveview renderer).

## Implement Suggestion

Add a new initial value property:
```rust
render! {
    input {
        intial_value: "red",
    }
}
```

### Add a initial value attribute to the html crate:
The HTML crate contains definitions for element/attribute the rsx macro uses.

We need to add the attribute to the [input element](https://github.com/DioxusLabs/dioxus/blob/563a2b09532da2bddaf26cc3fd957c3a5c2f4cdb/packages/html/src/elements.rs#LL1156C26-L1156C26). Unlike the value attribute, this attribute should *not* be volatile (written to the renderer even if the value has not changed).

### Add a initial value attribute to the interpreter:
The interpreter handles applying the element/attributes to the DOM. Most attributes will work automatically, but the initial value needs some special code to make it work.

We need to set the defaultValue variable of the input element manually when the attribute is set: [example](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_text_defaultvalue2)

Modify the `interpreter.js` file to add a new case [here](https://github.com/DioxusLabs/dioxus/blob/563a2b09532da2bddaf26cc3fd957c3a5c2f4cdb/packages/interpreter/src/interpreter.js#L171) matching on the attribute type.

Modify the same switch statement in the `sledgehammer_bindings.rs` file [here](https://github.com/DioxusLabs/dioxus/blob/563a2b09532da2bddaf26cc3fd957c3a5c2f4cdb/packages/interpreter/src/sledgehammer_bindings.rs#L72)

> If you are a new contributors and run into issues implementing this feature, don't hesitate to reach out here or ping me on [discord](https://discord.gg/XgGxMSkvUM) @Demonthos

DogeDark:
- Adds the ``initial_value`` attribute to set an input element's ``defaultValue`` property: [reference](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_text_defaultvalue2)
- Moves ``SetAttributeInner`` Javascript function into own file ``common.js`` as ``setAttributeInner()``
- ``setAttributeInner()`` binding available using interpreter feature ``minimal_bindings`` and located in the ``minimal_bindings`` module.

Closes #1061


DogeDark:
Desktop fails to render anything after these changes without any known difference in bindings or API.

jkelleyrtp:
Nice! I'm curious how the bundling situation works now since we have multiple JS files. Do we need a bundler or is the bindgen handling that for us?

DogeDark:
Currently, web uses an import statement and wasm bindgen. Desktop has a new route in protocol.rs that handles /common.js and it sends back the include_str of common.js

With or without release flag:
- [x] ``cargo run`` for desktop works.
- [x] ``trunk serve`` for web works.
- [x] ``dioxus serve`` on web with a project that uses released dioxus works.
- [ ] ``dioxus serve`` on web serves a blank index.html (no wasm, only websocket js, weird)
- [ ] ``dioxus serve`` on desktop gets stuck compiling indefinitely.
- [ ] ``dioxus serve`` on desktop with a project that uses released dioxus gets stuck compiling indefinitely.

Dioxus CLI is funky. Hard to tell what issues would be related to this PR. I believe at least the blank index.html issue is related though. I'm going to need a pointer or two to figure this one out.

Demonthos:
`dioxus serve` only works for dioxus web

DogeDark:
Dang, I don't know why I had the impression that it did.

DogeDark:
Turns out that ``dioxus serve`` wasn't working correctly because it doesn't import the necessary scripts into custom ``index.html`` files. I'm going to open a specific demand issue in the CLI repo for this.

This is still the case and from my testing everything works:
> Currently, web uses an import statement and wasm bindgen. Desktop has a new route in protocol.rs that handles /common.js and it sends back the include_str of common.js

DogeDark:
I created an issue in the CLI repo for the dioxus serve issue. This PR is good to go if everything checks out.

Demonthos (commented):
There was one area I think we can simplify the code a bit, but other than that this looks great!

Demonthos (approved):
Thank You!

packages/interpreter/src/interpreter.js
```diff
@@ -1,3 +1,5 @@
+import { setAttributeInner } from "./common.js";
```
goffrie, line 1: I believe this breaks liveview which injects the `WS_ADDR` prior to this file, causing "SyntaxError: import declarations may only appear at top level of a module"

DogeDark: I'll take a look when I get home.

DogeDark: #1073

PR #1064, opened 2023-06-02
Expanded fuzzing

Demonthos:
Expands the core fuzzing test to cover:
1) Adding and triggering Listeners
2) Randomly marking nodes as dirty
3) Borrowed props

(CI will likely fail initially because this uncovers some new issues)


jkelleyrtp:
Amazing, thank you!

Issue #1070, opened 2023-06-04
Attribute hidden always hides element (boolean attributes behave different than documentation says)


nmlt:
**Problem**

If I put `hidden: "false"` or `hidden: "true"` in my element, it will get the `hidden` attribute regardless. That's contrary to what the documentation says [here](https://dioxuslabs.com/docs/0.3/guide/en/describing_ui/special_attributes.html#boolean-attributes).

**Steps To Reproduce**

Here's a minimal example:

```rust
#![allow(non_snake_case)]
// import the prelude to get access to the `rsx!` macro and the `Scope` and `Element` types
use dioxus::prelude::*;

fn main() {
    // launch the web app
    dioxus_web::launch(App);
}

// create a component that renders a div with the text "Hello, world!"
fn App(cx: Scope) -> Element {
    cx.render(rsx! {
        div {
            hidden: "false",
            "text"
        }
    })
}

```

This results in the following HTML:

```html
<div hidden="false">text</div>
```

**Expected behavior**
(No attribute on the div element)
```html
<div>text</div>
```

**Environment:**
 - Dioxus version: "0.3.2"
 - Rust version: rustc 1.70.0 (90c541806 2023-05-31)
 - OS info: MacOS Ventura 13.3.1 (a) (22E772610a) (M2)
 - App platform: `web`

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Demonthos:
It appears this only happens with static text. If you make the hidden attribute dynamic, it works.

If you are interested in fixing this:

Setting static attributes is handled [here](https://github.com/DioxusLabs/dioxus/blob/0f684f056762a0dee05c321290b6cc4fc5299d72/packages/web/src/dom.rs#L132)

Currently that code unconditionally sets the attribute, we need to check if the attribute matches any of [these](https://github.com/DioxusLabs/dioxus/blob/0f684f056762a0dee05c321290b6cc4fc5299d72/packages/interpreter/src/sledgehammer_bindings.rs#L136) boolean attributes. If it does, and the text is `"false"`, we can just skip setting the attribute.

nmlt:
Ah, very cool! I'm looking into it.

nmlt:
[...]

PR #1072, opened 2023-06-04
Move bool_attrs to common.js

DogeDark:
Didn't realize ``bools_attrs`` object was referenced by ``setAttributeInner()``.


Demonthos:
Testing this, it does fixes the example in #1070

Demonthos (approved):

PR #1073, opened 2023-06-05
Fix liveview interpreter JS

DogeDark:
Previous PR #1063 changed how the javascript interpreter code is handled. Liveview didn't have the necessary changes to accommodate the new JS organization. (Hopefully this is the last thing that #1063 broke 👀)

### DIFF
--- a/packages/liveview/src/lib.rs
+++ b/packages/liveview/src/lib.rs
@@ -77,14 +77,14 @@ static INTERPRETER_JS: Lazy<String> = Lazy::new(|| {
         return;
       }
     }"#;
-    
+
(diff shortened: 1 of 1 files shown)


Demonthos (approved):
This is fine for now, but at some point it could be worth looking into getting a bundler working for liveview and desktop instead of just manually manipulating strings

DogeDark:
I've been thinking similarly. It's been pretty rough having all of these issues after the fact. It might even be worth adding a github action to check all JS code for issues such as the one earlier of bool_attrs. Something else I'm not super fond of is the duplicate JS code between sledgehammer and wasm-bindgen.

PR #1080, opened 2023-06-07
Convert use_eval to use send/recv system

### LINKED ISSUE #1026: Async Communication Between JavaScript and Rust
## Specific Demand

Currently, it is difficult to integrate with some web or JavaScript APIs because use_eval only allows you to return a single JavaScript value.

For example, the [intersection observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) uses callbacks to notify a script when the intersection has changed, but there is no easy way to send values to rust from that callback.

Dioxus can implement some web APIs as first party. This somewhat mitigates this problem, but it cannot provide first party bindings for all the JavaScript libraries users may want to integrate.

## Implement Suggestion

Change use_eval to include an incoming and outgoing channel for both the JavaScript and Rust side. In JavaScript, a `dioxus` variable would be set before the script is run with methods to communicate with rust.

```rust
let eval = use_eval(cx);
let result = eval(r#"dioxus.send("hello from JS");
const value = await dioxus.recv();
console.log(value);"#);
println!("{}", result.recv());
result.send("hello");
```

DogeDark:
Converts the ``use_eval`` hook to use a send/recv system.

**Example:**
```rs
    // Construct the evaluator
    let eval_provider = dioxus_html::prelude::use_eval(cx);
    let eval = eval_provider(r#"
        dioxus.send("Hi from JS!");
        let msg = await dioxus.recv();
        console.log(msg);
    "#);

    // Run the JS
    eval.run();

    // Send a message to JS
    eval.send("Hi from Rust!".into());

    // Get a receiver channel to receive messages from JS
    let receiver = eval.receiver();
    let future = use_future(cx, (), |_| async move {
        receiver.recv().await.unwrap()
    });
    
   // Act on received data
    match future.value() {
        Some(v) => {
            // Cleanup any eval artifacts
            eval.done();
            cx.render(rsx!(
                p { "{v}" }
            ))
        }
        _ => {
            cx.render(rsx!(
                p { "hello" }
            ))
        }
    }
}
```

**Todo:**
- [x] Web
- [x] Desktop
- [x] Liveview

Closes #1026


DogeDark:
Will fix the CI issues tomorrow.

packages/html/src/eval.rs
```diff
@@ -0,0 +1,85 @@
...
+        .expect("evaluator not provided");
+
+    let evaluator = eval_provider.new_evaluator(cx, js.to_string());
+
+    UseEval::new(evaluator)
+}
+
+/// A wrapper around the target platform's evaluator.
+pub struct UseEval {
+    evaluator: Box<dyn Evaluator + 'static>,
```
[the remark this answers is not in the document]
DogeDark: Done!

examples/eval.rs
```diff
@@ -5,29 +5,31 @@ fn main() {
...
-                    });
-                },
-                "Execute"
-            }
+    eval.run().unwrap();
+    eval.send("Hi from Rust!".into()).unwrap();
+
+    let receiver = eval.receiver();
+
+    let future = use_future(cx, (), |_| async move { receiver.recv().await.unwrap() });
```
Demonthos, line 22: Running this example on my M1 mac, it panics unwrapping at this line

packages/web/src/eval.js
```diff
@@ -0,0 +1,32 @@
+export class Dioxus {
+  constructor(sendCallback) {
+    this.sendCallback = sendCallback;
+    this.received = [];
+  }
+  // Receive message from Rust
+  recv() {
+    return new Promise((resolve, _reject) => {
+      // Ever 50 ms check for new data
+      let timeout = setTimeout(() => {
```
Demonthos, line 10: Instead of polling for new data could we first check if there is any currently resolved data. If there is we can resolve the promise immediately, otherwise we could set a pendingPromise field on the object to the resolve callback. That callback could then be called in the rustSend function.

examples/eval.rs
DogeDark: Does it say why?

DogeDark: Nevermind, I'm getting it too.

packages/web/src/eval.js
DogeDark: Much better solution! I knew a timeout didn't feel right.

examples/eval.rs
DogeDark: Should be fixed now.

packages/html/src/eval.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// Get a struct that can execute any JavaScript.
+///
+/// # Safety
+///
+/// Please be very careful with this function. A script with too many dynamic
+/// parts is practically asking for a hacker to find an XSS vulnerability in
+/// it. **This applies especially to web targets, where the JavaScript context
+/// has access to most, if not all of your application data.**
+pub fn use_eval<S: ToString>(cx: &ScopeState, js: S) -> &mut UseEval {
```
Demonthos, line 33: Returning &mut UseEval here where eval is not clone makes it [difficult to move around UseEval](https://dioxuslabs.com/docs/nightly/guide/en/interactivity/custom_hooks.html#hook-anti-patterns). I think you should still be able to start an eval within a future with this API, but you would need to make the javascript await for a start message after you run it which could be unintuitive

I also wonder if we could make the API a bit more similar to the previous API to make upgrading less breaking. What are your thoughts on this API?
- use_eval returns a Rc<dyn EvalProvider>
- You can call `.start` on the eval provider to start running the script
- The Eval returned from `.start` can either be turned into a future with [IntoFuture](https://doc.rust-lang.org/nightly/core/future/trait.IntoFuture.html) (which consumes it) like the old API or you can use the send/receive function like you currently implemented

DogeDark: I agree with the Rc but instead of a start function, how about implementing ``IntoFuture`` for the  ``UseEval`` struct? It would automatically run and await the first data returned through the channel. 

Additionally, any ``return data;`` JS could be replaced with ``dioxus.send(data);``? - it would add some string processing overhead during runtime though. If not that then we could add it as a step in the migration process - tell people to replace ``return`` with ``dioxus.send``?

```rs
// I could see this being useful outside of legacy code
let eval = use_eval(cx, "dioxus.send('hi from js');");
if let Ok(msg) = eval.await {
    // act on message
}
```
compared to the old:
```rs
let eval = use_eval(cx);
if let Ok(msg) = eval("return 'hi from js';").await {
    // act on message
}
```

Demonthos: > I agree with the Rc but instead of a start function, how about implementing `IntoFuture` for the `UseEval` struct? It would automatically run and await the first data returned through the channel.

This sounds good!

> Additionally, any `return data;` JS could be replaced with `dioxus.send(data);`? - it would add some string processing overhead during runtime though. (**On second thoughts, replacing the return statement could have false positives**)

Nested function make this difficult. I think it would be better to try to not modify the javascript script submitted (except the wrapping we already do).

> If not that then we could add it as a step in the migration process - tell people to replace `return` with `dioxus.send`?
> 
> ```rust
> // I could see this being useful outside of legacy code
> let eval = use_eval(cx, "dioxus.send('hi from js');");
> if let Ok(msg) = eval.await {
>     // act on message
> }
> ```
> 
> compared to the old:
> 
> ```rust
> let eval = use_eval(cx);
> if let Ok(msg) = eval("return 'hi from js';").await {
>     // act on message
> }
> ```

I could see the old system being useful when you have a chain of several different scripts like:

```rust
let eval = use_eval(cx);
use_future!(cx, |()| {
    to_owned![eval];
    async move {
        eval("return 'script 1'").await;
        // some rust processing code
        eval("return 'script 2'").await;
    }
});
```
Where script 1 and script 2 are from different libraries.

In the new system, that would need to look like this:

```rust
let eval1 = use_eval(cx, "script 1");
let eval2 = use_eval(cx, "script 2");
use_future!(cx, |()| {
    to_owned![eval1, eval2];
    async move {
        eval1.await;
        // some rust processing code
        eval2.await;
    }
});
```

If we force the code to be defined with the hook it could also hurt colocation:

```rust
let script = use_eval(cx, "script 1");

// 100 more lines of code setting up data

render!{
    button {
        onclick: move |_| async move {
            script.await
        }
    }
}
```

Looking at the onclick handler, it could be less clear what code script.await is running compared to:

```rust
let eval = use_eval(cx);

// 100 more lines of code setting up data

render!{
    button {
        onclick: move |_| async move {
            eval("script 1").await
        }
    }
}
```

DogeDark: > Looking at the onclick handler, it could be less clear what code script.await is running compared to: 
> ```rs
> let eval = use_eval(cx);
> 
> // 100 more lines of code setting up data
> 
> render!{
>     button {
>         onclick: move |_| async move {
>             eval("script 1").await
>         }
>     }
> }
> ```

I believe this shouldn't be as big of an issue with good variable naming. Developers could also move the use_eval call below the 100 lines.

> In the new system, that would need to look like this:
> ```rs
> let eval1 = use_eval(cx, "script 1");
> let eval2 = use_eval(cx, "script 2");
> use_future!(cx, |()| {
>     to_owned![eval1, eval2];
>     async move {
>         eval1.await;
>         // some rust processing code
>         eval2.await;
>     }
> });
> ```
I agree, chaining eval is useful. However, what if something like this was implemented which could effectively do the same:
```rs
let eval = use_eval(cx, 
    r#"
    script 1
    // wait for signal to continue
    await dioxus.signal(); // reads as "await dioxus signal"
    script 2
    "#);

use_future!(cx, |()| {
    to_owned![eval];
    async move {
        eval.run();
        // some rust processing code
        eval.signal(); // Signal the script to continue
    }
});
```
The dev could still pass data to the script when needed using the regular ``send()`` and ``recv()``.

I generally want to avoid using any closures or anonymous functions. In my opinion, they add complexity and make control flow a little confusing (dioxus-std was a pain).

DogeDark: I did the stuff @Demonthos
This works
```rs
let mut eval = use_eval(cx);

// 100 more lines of code setting up data

render!{
    button {
        onclick: move |_| async move {
            eval("script 1").await
        }
    }
}
```

packages/desktop/src/query.rs
```diff
@@ -56,6 +60,64 @@ impl QueryEngine {
...
+        if let Err(err) = webview.evaluate_script(&code) {
+            log::warn!("Query error: {err}");
+        }
+
+        let thread_receiver = self.sender.subscribe();
+        let thread_handle: JoinHandle<()> = tokio::spawn(async move {
+            let mut receiver = thread_receiver;
+            loop {
+                if let Ok(result) = receiver.recv().await {
+                    if result.id == request_id {
```
Demonthos, line 106: This ends up scaling poorly with large numbers of requests (explained more in https://github.com/DioxusLabs/dioxus/pull/1115). This shouldn't be any worse than the git version of use_eval. At this point I would like to get this PR merged, so that we can unblock the liveview router. Just something to think about for the future

playwrite-tests/web/src/main.rs
```diff
@@ -43,11 +48,16 @@ fn app(cx: Scope) -> Element {
...
-                let result = eval(r#"window.document.title = 'Hello from Dioxus Eval!';
-                return "returned eval value";"#.to_string());
-                if let Ok(serde_json::Value::String(string)) = result.get() {
-                    eval_result.set(string);
-                }
+                eval.run().unwrap();
+                let receiver = eval.receiver();
+                let setter = eval_result.setter();
+
+                wasm_bindgen_futures::spawn_local(async move {
```
Demonthos, line 55: Could we just return an async block here or use cx.spawn instead of pulling in another library?

Demonthos (changes requested):
API looks great! A few small things. I think we should wait on performance improvements so that we can get this into 0.4, but if you could remove `wasm_bindgen_futures::spawn_local` that would be great

playwrite-tests/web/src/main.rs
DogeDark: Turns out we don't even need async.

examples/eval.rs
Demonthos: This seems like it became an issue again. If the eval gets dropped it will panic

This fixes the issue, but if it is possible to make the original example work, that would probably be better:

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    let eval_provider = use_eval(cx);

    let future = use_future(cx, (), |_| {
        let mut eval = eval_provider(
            r#"
            dioxus.send("Hi from JS!");
            let msg = await dioxus.recv();
            console.log(msg);
        "#,
        );
        async move {
            eval.run().unwrap();
            eval.send("Hi from Rust!".into()).unwrap();

            let receiver = eval.receiver();
            receiver.recv().await.unwrap()
        }
    });

    match future.value() {
        Some(v) => cx.render(rsx!(
            p { "{v}" }
        )),
        _ => cx.render(rsx!(
            p { "hello" }
        )),
    }
}
```

DogeDark: Realistically the error should be handled, but we could create a wrapper type around the channel that has the same methods but instead of errors it returns Option<Value>

dr-oofenshmirtz:
Sorry to hijack. I'm using this PR and I am trying to get multiple messages from the channel. Having replicated your example successfully. I set out to tweak it for my use case. 

```rust
     let mut eval = eval_provider(
         r#"
           window.addEventListener("message", dioxus.send)
    "#,);
```

Which should in theory attach an event listener to the window. Listening for a postMessage from either a service worker or an iframe. (technically the same context works, but that's not the value add). The issue I'm having is the receiver is falling out of scope. In [the example you create a receiver](https://github.com/DioxusLabs/dioxus/pull/1080/files#diff-7c4e660449b789afbf10c871b80773475ec4eb758cfd01e7c0f50b6ce3f5bbabR20) and you use it as a future. If I grok the code this is specifically an [UnboundedReciever](https://docs.rs/futures-channel/latest/futures_channel/mpsc/struct.UnboundedReceiver.html#). Which leads me to believe that if I stuck it in a variable or a use_ref or some other form of maintaining a reference to the channel, I could read from it indefinitely. Which should work perfectly with window.addEventListener. 

Am I understanding the potential correctly? How would you recommend I maintain a reference to the channel to get multiple values out of it? And bonus points, if I make an event listener in the JS context in this manner, how could I clean it up?

DogeDark:
Yes, I believe that's possible.

> I could read from it indefinitely.

It sounds like you want [cx.spawn](https://docs.rs/dioxus/latest/dioxus/prelude/struct.ScopeState.html#method.spawn),  [cx.spawn_forever](https://docs.rs/dioxus/latest/dioxus/prelude/struct.ScopeState.html#method.spawn_forever) or maybe even [use_coroutine](https://docs.rs/dioxus-hooks/latest/dioxus_hooks/fn.use_coroutine.html). You would move the eval into the async block (solving the drop-ing issue) and then call ``eval.run()`` on it (so it only runs once and not every time the component re-renders).

In terms of cleaning up, you could have the script wait for a message before running additional code:
```rs
let mut eval = eval_provider(r#"
    window.addEventListener("message", dioxus.send);
    
    // Wait for any message from Rust
    await dioxus.recv();

    // Cleanup code
    window.removeEventListener("message", dioxus.send);
    "#);

// When you want to cleanup the event listener:
eval.send(serde_json::Value::Null);
```

dr-oofenshmirtz:
> @dr-oofenshmirtz Yes, I believe that's possible.
> 
> > I could read from it indefinitely.
> 
> It sounds like you want [cx.spawn](https://docs.rs/dioxus/latest/dioxus/prelude/struct.ScopeState.html#method.spawn), [cx.spawn_forever](https://docs.rs/dioxus/latest/dioxus/prelude/struct.ScopeState.html#method.spawn_forever) or maybe even [use_coroutine](https://docs.rs/dioxus-hooks/latest/dioxus_hooks/fn.use_coroutine.html). You would move the eval into the async block (solving the drop-ing issue) and then call `eval.run()` on it (so it only runs once and not every time the component re-renders).
> 
> In terms of cleaning up, you could have the script wait for a message before running additional code:
> 
> ```rust
> let mut eval = eval_provider(r#"
>     window.addEventListener("message", dioxus.send);
>     
>     // Wait for any message from Rust
>     await dioxus.recv();
> 
>     // Cleanup code
>     window.removeEventListener("message", dioxus.send);
>     "#);
> 
> // When you want to cleanup the event listener:
> eval.send(serde_json::Value::Null);
> ```

Clever use of await!

Demonthos (approved):
Looks great! (waiting for a second review before merging)

Demonthos:
This now works with Liveview, and it implements the more efficient non-broadcast method of channels

PR #1084, opened 2023-06-10
fix: Properly drop UseFuture's values to avoid leaking memory

marc2332:


jkelleyrtp:
Thanks @marc2332 :D 

Rolling this into #1131

PR #1094, opened 2023-06-13
hooks.md and walkthrough_readme.md updates.

echochamber:


docs/guide/src/en/contributing/walkthrough_readme.md
```diff
@@ -63,64 +63,74 @@ pub struct VirtualDom {
...
+>
+> Dioxus uses "synchronized slabs" to communicate between the renderer and the VDOM. When a node is created in the Virtual DOM, an (elementId, mutation) pair is passed to the renderer to identify that node, which the renderer will then render in actual DOM. These ids are also used by the Virtual Dom to reference that node in future mutations, like setting an attribute on a node or removing a node. When the renderer sends an event to the Virtual Dom, it sends the ElementId of the node that the event was triggered on. The Virtual DOM uses this id to find that node in the slab and then run the necessary event handlers (Question - what are "necessary event handlers?").
 
-The virtual dom is a tree of scopes. A new scope is created for every component when it is first rendered and recycled when the component is unmounted.
+The virtual DOM is a tree of scopes. A new `Scope` is created for every component when it is first rendered and recycled when the component is unmounted (JSNote: unmounted and recycled are terms I don't fully understand in this context. Best guess: Unmounted = element a scope is for is deleted from the virtual dom? Recycled = a scope having its slab dealocated?)
```
echochamber: > recycled when the component is unmounted

Unmounted and recycled are terms I don't fully understand in this context. Best guess: Unmounted = element a scope is for is deleted from the virtual dom? Recycled = a scope having its slab dealocated?

docs/guide/src/en/contributing/walkthrough_readme.md
```diff
@@ -63,64 +63,74 @@ pub struct VirtualDom {
...
 
-Next, the Virtual Dom will compare the new VNode with the previous VNode and only update the parts of the tree that have changed.
+Next, the Virtual DOM will compare the new VNode with the previous VNode and only update the parts of the tree that have changed. Because of this approach, when a component is re-rendered only the parts of the tree that have changed will be updated in the DOM by the renderer.
 
-When a component is re-rendered, the Virtual Dom will compare the new VNode with the previous VNode and only update the parts of the tree that have changed.
+The diffing algorithm goes through the list of dynamic attributes and nodes and compares them to the previous VNode. If the attribute or node has changed, a mutation that describes the change is added to the mutation list. JSNOTE (add a note/comment about why `#[derive(PartialEq)]` on props improves performance here?).
```
echochamber, line 122: Maybe add a mention of how `#[derive(PartialEq, Eq)]` on props/Components can affect performance here?

docs/guide/src/en/contributing/walkthrough_readme.md
```diff
@@ -63,64 +63,74 @@ pub struct VirtualDom {
...
-> When the renderer sends an event to the Virtual Dom, it sends the ElementId of the node that the event was triggered on. The Virtual Dom uses this id to find the node in the slab and then run the necessary event handlers.
+>
+> Dioxus uses "synchronized slabs" to communicate between the renderer and the VDOM. When a node is created in the Virtual DOM, an (elementId, mutation) pair is passed to the renderer to identify that node, which the renderer will then render in actual DOM. These ids are also used by the Virtual Dom to reference that node in future mutations, like setting an attribute on a node or removing a node. When the renderer sends an event to the Virtual Dom, it sends the ElementId of the node that the event was triggered on. The Virtual DOM uses this id to find that node in the slab and then run the necessary event handlers.
```
echochamber, line 71: I don't know what this means exactly here and didn't feel like I could glean enough context to reword it well myself.

docs/guide/src/en/contributing/walkthrough_readme.md
```diff
@@ -63,64 +63,74 @@ pub struct VirtualDom {
...
 
-The virtual dom is a tree of scopes. A new scope is created for every component when it is first rendered and recycled when the component is unmounted.
+The virtual DOM is a tree of scopes. A new `Scope` is created for every component when it is first rendered and recycled when the component is unmounted.
```
echochamber, line 73: Unmounted and recycled are terms I don't fully understand in this context. Best guess: Unmounted = element a scope is for is deleted from the virtual dom? Recycled = a scope having its slab dealocated?

echochamber (commented):
My comments in this review are places where I had questions for the owners.

docs/guide/src/en/contributing/walkthrough_readme.md
Demonthos: Recycled means that the scope id is [removed](https://docs.rs/slab/latest/slab/struct.Slab.html#method.remove) from the slab which deallocates the memory and allows us to reuse the id

docs/guide/src/en/contributing/walkthrough_readme.md
Demonthos: Mutations are the way dioxus communicates what should be rendered. When the virtual DOM creates an element, it generates a mutation with an unique id for the element. The id can both:
1. Identify the element in the virtual DOM. This is used to identify where to trigger event listeners when the renderer sends an event back to the virtual DOM. ([desktop renderer events with element id](https://github.com/DioxusLabs/dioxus/blob/master/packages/interpreter/src/interpreter.js#L461))
2. Identify the element in the renderer. This is used to modify or remove the element later.

I called it a "synchronized slab" because:
- The virtual DOM has the original slab. New elements are created and removed in that slab.
- The renderer has a mirrored slab that changes with mutations to reflect the original slab but with platform-specific elements instead of virtual nodes. In the desktop renderer, this is the [nodes](https://github.com/DioxusLabs/dioxus/blob/master/packages/interpreter/src/interpreter.js#L69) variable. Elements being created, [sets an element at the index of the element id](https://github.com/DioxusLabs/dioxus/blob/master/packages/interpreter/src/interpreter.js#L131). Elements will be reused when the virtual DOM removes the node id and then assigns the element id to a new element (overriding the previous element).

Issue #1092, opened 2023-06-13
`onmounted` is unreliably triggered


jmsfltchr:
**Problem**

When making an async function to respond to `onmounted` events, the event is often not triggered. I find it to be quite reliable if we are using `onmounted` on <=10 divs (but still fails about 5% of the time), but consistently fails for >= 20 divs.

I don't know any workarounds for this issue.

**Steps To Reproduce**

Dioxus commit `b476c6e8a73bd6591ec4827c6f9b196d7c270c3d`.

Try this code: 

```
#![allow(clippy::await_holding_refcell_ref)]
use std::rc::Rc;

use dioxus::{html::geometry::euclid::Rect, prelude::*};
use dioxus_desktop::use_eval;

fn main() {
    dioxus_desktop::launch_cfg(
        app,
        dioxus_desktop::Config::default().with_custom_head(
            r#"
<style type="text/css">
    html, body {
        height: 100%;
        width: 100%;
        margin: 0;
    }
    #main {
        height: 100%;
        width: 100%;
    }
</style>
"#.to_owned(),
        ),
    );
}

fn js_get_element_dimensions(id: u64) -> String {
    format!(r#"
    const el = document.getElementById("{id}");
    let width = el.offsetWidth;
    let height = el.offsetHeight;
    return [width, height];
    "#).to_string()
}

fn app(cx: Scope) -> Element {
    let eval = use_eval(cx);
    let ndvis = 50;
    cx.render(rsx!(
        div {
            position: "relative",
            width: "1000px",
            height: "10ß0px",
            (0..ndvis).map(|id| {
                let l = id * 50;
                let t = id * 50;
                rsx!(div {
                    id: "{id}",
                    position: "absolute",
                    top: "{t}px",
                    left: "{l}px",
                    background_color: "red",
                    onmounted: move |_| {
                        to_owned![eval];
                        cx.spawn(async move {
                            if let Ok(res) = eval(js_get_element_dimensions(id as u64)).await {
                                let width = res.as_array().unwrap().get(0).unwrap().as_f64().unwrap();
                                let height = res.as_array().unwrap().get(1).unwrap().as_f64().unwrap();
                                println!("Got width {} height {} for id {}", width, height, id);
                            }
                        });
                    },
                    p{
                        "This element is {id}"
                    }
                })
            })
        }
    ))
}
```

**Expected behavior**

With `ndivs = 10` I see the following in std output:
```
Got width 110 height 50 for id 0
[...]

PR #1095, opened 2023-06-14
Increase the size of the queue for the query engine

jmsfltchr:
# Description

Increase the number of tokio channels for the query engine from 8 to 1000. This fixes issues where making many calls to `eval` or `unmounted` would cause said calls to fail with error `Error receiving query result: channel lagged by 8`.

Solves https://github.com/DioxusLabs/dioxus/issues/1092


Demonthos (approved):
Thank You!

Issue #1086, opened 2023-06-12
special attributes cannot be set properly with boolean values


Busyvar:
When a special attribute (like disabled, hidden, ...) is set with a boolean value, special attribute is not positional anymore.

For example :
_definition_
`disabled: X as bool,`
_results_
`<div disabled="true" ></div>` or
`<div disabled="false" ></div>` 
_What i expected_
`<div disabled></div>` or 
`<div></div>`

Here is a correct definition
`disabled: if X {"true"}else{"false"},`
_results_
`<div></div>` or
`<div disabled></div>`

Do you think current syntax can be enhanced to take boolean value as input?

**Environment:**
 - Dioxus version: 0.3.2
 - Rust version:  1.72.0 nightly
 - OS info: Arch x64
 - App platform: desktop

onweru:
@Busyvar, 

> Do you think current syntax can be enhanced to take boolean value as input?

Assuming by `'input'` you mean props / local variable. If so,

Maybe:

```rust
fn Button(cx: Scope) -> Element {
    ...
    let is_disabled = if cx.props.disabled {
        true
    } else {
        false
    };
    cx.render(rsx!(
        button {
            disabled: is_disabled,
            ...
        }
    ))
    ...
}
```

For additional context

- [Boolean attributes](https://dioxuslabs.com/docs/0.3/guide/en/describing_ui/special_attributes.html#boolean-attributes)
- [Conditional rendering](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/dynamic_rendering.html#conditional-rendering)

onweru:
@Busyvar, 

First I want to clarify something, just in case I misunderstood your original question.

If you're using the button component as defined by under [dioxus html](https://docs.rs/dioxus-html/latest/src/dioxus_html/elements.rs.html#1022) you only need to specify the disabled prop value when its value is `true`

```rust
// same as setting `disabled: false`
button {
    "My button" 
}
// when disabled: true
button {
    disabled: true,
    "My button"
}
```  

👆🏼  This code will generate this html

```html
<button>My button</button> <!-- disabled is false by default i.e when the disabled attribute is missing ->
<button disabled="true">My button</button>
```

<img width="1109" alt="SCR-20230613-lpcs" src="https://github.com/DioxusLabs/dioxus/assets/16350351/a181e3cc-f61d-406c-acfa-247167bf8cde">

onweru:
If you need a custom button component,

[...]

PR #1104, opened 2023-06-19
Fix boolean attributes with raw boolean values

Demonthos:
Checks if boolean attributes are equal to either the value `true` or the string `"true"` before choosing to either add or remove the attribute

fixes https://github.com/DioxusLabs/dioxus/issues/1086


jkelleyrtp:
Thanks, this rolls up into #1131

PR #1107, opened 2023-06-20
Add iOS to CI

jkelleyrtp:


Demonthos:
> * global shortcuts are not supported in iOS, apparently, in Tao/Wry.
> * RFD being prevents dioxus-desktop from compiling for iOS. Needs to be fixed in RFD, I think.
> * RFD in the dioxus-html crate is probably not the best
It looks like RFD does not have mobile support. 

The intention was for RFD to be used in native crates like Blitz down the line, so the html crate was a good temporary place to put a common native implementation of the file events. We could pull native and web bindings into separate crates or just disable the rfd dependency when targeting mobile

Demonthos:
Relevant RFD issue: https://github.com/PolyMeilex/rfd/issues/82

Issue #1097, opened 2023-06-16
Introduction example in docs is broken


RobinBoers:
**Problem**

The example in the docs shows an error when clicking run.

**Steps To Reproduce**

- Open docs (<https://dioxuslabs.com/docs/0.3/guide/en/>)
- Click play button

**Expected behavior**

The example should work

**Screenshots**

![image](https://github.com/DioxusLabs/dioxus/assets/60298132/29b43b8d-edd3-46d4-9e63-c88c80876644)

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [x] I don't have time to fix this right now, but maybe later

marc2332:
Introduction examples on documentation are sometimes kept simple on purpose.
You just need to import dioxus like this to make it work:

```rs
use dioxus::prelude::*;
```

RobinBoers:
The problem isn't that the example fails. I can see why that happens. I just expect the example to work if I run play. This doesn't seem professional.

marc2332:
> The problem isn't that the example fails. I can see why that happens. I just expect the example to work if I click play. This doesn't seem professional.

These examples are not interactive, I'll open a PR later so the play button doesnt show up.

PR #1118, opened 2023-06-22
Fix: Remove play button from guide example

marc2332:
Context https://github.com/DioxusLabs/dioxus/issues/1097#issuecomment-1603162527


Demonthos (commented):
Thank you, could you apply the change to the other code snippets in the guide as well?

[This config](https://github.com/DioxusLabs/dioxus/blob/master/docs/guide/book.toml#L28) should mark all code snippets as no_run by default, but it looks like it doesn't work

Demonthos (approved):
Looks great! Thank you!

Issue #1124, opened 2023-06-24
Bump Dioxus Liveview Salvo Version


xinglixing:
**Problem**

dioxus_liveview::salvo_socket(ws) - mismatched types 'WebSocket' and 'salvo_extra::ws::WebSocket'.

**Steps To Reproduce**

- Change salvo's version to 0.44 in Cargo.toml

**Environment:**
 - Dioxus version: [v0.3.2 `master`]
 - Rust version: [v1.70.0]
 - OS info: [MacOS]

Demonthos:
Dioxus liveview uses salvo version `0.37.7`. If you downgrade the your salvo version to `0.37`, it should fix the error. If you are interested in contributing to Dioxus, you could bump the salvo version [here](https://github.com/DioxusLabs/dioxus/blob/dfa4d8d989385ceae0233ed6822b21d5e99869bc/packages/liveview/Cargo.toml#L40) and fix any breaking changes in the salvo adapter [here](https://github.com/DioxusLabs/dioxus/blob/dfa4d8d989385ceae0233ed6822b21d5e99869bc/packages/liveview/src/adapters/salvo_adapter.rs)

Demonthos:
Closed by #1130

PR #1130, opened 2023-06-28
Fix: bump liveview salvo version to 0.44.1. (#1124)

xinglixing:
bump liveview salvo version to 0.44.1


Demonthos (approved):
Great, thank you!

PR #1135, opened 2023-06-29
fix: Remove conflicting rustfmt config for cli

marc2332:
Looks like the CLI was using some nightly rustfmt options.


Demonthos (approved):
Thank you!

PR #1138, opened 2023-06-30
Fix CLI testing issue

DogeDark:
The ``spawn_properly`` test requires the ``dioxus`` command to be available. In our CI system, it isn't. This PR comments out ``spawn_properly`` and adds a ``test_auto_fmt`` test. 

We could get the ``spawn_properly`` test working in CI but I think it'd be a little tricky and for now, it'd be nice to get CI working correctly.

\- Also fixes CLI clippy complaints


jkelleyrtp:
Thanks @DogeDark, this is super helpful.

PR #1142, opened 2023-07-02
Change missing trailing comma span to be more helpful

torsteingrindvik:
First time contributing, let me know if something should be different e.g. where I place the test and so on (or if you disagree with the PR altogether).

I noticed if I forget a trailing comma, the span info pointed to the span _before_ the one I expected.

```rust
// user code
rsx!{
  p { class: "foo bar" 
    "hi"
  }
}

// stderr before
error: missing trailing comma
 --> tests/rsx/trailing-comma-0.rs:9:20
  |
9 |             class: "foo bar"
  |             ^^^^^

// stderr after
error: missing trailing comma
 --> tests/rsx/trailing-comma-0.rs:9:20
  |
9 |             class: "foo bar"
  |                    ^^^^^^^^^
```

So I changed the span to the one after, and added a test to verify the new stderr.

### DIFF
--- a/packages/core-macro/Cargo.toml
+++ b/packages/core-macro/Cargo.toml
@@ -22,6 +22,7 @@ dioxus-rsx = { workspace = true }
 
 # testing
 [dev-dependencies]
+dioxus = { workspace = true }
 rustversion = "1.0"
 trybuild = "1.0"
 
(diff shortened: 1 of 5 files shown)


jkelleyrtp:
This is great - even comes with a nice test. Thank you so much!

jkelleyrtp (approved):

PR #1146, opened 2023-07-03
Add a helper macro for properties & async

traxys:
The to_owned macro is helpful to handle hooks with async components, but when using properties it can be problematic because `cx` will be moved in the async block, and will result in a compilation error. This adds a new helper that allows to make either hooks or properties owned.

Note that the macro could be merged with `to_owned`, but having two macros may be better for discoverability (and the to_owned_props may be more expensive at compile time).

### DIFF
--- a/packages/hooks/src/lib.rs
+++ b/packages/hooks/src/lib.rs
@@ -1,24 +1,53 @@
 #[macro_export]
-/// A helper macro for using hooks in async environements.
+/// A helper macro for using hooks and properties in async environements.
 ///
 /// # Usage
 ///
 ///
-/// ```ignore
+/// ```
+/// # use dioxus::prelude::*;
+/// #
+/// # #[derive(Props, PartialEq)]
+/// # struct Props {
+/// #    prop: String,
+/// # }
+/// # fn Component(cx: Scope<Props>) -> Element {
+///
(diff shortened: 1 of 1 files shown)


Demonthos:
Could we just modify the to_owned macro to accept `$first:ident $(remaining:ident)*` instead and set the cloned variable to the last segment.

```rust
to_owned![cx.props.test, cx.props.other.nested, x];
```
Would expand to:
```rust
let test = cx.props.test.clone();
let nested = cx.props.other.nested.clone();
let x = x.clone();
```

traxys:
Would need to try, but this may not work with macro hygene

traxys:
It may be possible to merge the handling of paths & not paths, but I think it needs a bit more macro trickery than what I am able to do.

Demonthos:
I like that approach a lot better! If we need to, we could use a [proc macro](https://doc.rust-lang.org/reference/procedural-macros.html) to parse the macro, but the macro you posted seems to work well without it.

traxys:
I edited the example a bit to make it compile (while hiding the boilerplate) so that it can test the macro

jkelleyrtp (approved):
Looks great, thank you! This is very cool :)

Issue #1122, opened 2023-06-24
Don't close app when last window is closed on macOS


terhechte:
On macOS, the default behaviour for apps is to keep running even if the user closes all windows. If I'm in Safari, and I close all windows, the app is still running and I can go back and open new windows.

Single Window apps usually don't close the window, they just hide it, so that going back to the app opens the window again.

Dioxus, on the other hand, closes the app as soon as no Window is running anymore. This makes sense on other platforms, but on macOS this feels alien. I'd like to propose an addition to the `WindowBuilder` API to allow setting the close behaviour between:

``` rs
enum WindowCloseBehaviour {
  /// Default behaviour, closing the last window exits the app
  LastWindowExitsApp,
  /// Closing the last window will not actually close it, just hide it
  LastWindowHides,
  /// Closing the last window will close it but the app will keep running so that new windows can be opened
  CloseWindow,
}
```

I'd be happy to implement this if this is desired functionality.

Demonthos:
This would be great! On windows, this could also be useful in once we implement [system tray support](https://docs.rs/wry/0.29.0/wry/application/system_tray/index.html).

PR #1150, opened 2023-07-04
Window Close Behaviour

terhechte:
This implements the functionality outlined in the following issue:

https://github.com/DioxusLabs/dioxus/issues/1122

For macOS, it adds the `objc` dependency as that is required to call the correct selector, `[NSApplication hide:]`. This will hide the window in a way that it appears again when the user switches back to the app.

On Linux or Windows clicking the icon in the tab bar should make it visible again.

### DIFF
--- a/examples/window_focus.rs
+++ b/examples/window_focus.rs
@@ -2,9 +2,12 @@ use dioxus::prelude::*;
 use dioxus_desktop::tao::event::WindowEvent;
 use dioxus_desktop::use_wry_event_handler;
 use dioxus_desktop::wry::application::event::Event as WryEvent;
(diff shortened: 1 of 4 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Looks great! Thank you!

Two things:
- Clippy is failing, something related to unused deps on Unix
- I think the example that you show this behavior might be better with `LastWindowHidesApp`

Otherwise, looks great and is something we've been meaning to add for a while!

terhechte:
Thanks for fixing the clippy issues @jkelleyrtp ! I only have time for open source every couple of days, so I couldn't have a look any earlier.

PR #1154, opened 2023-07-04
Fix playwright fullstack clippy

DogeDark:
(It is used, just behind feature flags)
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,7 @@ on:
       - lib.rs
       - Cargo.toml
       - Makefile.toml
+      - playwrite-tests/**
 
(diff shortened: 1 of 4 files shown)


DogeDark:
Not sure why ``Rust CI / Clippy`` action isn't running but that's what this PR needs to verify everything is working.

Demonthos:
> Not sure why `Rust CI / Clippy` action isn't running but that's what this PR needs to verify everything is working.

The playwrite tests are not included in this list of files that trigger the tests: https://github.com/DioxusLabs/dioxus/blob/master/.github/workflows/main.yml#L7

jkelleyrtp:
Thanks!

PR #1156, opened 2023-07-05
Add iOS + Android example to mainline

jkelleyrtp:
Similar to the PWA example, we now integrate iOS into the mainline.

This also features a fix to Dioxus Desktop that broke touch interactions.


jkelleyrtp:
Android is broken for me :(

For whatever reason the `unbounded_send` method on futures-channel just hangs.

All the dioxus stuff on android works fine, but I can't for the life of me debug why that's happening.

It means we can handle events, but we can't seem to call "needs_update" - or any number of other methods. `use_future` also just hangs the rendering entirely.

 Maybe someone with more android knowledge knows more than me here? Do people ship Rust on Android and not run into this?

PR #1159, opened 2023-07-05
Fix waker issue on Android

### LINKED ISSUE #1158: `unbounded_send` in core scheduler hangs in Android
**Problem**

Android is broken for me :(

For whatever reason the unbounded_send method on futures-channel just hangs.

All the dioxus stuff on android works fine, but I can't for the life of me debug why that's happening.

It means we can handle events, but we can't seem to call "needs_update" - or any number of other methods. use_future also just hangs the rendering entirely.

Maybe someone with more android knowledge knows more than me here? Do people ship Rust on Android and not run into this?

**Environment:**
 - Dioxus version: [`master`]
 - Rust version: [1.70]
 - OS info: [macOS]
 - App platform: [`android`]

**Questionnaire**

jkelleyrtp:
Attempts to fix #1158 .

Doing the async + waker hack on that we do on desktop seems to behave weird on Android. I haven't figured it out yet.

This branch attempts to fix the issue.

### DIFF
--- a/examples/mobile_demo/Cargo.toml
+++ b/examples/mobile_demo/Cargo.toml
@@ -40,6 +40,9 @@ dioxus = { path = "../../packages/dioxus" }
 dioxus-desktop = { path = "../../packages/desktop", features = [
     "tokio_runtime",
 ], default-features = false }
+futures-channel = "0.3.28"
+futures-util = "0.3.28"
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Issue ended up being inside tao

https://github.com/tauri-apps/tao/pull/761

PR #1182, opened 2023-07-11
CLI HTTPS

### LINKED ISSUE #1168: Dioxus CLI HTTPs mode
## Specific Demand

Currently when you want to test or develop with certain Web APIs which are only available through HTTPS this isn't possible through Dioxus without perhaps a very complex / painful setup, if Dioxus CLI could ship with HTTPs say through MKCert, this process would be much more straight forward.

## Implement Suggestion

Use MkCert to create a HTTPs and WSS connections as an opt in feature through `Dioxus.toml`

DogeDark:
Adds support for localhost HTTPs using either custom or [mkcert](https://github.com/FiloSottile/mkcert) auto-generated certificates.

This PR adds a few new ``Dioxus.toml`` options:
```toml
[web.https]
enabled = true
mkcert = false
cert_path = "./cert.pem"
key_path = "./key.pem"
```

mkcert generated certificates are stored in an ``ssl`` folder:
image

Closes #1168

### DIFF
--- a/packages/cli/src/config.rs
+++ b/packages/cli/src/config.rs
@@ -146,8 +146,8 @@ pub struct WebDevResourceConfig {
 pub struct WebHttpsConfig {
     pub enabled: Option<bool>,
     pub mkcert: Option<bool>,
-    pub key_path: Option<PathBuf>,
-    pub cert_path: Option<PathBuf>,
+    pub key_path: Option<String>,
+    pub cert_path: Option<String>,
 }
 
 #[derive(Debug, Clone)]
--- a/packages/cli/src/config.rs
+++ b/packages/cli/src/config.rs
@@ -107,6 +107,7 @@ pub struct WebConfig {
     pub proxy: Option<Vec<WebProxyConfig>>,
     pub watcher: WebWatcherConfig,
     pub resource: WebResourceConfig,
+    #[serde(default)]
     pub https: WebHttpsConfig,
 }
 
@@ -141,7 +142,7 @@ pub struct WebDevResourceConfig {
     pub script: Option<Vec<PathBuf>>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct WebHttpsConfig {
     pub enabled: Option<bool>,
     pub mkcert: Option<bool>,
--- a/packages/cli/src/config.rs
+++ b/packages/cli/src/config.rs
@@ -77,6 +77,12 @@ impl Default for DioxusConfig {
                     style: Some(vec![]),
                     script: Some(vec![]),
                 },
+                https: WebHttpsConfig {
+                    enabled: None,
+                    mkcert: None,
+                    key_path: None,
+                    cert_path: None,
+                },
             },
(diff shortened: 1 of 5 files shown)


DogeDark:
This PR also cleans up the ``dx serve`` code, reducing code duplication and separating some stuff into separate files. There's still more to do, but that would be best another time.

packages/cli/src/config.rs
```diff
@@ -77,6 +77,12 @@ impl Default for DioxusConfig {
                     style: Some(vec![]),
                     script: Some(vec![]),
                 },
+                https: WebHttpsConfig {
```
Demonthos, line 80: It looks like this config is unused, and `https://localhost:8080` does not appear to work locally with this branch.

Maybe you forgot to push some changes or some changes that used this got overriten?

DogeDark: I hate to say this but, are you sure it's not something on your end?

Config is used:
https://github.com/DioxusLabs/dioxus/blob/d83b00ff0f4a51e27e4fab33fd1c6d25a7caac0f/packages/cli/src/server/mod.rs#L214

Localhost HTTPs: 
image

No missing commits on my end.

Demonthos: Sorry, I was searching the git diff, but one file was collapsed

DogeDark: If you need it, this cargo install command works targeting my branch.
```
cargo install --debug --git https://github.com/DogeDark/dioxus.git --branch cli-https dioxus-cli
```

packages/cli/src/config.rs
```diff
@@ -101,6 +107,7 @@ pub struct WebConfig {
     pub proxy: Option<Vec<WebProxyConfig>>,
     pub watcher: WebWatcherConfig,
     pub resource: WebResourceConfig,
+    pub https: WebHttpsConfig,
```
Demonthos, line 110: Could this use [#[serde(default)]](https://serde.rs/attr-default.html) so that if no `[web.http]` section is found, it will use the default config? (This would make the PR a non-breaking change)

DogeDark: Absolutely!

Demonthos (approved):
This looks great! Https support will be very useful

itsezc:
It'd be cool if you could configure where mkcert autogenerated files end up, as a configurable path (not sure if thats too big of an ask tho), but once again thanks for working on this! 😊

DogeDark:
Not a bad idea. We could reuse the existing ``cert_path`` and ``key_path`` configs.

DogeDark:
Some examples of config:
```toml
# Enable https with custom certs
[web.https]
enabled = true
cert_path = "mypath/cert.pem"
key_path = "mypath/key.pem"

# Enable https with autogenerated certs, default path
[web.https]
enabled = true
mkcert = true

# Enable https with autogenerated certs, custom path
[web.https]
enabled = true
mkcert = true
cert_path = "mypath/cert.pem"
key_path = "mypath/key.pem"
```

itsezc (approved):

DogeDark:
Found an issue: full-page reloading on changes isn't working. I believe I know the cause. I assume the same issue exists for hot reload.

jkelleyrtp (approved):

PR #1187, opened 2023-07-11
better diagnostics for use_shared_state

Niedzwiedzw:
Currently it is quite hard to debug what causes the panics at runtime, this is an attempt making it a bit less intimidating for new users.
I also propose exposing the `try_*` methods in case someone wanted to handle the errors somehow (current implementation crashes and leaves the application frozen which might be really confusing to end user)

changelist: 
 - expose `try_*` methods on `UseSharedState<T>`
 - add meaningful panic information on `read()`/`write()` errors
 - unify `thiserror`'s versions across all packages and make it a `workspace`-wide dependency
 - above mentioned `read()` / `write()` methods also track callers in debug mode, providing specific location at which borrow error occurred in the panic message


packages/hooks/src/use_shared_state.rs
```diff
@@ -121,30 +121,98 @@ pub struct UseSharedState<T> {
     pub(crate) inner: Rc<RefCell<ProvidedStateInner<T>>>,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum UseSharedStateError {
```
Demonthos, line 125: It would be nice to provide information about the location where the state was previously read and written to in debug mode to provide nicer debug information. (This would use https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute to track the file location of the caller of read and write in debug mode)

(It doesn't need to implemented in this PR)

Niedzwiedzw: oooh I've never seen that one - nice. (pushed)

Niedzwiedzw: pushed another one, so that it's only there in `debug` mode

Niedzwiedzw: oh yeah, tracking itself should also be debug-only - pushed

Demonthos: Apparently what I was thinking of is already implemented in https://docs.rs/debug-cell/latest/debug_cell. It tracks the current borrows and logs the location of any relevant borrows on errors

Niedzwiedzw: hmmm this could almost be a separate feature behind a specific feature flag, idk your call

Demonthos: It is probably better implemented in a different PR, but I think we could enable it in debug mode by default

Niedzwiedzw: want me to do it?

Demonthos: If you are up for it, that would be great! We could also use the same method for the use_ref hook

Niedzwiedzw: ok soo I've done something a bit different instead but with the same result... let me know what you think

Niedzwiedzw: soo the problem is... it doesn't work

```rust
Writing to shared state failed: [/home/niedzwiedz/.cargo/git/checkouts/dioxus-7f2b82817ca79561/760d937/packages/hooks/src/use_shared_state.rs:296:27] dioxus_hooks::use_shared_state::UseSharedState<editor::app::App> is already borrowed at [None], so it cannot be borrowed mutably.
```

as you can see it shows the spot inside `hooks/src/use_shared_state.rs` ...

Niedzwiedzw: ok it doesn't actually work, it only shows the thing that was previously borrowed but it could've been dropped in the meantime. I guess there's no way around it other than rolling a custom RefCell that tracks borrows but also when they are released

Niedzwiedzw: https://github.com/alexcrichton/debug-cell/pull/5 made a PR to `debug-cell` to have it expose the `try_*` methods on debug RefCell, let's hope it's not too much to ask 🤞🏿

Demonthos (approved):
This looks great! The new error is so much more helpful.

We need to either wait for the PR you opened to be merged or publish a forked version before this PR is merged. (I don't think we can have git dependancies in the crates.io version of Dioxus)

Niedzwiedzw:
I would suggest forking debug-cell into DioxusLab for the time being, this repository doesn't have much traffic sadly

Niedzwiedzw:
@jkelleyrtp  it still points to my fork of debug-cell... is this a good idea? I don't want to set a precedence like that...

PR #1197, opened 2023-07-14
Rip out async components in favor of a simpler suspense

jkelleyrtp:
This PR rips out async components, once again enabling us to use anything that implements `IntoIterator<T: IntoTemplate>`. It makes our suspense system simpler, with the limitation that it's really only useful for SSR.

Loading states are not implemented, so components will be blank until their data is ready.

This is being done in a forward-compatible way, so we can add loading states and tree-based suspense later into the 0.4 release.

### DIFF
--- a/packages/core/src/scope_arena.rs
+++ b/packages/core/src/scope_arena.rs
@@ -25,10 +25,10 @@ impl VirtualDom {
             name,
             props: Some(props),
             tasks: self.scheduler.clone(),
-            placeholder: Default::default(),
             node_arena_1: BumpFrame::new(0),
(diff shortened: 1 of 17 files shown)


Demonthos (approved):

packages/core/src/scope_arena.rs
```diff
@@ -144,6 +83,12 @@ impl VirtualDom {
             id: scope.id,
         });
 
+        if scope.suspended.get() {
```
Demonthos, line 86: Should this only run when the rendered value is none?

jkelleyrtp: Good point, yes

PR #1198, opened 2023-07-15
CLI Bin Flag

### LINKED ISSUE #1190: Use Cargo workspace member from CLI
## Specific Demand

Monorepos are pretty popular currently many people (like myself), would like to setup multiple dioxus projects within the same repository, Cargo does this via the `--bin` flag

## Implement Suggestion

Given that a repository might have multiple Dioxus projects, we should be able to run them via `dioxus serve --bin <name>` and fallback to the current directroy if its not a workspace or a workspace member.

DogeDark:
Adds a ``--bin`` flag to the ``clean``, ``build``, and ``serve`` commands.
E.g. ``dx serve --bin desktop``

This PR also un-folder-nests the commands.

Closes #1190


packages/cli/src/config.rs
```diff
@@ -20,8 +20,19 @@ fn default_plugin() -> toml::Value {
 }
 
 impl DioxusConfig {
-    pub fn load() -> crate::error::Result<Option<DioxusConfig>> {
-        let Ok(crate_dir) = crate::cargo::crate_root() else { return Ok(None); };
+    pub fn load(bin: Option<PathBuf>) -> crate::error::Result<Option<DioxusConfig>> {
```
Demonthos, line 23: I'm somewhat worried this will be confused with how `cargo run --bin` works. With cargo run --bin, it looks at the workspace and finds the binary with the specified name instead of accepting a path

DogeDark: We'd have to intentionally check if the bin exists in the workspace members otherwise accepting a String would have the same result: ``path.join(bin)``

Would you want to:
- Fix the behavior now (I would probably need to pull in cargo_toml crate)
- Wait until later (I have wip stuff that will probably reiterate on this but it'll be some time)
- Push it out with current behavior and wait for next PR that reiterates on it
- Other?

Demonthos: Great! API looks good now. One last thing, could you add something about the command to the docs (here)[https://github.com/DioxusLabs/dioxus/blob/master/packages/cli/docs/src/cmd/serve.md]

DogeDark: Done 👍

Demonthos (approved):
Thank You!

jkelleyrtp:
Yeet

PR #1200, opened 2023-07-15
fix(cli): improve error message for invalid config

eventualbuddha:
Previously an invalid `Dioxus.toml` file would yield this error when running e.g. `dx serve`:

```
Error: Failed to load `Dioxus.toml` because: Dioxus.toml parse failed
```

This doesn't give any indication why it failed to parse. This commit updates it to include why the parsing failed, e.g.:

```
Error: Failed to load Dioxus config because: Dioxus.toml missing field `name` for key `application` at line 38 column 1
```

I initially had it format the message to include `Dioxus.toml:38:1` to be clickable in some terminals, but the location specified didn't always seem particularly relevant to the actual problem so I left it as-is.

This also fixes what I believe would be an issue on case-sensitive file systems where if `dioxus.toml` existed it would try to read `Dioxus.toml`.

I'm still fairly new to Rust, so the `.as_path()` calls may not be the best way to deal with the fact that the borrow checker wouldn't let me re-use `crate_dir` and `dioxus_conf_file`. I'm open to suggestions!

### DIFF
--- a/packages/cli/src/config.rs
+++ b/packages/cli/src/config.rs
@@ -1,6 +1,9 @@
 use crate::error::Result;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct DioxusConfig {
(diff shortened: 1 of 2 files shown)


Demonthos (approved):
This is much better! `as_path` works here, you could also use `&*path` to get the same effect because PathBuf implements [Deref<Path>](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#deref-methods-Path)

Thank You!

PR #1203, opened 2023-07-16
chore: rename `playwrite` to `playwright`

eventualbuddha:
I assume this was a mistake and not intentional.


Demonthos (approved):
Not intentional, Thank You!

PR #1208, opened 2023-07-17
Fixed typo in Router guide getting-started.md

leahiel:
PR is largely self-explanatory.

The [Getting Started page for the Dioxus Router](https://dioxuslabs.com/docs/0.3/router/guide/getting-started.html) says we should use:
`rustup target add wasm32-unkown-unknown`, however the correct target is
`rustup target add wasm32-unknown-unknown`.

This PR adds the missing `n` to the first `unknown`.


Demonthos (approved):
Thank You!

PR #1209, opened 2023-07-18
Make README less inadvertently hilarious

squidsoup:
"no-std" and "non-std" have _very_ different meanings (insert crabs joke here 🦀 ).


jkelleyrtp:
I think it might be better to maybe make it explicit that `no-std` is the Rust no-std, perhaps by putting the word `no-std` in brackets/codeblock:

`#[no-std]`

PR #1210, opened 2023-07-18
Move desktop hot reload into the CLI

Demonthos:
This PR moves hot reloading from the hot_reload_init macro to the CLI.

This has a few advantages:
- Hot reloading can detect changes when rebuilding the application
- All platforms use the CLI to serve applications


jkelleyrtp:
Does this remove the old macros?

Demonthos:
The old macros are still enabled to allow Freya to use hot reloading. The new CLI support assumes that you use the HTML elements defined in the `dioxus-html` crate. Future PRs could add support for defining a schema for the CLI to use or shift decoding attributes to the renderer.

PR #1215, opened 2023-07-18
feat(check): adds `dx check`

### LINKED ISSUE #1202: Create a lint for the rules of hooks
## Specific Demand

It would be nice if the Dioxus CLI could report when the [rules of hooks](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/hooks.html) are not being followed. Eventually we could expand this to lint other [antipaterns](https://dioxuslabs.com/docs/0.3/guide/en/best_practices/antipatterns.html) like using an index as an iterator key.

## Implement Suggestion

Eventually adopting some of the tooling from the [Clippy tool](https://github.com/rust-lang/rust-clippy) would be helpful to add lints.

For the initial implementation we can use [syn](https://docs.rs/syn/latest/syn/) to search for any functions that start with `use_` that are being used within a if statement of loop. (Specifically the [Visit](https://docs.rs/syn/latest/syn/visit/trait.Visit.html) trait could be useful here).

That will only allow us to detect issues with the current crate, but that is ok for a MVP.

We could run linting before the build and serve commands and also add a new `dioxus check` command that just runs the linting.

eventualbuddha:
At the moment this only checks the [Rules of Hooks](https://dioxuslabs.com/docs/0.3/guide/en/interactivity/hooks.html
), ensuring that hook functions (i.e. `use_*`) are being called as expected.

TODO:
- [x] more tests
- [ ] add Rules of Hooks link to issue report output?

image

Closes #1202


Demonthos (approved):
This looks amazing, thank you!

There are a few things we can improve on in the future, but I think they are better addressed in future PRs so that we can publish this in 0.4:
- Reading mod statements to traverse only paths that cargo will use in the build
- VScode integration to show lints inline
- Allowing closures that start with use_* and act like normal hooks. These should be the same as normal hooks. Depending on how difficult this is with the current system, it could be part of this PR or not.

(Waiting for a second review before merging)

eventualbuddha:
> * Allowing closures that start with use_* and act like normal hooks.

This is done. I agree the other things should be separate PRs.

jkelleyrtp:
No issues, this is awesome. Clean, concise, reusable, and we can likely pop some of the functionality out into the vscode extension later on.

Thanks so much!

PR #1219, opened 2023-07-19
Halve Build Times

Demonthos:
This roughly halfs build times for a hello world example.

Relies on https://github.com/DioxusLabs/prettier-please/pull/1 being published as the `0.2` version


packages/autofmt/src/component.rs
jkelleyrtp, line 1: This should be unparse_expr

jkelleyrtp (changes requested):

jkelleyrtp:
Errr... looks like the multine pr #1224 has a merge conflict here

PR #1223, opened 2023-07-19
Fix race condition in use_future by using use_state directly.

jkelleyrtp:
We have a lot of vestigial code from when use_future was meant to work with async components.

This cleans all that up and relies on use_state directly, since use_future is basically a glorified use_state.


jkelleyrtp:
This needs a test before we can merge

PR #1224, opened 2023-07-20
Improve auto formatting multiline raw attributes/event handlers

Demonthos:
Fixes a few issues I ran into while trying to get autofmt to work for the docsite. Mainly handlers with multiple lines and multi-line raw attributes were not formatted correctly


jkelleyrtp:
Nice!

PR #1225, opened 2023-07-20
Fix event bubbling without phantom nodes

jkelleyrtp:
Note that I haven't tested this too much, but it seems like old me meant to use this approach but forgot to finish implementing it.

Note that this does not currently pass miri as the vnode pointer gets invalidated between renders. But it's a start...


Demonthos:
Does this handle cases where you have nested Vnodes being created in one scope? (like bubbling through iterators):

```rust
rsx!{
    div { // div 1
        for _ in 0..10 {
            div { // div 2
                onclick: |evt| {}
            }
        }
    }
}
```
Will an event in `div 2` bubble to `div 1`

PR #1226, opened 2023-07-20
Fix: dont use bumpslab anymore, just box scopestates

jkelleyrtp:
While working on #1225 I noticed that miri isn't passing in one of our tests. I think it's a new regression.

This change expands our CI to run miri on every core test and removes the bumpslab in favor of simply boxing each scope state.

We take a slight performance hit but I don't think it matters too much. And now, we know that scopestates properly drop their contents.


Demonthos (approved):

jkelleyrtp:
Note that the only reason we use stable pointers is because we need to traverse the context tree. In theory, we could get away allowing contexts via threadlocals and not need stable pointers anyways.

PR #1228, opened 2023-07-20
Update license field following SPDX 2.1 license expression standard

frisoft:
The new recommendation is to follow the SPDX 2.1 standard. This allows automatic license verification software to work properly. Reference: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields


jkelleyrtp:
Thanks for the PR!

PR #1233, opened 2023-07-21
Log errors when a component fails to render

Demonthos:
When a component fails to render on desktop/ssr, dioxus will catch the error and render a placeholder. This PR adds a log::error to let the developer know what component failed to render and what the error was

### DIFF
--- a/packages/core/src/any_props.rs
+++ b/packages/core/src/any_props.rs
@@ -67,7 +67,7 @@ unsafe impl<'a, P> AnyProps<'a> for VProps<'a, P> {
             Ok(Some(e)) => RenderReturn::Ready(e),
             Ok(None) => RenderReturn::default(),
             Err(err) => {
-                let component_name = std::any::type_name::<P>();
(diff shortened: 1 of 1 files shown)


packages/core/src/any_props.rs
```diff
@@ -65,7 +65,12 @@ unsafe impl<'a, P> AnyProps<'a> for VProps<'a, P> {
 
         match res {
             Ok(Some(e)) => RenderReturn::Ready(e),
-            _ => RenderReturn::default(),
+            Ok(None) => RenderReturn::default(),
+            Err(err) => {
+                let component_name = std::any::type_name::<P>();
```
jkelleyrtp, line 70: This will render the props name, not the component name. Usually the same, but not always. Every component has a name though, available via scopestate.

Screenshot 2023-07-21 at 2 17 31 PM

jkelleyrtp (changes requested):

PR #1238, opened 2023-07-23
fix: Use correct cfg for file_watcher feature in dioxus-hot-reload

marc2332:
- Made `Config` public again
- Fix import in file_watcher.rs
- Use correct cfg syntax for `file_watcher` feature


Demonthos (approved):
Thank you

PR #1250, opened 2023-07-25
Fix root component suspense

Demonthos:
Currently, the Vdom panics when the root component is suspended. This PR treats the root component like any other aborted component instead.

This should let us simplify our examples and docs a bit


jkelleyrtp:
Easy fix, thanks

PR #1254, opened 2023-07-26
Create cli bundle command

Demonthos:
Creates a bundle command for the CLI desktop targets.

Currently this has only been confirmed to work for Windows (x64) and MacOs (arm). Linux targets have not been tested.


jkelleyrtp:
Sweet!

PR #1255, opened 2023-07-26
Make more of core private

Demonthos:
There are a few places in core where we expose more of the API of the nodes than we need to for the renderers. This PR makes a few more things private to reduce breaking changes in the future


jkelleyrtp:
Great, thank you!

Issue #1081, opened 2023-06-09
Expose `web_sys::File` via `FileEngine`


wooden-worm:
## Specific Demand

On wasm I need to access the raw `web_sys::File` object in order to deal with big files (bigger than available memory), however, `FileEngine` only provides `read_to_string` and `read_to_bytes`.

## Implement Suggestion

- Gated `#[cfg...]` on `wasm` to return `web_sys::Filie`
- Extension trait as how Rust handle platform specific functionalities - example: https://doc.rust-lang.org/std/os/windows/io/trait.AsHandle.html

Demonthos:
This would be a great addition!

Dioxus does not include `#[cfg...]` statements for each platform so that other external renderer can be implemented. If we did take the config approach it would prevent a renderer like [freya](https://github.com/marc2332/freya) from using the html crate with it's own file engine type.

Instead we could expose the inner type as Any type and allow users to downcast it to a platform specific type. We do something similar to this with the [onmounted event](https://github.com/DioxusLabs/dioxus/blob/master/packages/html/src/events/mounted.rs#L19). We could expose an extension trait on each platform to make downcasting to the platform type more easy:
```rust
#[async_trait::async_trait(?Send)]
pub trait FileEngine {
    // get a list of file names
    fn files(&self) -> Vec<String>;

    // read a file to bytes
    async fn read_file(&self, file: &str) -> Option<Vec<u8>>;
    
    async fn read_native_file(&self, file: &str) -> Option<Box<dyn Any>>;

    // read a file to string
    async fn read_file_to_string(&self, file: &str) -> Option<String>;
}

trait WebFileEngineExt {
    fn read_web_file(&self, file: &str) -> Option<web_sys::File>;
}

impl WebFileEngineExt for std::sync::Arc<dyn FileEngine> {
    fn read_web_file(&self, file: &str) -> Option<web_sys::File> {
        self.read_native_file().map(|file| *file.downcast())
    }
}
```

Demonthos:
also related to #93

wooden-worm:
I can work on a PR if the above approach is acceptable, let me know.

Demonthos:
> I can work on a PR if the above approach is acceptable, let me know.

That would be great!

PR #1258, opened 2023-07-27
FileEngine: expose native file types on different platforms

wooden-worm:
Impl `get_native_file` to return the native file type on different platforms: `web_sys::File` on wasm and `tokio::fs::File` on desktop.

Fixes https://github.com/DioxusLabs/dioxus/issues/1081


Demonthos (approved):
Thank you, this looks great!

PR #1280, opened 2023-08-03
Added use_read_silent fn

serzhiio:


packages/fermi/src/hooks/read.rs
```diff
@@ -34,3 +34,26 @@ pub fn use_read_rc<V: 'static>(cx: &ScopeState, f: impl Readable<V>) -> &Rc<V> {
...
+///
+/// Example:
+/// ```ignore
+/// static ATOM: Atom<bool> = Atom(|_| false);
+/// fn App(cx: Scope) {
+///     use_init_atom_root(cx);
+///     cx.render(rsx!(
+///         button{
+///             onclick: || {
+///                 log::info!("Atom's value is {}", use_read_silent(cx, &ATOM))
```
Demonthos, line 51: You cannot use hooks inside of an event listener because the hook will not be run every time the component runs. Instead it will only be run when the event triggers. This violates the [rules of hooks](https://dioxuslabs.com/learn/0.4/guide/state#the-rules-of-hooks)

PR #1288, opened 2023-08-03
Make launch functions pub

CodedNil:
I am currently working on a project where I need to run server-side code concurrently with Dioxus Fullstack. However, I noticed that there is no way to reuse the Tokio runtime created in the launch function. As a result, the only option available to me right now is to spawn a new thread and runtime, which seems inefficient.

I propose making the launch functions public. This would allow developers like me to copy the launch function's code into our codebases and manipulate the Tokio runtime as needed, such as spawning new functions onto it. This change would provide more flexibility and efficiency in running concurrent server-side code.

### DIFF
--- a/packages/fullstack/src/launch.rs
+++ b/packages/fullstack/src/launch.rs
@@ -119,13 +119,13 @@ impl<Props: Clone + serde::Serialize + serde::de::DeserializeOwned + Send + Sync
     }
 
     #[cfg(feature = "web")]
-    fn launch_web(self) {
+    pub fn launch_web(self) {
         let cfg = self.web_cfg.hydrate(true);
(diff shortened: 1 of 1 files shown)


packages/fullstack/src/launch.rs
```diff
@@ -119,21 +119,21 @@ impl<Props: Clone + serde::Serialize + serde::de::DeserializeOwned + Send + Sync
     }
 
     #[cfg(feature = "web")]
-    fn launch_web(self) {
+    pub fn launch_web(self) {
```
Demonthos, line 122: If we make these public, we need to add documentation to the function. Other than that, this looks like a good change!

Demonthos (approved):

PR #1289, opened 2023-08-04
Minor Proxy Improvements

eventualbuddha:
Updates a mistake in the docs and provides a better error message for a pathless proxy URL. Without this, `dx serve` panics with this message:

```
Paths must start with a `/`. Use "/" for
 root routes
 ```

 That's not very clear. Instead, we can detect this situation and provide a better error message:

```
Error: 🚫 Serving project failed: Failed to establish proxy: Proxy backend URL must have a non-empty path, e.g. http://localhost:8080/api instead of http://localhost:8080
```


Demonthos (approved):
thank you

PR #1299, opened 2023-08-04
docs(hooks): fix typos

eventualbuddha:
I'm not totally sure how the sentence was supposed to read, but removing "state" from "It state" seemed to work. Also, "it's" is "it is" and not possessive, so I fixed that.


Demonthos (approved):
Good catch, this is what is should have said

PR #1300, opened 2023-08-04
Complete Signals implementation

Demonthos:
This is a more complete implementation of Copy state with local subscriptions (signals).

There are three main components that come together to make signals work:

# Copy State Management

This PR creates a new crate for copy state management. It essentially makes lifetimes dynamic. The implementation is kind of like a global generational allocator for rust. It is a bit complex, but the code contains no `unsafe`!

```rust
// This stores the actual data
let store = Store::new();

// This handles dropping any copy data created by this owner when it goes out of scope
let owner = Owner::new();

// This data is Copy, even though the inner data is not. When Owner is dropped, this data will be removed
let data = owner.insert(String::from("Look ma, no clones!"));
let first = *data;
let second = *data;

println!("{}", first.read());

*second.write() += 1;

println!("{}", first.read());

drop(owner);
// This will panic
println!("{}", first.read());
```

# Local Subscriptions and Selectors

Currently, state management is somewhat complex:

Local state:
- use_state: basic local state primitive because of how the hook is structured it allows for Deref<T> for copy data unlike use_ref

- use_ref: very similar to use state, but uses generations instead of PartialEq for diffing when passed as a prop, or used in a memorized hook which can be more performant with large structures

Global state management:
- use_context: global immutable state or tree-local state, but does not allow for changing the data
- Fermi: global mutable state. This will automatically subscribe components when you read the data, unlike use_state 

None of these options are performant with larger, global data (like [this](https://github.com/Satellite-im/Uplink/blob/90f5a0c83e3a104876fc0251824814734932078c/common/src/state/mod.rs#L67)). When you write to any of these states, all components that read from any part of the state will rerun.

## Solutions

There are essentially two different approaches different frameworks have tried to address this issue:

Selectors:

([this PR](https://github.com/DioxusLabs/dioxus/pull/562), [use_selector](https://docs.rs/dioxus-hooks/latest/dioxus_hooks/fn.use_selector.html))

Selectors are a smarter version of [use_memo](https://docs.rs/dioxus-hooks/latest/dioxus_hooks/fn.use_memo.html). They allow a component to subscribe to a specific part of a broader state:

```rust
use dioxus::prelude::*;

fn Parent(cx: Scope) -> Element {
   let counts = use_tracked_state(cx, || vec![0]);

   render! {
       Child {
           counts: counts.clone(),
       }
   }
}

#[inline_props]
fn Child(cx: Scope, counts: Tracked<usize>) -> Element {
   // *counts[0] < 5 will rerun whenever counts is changed, but the component will only rerun when the result is different
   let less_than_five = use_selector(cx, counts, |counts| *counts[0] < 5);

   render! {
       "{less_than_five}"
   }
}
```

Signals:

Signals are like selectors, but they allow you to only rerun a selector when a specific part of your state changes

```rust
use dioxus::prelude::*;

fn Parent(cx: Scope) -> Element {
   let counts = use_signal(cx, || vec![Signal::new(0)]);

   render! {
       Child {
           counts: counts.clone(),
       }
   }
}

#[inline_props]
fn Child(cx: Scope, counts: Tracked<usize>) -> Element {
   // *counts[0].read() < 5 will rerun whenever counts[0] is changed (because we read from the first element), but the component will only rerun when the result is different
   // It will not rerun if a new element is added to counts or counts[1] is changed
   let less_than_five = use_selector(cx, counts, |counts| *count[0].read() < 5);

   render! {
       "{less_than_five}"
   }
}
```

This PR completes the dioxus-signals crate. If you use signals, your state management is simplified to:

Local state:
- use_signal: basic local state that is copy

Global state management:
- signals through the context API
- signals in a [LazyCell](https://doc.rust-lang.org/std/cell/struct.LazyCell.html)

# Global Scope APIS

To make local subscriptions work, we need to make some scope subscription APIs global.

Currently, you can subscribe to a scope with [scope_id](https://docs.rs/dioxus-core/latest/dioxus_core/struct.ScopeState.html#method.scope_id), and [schedule_update_any](https://docs.rs/dioxus-core/latest/dioxus_core/struct.ScopeState.html#method.schedule_update_any) or just [schedule_update](https://docs.rs/dioxus-core/latest/dioxus_core/struct.ScopeState.html#method.schedule_update). Generally hooks need access to the scope when they are crated:

```rust
use_state(&cx, || 0)
```

We need to be able to create signals without scopes to allow signals to be created in futures:

```rust
// Requiring cx here is OK
let state = use_signal(&cx, || Vec::new());

use_future(&cx, (), |_| async move {
    // We cannot access cx here
    state.push(Signal::new(0));
});
```

Because of this we need to make methods like `schedule_update`, `consume_context`, and `provide_context` global.

TODO:
- [x] Handle global context
  - [x] In Components
  - [x] In Futures
  - [x] In Event Handlers
- [x] Handle switching runtimes when waiting for suspense
- [x] Handle switching runtimes in liveview
- [x] Create Effects
- [x] Create Selectors
- [x] Document Final API


jkelleyrtp:
Going to merge this but then test drive it for a little while before we do a full release.

PR #1301, opened 2023-08-04
chore(cli): remove unused function

eventualbuddha:
I believe that technically this function is part of the public API via `dioxus_cli::cli::cfg::parse_public_url`. However, I don't see any evidence that this function is actually used inside or outside this crate. The last use within the crate was removed in January 2022 (9bb10c65a3cbabef127f6af5c0d1eadab1d01531).

### DIFF
--- a/packages/cli/src/cli/cfg.rs
+++ b/packages/cli/src/cli/cfg.rs
@@ -101,13 +101,6 @@ pub enum Platform {
     Desktop,
 }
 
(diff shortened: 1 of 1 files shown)


Demonthos (approved):
Thank you! (This is technically a breaking change, waiting for a second review before merging)

Esales301080:
Good

PR #1303, opened 2023-08-04
docs(cli): update subcommand comments

eventualbuddha:
A bunch of these were copy-pasted from `build`. This commit updates them to be the same as what their `dx help` string is.


Demonthos (approved):
Thank you!

PR #1305, opened 2023-08-05
fix(cli): prevent "zip slip" vulnerability in tool extract

eventualbuddha:
Using the name directly from a zip archive's entry and writing to it is a potential security vulnerability. More information about the vulnerability can be found here: https://security.snyk.io/research/zip-slip-vulnerability and here https://docs.rs/zip/latest/zip/read/struct.ZipFile.html#warnings.

It looks like the previous version maybe tried to remove `\` characters to prevent this, but only in directory paths?

For posterity, I created the zip files like so:

```python
import zipfile
z = zipfile.ZipFile('dangerous.zip', 'w')
z.writestr('/etc/passwd', 'danger!')
z.close()
```


jkelleyrtp (approved):

jkelleyrtp:
Thanks for finding this. Really appreciate the security analysis!

Will release soon.

PR #1307, opened 2023-08-05
refactor(cli): use more semantic `Instant::elapsed`

eventualbuddha:
Plus, one of these was incorrectly subtracting the end time from the start time.


Demonthos (approved):
Thank you!

PR #1308, opened 2023-08-05
incorrect word spelling

dylanblokhuis:
futher -> further according to https://word.tips/spelling/further-vs-futher/


Demonthos (approved):
Thank you

Issue #1291, opened 2023-08-04
The default behavior of `index_on_404` varies depending on if a `Dioxus.toml` is present


swpecht:
**Problem**

The default behavior of `index_on_404` varies depending on if a `Dioxus.toml` is present.

This causes multi-page apps to stop working when implementing the tailwind example.

**Steps To Reproduce**

Run `dx serve` for a project without a `Dioxus.toml`

Add the `Dioxus.toml` from the [tailwind example](https://dioxuslabs.com/learn/0.4/cookbook/tailwind) and run `dx serve` again. 

**Expected behavior**

Expect the `index_on_404` setting to be consistent across runs.

**Screenshots**

**Environment:**
 - Dioxus version: 0.4
 - Rust version: 1.71.0
 - OS info: Ubuntu
 - App platform: `web`

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

PR #1326, opened 2023-08-08
Use serde default instead of options in the CLI

ealmloff:
Changes several fields in the CLI that were optional only to make serde work correctly. Instead this PR uses the `serde(default)` flag to insert a default value when no other value is found

Closes https://github.com/DioxusLabs/dioxus/issues/1291


ealmloff:
This has been merged into https://github.com/DioxusLabs/dioxus/pull/1622

PR #1343, opened 2023-08-12
DOC: Add note about form prevent_default behavior.

vgobbo:
Starting on version 0.4 (PR #926), the `prevent_default` behavior inverted for forms in `dioxus-web`. This just adds a note to the documentation to document this.

### DIFF
--- a/docs/guide/src/en/interactivity/event_handlers.md
+++ b/docs/guide/src/en/interactivity/event_handlers.md
@@ -51,6 +51,8 @@ Any event handlers will still be called.
 
 > Normally, in React or JavaScript, you'd call "preventDefault" on the event in the callback. Dioxus does _not_ currently support this behavior. Note: this means you cannot conditionally prevent default behavior based on the data in the event.
 
+> Note for web developers: on Javascript the default behavior is to **submit the form**, while on Dioxus the default behavior is **to not submit the form**. Consequently, specifying `prevent_default: "onsubmit"` will cause the form to be submitted.
+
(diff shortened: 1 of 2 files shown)


docs/guide/src/en/interactivity/event_handlers.md
```diff
@@ -51,6 +51,8 @@ Any event handlers will still be called.
 
 > Normally, in React or JavaScript, you'd call "preventDefault" on the event in the callback. Dioxus does _not_ currently support this behavior. Note: this means you cannot conditionally prevent default behavior based on the data in the event.
 
+> Note for web developers: on Javascript the default behavior is to **submit the form**, while on Dioxus the default behavior is **to not submit the form**. Consequently, specifying `prevent_default: "onsubmit"` will cause the form to be submitted.
```
Demonthos, line 54: Note that this should only prevent the default form behavior if a listener is attached to the element. If there is no listener, the form should act as normal

vgobbo: Thank you for noticing and point it out, and sorry for missing it.

I rephrased the note in both languages. Hope it is correct and clear now.

Thanks for reviewing it!

Demonthos (approved):
Looks good, thank you!

Demonthos (approved):

arqalite:
IMO this needs to be added to the public documentation ASAP, it's a breaking change that will cause issues for anyone trying to port their apps from 0.3 to 0.4.

PR #1344, opened 2023-08-12
Use `--profile` to identify cargo build artifact if present for dx cli web builds

### LINKED ISSUE #1338: dioxus-cli doesn't support `--profile` flag for web builds
**Problem**

Running `dx build --profile my-profile` does not copy the the correct output to the dist folder.

The build command seems to be properly run with the profile ([source](https://github.com/DioxusLabs/dioxus/blob/f80c3c9ffb2e7d28a8c409dbbb58530d6d7d929c/packages/cli/src/builder.rs#L71C1-L76C7)).

But the cli will only copy artifacts from the release or debug targets: [source](https://github.com/DioxusLabs/dioxus/blob/f80c3c9ffb2e7d28a8c409dbbb58530d6d7d929c/packages/cli/src/builder.rs#L101C4-L109C7)

**Steps To Reproduce**

Run `dx build` with a custom profile.

**Expected behavior**

The artifact built with the custom profile ends up in the dist folder. Instead, the debug build is used.

**Screenshots**

**Environment:**
 - Dioxus version: v0.4, nightly
 - Rust version: 1.71
 - OS info: Ubuntu
 - App platform: `web`

**Questionnaire**

swpecht:
Currently, `dx build` for web will only pull artifacts from `wasm32-unknown-unknown/release` or `wasm32-unknown-unknown/debug`.

This change updates the dx cli to use the `--profile` argument if present to pull the correct build: `wasm32-unknown-unknown/{profile}`.

closes #1338

### DIFF
--- a/packages/cli/src/builder.rs
+++ b/packages/cli/src/builder.rs
@@ -93,18 +93,21 @@ pub fn build(config: &CrateConfig, quiet: bool) -> Result<BuildResult> {
     // [2] Establish the output directory structure
     let bindgen_outdir = out_dir.join("assets").join("dioxus");
 
-    let release_type = match config.release {
-        true => "release",
-        false => "debug",
+    let build_profile = if config.custom_profile.is_some() {
+        config.custom_profile.as_ref().unwrap()
+    } else if config.release {
+        "release"
+    } else {
+        "debug"
     };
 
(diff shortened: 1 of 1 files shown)


Demonthos (approved):
Thank you!

PR #1349, opened 2023-08-13
feat: props spread

### LINKED ISSUE #1179: Extensible Element & EventHandler generics
## Specific Demand

For people building design systems, it's useful to allow users to pass event handlers and other relevant attributes on to your components for instance a `button` might have an `onClick`, `disabled`, `onMouseover` among many other events, which are different for every type of element. This is very important for the DX of both end users and design system engineers alike.

Users of components with such generic, should be able to pass any event that extends the elements possible event types. This functionality could also be extended onto Blitz (down the line), allowing for a feature parity with the web platform.

## Implement Suggestion

Dioxus should provide an abstraction to easily allow one or multiple EventHandlers onto the underlying component easily, this maybe through providing EventHandlers per element type i.e. `use dioxus::elements::ButtonAttributes`.

React is an amazing example to look at when implementing this to Dioxus, a good example:

```ts
import React from 'react';

interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  title: string;
  showIcon: boolean;
}

const Button: React.FC<ButtonProps> = ({ title, showIcon, ...props }) => {
  return (
    <button {...props}>
      {title}
      {showIcon && <Icon/>}
    </button>
  );
};
```

atty303:
(I'll write more details later.)

fixes #1179


packages/dioxus/tests/props_spread.rs
```diff
@@ -0,0 +1,121 @@
...
+            // rsx! {
+            //     audio {
+            //         muted: muted,
+            //     }
+            // }
+            ::dioxus::core::LazyNodes::new(move |__cx: &::dioxus::core::ScopeState| -> ::dioxus::core::VNode   {
+                static TEMPLATE: ::dioxus::core::Template = ::dioxus::core::Template { name: concat!(file!(), ":", line!(), ":", column!(), ":", "" ), roots: &[::dioxus::core::TemplateNode::Element { tag: dioxus_elements::audio::TAG_NAME, namespace: dioxus_elements::audio::NAME_SPACE, attrs: &[::dioxus::core::TemplateAttribute::Dynamic { id: 0usize }], children: &[] }], node_paths: &[], attr_paths: &[&[0u8]] };
+                let mut attrs = vec![__cx.attr(dioxus_elements::audio::muted.0, muted, dioxus_elements::audio::muted.1, dioxus_elements::audio::muted.2)];
+                for attr in attributes {
+                    attrs.push(__cx.attr(attr.name, attr.value.into_value(__cx.bump()), attr.namespace, attr.volatile));
```
atty303, line 101:
```
error[E0161]: cannot move a value of type `dyn dioxus::prelude::IntoAttributeValue<'_>`
   --> packages\dioxus\tests\props_spread.rs:101:53
    |
101 |                     attrs.push(__cx.attr(attr.name, attr.value.into_value(__cx.bump()), attr.namespace, attr.volatile));
    |                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the size of `dyn dioxus::prelude::IntoAttributeValue<'_>` cannot be statically determined
```

I am stuck with this problem and can't solve it.
Is there a good way to create dynamic_attrs from the list of attributes that the Builder holds?

Demonthos: Instead of creating a new AttributeBox type can we just use the existing Attribute struct? You will need to modify the builder to take `cx: &ScopeState` as a parameter when you first create it. Then you can use it in the `push_attribute` function to create an attribute like the attr function

Demonthos: We would need to change the properties trait to take ScopeState as well which would be breaking. All of the implementers of that trait should be through the derive macro so I don't think breaking that API is an issue

atty303: Thanks for the advice, I was able to build `Vec<Attribute<'a>>` with Builder!

Now I am having trouble with the `VNode` where I set `dynamic_attrs: &'a [Attribute<'a>]`.

```
rsx! {
    audio {
        muted: dynamic_value,.
        ... .cx.props.attributes
    }
}
```

I think I need to merge `muted` and `cx.props.attributes` in the above macro into one slice and set it to `dynamic_attrs`. I know this is possible by implementing `Copy` on `Attribute`, but I want to implement it without copying. Any ideas? I will continue to investigate bumpalo and others.

Demonthos: The order of the `dynamic_attrs` array determines what element the attribute is attached to in the template ([docs](https://dioxuslabs.com/learn/0.4/contributing/walkthrough_readme#the-rsx-macro)). Because of this we need the size and order to be fixed at compile time.

We need to change [dynamic_attrs](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L63C34-L63C34) from `dynamic_attrs: &'a [Attribute<'a>]` to `dynamic_attrs: &'a [AttributeEnum<'a>]` where `AttributeEnum` is either `SingleAttribute` or `Vec<Attribute<'a>>`.

Then when we [create](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/create.rs#L321) and [diff](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/diff.rs#L166) attributes, we need to handle that new case by looping over the spread attributes and setting and diffing each one individually.

atty303: Ah, I understand now! It will take some time because of the large impact, but we will proceed with the implementation in that direction. Thanks!

ealmloff: I would like to get this merged in the next few weeks to unblock some event handler changes. @atty303 is this something you are still interested in working on? If not, that is completely understandable, would you be ok with me finishing this PR?

tigerros:
This would be really useful for #1483

ealmloff:
This should be working now with a few caveats:
- The builder macro is getting very large. It works, but it is difficult to work in. We should pull it out into multiple files. Because this PR unblocks a UI toolkit, I don't think this should be changed in this PR. I can refactor the macro in a new PR
- This PR does not handle overlapping attributes between elements you extend (eg. input and global attribute both have a `width` attribute). We need some form of specialization here. We also need an owned version to the builder. Deref specialization only gives you an immutable reference to the object. I remember some marker trait based specialization, but I can't find the article now. Any ideas here are welcome!

atty303:
@ealmloff
 I don't have enough time to contribute now, so I'll leave the rest to you. Thanks and sorry.

jkelleyrtp:
@atty303 Thanks for the PR! Excited to get this out in 0.5!

PR #1351, opened 2023-08-13
Fix the context API in lists

### LINKED ISSUE #1350: Context API is broken on master
**Problem**

`use_shared_state` sometimes fails to find the state object. I bisected the error to have appeared in https://github.com/DioxusLabs/dioxus/commit/0be5dbe71508bf33903ddf10258f933db6f7183b . The bisect might be a couple commits off as some had unrelated build errors.

**Steps To Reproduce**

Minimal reproduction, using current master (commit 1ab5a03aeffa332354080ed4b625c3ccbc9a29c1)
```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

    use_shared_state_provider(cx, || false);
    let x = use_shared_state::<bool>(cx).unwrap();
    cx.render(rsx! {
        div {
            button {
                onclick: move |_| {
                    *x.write() = true;
                },
                "Crash"
            }
            if *x.read() {
                rsx! { Foo {} }
            }
        }
    })
}

fn Foo(cx: Scope) -> Element {
    let _x = use_shared_state::<bool>(cx).unwrap();
    cx.render(rsx! { div { } })
}
```

**Expected behavior**

There should be no panic logged when the button is pressed.

**Current behavior**

Panic.
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/main.rs:27:43
stack backtrace:
   ...
   3: core::option::Option<T>::unwrap
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/option.rs:935:21
   4: dioxus_repro::Foo
             at ./src/main.rs:27:14
```

**Environment:**
 - Dioxus version: master
 - Rust version: 1.71.0
 - OS info: Linux
 - App platform: desktop

**Questionnaire**

Demonthos:
The scope stack was not being set correctly when diffing a scope which causes the parent of a scope to be incorrect. This PR pushes and pops the scope id into the scope_stack when diffing a scope

Fixes #1350, https://github.com/DioxusLabs/dioxus/issues/1356

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -15,6 +15,7 @@ use DynamicNode::*;
 
 impl<'b> VirtualDom {
     pub(super) fn diff_scope(&mut self, scope: ScopeId) {
+        self.runtime.scope_stack.borrow_mut().push(scope);
         let scope_state = &mut self.get_scope(scope).unwrap();
         unsafe {
             // Load the old and new bump arenas
@@ -45,6 +46,7 @@ impl<'b> VirtualDom {
                 (Aborted(l), Ready(r)) => self.replace_placeholder(l, [r]),
             };
         }
+        self.runtime.scope_stack.borrow_mut().pop();
     }
 
     fn diff_ok_to_err(&mut self, l: &'b VNode<'b>, p: &'b VPlaceholder) {


serzhiio:
Any reason for this to not being merged?

PR #1353, opened 2023-08-13
Add 'with' and 'with_mut' to 'use_shared_state'

DanielAlbl:
I was wanting to use the `with` and `with_mut` functions in `use_shared_state` in the same way you can with `use_ref`. Is it OK to add these?


Demonthos (approved):
This looks great. It is nice to get the APIs more similar between different hooks

Issue #1331, opened 2023-08-09
Pre-built Dioxus CLI binaries for GitHub Actions


esimkowitz:
## Specific Demand

As the Dioxus CLI becomes a one-stop shop for development of Dioxus apps, it should also be the primary tool for automating CI/CD of Dioxus apps.

The default GitHub Actions agent does not have enough RAM to compile Dioxus CLI. I've managed to get it working by increasing the page file size of the agent, but it still takes over 30 minutes to build just the CLI. The GitHub Releases should include precompiled versions of the Dioxus CLI for x86 Linux, macOS, and Windows targets to support the default GitHub Actions agents. Dioxus should also publish a GitHub Action to wrap fetching of the latest Dioxus CLI binary from the releases feed.

### Quick and easy

As it seems like releases are a manual process currently, I think that a simple build script could work as a temporary solution. This would involve installing the necessary targets via RustUp and bundling and archiving the CLI for the targets, then they can be manually copied into the Release assets when creating it.

### More involved

I think the long-term goal should be to automate Release workflow so that when new tags are created, a GitHub Actions workflow is kicked off to run the building and bundling of the CLI. This could also be used to compile changelogs and publish releases to Crates.io. 

## Implementation Suggestion - Dioxus Action

This should be very straightforward and can copy the mechanisms used by this third-party [Trunk action](https://github.com/jetli/trunk-action). It just goes to the Trunk release feed and grabs the latest binary archive matching the target agent. The end setup would be the same regardless of the mechanism used to build and publish the CLI binaries.

esimkowitz:
I'm happy to take this on

Demonthos:
This would be great! Instead of creating our own way to install a prebuild version of a CLI, we can use `cargo-bininstal`? They already have an [action](https://github.com/taiki-e/install-action) setup to install rust CLI's in CI.

You can see the setup instructions for `cargo-bininstall` [here](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md)

esimkowitz:
Agreed, this looks even better! I'll play around with this and see if I can integrate it into the existing CI

PR #1358, opened 2023-08-14
Publish CLI binaries for cargo-binstall

esimkowitz:
# Summary
This PR adds a workflow that will run on every new release to generate binaries of the CLI for Windows, Linux, and macOS. It will then upload these binaries to the release, along with a SHA. This also adds new metadata to the dioxus-cli crate so that [cargo-binstall](https://github.com/cargo-bins/cargo-binstall/tree/main) can find the published artifacts.

# Reason
The goal of this PR is to publish binaries that can be used in downstream partners' GitHub Actions workflows. By using the [install-action](https://github.com/marketplace/actions/install-development-tools), which uses cargo-binstall under the hood, developers can quickly install the latest CLI binary.

Here's a shorthand for installing the CLI using that action:
```yml
- uses: taiki-e/install-action@dioxus-cli
```

# Caveats
Note that I have not tested the install via cargo-binstall yet as I do not have permissions to publish a new version of the package, though I have tested that it publishes the binaries to the GitHub Release.

### DIFF
--- a/.github/workflows/cli_release.yml
+++ b/.github/workflows/cli_release.yml
@@ -7,6 +7,8 @@ on:
 
 jobs:
   build:
+    permissions:
+      contents: write
     runs-on: ${{ matrix.platform.os }}
     strategy:
       matrix:
--- a/.github/workflows/cli_release.yml
+++ b/.github/workflows/cli_release.yml
@@ -9,7 +9,7 @@ jobs:
   matrix_build:
     runs-on: ${{ matrix.platform.os }}
     env:
-      WORKING_DIR: ${{github.}}/source/packages/cli
+      WORKING_DIR: ${{github.workspace}}/source/packages/cli
     strategy:
       matrix:
         platform:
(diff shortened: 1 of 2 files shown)


esimkowitz:
this resolves #1331

esimkowitz:
I found a simpler approach that uses the [upload-rust-binary-action](https://github.com/taiki-e/upload-rust-binary-action), which is from the same dev as the install-action mentioned above. This can handle the build and upload in a single action. I tried to add support for ARM Mac binaries as well but ran into a cross-compilation issue due to git2-rs's dependency on system installations of openssl (see https://github.com/rust-lang/git2-rs/issues/623 for more info). It looks like GitHub is [adding support for ARM Mac runners](https://github.com/github/roadmap/issues/528) by the end of the year though, so we can add support for those then.

esimkowitz:
I've also updated the toolchain install step to use a new action as the actions-rs organization and repos appear to be unmaintained since 2020. I've opened another issue #1360 to track replacing the action in the other pipelines.

Demonthos (approved):
This looks great! This should make the docsite builds a lot shorter

Thank you!

Issue #1360, opened 2023-08-15
Replace unmaintained actions-rs actions


esimkowitz:
## Specific Demand

[actions-rs/toolchain](https://github.com/actions-rs/toolchain) and the rest of the actions-rs org have been unmaintained for over two years. See [this Reddit thread](https://www.reddit.com/r/rust/comments/z1mlls/actionsrs_github_actions_need_more_maintainers_or/) for more info. I'm also tagging the [issue of shame](https://github.com/actions-rs/toolchain/issues/216). We should update the pipelines to use a different approach for installing Rust toolchains.

## Implement Suggestion

One that keeps coming up in the forums as a good replacement is [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain). I've successfully integrated this in my [personal project](https://github.com/esimkowitz/dev-widgets/pull/123) and in my PR #1358 for the new CLI release pipeline. I think this is a good move for the other pipelines in this project to avoid any vulnerabilities or disruptions (such as the [eventual `set-output` deprecation](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)).

Demonthos:
Closed by #1366

striezel:
Just for laughs take a look at https://github.com/DioxusLabs/dioxus/pull/1099.

So the issue described here was already fixed in #1099 on June 19th, but both `actions-rs/toolchain` and `actions-rs/cargo` were re-introduced in #1222 on July 20th, a month later. PR #1222 was also merged by its creator without review, so you might possibly want to put some checks in place to avoid that in the future. Otherwise you might see yourselves in the position where you need to address this very issue _again_.

(And then there is #1077 which also uses / re-introduced the unmaintained `actions-rs` stuff, but since the issue was created before #1099 and was only merged after #1099 one cannot really blame its creator. It's just an unfortunate chain of events. Had it been merged before #1099, then #1099 would have picked that up.)

PR #1366, opened 2023-08-15
Update all uncommented commands to remove actions-rs

esimkowitz:
This commit addresses #1360 by removing references to [actions-rs/toolchain](https://github.com/actions-rs/toolchain/issues/216) and [actions-rs/cargo](https://github.com/actions-rs/cargo/issues/216#issuecomment-1329850830), both of which are unmaintained. 

Some parts of the workflows already use [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) and this is the one most of the community has moved to, so I've changed remaining references to actions-rs/toolchain to dtolnay/rust-toolchain. There's no good replacement for actions-rs/cargo so I've instead moved to just running the commands directly. In cases where Cross is required instead of Cargo to run CI tests, I've added a new action [taiki-e/install-action](https://github.com/taiki-e/install-action) to install Cross and then use a dynamic environment variable to select the correct command to use.

There were a few references to actions-rs that were commented out, I have not touched these.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,52 +85,56 @@ jobs:
 
   matrix_test:
     runs-on: ${{ matrix.platform.os }}
+    env:
+      RUST_CARGO_COMMAND: ${{ matrix.platform.cross == true && 'cross' || 'cargo' }}
     strategy:
       matrix:
         platform:
           - {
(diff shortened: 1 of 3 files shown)


Demonthos (approved):
Thank you!

PR #1378, opened 2023-08-20
deps: Update keyboard-types to 0.7.

waywardmonkeys:
This brings a use of bitflags 2 and a couple of minor features.


waywardmonkeys:
This will require `blitz` to be updated as well since the dependency is used in the interface of `dioxus-html`. (Perhaps a re-export of `keyboard-types` from somewhere and using it via the re-export from `blitz-core` would be better...)

Demonthos (approved):
Thank you!

waywardmonkeys:
The original usage is from here in crates in this repo, so the `pub use` should be here, so that `blitz-core` can just pick it up when it picks up the structs that it is importing ... I'll take a look in the next day or three.

PR #1384, opened 2023-08-21
Expose new_in_scope for signals

ealmloff:
Exposes the new_in_scope function for signals which makes it possible for users to make a signal last longer than the current scope.

For example:
```rust
let signals = use_signal(cx, || vec![Signal::new(0)]);
// ...In a Child component...

// Because we use the origin scope of the array, even if the child scope is dropped, the signal will still exist as long as the array does
signals.push(Signal::new_in_scope(123, signals.origin_scope()));
```

### DIFF
--- a/packages/signals/src/signal.rs
+++ b/packages/signals/src/signal.rs
@@ -145,6 +145,21 @@ impl<T: 'static> Signal<T> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


JaniM:
This is particularly important for me. Could we get this merged so I could stop using the branch and enjoy whatever master has? :smile:

PR #1389, opened 2023-08-22
Use a `gen` in `PartialEq` for `UseSharedState`

DanielAlbl:
So I had tried using a `UseSharedState` as a dependency for `use_future`, and it wasn't re-running when it changed. I think this was because it was comparing the `.inner` values which still were pointing to the same data. 
I implemented a special `UseFutureDep` so it would compare a `gen` value, sort of following the example of `UseRef`. I did have to remove the `PartialEq` here to avoid conflicting with this: `impl<A: Dep> UseFutureDep for &A`, but this might not be the right thing to do.

### DIFF
--- a/packages/hooks/src/use_shared_state.rs
+++ b/packages/hooks/src/use_shared_state.rs
@@ -197,7 +197,8 @@ pub struct UseSharedState<T> {
 
 impl<T> UseSharedState<T> {
     fn new(inner: Rc<RefCell<ProvidedStateInner<T>>>) -> Self {
-        Self { inner, gen: 0 }
+        let gen = inner.borrow().gen;
+        Self { inner, gen }
     }
 
     /// Notify all consumers of the state that it has changed. (This is called automatically when you call "write")
--- a/packages/hooks/src/use_shared_state.rs
+++ b/packages/hooks/src/use_shared_state.rs
@@ -314,7 +314,7 @@ impl<T> Clone for UseSharedState<T> {
     }
 }
 
-impl<T: PartialEq> PartialEq for UseSharedState<T> {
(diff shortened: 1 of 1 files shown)


packages/hooks/src/use_shared_state.rs
```diff
@@ -306,11 +309,26 @@ impl<T> Clone for UseSharedState<T> {
     }
 }
 
-impl<T: PartialEq> PartialEq for UseSharedState<T> {
-    fn eq(&self, other: &Self) -> bool {
-        let first = self.inner.borrow();
-        let second = other.inner.borrow();
-        first.value == second.value
+impl<T> UseFutureDep for &UseSharedState<T> {
```
Demonthos, line 312: Can we make this the partialeq impl instead like use state does?

UseFutureDep is implemented for all tuples with partial eq elements. This may  fix your problem for a single dependency, but it will make use shared state unusable with multiple dependencies in a tuple

Demonthos (changes requested):

packages/hooks/src/use_shared_state.rs
DanielAlbl: OK this makes sense. I put a `gen` in `UseSharedState` itself to use in `PartialEq`, since the `gen`s in `ProvidedStateInner` will always be the same. I have them getting synced up in `use_shared_state`, similar to how `use_state` updates `current_val`. I'm not sure if this is the simplest solution, but it does seem to be working in tuples now.

ealmloff (approved):
Sorry I lost track of this. Looks good, thank you!

ealmloff:
Clippy failure is unrelated. This change can potentially be breaking (this could fail to memorize some props). Waiting for a second review before merging

PR #1394, opened 2023-08-23
Extend MouseData functions to PointerData via a common trait

### LINKED ISSUE #1269: Bring pointer event data to parity with mouse event data
## Specific Demand

The pointer events such as onpointerdown, etc. propagate the separate PointerData struct, which is missing the helper functions that MouseData has. To better support cross-platform applications and sites, PointerData should be updated with the same impls as MouseData for things like getting the pointer coordinates in reference to different spaces.

## Implement Suggestion

Create a common trait for both PointerData and MouseData that exposes their common values and holds the implementation logic for the helper functions.

esimkowitz:
# Summary

This PR moves the fields of the MouseData struct to a separate private struct and exposes the set of functions for accessing the data stored in the new struct as a trait. It also extends this new trait to the PointerData struct, bringing the support of the pointer events to parity with the mouse events.

This also deprecates public fields for the pointer-specific event data and establishes new functions for accessing that data. Eventually, this too will be moved into a private struct.

This also closes some documentation gaps in the PointerEvent and related types.

# Justification

Bringing the PointerEvent support to parity with MouseEvent allows developers to support additional device types as it exposes events not just for mouse, but also for touchscreens. It can also support multiple simultaneous pointers and can enable more-advanced scenarios such as capturing the pointer event even if the pointer leaves the scope of the element that is listening.

# Remarks

A side-effect of moving the MouseData fields to a private struct is that it completes the removal of public access to these fields, which have been deprecated since 0.3.0.

Closes #1269


esimkowitz:
I'm wondering if it would make sense to host the new PointerInteraction code in the input_data file

esimkowitz:
I am marking this as ready for review and seeking feedback on overall project structure for this change. I am not super confident on the naming of the new structs/traits or where I've placed them within the project. With this change, a new trait must be imported into any code that wants to access functions that were previously available directly. Ideally, I would like these traits to be implicitly imported, say through `dioxus::prelude::*`, but I wasn't sure the best way to do this.

packages/dioxus-tui/tests/events.rs
```diff
@@ -131,13 +134,9 @@ fn mouse_up() {
...
         cx.render(rsx! {
-            div {
-                width: "100%",
-                height: "100%",
-                onmouseup: move |_| {
+            div { width: "100%", height: "100%", onmouseup: move |_| {
                     tui_ctx.quit();
-                },
-            }
+                } }
```
ealmloff, line 139: It looks like you ran the `dx fmt` command and it is having a few issues?

esimkowitz: The workspace settings for vscode automatically run the built-in formatter, I have mine configured to [Prettier](https://prettier.io/). That might be the issue... Let me try to fix it

packages/html/src/events/pointer.rs
```diff
@@ -1,33 +1,74 @@
...
 
+use crate::point_interaction::{PointData, PointInteraction};
+
+/// A synthetic event that wraps a web-style [`PointerEvent`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent)
 pub type PointerEvent = Event<PointerData>;
+
+/// Data associated with a pointer event, aside from the data shared with mouse events
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq)]
+pub struct PointerEventData {
```
ealmloff, line 13: Usimg `PointerEventData`, `PointerEvent`, and `PointerData` inside an `Event<>` could get confusing quickly. Instead of having this as a seperate type, can we just allow users to pass in all of the parameters in the new function, or even better expose `.with_pointer_id`, `.with_width`, etc. builder methods on the event?

If not, maybe we could call this `PointerOnlyData`, or `PointerMetadata`? Those are not a lot better, but could be more clear because they are not related to an event

packages/dioxus-tui/tests/events.rs
esimkowitz: I fixed this by manually overriding the workspace setting so it doesn't auto-format. I think it may be worth removing the workspace setting altogether though unless it specifies which formatter to use, otherwise it could create some thrashing depending on the contributor.

packages/html/src/lib.rs
```diff
@@ -35,6 +36,7 @@ pub use transit::*;
 pub use elements::*;
 pub use events::*;
 pub use global_attributes::*;
+pub use point_interaction::PointInteraction;
```
ealmloff, line 39: I agree this would be nice in the prelude. We could add this to the html prelude and it will be re-exported in the dioxus prelude automatically

packages/dioxus-tui/tests/events.rs
ealmloff: Interesting! I'm didn't know prettier would try to format macros

packages/html/src/events/pointer.rs
esimkowitz: I added the PointerEventData struct because I was getting a clippy warning about too many parameters on the ctor. I like the idea of a builder, let me try that

packages/dioxus-tui/tests/events.rs
esimkowitz: I've added a `dxfmt::skip` attribute to the parts that were causing issues. I've also updated the settings.json to specify a formatter and added some workspace recommendations. I can move these to a separate PR though if you think that bloats this one

ealmloff: >I fixed this by manually overriding the workspace setting so it doesn't auto-format. I think it may be worth removing the workspace setting altogether though unless it specifies which formatter to use, otherwise it could create some thrashing depending on the contributor.

Rust analyzer or cargo fmt seems pretty standard for formatting. If a different formatter messes it up a bit, that is probably fine. The RSX formatting isn't consistent throughout the repo. The real solution here is to get `dx fmt` running in CI which should consistently. I'm not opposed to removing it, but I think it is fine for most people. For anything outside of macros, the CI will catch it

packages/html/src/events/pointer.rs
esimkowitz: I ended up just suppressing the too_many_arguments clippy rule as it seems there's no real reason for it other than style

packages/html/src/lib.rs
esimkowitz: added to prelude

packages/dioxus-tui/tests/events.rs
esimkowitz: Yeah i am going to revert the settings changes for this PR. There's some weird inconsistencies with how the formatting is being applied, I'll put that as a follow-up issue

esimkowitz:
Closing in favor of #1402

PR #1396, opened 2023-08-24
Implicit atoms using global context and local subscriptions

jkelleyrtp:
This PR uses the new signals API and global context API to allow atoms to be used from anywhere inside your app with implicit local subscriptions powering it all.

This means no more use_read use_write. Instead you just use your values like normal and the component is automatically subscribed.

It's kinda like magic.

```rust
#[inline_props]
pub fn MessageListItem(cx: Scope, idx: usize) -> Element {
    let msg = MESSAGES[*idx];
    render! {
        div {
            onclick: move |_| SELECTED.set(Some(*idx)),
            class: "text-bold font-sm overflow-hidden truncate flex flex-row cursor-default rounded {class}",
            class: if SELECTED == Some(*idx) { "bg-blue-400" }
            div { class: "flex-1 overflow-hidden ml-4 user-select-none", "{msg.name}" }
            div { class: "flex-1 overflow-hidden ml-4 user-select-none", "{msg.snippet}" }
            div { class: "flex-1 overflow-hidden ml-4 user-select-none", "{msg.sent}" }
            div { class: "flex-1 overflow-hidden ml-4 user-select-none", "{msg.received}" }
            "{MESSAGES.with(|f| f.len())}"
        }
    }
}

```


jkelleyrtp:
Punting to integrate with #1791

PR #1398, opened 2023-08-25
Add ScopeId::ROOT

JaniM:
Currently `ScopeId(0)` is the only way to directly refer to the root scope (useful for signals). This isn't documented anywhere afaict and it's somewhat surprising as usually ids shouldn't be constructed by the user.

This PR adds a constant that provides the root ScopeId.

### DIFF
--- a/packages/core/src/scopes.rs
+++ b/packages/core/src/scopes.rs
@@ -63,6 +63,21 @@ impl<'a, T> std::ops::Deref for Scoped<'a, T> {
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
 pub struct ScopeId(pub usize);
 
+impl ScopeId {
+    /// The root ScopeId.
+    ///
+    /// This scope will last for the entire duration of your app, making it convenient for long-lived state
+    /// that is created dynamically somewhere down the component tree.
+    ///
+    /// # Example
+    ///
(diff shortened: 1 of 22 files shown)


packages/core/src/scopes.rs
```diff
@@ -63,6 +63,21 @@ impl<'a, T> std::ops::Deref for Scoped<'a, T> {
...
+    /// This scope will last for the entire duration of your app, making it convenient for long-lived state
+    /// that is created dynamically somewhere down the component tree.
+    ///
+    /// # Example
+    ///
+    /// ```rust, ignore
+    /// use dioxus_signals::*;
+    /// let my_persistent_state = Signal::new_in_scope(ScopeId::ROOT, String::new());
+    /// ```
+    pub const ROOT: ScopeId = ScopeId(0);
```
ealmloff, line 78: There are several places today we use ScopeId(0) to refer to the root in the core crate. Can you replace those with ScopeId::ROOT to make it more consistent?

JaniM: I can, will take a moment.

ealmloff (approved):
Much more clear, thank you!

ealmloff:
A new rust version was released, which bumps the version of clippy that CI is running. Just ignore those for this PR. I will fix it in a separate PR

Issue #93, opened 2022-01-08
Access to native event type in web


jkelleyrtp:
It should be possible to access the native event that caused the event listener to fire in Dioxus web.

Unfortunately there's no way to access the event natively in Dioxus Desktop (webview) - though we *could* hold the event in stasis and provide some wrappers through FFI.

Issue #793, opened 2023-01-22
Access to raw web_sys::Event


theduke:
Is there a way to get access to a raw web_sys::Event?

My concrete need is that I need to work with files in the browser, and the API provided by `FileEngine` is way too limited.

jkelleyrtp:
Not yet, but I'm working on a change that will add this to be released in the next bump of Dioxus.

gbj:
Duplicate of #93 if it’s helpful to know — always feels good to close two issues at once!

ealmloff:
done in https://github.com/DioxusLabs/dioxus/pull/1402

PR #1402, opened 2023-08-25
Rework Event System

### LINKED ISSUE #1163: Improve Event System Code Snipping
## Specific Demand

Currently, the event system doesn't get snipped correctly because we send dynamic names as the event type from the core crate and receive the dynamic name in the renderers. Because of this, the renderer code that handles each specific event cannot get snipped if the event is never used.

As we implement more events on top of specific elements (eg. #1103) this will become an issue for the web binary size.

## Implement Suggestion

We could move the code that converts from a system event to a html event from the renderers to integrate with html crate:

```rust
static EVENT_CONVERTER: Option<Box<dyn HtmlEventConverter>> = None;
trait HtmlEventConverter {
    fn convert_onmounted_event(&self, event: Box<dyn Any>) -> Box<dyn Any>;
}

fn onmounted(event_handler: ...) -> EventHandler {
    EventHandler {
        convert: convert_onmounted_event,
        // ...
    }
}
```

Each renderer would set EVENT_CONVERTER and if the constructor for the event handler is never called, the function that converts the event would never be included in the final binary.

ealmloff:
There are three main goals here:
- Access to native event types (#793)
- Binary size improvements (https://github.com/DioxusLabs/dioxus/issues/1163)
- Lazily convert events for performance

This is built on top of the better pointer events introduced in https://github.com/DioxusLabs/dioxus/pull/1394

closes #1163 
closes https://github.com/DioxusLabs/dioxus/issues/93


esimkowitz:
This is freaking awesome! Curious for my own learning: does this place an additional burden on code that clones/copies an event to allocate additional space on the heap? I'm assuming the Box can't be reused by copied instances right?

esimkowitz:
> This is freaking awesome! Curious for my own learning: does this place an additional burden on code that clones/copies an event to allocate additional space on the heap? I'm assuming the Box can't be reused by copied instances right?

Or building on that, does this place an additional burden on the transmission system to allocate additional space for every event?

ealmloff:
> This is freaking awesome! Curious for my own learning: does this place an additional burden on code that clones/copies an event to allocate additional space on the heap? I'm assuming the Box can't be reused by copied instances right?

The virtual dom will receive one non-specific web event. That event is wrapped in an Rc so that it will not be cloned. The virtual dom will bubble that event to event handlers. Each event handler will downcast it to the specific event that is needed. This work will be duplicated if the event bubbles, but it should be fairly cheap. Downcasting in the event handler is what lets us do bundle splitting better.

Once the event is downcast, calling any methods on it will get the needed data from it. For example if you call modifiers, it will build a modifiers struct every time instead of caching it. This is slightly inefficient, but because the old implementation used a depreciated old store of data we were doing this anyway.

> Or building on that, does this place an additional burden on the transmission system to allocate additional space for every event?

On the web renderer, we should allocate less in general because we don't get all of the data eagerly when we receive an event. For native/liveview events, we just use the old serialized form of data so it should be the same.

packages/html/src/events/pointer.rs
```diff
@@ -1,33 +1,87 @@
...
+    inner: Box<dyn HasPointerData>,
+}
+
+impl<E: HasPointerData> From<E> for PointerData {
+    fn from(e: E) -> Self {
+        Self { inner: Box::new(e) }
+    }
+}
+
+impl std::fmt::Debug for PointerData {
```
esimkowitz, line 19: this should probably also return the data from the underlying PointData

esimkowitz: same for eq

packages/html/src/events/pointer.rs
```diff
@@ -62,3 +116,237 @@ impl_event![
...
+    }
+
+    fn trigger_button(&self) -> Option<MouseButton> {
+        self.inner.trigger_button()
+    }
+}
+
+#[cfg(feature = "serialize")]
+#[derive(serde::Serialize, serde::Deserialize)]
+struct SerializedPointData {
```
esimkowitz, line 204: SerializedPointerData?

ealmloff:
This can make binary sizes a lot smaller, but you need to enable some rustc flags to get the full potential out of this PR. Specifically this PR relies on https://github.com/rust-lang/rust/pull/96285 to eliminate the dead code in the `dyn EventConverter`

```toml
[unstable]
build-std = ["std", "panic_abort", "core", "alloc"]
build-std-features = ["panic_immediate_abort"]

[build]
rustflags = ["-Zvirtual-function-elimination", "-Clto"]
```

packages/html/src/events/mouse.rs
```diff
@@ -150,131 +106,91 @@ impl_event! {
...
 
-impl Debug for MouseData {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("MouseData")
-            .field("coordinates", &self.coordinates())
-            .field("modifiers", &self.modifiers())
-            .field("held_buttons", &self.held_buttons())
-            .field("trigger_button", &self.trigger_button())
-            .finish()
+impl PartialEq for MouseData {
```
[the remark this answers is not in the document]
ealmloff: Sorry I missed this initially, there isn't an underlying type for MouseData that we can compare. This could be comparing a web event to a web event. On desktop, liveview and TUI the type is SerializedPointInteraction, but on the web renderer, that struct is not included in the binary because the serde feature is turned off (to make the binary smaller).

This will be a more expensive `PartialEq` than a simple struct but not because of the derefs. That would be very cheap. Converting the underlying event to the Dioxus types in each of the functions in the trait is expensive. There is a tradeoff here between performance and binary size. If we know that we will read every field in the event, then we should convert the event once and then use the serialized event for PartialEq, Debug, etc. (this is what the 0.4 release of dioxus does). However, most of the time we only use some of the fields from the event so we are converting some data we never use. If the whole application never uses a field, we also include more code in the final binary than we need to. In this PR, I convert each field when you need the data which can help with performance and binary size if you only use a few fields, but if you read a field twice it will be slower. The middleground here is to lazily convert each field which this PR doesn't implement yet, but with the changes in this PR it should be a non-breaking change in a future PR.

ealmloff:
Marking this as a draft until the pointer trait handles touch events better (mainly multi-touch events)

esimkowitz:
> Marking this as a draft until the pointer trait handles touch events better (mainly multi-touch events)

What is the issue here? Do you have an issue tracking this work?

ealmloff:
> I see, you are referencing https://github.com/DioxusLabs/dioxus/issues/1439#issuecomment-1712507695. As it is right now, the pointer events will operate [as expected by web developers](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Multi-touch_interaction), which is to say that there isn't much hand-holding for multi-touch. The developer is responsible for ensuring their app caches and groups multiple pointer events into a single multi-touch event. If we want to extend that and provide some mechanisms for automatically grouping multi-touch events in PointerEvent so devs can just track one event I think that would be cool, but maybe we should handle that in a separate PR to prevent feature creep.

> The [Touch Event](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent) offers a better mechanism for handling multi-touch as it already groups touch targets together. It's not a perfect match though to the PointInteraction trait as we've set it up. In Touch Events, modifier keys are handled at the top-level event, while point data is handled in the list of touch targets. Things like trigger button and held buttons also don't make sense here. It may be worth breaking out the trait into three: PointTarget for coordinate data, PointTrigger for trigger button and held buttons, and PointModifier for modifier keys.

> Then, the MouseEvent and PointerEvent would implement all three traits, while the TouchEvent can implement PointModifier on the top-level event and PointTarget on the individual touch targets.

For this PR, I just want to get the pointer interaction trait into a place that can be used for touch events as well. For now we can just target web parity with some nice APIs on top of that. Stateful events should definitely be a different PR. (and more discussion in a separate issue). Separating the traits sounds reasonable with some extra functions for pointer events for target touches, changed touches, etc.

packages/html/src/events/touch.rs
```diff
@@ -1,17 +1,147 @@
...
+impl<E: HasTouchData> From<E> for TouchData {
+    fn from(e: E) -> Self {
+        Self { inner: Box::new(e) }
+    }
+}
+
+impl std::fmt::Debug for TouchData {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TouchData")
+            .field("alt_key", &self.alt_key())
```
esimkowitz, line 17: touch event can also use the InteractionModifiers trait

packages/html/src/events/touch.rs
```diff
@@ -1,17 +1,147 @@
...
 pub type TouchEvent = Event<TouchData>;
-#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TouchData {
-    pub alt_key: bool,
-    pub ctrl_key: bool,
-    pub meta_key: bool,
-    pub shift_key: bool,
-    // get_modifier_state: bool,
-    // changedTouches: DOMTouchList,
```
esimkowitz, line 12: were you thinking of adding these back in a later PR or are you planning to add these here?

ealmloff: I would like to add it back in this PR

packages/html/src/events/touch.rs
ealmloff: Yes, I appreciate the review, but I'm still working on implementing touch events, so I wouldn't spend too much time reviewing this until I open the PR again

esimkowitz (approved):
Awesome work!!

sdwoodbury:
this branch needs the `tray` feature. it was added to the `fix-event-bubbling` branch but not this one.

ealmloff:
> this branch needs the `tray` feature. it was added to the `fix-event-bubbling` branch but not this one.

The branch was a few commits behind the main branch. I just pulled in the latest changes. (this happens automatically when the PR is merged)

esimkowitz:
@ealmloff are you holding this for the 0.5 release? May be good to add a milestone tag for it

ealmloff:
> @ealmloff are you holding this for the 0.5 release? May be good to add a milestone tag for it

Yes, everything breaking is for 0.5

PR #1406, opened 2023-08-30
Fix web query segments

ealmloff:
Fixes decoding the query segment on the web router. We currently parse the route from only the `pathname` which does not include the query. This PR adds the query segment before parsing the route which should fix query parameters always being empty on the web router.


Stygmates:
Is there something preventing this from being merged?

jkelleyrtp (approved):

Issue #1403, opened 2023-08-25
Dioxus SSR can no longer compile to wasm32-wasi


skabber:
**Problem**

In dioxus 0.3 I could compile to wasm32-wasi. In dioxus 0.4 this is now broken.

**Steps To Reproduce**

- Create a Cargo.toml file that contains both `dioxus` and `dioxus-ssr` version 0.4
- Compile with `cargo build --target wasm32-wasi`
- See compile error about tokio
`compile_error!("Only features sync,macros,io-util,rt,time are supported on wasm.");`

**Expected behavior**

I would expect to be able to compile dioxus-ssr 0.4 to run on wasm-wasi

**Environment:**
 - Dioxus version: 0.4
 - Rust version: rustc 1.74.0-nightly (249595b75 2023-08-23)
 - OS info: MacOS
 - App platform: `web`

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

In Discord @ealmloff mentioned that `We need to move incremental rendering behind a feature flag. It just uses tokio::io for filesystem access`

ealmloff:
As I understand it, the issue here is in the `ssr` crate. dioxus `web` does compile on wasm correctly

ealmloff:
Fixed by #1409

PR #1409, opened 2023-08-30
Make incremental a cfg feature.

skabber:
#1403


packages/ssr/Cargo.toml
```diff
@@ -29,3 +29,6 @@ argh = "0.1.4"
 serde = "1.0.120"
 serde_json = "1.0.61"
 fs_extra = "1.2.0"
+
+[features]
+incremental = ["dep:tokio"]
```
ealmloff, line 34: can we enable the feature by default so that this is non-breaking?
```
default = ["incremental"]
```

skabber: Absolutely.

PR #1422, opened 2023-09-03
feat: Simplify `use_tracker_state` and `use_selector`

marc2332:
Feel free to close if I am wrong but I was casually reading the source code of `use_tracked_state` and `use_selector` and it seemed like `use_selector` was doing unnecessary work, like creating a whole subscriber vector that only was subscribed itself.

I have tested it with my changes, and it seems to work fine, thoughts?

### DIFF
--- a/packages/hooks/src/computed.rs
+++ b/packages/hooks/src/computed.rs
@@ -11,7 +11,7 @@ use std::{
 /// Create a new tracked state.
 /// Tracked state is state that can drive Selector state
 ///
-/// It will efficiently update any Selector state that is reading from it, but it is not readable on it's own.
+/// It will efficiently update any Selector state that is reading from it, but it is not readable on its own.
 ///
 /// ```rust
 /// use dioxus::prelude::*;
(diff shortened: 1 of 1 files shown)


packages/hooks/src/computed.rs
```diff
@@ -74,14 +73,13 @@ impl<I> Tracked<I> {
     pub fn compute<O: PartialEq + 'static>(
```
ealmloff, line 73: For places where a selector is used in multiple places you could previously create the selector only once, now you will need to create it in each child component and the callback will need to be rerun for each component when the tracked state changes instead of just once.

This API works, but it doesn't make much sense for compute to be public in this version. I think most users shouldn't need to know what a scope id is and now that there is no public API for Selector, I don't think anyone could get useful information out of this

I think we should depreciate selectors in favor of signals. The [use_selector](https://github.com/DioxusLabs/dioxus/blob/master/packages/signals/examples/selector.rs#L12) hook in the signals crate is much more powerful because you can directly read the signal's value and create selectors that read from multiple signals.

marc2332: _now you will need to create it in each child component_ oh, that might make sense

Sorry for taking your time!

Issue #1430, opened 2023-09-05
cargo-binstall for Dioxus CLI fails if release tag is not consitent


esimkowitz:
**Problem**

The current release of the Dioxus CLI uses a different release tag naming scheme than that used for the previous GitHub Releases. This breaks the pattern set in the [Cargo.toml](https://github.com/DioxusLabs/dioxus/blob/master/packages/cli/Cargo.toml), meaning cargo-binstall cannot find the binaries and falls back to the standard cargo install procedure.

**Steps To Reproduce**

- Create a new GitHub Actions workflow with the following:
```yml
name: Test cargo-binstall

on:
  push:
    branches: ["main"]
  workflow_dispatch:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repo
        uses: actions/checkout@v3
      
      # Install toolchain
      - name: Install Rust Toolchain
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: stable
          targets: wasm32-unknown-unknown
      - name: Install Dioxus CLI
        uses: taiki-e/install-action@v1
        with:
          tool: dioxus-cli@0.4.1
```
- Check this into the main branch of a dummy repository and the action should automatically run
- You will see that rather than install the pre-built binaries, the action compiles from source code.

**Expected behavior**

When running cargo-binstall or a GitHub Action like [taiki-e/install-action](https://github.com/taiki-e/install-action) which uses cargo-binstall under the hood to install Dioxus CLI v0.4.1 or later, the pre-built binaries should be installed rather than re-compiling the CLI from source. 

**Screenshots**

**Environment:**
 - Dioxus version: v0.4.1
 - Rust version: 1.71.0
 - OS info: Ubuntu 22.04
 - App platform: web

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

esimkowitz:
To get around this, the [Build CLI for release workflow](https://github.com/DioxusLabs/dioxus/actions/workflows/cli_release.yml) could update the value of the `pkg-url` with the exact title of the release, rather than using a pattern. It could then commit that change back to the release tag.

esimkowitz:
I have a potential fix ready: #1431

PR #1431, opened 2023-09-05
Fix to cargo-binstall not working for Dioxus CLI

esimkowitz:
## Summary
This PR is an attempt to fix cargo-binstall's ability to find the correct pre-built binary regardless of the name of the release tag, identified in #1430. Rather than using the crate version to fill the `pkg-url` field in the Dioxus CLI Cargo.toml file, now the workflow that publishes the binaries to the release will also commit a change to the release tag that updates the pkg-url directly with the name of the tag.

### The workflow must run before the crate is published
For this to work, the workflow must be run before the new package version is published to Crates.io, otherwise the pkg-url will be malformed and cargo-binstall will fall back to `cargo install`.

A future improvement here would be to automate Crates.io publishing into a workflow as well.

### The commit will appear from a spoofed author
I needed to provide an author for the commit so I used a fake email and name that I took from a similar project [git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action). This is a bit janky. Ideally, we should provision a proper service account to publish and sign the commit. See [Machine users](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#machine-users) for more info.

### DIFF
--- a/.github/workflows/cli_release.yml
+++ b/.github/workflows/cli_release.yml
@@ -47,7 +47,7 @@ jobs:
         if: ${{matrix.platform.target == 'x86_64-unknown-linux-gnu'}}
         run: |
           # Replace the tag for the cargo-binstall record with the current release tag
-          sed -i 's/\/{ tagName }\//${{ github.event.release.tag_name }}/g' packages/cli/Cargo.toml
+          sed -i 's/\/{ tagName }\//\/${{ github.event.release.tag_name }}\//g' packages/cli/Cargo.toml
 
           # Commit and push the changes to the tag
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com" # This is a dummy email, copied from https://github.com/stefanzweifel/git-auto-commit-action
--- a/.github/workflows/cli_release.yml
+++ b/.github/workflows/cli_release.yml
@@ -50,8 +50,8 @@ jobs:
           sed -i 's/\/{ tagName }\//\/${{ github.event.release.tag_name }}\//g' packages/cli/Cargo.toml
 
           # Commit and push the changes to the tag
-          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com" # This is a dummy email, copied from https://github.com/stefanzweifel/git-auto-commit-action
-          git config --local user.name "github-actions[bot]"
+          # git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com" # This is a dummy email, copied from https://github.com/stefanzweifel/git-auto-commit-action
+          # git config --local user.name "github-actions[bot]"
           git add packages/cli/Cargo.toml
           git commit -m "Update pkg-url to ${{ github.event.release.tag_name }}"
           git tag -fa ${{ github.event.release.tag_name }} -m "Update pkg-url to ${{ github.event.release.tag_name }}"
--- a/.github/workflows/cli_release.yml
+++ b/.github/workflows/cli_release.yml
@@ -42,6 +42,13 @@ jobs:
         with:
           workspaces: packages/cli -> ../../target
 
+      # Replace the tag for the cargo-binstall record with the current release tag
+      - name: Replace pkg-url in Cargo.toml
+        run: |
+          sed -i 's/{ tagName }/${{ github.event.release.tag_name }}/g' packages/cli/Cargo.toml
+      - name: Commit and push changes to tag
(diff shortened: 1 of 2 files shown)


esimkowitz:
The Miri pipeline is failing but I don't see anything that would be causing it to fail, any thoughts?

.github/workflows/cli_release.yml
```diff
@@ -42,6 +42,21 @@ jobs:
...
+          # Replace the tag for the cargo-binstall record with the current release tag
+          sed -i 's/\/{ tagName }\//\/${{ github.event.release.tag_name }}\//g' packages/cli/Cargo.toml
+
+          # Commit and push the changes to the tag
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com" # This is a dummy email, copied from https://github.com/stefanzweifel/git-auto-commit-action
+          git config --local user.name "github-actions[bot]"
+          git add packages/cli/Cargo.toml
+          git commit -m "Update pkg-url to ${{ github.event.release.tag_name }}"
+          git tag -fa ${{ github.event.release.tag_name }} -m "Update pkg-url to ${{ github.event.release.tag_name }}"
+          git push origin --tags --force
```
ealmloff, line 58: It looks like you update the tag that caused the CI to run and override the description. Is there a reason we need to override the message for cargo-bininstall. If we don't do that, can we avoid force pushing to the main branch. It might be fine, but I'm a bit worried about force pushing in CI

ealmloff:
> The Miri pipeline is failing but I don't see anything that would be causing it to fail, any thoughts?

All of the CI pipelines have been flakey lately. Github might be having some issues

.github/workflows/cli_release.yml
esimkowitz: So we need to pass it some tag description in order to generate the tag. If the release notes are generated in GitHub, they actually aren't added to the tag annotations. I think this is an optimization to keep the .git index smaller. GitHub will only show the description from the tag annotations if the release body is empty. I have tested to make sure that the release body is not overwritten when the tag is force pushed. I could copy the contents of the tag annotation into the new tag, it will just likely be empty if the tag is created in GitHub, as opposed to in the git client.

esimkowitz: I am hesitant to copy the contents of the release body into the tag description, though, as that could bloat the repo index.

esimkowitz: If I use a lightweight tag I can avoid the message altogether, but then there's no record of who created the tag

ealmloff: I think the empty tag github creates automatically is ok if we can avoid force pushing. If we want a more detailed git tag we can create that in the CLI before we publish a release. Or CI could create it when we create a workflow for creating releases

esimkowitz: I've changed the tag so it now creates a lightweight tag without a description. This will still overwrite any tag annotations that were added to the previous tag. That means that release notes must be written via the GitHub release instead of in the tag message, but from what I can tell, all recent releases already do this so it shouldn't be a problem.

ealmloff: Is the force push still necessary with the lightweight tag?

esimkowitz: Yes, the force push is still necessary as we are replacing an existing tag in the remote. Without the force push, we get the following error:
```txt
To https://github.com/esimkowitz/dioxus
 ! [rejected]        CLI-v0.4.1.20 -> CLI-v0.4.1.20 (already exists)
error: failed to push some refs to 'https://github.com/esimkowitz/dioxus'
hint: Updates were rejected because the tag already exists in the remote.
```

esimkowitz:
@ealmloff @jkelleyrtp any concerns with merging this? It's been in a ready state for a month now..

ealmloff:
> @ealmloff @jkelleyrtp any concerns with merging this? It's been in a ready state for a month now..

I'm still not comfortable force pushing to the master branch with CI. I would much rather just set the tag name in a consistent format if force pushing is the only way we can get this to work

esimkowitz:
Maybe this effort should be expanded to a more managed release process that can handle SemVer and publishing on-demand. There's some projects like [release-plz](https://github.com/MarcoIeni/release-plz) that automate these steps.

esimkowitz:
Closing in favor of a more managed release experience: #1539 and #1538

PR #1436, opened 2023-09-06
Switch to tracing for logging

ealmloff:
Switches from the log crate to tracing and adds a bit more logging in some places. This makes some logging in places cleaner with spans and future integration. In the future, we can use a custom subscriber to create a dashboard for a dioxus application


Cargo.toml
```diff
@@ -83,6 +83,8 @@ dioxus-hot-reload = { path = "packages/hot-reload", version = "0.4.0" }
 dioxus-fullstack = { path = "packages/fullstack", version = "0.4.1"  }
 dioxus_server_macro = { path = "packages/server-macro", version = "0.4.1" }
 log = "0.4.19"
```
esimkowitz, line 85: Is log still necessary after this shift?

ealmloff: Not anymore, a few example packages were using this but it should be fixed now

jkelleyrtp:
Sweet!! Excited for the custom subscriber.

PR #1441, opened 2023-09-08
Expose public methods for manipulating the global runtime

ealmloff:
Exposes a few new functions to manipulate the global runtime of Dioxus. This fixes dynamic linking of components as a dynlib. When a dynlib is linked, the global scope list in the dynlib is different from the main application you are running the VirtualDom in which breaks the context API. With this PR, you can pass the runtime from the main application to the dynlib and run `override_runtime` to set the dynlib runtime to be the same as the main application.


jkelleyrtp:
I feel like this should be handled automatically. I spun out an issue for us to circle back to - but if tokio has this solved then we should also have it solved.

jkelleyrtp (approved):

Issue #1041, opened 2023-05-24
Hot reload causes browser to refresh page multiple times per change


elliotwaite:
**Problem**

Sometimes when using hot reloading, making a single change will cause the browser to refresh the page multiple times.

**Steps To Reproduce**

- Create the minimal "Hello, world!" app.
- Run: `dioxus serve --hot-reload`
- Change an element.

**Expected behavior**

For it to only refresh the page once per change.

**Screenshots**

![ezgif-1-f4a8b6fd4f](https://github.com/DioxusLabs/dioxus/assets/1767836/a912236c-9490-4e3b-b3e6-a4ea1d5cd14b)

The terminal shows this warning right before it refreshes the page. I'm not sure if it's related:
```
[WARN] failed to parse `name` custom section Invalid name type (at offset 31788206)
```

My Cargo.toml:
```
[package]
name = "dioxus-test"
version = "0.1.0"
edition = "2021"

[dependencies]
dioxus = { git = "https://github.com/DioxusLabs/dioxus", rev = "b476c6e8a73bd6591ec4827c6f9b196d7c270c3d" }
dioxus-web = { git = "https://github.com/DioxusLabs/dioxus", rev = "b476c6e8a73bd6591ec4827c6f9b196d7c270c3d" }
```

**Environment:**
 - Dioxus version: 0.3.2 (the latest commit: `{ git = "https://github.com/DioxusLabs/dioxus", rev = "b476c6e8a73bd6591ec4827c6f9b196d7c270c3d" }`)
 - Rust version: 1.71.0-nightly (8b4b20836 2023-05-22)
 - OS info: macOS Ventura 13.2.1 (22D68)
 - App platform: web

**Questionnaire**
- [x] I'm new to Dioxus so I don't think I know enough to try to fix this on my own yet.

Demonthos:
What version of the Dioxus CLI are you using?
Testing this with the git version of the dioxus-cli (installed with `cargo install --git https://github.com/DioxusLabs/cli`). I don't see the browser reloading at all, so this issue may already be fixed. The CLI should be able to hot reload changes to the static rsx without a recompile (and browser refresh).

elliotwaite:
Hi, sorry, I currently don't have time to investigate this issue further. Feel free to close this issue if you believe it has been resolved.

@linulas, I saw you liked the original post. If that means that you were also running into this issue and it hasn't been resolved for you yet, feel free to take over trying to debug this issue with @Demonthos.

[...]

PR #1446, opened 2023-09-09
Deduplicate serve code with hot reloading in the CLI crate

ealmloff:
Combines some code in the CLI that serves the hot reload and non-hot reload version of the cli.

This fixes some issues that were applied to only the non-hot-reloading version of the code. Fixes https://github.com/DioxusLabs/dioxus/issues/1041.

This will also fix hot rsx reloading after any changes that require a full rebuild happens. For example after this change:
```rust
p { "{10 * 10}" }
->
p { "{10 * 20}" }
```
The page reloads but this change currently fails to hot reload:
```rust
p { "{10 * 20}" }
->
p { "{10 * 20}" }
"This should be hot reloadable"
```
This PR makes the second change hot reloadable.

### DIFF
--- a/packages/cli/src/server/desktop/mod.rs
+++ b/packages/cli/src/server/desktop/mod.rs
@@ -1,7 +1,7 @@
 use crate::{
     server::{
         output::{print_console_info, PrettierOptions},
-        setup_file_watcher, setup_file_watcher_hot_reload,
+        setup_file_watcher,
     },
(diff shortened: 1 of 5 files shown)


jkelleyrtp:
Perfect - I think this was the issue I was seeing with hotreloading before!

jkelleyrtp (approved):

PR #1448, opened 2023-09-10
Add system for creating component attributes + new `#[component]` attribute

### LINKED ISSUE #1445: `#[inline_props]` macro fails when `use dioxus::core;` is present
**Problem**

The `#[inline_props]` macro fails when the `use dioxus::core;` import is present. Error message:
```
failed to resolve: could not find `default` in `core` [E0433]
Help: consider importing one of these items
Help: if you import `Default`, refer to it directly
```

I know this is very insignificant, as I don't think anybody would import `dioxus::core`, but I figured I would report it anyway.

**Steps To Reproduce**

- Add `use dioxus::core;` to your file.
- Create a component with an additional argument.
- Apply the `#[inline_props]` macro

**Expected behavior**

It should work as if there was no `dioxus::core` import.

**Environment:**
 - Dioxus version: `0.4.0`
 - Rust version: `1.73.0-nightly`
 - OS info: Windows 10
 - App platform: `web`

**Questionnaire**

I'm working on a PR that improves the component macro system, so I'll try and fix this as well.

tigerros:
I'm sorry in advance for potentially bad code! I'm still not that comfortable with Rust.

The details of the actual functionality can be found here: [packages/core-macro/src/component_body/mod.rs](https://github.com/tigerros/dioxus/blob/master/packages/core-macro/src/component_body/mod.rs).

Here's an overview:
* New `#[component]` attribute. What it does:
  * Renames component function name to `PascalCase` from `snake_case`, without generating any warnings. However, everything else other than the function name will still raise a warning if it's not `snake_case`
  * Automatically uses `#[inline_props]` if there's more than 1 parameter in the function.
  * Maybe more stuff in the future. What I wanted is to have a more "standardized" way of creating components.
* New system for making component macros. Both `#[component]` and `#[inline_props]` use this. It allows macros to take arguments, use each other, and eliminates repetition. It also verifies the validity of the component function. Right now that only includes checking if:
  * The component has a `Scope` argument.
  * The component has an `Element` return type.
* Add appropriate spans for errors. That means that errors will highlight the section of code that is causing the error, instead of just highlighting the macro. This doesn't apply everywhere. The only macros I touched were `#[inline_props]`, `#[derive(Props)]` and `#[component]`.
* Add global paths in `quote!`s (`::` before each path) in [packages/core-macro/src/props/mod.rs](https://github.com/tigerros/dioxus/blob/master/packages/core-macro/src/props/mod.rs). This should fix #1445.
* Fix some typos.


tigerros:
The failing checks all seem to be from different packages than `core-macro`. So not really my fault.

ealmloff:
> The failing checks all seem to be from different packages than `core-macro`. So not really my fault.

Core macro is re-exported in the dioxus crate which is used in some tests. This PR doesn't cause any errors within the core macro crate but it does cause errors in users code when you use the core macro crate. It look like several things are broken in the macro. One of these issue is this PR doesn't apply type bounds to generic the inline props macro (causing [this example](https://github.com/DioxusLabs/dioxus/blob/master/examples/rsx_usage.rs#L289) to break)

tigerros:
> Core macro is re-exported in the dioxus crate which is used in some tests. This PR doesn't cause any errors within the core macro crate but it does cause errors in users code when you use the core macro crate. It look like several things are broken in the macro. One of these issue is this PR doesn't apply type bounds to generic the inline props macro (causing [this example](https://github.com/DioxusLabs/dioxus/blob/master/examples/rsx_usage.rs#L289) to break)

My bad. I guess at some point I removed the where clause parsing and forgot to add it back. I assume that will clear up a lot of the errors. However, before I start fixing the errors, is this a feature that you would merge?

tigerros:
Yeah, that's all the errors gone. The only check not passing is Clippy, because `LinkProps` is apparently too complex, but I certainly didn't change it. However, I did change the `Props` macro, but I only added fully qualified paths, so that bugs like the mentioned #1445 wouldn't happen. Could that be why?

ealmloff (commented):
I like the idea of a `component` macro instead of just `inline_props`. It would let us integrate `dx check` (for the rules of hooks) into the macro itself. I agree linting and allowing errors feels out of place in the inline props macro.

> Renames component function name to PascalCase from snake_case, without generating any warnings. However, everything else other than the function name will still raise a warning if it's not snake_case

I don't know if renaming the function to PascalCase is the right choice here. It makes the real component name more difficult to reason about. It also makes it more difficult to find components visually. I understand not wanting to enable the `allow_snake_case` lint globally (or even within the function), but we could keep the function name the same without disabling the lint for the inner function.

We could expand:
```rust
#[component]
fn MyFunction(cx: Scope) -> Element {
    todo!()
}
```

into something like:
```rust
#[allow(non_snake_case)]
fn MyFunction(cx: Scope) -> Element {
    #[warn(non_snake_case)]
    fn __inner_dioxus_component(cx: Scope) -> Element {
        fn SnakeCaseIsEnforced() {}
        todo!()
    }
    __inner_dioxus_component(cx)
}
```

> New system for making component macros. Both #[component] and #[inline_props] use this. It allows macros to take arguments, use each other, and eliminates repetition. It also verifies the validity of the component function. Right now that only includes checking if:
> - The component has a Scope argument.
> - The component has an Element return type.

I love the lints! but can stop some valid code from compiling if users use type aliases. We should add a way to opt out of these lints

```rust
use dioxus::{core_macro::component, prelude::*};

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        DisplayTLDR {}
    })
}

type MyType<'a> = Element<'a>;

#[component]
fn display_t_l_d_r(cx: Scope) -> MyType {
    render! {
        div {}
    }
}
```

> Add appropriate spans for errors. That means that errors will highlight the section of code that is causing the error, instead of just highlighting the macro. This doesn't apply everywhere. The only macros I touched were #[inline_props], #[derive(Props)] and #[component].
> Add global paths in quote!s (:: before each path) in [packages/core-macro/src/props/mod.rs](https://github.com/tigerros/dioxus/blob/master/packages/core-macro/src/props/mod.rs). This should fix https://github.com/DioxusLabs/dioxus/issues/1445.
> Fix some typos.

❤️

ealmloff:
> Yeah, that's all the errors gone. The only check not passing is Clippy, because `LinkProps` is apparently too complex, but I certainly didn't change it. However, I did change the `Props` macro, but I only added fully qualified paths, so that bugs like the mentioned #1445 wouldn't happen. Could that be why?

Maybe the extra few characters trigger the lint? We can just allow(complex_types) for the link props

tigerros:
> I don't know if renaming the function to PascalCase is the right choice here.

You're probably right. I was just thinking that PascalCase function = bad, but I suppose with the warning system it doesn't matter.

> I love the lints! but this can stop some valid code from compiling if users use type aliases. We should add a way to opt out of these lints.

Also true. I didn't really consider this because I don't think anyone would create a type alias for `Scope` or `Element`, but it wouldn't hurt to add an argument that disables it, just in case.

**Tasks**
- [ ] Remove `snake_case -> PascalCase` converter, but rather enforce PascalCase.
- [ ] Add argument that disables *type* lints. There's already an argument that disables the snake_case enforcer lint and conversion to PascalCase, so I think it should be kept separate.
- [ ] Put the second function inside the main one instead of besides it. Right now this:
```rs
#[component]
fn MyFunction(cx: Scope) -> Element {
    todo!()
}
```
Expands to this:
```rs
#[allow(non_snake_case)]
fn MyFunction(cx: Scope) -> Element {
    todo!()
}

#[inline(always)]
fn __my_function(cx: Scope) -> Element {
    todo!()
}
```
but putting the second one inside the first one would be better.

tigerros:
Perhaps a lint + argument system would also be good.

tigerros:
Adding `#[allow(clippy::type-complexity)]` doesn't satisfy Clippy. I assume it's because `LinkProps` isn't the problem, but rather a type generated by `derive(Props)` is. I'll allow the lint for the entire file. Not a fan, but I don't know how else to do it.

ealmloff:
Snake case components that are allowed in rsx because they have at least one `_` now require `no_case_check`:
```rust
#[component(no_case_check)]
fn valid_component(cx: Scope) -> Element {
    render! {
        div{}
    }
}
```

tigerros:
Actually, that might not be final. I think it might be a better idea to unify this under a lint system which the `ComponentBody` uses. That way every macro that uses `ComponentBody` has the same lints. No documentation repetition, and it's going to be suuper easy to create new lints. Right now, the "lints" are just randomly scattered in the code.

tigerros:
Here's what I mean:
```rs
use syn::{ItemFn, Error};

pub trait ComponentLint {
    fn passes(&self, source: &ItemFn) -> Option<Error>;
}

pub struct ComponentLinter {
    pub lints: Vec<Box<dyn ComponentLint>>,
}

impl ComponentLinter {
    pub fn passes(&self, source: &ItemFn) -> Option<Error> {
        for lint in &self.lints {
            if let Some(error) = lint.passes(source) {
                return Some(error);
            }
        }

        None
    }
}

struct CaseCheckLint;

impl ComponentLint for CaseCheckLint {
    fn passes(&self, source: &ItemFn) -> Option<Error> {
        let ident = &source.sig.ident;

        if !is_pascal_case(&ident.to_string()) {
            Some(Error::new(ident.span(), "lint::ComponentLint - This component does not use PascalCase."))
        } else {
            None
        }
    }
}

fn is_pascal_case(input: &str) -> bool {
    // ...
}
```

ealmloff:
That looks nice! Reusing the `allow(snake_case_lint_name)`, `warn(snake_case_lint_name)` (using [this pattern](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3807a00933a84c96b802885950e218d5)), and `deny(snake_case_lint_name)` syntax from normal rust could it a bit easier to learn. There is a lot of overlap here with the [dioxus-check crate](https://github.com/DioxusLabs/dioxus/tree/master/packages/check). With that architecture it seems fairly easy to reuse the visitor in the check crate inside of the component macro for linting.

tigerros:
I never implemented lints that can be used like normal warnings with `allow`, so I'm not sure how to go about it. However, even then, wouldn't the user need to run a command to check the lints, like `cargo clippy`? The macro does it as you type, although I don't think there's a way to report a warning rather than an error. And about dioxus-check, what does it do exactly? On GitHub, it says it reports errors in Dioxus code, such as the rules of hooks. On crates.io, it's listed as dioxus-autofmt, stating that it formats `rsx!` code, but the CLI already has `dx fmt`. I'm confused :sweat_smile:

tigerros:
> 1 change: It would let us integrate dx check (for the rules of hooks) into the macro itself

> 2 change: Reusing the allow(snake_case_lint_name), warn(snake_case_lint_name), and deny(snake_case_lint_name) syntax

So, from this I'm getting the notion that something like this is the desired result:
```rs
type ElemAlias<'a> = Element<'a>;

// 2 change:
#[component(allow(no_element_return))]
pub fn App(cx: Scope) -> ElemAlias {
    render! { "..." }
}

// 2 change:
#[component(allow(no_pascal_case_name), warn(no_element_return))]
pub fn snake_case(cx: Scope) -> Element {
    let you_are_happy = true;
    let you_know_it = true;

    if you_are_happy && you_know_it {
      let something = use_state(cx, || "hands");
      //              ^^^^^^^^^
      // hook called conditionally: `use_state` (inside `if`)
      // note: `if you_are_happy && you_know_it { … }` is the conditional

      // 1 change: The above error is caused by the macro, not a command like `dx check`
      println!("clap your {something}");
    }
    
    render! { "..." }
}
```
Or a separate `#[component_lint]` macro so that it's easier to understand what's being allowed/other controls.

tigerros:
The problem is that proc macros are quite limited, since they run before rustc. That's why, for example, aliases cause issues. Maybe it would be better to ditch the macro lints and instead make a Dioxus fork of Clippy.

ealmloff:
> So, from this I'm getting the notion that something like this is the desired result:
> ...
> The problem is that proc macros are quite limited, since they run before rustc. That's why, for example, aliases cause issues. Maybe it would be better to ditch the macro lints and instead make a Dioxus fork of Clippy.

Yes, `dx check` could use some of the internals of clippy's analysis. Right now, the analysis is very local which makes it possible to integrate with the component macro for users that do not have the CLI installed/want quicker feedback from rust analyzer. If we switch to a more global form of analysis like clippy, integrating with the component macro makes less sense. Alases are not a huge issue, it would just be nice if there was some way to opt out of part of the linting when aliases cause issues

tigerros:
> If we switch to a more global form of analysis like clippy, integrating with the component macro makes less sense.

Of course, but it would be more reliable and powerful. There's things which the component macro can do just fine, but it is limited. With Clippy, we could use it for basically everything. As I mentioned, warnings for one are something that macros can't do, but warnings are something Dioxus would benefit from, with it's custom rsx syntax.

ealmloff:
> The problem is that proc macros are quite limited, since they run before rustc. That's why, for example, aliases cause issues. Maybe it would be better to ditch the macro lints and instead make a Dioxus fork of Clippy. That would also allow us to make warning lints, since I'm not aware of a way to raise warnings in proc macros.

You can create an very ugly, much less readable version of warnings in a macro:

```rust
#[deprecated(note = "this is a custom warning")]
#[allow(unused)]
struct CustomWarning;

const _: CustomWarning = CustomWarning;
```

tigerros:
> That's why I'm suggesting Clippy

Any opinion on this? If you don't want to use Clippy, I can make the lint system in the macro, but right now I'm just waiting on your input.

ealmloff:
> > That's why I'm suggesting Clippy
> 
> @ealmloff Any opinion on this? If you don't want to use Clippy, I can make the lint system in the macro, but right now I'm just waiting on your input.

Clippy would definitely be better. If we can get Clippy working, linting in the macro isn't needed

tigerros:
> Not including a lint system in this PR, and working on a Clippy-backed system in another PR might be the best option.

Okay then. I'll remove type checking (cx: Scope, -> Element), but I'll keep the functionality that allows a PascalCase component with no warnings, since that was why made this PR in the first place.

tigerros:
Clippy is failing because I made the `#[inline_props]` macro deprecated. I can remove it if requested, but I think it makes sense to deprecate it, since `#[component]` automatically inlines props if there's more than 1 argument, along with some additional features.

ealmloff:
> For the reasons above, I would also like to remove `#[inline_props]` completely in the future, but that's too drastic for now. Otherwise, it's ready to merge!

Depreciating `#[inline_props]` makes sense. We should keep it in for at least one major release before completely removing it because of how common it is

> I'll update the docs and occurrences of `#[inline_props]` when it's merged.

I try to avoid merging PRs that have checks failing. I understand why clippy is failing and I agree it isn't much of an issue, but if I merge this without checks passing, all new PRs before you create your fix PR will have tests failing for a unrelated reason which makes finding PRs that are ready for a review more difficult

tigerros:
> I try to avoid merging PRs that have checks failing

I understand. I'll replace it then. I'll also update the docsite and create a PR there. Otherwise, does the state of this look good?

packages/core-macro/src/component_body/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Struct field guidelines:
+/// * Must be public, so that other deserializers can utilize them.
+/// * Should usually be [`Item`]s that you then simply combine in a [`quote!`]
+/// in the [`ComponentBodyDeserializer::output_to_token_stream2`] function.
+/// * If an [`Item`] might not be included, wrap it in an [`Option`].
+/// * Must be at the component function "level"/"context".
+/// For example, the [`InlinePropsDeserializer`](crate::component_body_deserializers::inline_props::InlinePropsDeserializer)
+/// produces two [`Item`]s; the function but with arguments turned into props, and the props struct.
+/// It does not return any [`Item`]s inside the struct or function.
+pub trait DeserializerOutput: ToTokens {}
```
ealmloff, line 18: Does this need to be a seperate trait from ToTokens? Are there any specific structural bounds that an implementer must have that make implementing this trait different from ToTokens?

The guidelines here are good, but that can be documented at the module level or when you use this trait

packages/core-macro/src/component_body/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// The args passed to a [`ComponentBody`] when deserializing it.
+///
+/// It's also the struct that does the deserializing.
+/// It's called "DeserializerArgs", not "Deserializer". Why?
+/// Because "args" makes more sense to the caller of [`ComponentBody::deserialize`], which
+/// takes an [`DeserializerArgs`] argument. However, you can think of "DeserializerArgs" as the deserializer.
+pub trait DeserializerArgs<TOutput>: Clone
```
ealmloff, line 26: Deserialize might be a bit consusing for tokens because of how common it is with `serde`. Syn tends to use Parse for tokens maybe we could use ParseArgs?

packages/core-macro/src/component_body/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+where
+    TOutput: DeserializerOutput,
+{
+    // There's a lot of Results out there... let's make sure that this is a syn::Result.
+    // Let's also make sure there's not a warning.
+    /// Creates a [`DeserializerOutput`] from the `self` args and a [`ComponentBody`].
+    /// The [`ComponentBody::deserialize`] provides a cleaner way of calling this function.
+    /// As a result, don't make this public when you implement it.
```
ealmloff, line 34: function items are declared private or public when you declare the trait (here). When you implement the trait, marking the function as public doesn't change anything

ealmloff (commented):
Overall looks good, a few questions about some function/trait names. Errors and imports look much nicer now! The name `#[component]` makes more sense as we expand the functionality in the future

packages/core-macro/src/component_body/utils.rs
tigerros: I think deserialize makes more sense. Maybe not in this case, but `ComponentBody::deserialize()` is IMHO better than `ComponentBody::parse()`. It also doesn't parse/deserialize tokens, but rather the `ComponentBody`.

packages/core-macro/src/component_body/utils.rs
tigerros: > Does this need to be a seperate trait from ToTokens?

Not really, but if I removed `DeserializerOutput`, you could use any struct that implements `ToTokens`, as an output struct. This way, it has to explicitly be a `DeserializerOutput`. And in the future, `DeserializerOutput` might implement more traits or provide a function.

packages/core-macro/src/component_body/utils.rs
tigerros: My bad. Removed that comment.

packages/core-macro/src/component_body/utils.rs
ealmloff: > Not really, but if I removed DeserializerOutput, you could use any struct that implements ToTokens, as an output struct. This way, it has to explicitly be a DeserializerOutput.

Unless I am missing something about what `DeserializerOutput` is communicating, I think it should be removed. It seems like the documentation is just talking about how to make something easy to use with quote which is also true for anything that implements ToTokens.

To me this looks similar to adding a blank trait to the built in `Into` trait. You can add extra docs there if you want to, but it makes the code less clear:

```rust
/// Something that can be converted into. To make a struct easy to convert to, you should avoid data that is mixed between multiple fields and unclear names for fields
trait Convertible {}

trait Into<O: Convertible> {
    fn into(self) -> O;
}
```

> in the future, DeserializerOutput might implement more traits or provide a function.

If we need it in the future we could always re-add it.

tigerros: I think it's nice because it has the documentation for it, whereas you can't make documentation for `ToTokens`. I can put the documentation in the module, but having it right in the struct that the `DeserializerArgs::to_output` returns.

ealmloff: If you want to keep it, you can make it an automatically implemented by adding:
```rust
impl<T: ToTokens> DeserializerOutput for T {}
```

tigerros:
I changed all references of `#[inline_props]` in `examples/` and `packages/` to `#[component]`. I also sometimes removed `#![allow(non_snake_case)]` but not always. Anyway, there's still references to `#[inline_props]` in the `docs/` directory, but that's going to be removed, so I'm not sure what to do there. Should I just remove the folder? I can copy the changes from #1408 then push them here to fix Clippy.

ealmloff:
> I changed all references of `#[inline_props]` in `examples/` and `packages/` to `#[component]`. I also sometimes removed `#![allow(non_snake_case)]` but not always. Anyway, there's still references to `#[inline_props]` in the `docs/` directory, but that's going to be removed, so I'm not sure what to do there. Should I just remove the folder? I can copy the changes from #1408 then push them here to fix Clippy.

Either one works. Both removing the folder or switching inline_props to components works should fix it. If you remove the docs folder, I can just close my original PR

tigerros:
Removed the folder. Did all the tests on my machine, so it should be good.

ealmloff (approved):
Looks good, thank you!

PR #1453, opened 2023-09-11
Remove stray log statement in `onmounted`

oxkitsune:
I think this log statement wasn't caught in code review :shipit:

### DIFF
--- a/packages/web/src/dom.rs
+++ b/packages/web/src/dom.rs
@@ -249,7 +249,6 @@ impl WebsysDom {
         for id in to_mount {
             let node = get_node(id.0 as u32);
             if let Some(element) = node.dyn_ref::<Element>() {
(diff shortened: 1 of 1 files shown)


packages/web/src/dom.rs
```diff
@@ -249,7 +249,6 @@ impl WebsysDom {
         for id in to_mount {
             let node = get_node(id.0 as u32);
             if let Some(element) = node.dyn_ref::<Element>() {
-                log::info!("mounted event fired: {}", id.0);
```
oxkitsune, line 252: Maybe it should be logged as `trace` or `debug` instead

ealmloff (approved):
This wasn't intentionally added. Element ids are not very useful for debugging purposes here

PR #1461, opened 2023-09-13
`Routable` improvements

### LINKED ISSUE #1451: `Routable::static_routes()` always produces an empty result.
**Problem**

The `Routable::static_routes()` function always produces an empty result. Maybe I'm just using it wrong, but I don't think so.

**Steps To Reproduce**

- Serve this code:
```rs
    render! { Router::<Route2> {} }
}

#[derive(Routable, Clone)]
#[rustfmt::skip]
enum Route2 {
    #[route("/")]
    Index {},
    #[route("/foo")]
    Foo {},
}

impl Debug for Route2 {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        f.write_str(&self.to_string())
    }
}

#[inline_props]
fn Index(cx: Scope) -> Element {
    let routes: Vec<Route2> = Route2::static_routes();

    render! {
        h1 { "Routeable::static_routes(). Length: {routes.len()}. Content:" }
        pre { "{routes:?}" }
    }
}

#[inline_props]
fn Foo(cx: Scope) -> Element {
    render! { "Foo" }
}
```
- Static routes length is 0.

**Expected behavior**

It should display `/` and `/foo`.

**Environment:**
 - Dioxus version: 0.4.0
 - Rustc version: 1.73.0-nightly
 - OS info: Windows 10
 - App platform: `web`

**Questionnaire**
- [x] I would like to fix and I have a solution.
I fixed this by tweaking the `static_routes()` source code, but I think using the default `from_str` function is the issue, because the router is a very complex enum. Instead, I think `Routable` should implement `FromStr` and also `Display` manually, along with documentation of how the input string should be structured. The `FromStr` and `Display` implementations should "mirror", as in:
```rs
// to_string() provided by `Display`
let str = Routable::to_string();
let from_str = Routable::from_str(str);

assert_eq!(str, from_str);
```

tigerros:
- Fixes #1451. Should also be a bit faster now.
- Adds:
  ```rs
  type SiteMapFlattened<'a> = FlatMap<
      Iter<'a, SiteMapSegment>,
      Vec<Vec<SegmentType>>,
      fn(&SiteMapSegment) -> Vec<Vec<SegmentType>>,
  >;
  ```
  This is used in the methods below.
- Adds more methods to `Routable`. These are:
  - Similar methods to `Routable::static_routes()`:
    - `all_routes() -> Vec<Routable>`
    - `dynamic_routes() -> Vec<Routable>`
    - `catch_all_routes() -> Vec<Routable>`
  - Utility methods:
    - `flatten_site_map<'a>() -> SiteMapFlattened<'a>` - Returns a flattened version of `Self::SITE_MAP`.
    - ```rs
      /// Calls a [`Iterator::filter_map`] on [`SiteMapFlattened`].
      fn filter_map_routes<'a, B, F>(f: F) -> FilterMap<SiteMapFlattened<'a>, F>
      where F: FnMut(Vec<SegmentType>) -> Option<B>
      ```

### DIFF
--- a/packages/router/src/routable.rs
+++ b/packages/router/src/routable.rs
@@ -3,7 +3,7 @@
 #![allow(non_snake_case)]
 use dioxus::prelude::*;
 
-use std::iter::{FilterMap, FlatMap};
+use std::iter::FlatMap;
 use std::slice::Iter;
 use std::{fmt::Display, str::FromStr};
 


ealmloff:
>   This is used in the methods below.
> * Adds more methods to `Routable`. These are:
>   
>   * Similar methods to `Routable::static_routes()`:
>     
>     * `all_routes() -> Vec<Routable>`
>     * `dynamic_routes() -> Vec<Routable>`
>     * `catch_all_routes() -> Vec<Routable>`

I can't think of a good usecase for these methods. The route enum is an instance of the route. It is a specific route, not the template from which routes can be made from (the site map). All of these methods return a instance of the route, not the route definition. For static routes, both the instance of a specific route and the route template are the same because there are no dynamic parameters.

The way the code is structured currently, it looks like the dynamic parameters are filled in with the name of the dynamic parameter. This can cause parsing the route to behave unexpectedly. For example, this code:

```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;
use dioxus_router::prelude::*;

fn main() {
    dioxus_web::launch(App);
}

#[derive(Routable, Clone, Debug)]
enum Route {
    #[route("/")]
    Home {},
    #[route("/:id")]
    TakesU64 { id: u64 },
    #[route("/:string")]
    TakesString { string: String },
    #[route("/:..route")]
    TakesSpread {
        route: Vec<String>,
    },
}

    render! {
        Router::<Route> {}
    }
}

fn Home(cx: Scope) -> Element {
    render! {
        h1 { "Welcome to the Dioxus Blog!" }
        pre {
            "all_routes:\n{Route::all_routes():?}"
        }
        pre {
            "static_routes:\n{Route::static_routes():?}"
        }
        pre {
            "dynamic_routes:\n{Route::dynamic_routes():?}"
        }
        pre {
            "catch_all_routes:\n{Route::catch_all_routes():?}"
        }
    }
}

fn Blog(cx: Scope) -> Element {
    todo!()
}

#[inline_props]
fn TakesString(cx: Scope, string: String) -> Element {
   todo!()
}

#[inline_props]
fn TakesU64(cx: Scope, id: u64) -> Element {
    todo!()
}

#[inline_props]
fn TakesSpread(cx: Scope, route: Vec<String>) -> Element {
    render! {
        h1 { "Page not found" }
        p { "We are terribly sorry, but the page you requested doesn't exist." }
        pre {
            color: "red",
            "log:\nattemped to navigate to: {route:?}"
        }
    }
}
```
Displays this site:
Screenshot 2023-09-13 at 9 49 13 AM

>   * Utility methods:
>     
>     * `flatten_site_map<'a>() -> SiteMapFlattened<'a>` - Returns a flattened version of `Self::SITE_MAP`.
>     * ```rust
>       /// Calls a [`Iterator::filter_map`] on [`SiteMapFlattened`].
>       fn filter_map_routes<'a, B, F>(f: F) -> FilterMap<SiteMapFlattened<'a>, F>
>       where F: FnMut(Vec<SegmentType>) -> Option<B>
>       ```

`flatten_site_map` and `SiteMapFlattened` is nice! `filter_map_routes` seems a bit redundant if you can use `Route::flatten_site_map().filter_map(f)` directly

tigerros:
> All of these methods return a instance of the route, not the route definition.

I've changed them to instead return the definition. I can't test it because I get an error for `derive(Routable)`:
```
method `render` has an incompatible type for trait

Note: expected signature `fn(&Route, &'a dioxus_core::scopes::ScopeState, _) -> Option<dioxus_core::nodes::VNode<'a>>`
found signature `fn(&Route, &ScopeState, _) -> Option<VNode<'_>>
```
I'm surprised it works for you. I have the same exact code in a clean project as you, but still.

ealmloff:
> > All of these methods return a instance of the route, not the route definition.
> 
> I've changed them to instead return the definition. I can't test it because I get an error for `derive(Routable)`:
> 
> ```
> method `render` has an incompatible type for trait
> 
> Note: expected signature `fn(&Route, &'a dioxus_core::scopes::ScopeState, _) -> Option<dioxus_core::nodes::VNode<'a>>`
> found signature `fn(&Route, &ScopeState, _) -> Option<VNode<'_>>
> ```
> 
> I'm surprised it works for you. I have the same exact code in a clean project as you, but still.

You need to make the source of the dioxus, dioxus-web, and dioxus-router crates to all match. (in this case have them all point to the same git repo or local workspace)

tigerros:
Thanks. It returns the name of the parameter, not the actual route.
```
Welcome to the Dioxus Blog!/

all_routes:
["/", "/id", "/string", "/route"]

static_routes:
[Home]

dynamic_routes:
["/id", "/string"]

catch_all_routes:
["/route"]
```
Not sure if that's useful...

ealmloff:
Yeah, I don't think it is very useful. We could either remove those functions or return Vec<Vec<RouteSegment>> for all of the non-static-route filters

tigerros:
I'll try to return `Vec<Vec<RouteSegment>>` then, but that has to be done in the macro right? `SITE_MAP` doesn't contain any `RouteSegments`s.

ealmloff (approved):

PR #1463, opened 2023-09-14
Implement multiple optional attributes

### LINKED ISSUE #1246: Multiple optional attributes
## Specific Demand

When building applications you may want to optionally add different parts of an attribute. This is very common when using tailwind specifically.

```rust
div {
    class: "boxed",

## Implement Suggestion

<!--
If you have any suggestion for complete this feature, you can tell us.
-->

### PR DESCRIPTION
**Pull request by ealmloff**
Introduces a new syntax for creating and combining optional `class` and `style` attributes:
```rust
div {
    class: if true { "m-2" },
    class: if false { "p-2" },
}
```

This is breaking because we previously accepted:
```rust
div {
    class: "m-2",
    class: "p-2",
}
```
and rendered it as 
```rust
div {
    class: "p-2",
}
```
It now renders as:
```rust
div {
    class: "m-2 p-2",
}
```

Closes #1246


kaivol:
Maybe I'm just bad at googling or documentation reading, but I didn't find this feature until I came across this PR.
I would have expected it to be mentioned [here](https://dioxuslabs.com/learn/0.5/reference/rsx).

ealmloff:
> Maybe I'm just bad at googling or documentation reading, but I didn't find this feature until I came across this PR. I would have expected it to be mentioned [here](https://dioxuslabs.com/learn/0.5/reference/rsx).

Documentation for this feature is added in https://github.com/DioxusLabs/docsite/pull/219/commits/f63e9e083a9c37e42ba20820e841a55b08bf83fa

Issue #1465, opened 2023-09-16
We should not expose global runtime manipulation.


jkelleyrtp:
In #1441, we introduced a way to mess with the stack of global runtimes.

We should prefer a system like Tokio's guard where we can manually enter a runtime for a duration of a scope.

PR #1467, opened 2023-09-17
Make RuntimeGuard public

ealmloff:
This is a different approach to https://github.com/DioxusLabs/dioxus/pull/1441 that just makes the runtime guard public. This API is slightly more complex because it uses drop handlers, but can be more flexible

Closes https://github.com/DioxusLabs/dioxus/issues/1465

### DIFF
--- a/packages/core/src/lib.rs
+++ b/packages/core/src/lib.rs
@@ -32,7 +32,7 @@ pub(crate) mod innerlude {
     pub use crate::nodes::RenderReturn;
     pub use crate::nodes::*;
     pub use crate::properties::*;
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
I don't see a drop handler here?

ealmloff:
> I don't see a drop handler here?

That isn't changed in this PR. In https://github.com/DioxusLabs/dioxus/blob/8b411b8c6da0707fc618985c8b35201e00d638da/packages/core/src/runtime.rs#L157:

```rust
impl Drop for RuntimeGuard {
    fn drop(&mut self) {
        pop_runtime();
    }
}
```

jkelleyrtp (approved):

PR #1481, opened 2023-09-21
liveview: Add `interpreter_glue_relative_uri

kang-sw:
By utilizing `window.location.host` in the client-side JavaScript, we can easily derive the WebSocket URI from a relative path URI. This approach obviates the need for host address retrieval on the server side, unlike the method of serving glue code in liveview using `interpreter_glue`.

### DIFF
--- a/packages/liveview/src/lib.rs
+++ b/packages/liveview/src/lib.rs
@@ -110,7 +110,7 @@ static MAIN_JS: &str = include_str!("./main.js");
 /// ```
 pub fn interpreter_glue(url_or_path: &str) -> String {
     // If the url starts with a `/`, generate glue which reuses current host
-    let get_ws_url = if url_or_path.starts_with("/") {
+    let get_ws_url = if url_or_path.starts_with('/') {
         r#"
   let loc = window.location; 
   let new_url = "";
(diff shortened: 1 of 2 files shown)


packages/liveview/src/lib.rs
```diff
@@ -109,3 +109,46 @@ pub fn interpreter_glue(url: &str) -> String {
...
     )
 }
+
+/// This function behaves similarly to [`interpreter_glue`], but interprets the provided URI as a
+/// relative path, automatically prefixing the protocol and host endpoint.
+///
+/// For instance, if you call `interpreter_glue("ws://localhost:1234/ws/liveview")` and the root of
+/// the liveview is being served at `localhost:1234`, you can replace it with
+/// `interpreter_glue_relative_uri("/ws/liveview")`.
+pub fn interpreter_glue_relative_uri(mut relative_uri: &str) -> String {
```
ealmloff, line 119: The option to host the url on a relative path seems useful to make debug and production builds automatically point to a different host

Does this need to be a separate function from interpreter_glue? Or could we just check if the path start with a `/` (as you do bellow) and choose to interpret it as a relative or absolute URL from that.

kang-sw: Oh, that's better, I'll fix the code in the direction you said and re-commit.

kang-sw: I've merged `interpreter_glue_relative_uri` functionality into existing API, and editted `axum` example a bit to test if this works.

ealmloff (approved):
Thank you!

PR #1494, opened 2023-09-25
Register server functions with custom context for Axum adapter 

Calastrophe:
Allows for registering server functions with the default handler, but only modifies the DioxusServerContext with a closure.

I tried to make this more modular than just copy pasting `register_server_fns()`, but this is the approach I landed on.

I'm looking for opinions, suggestions, or goals from the core team on what they'd be looking for to make this adapter better. I had tried to create a default handler, but it resulted in a bunch of Box<dyn ...>.

### DIFF
--- a/packages/fullstack/src/adapters/axum_adapter.rs
+++ b/packages/fullstack/src/adapters/axum_adapter.rs
@@ -188,6 +188,38 @@ pub trait DioxusRouterExt<S> {
     /// ```
     fn serve_static_assets(self, assets_path: impl Into<std::path::PathBuf>) -> Self;
 
+    /// Registers server functions with the default handler but allows for modifying the [`DioxusServerContext`] with a closure, which you can add shared state across all server functions.
+    ///
+    /// # Example
+    /// ```rust
+    /// use dioxus::prelude::*;
+    /// use dioxus_fullstack::prelude::*;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let addr = std::net::SocketAddr::from(([127, 0, 0, 1], 8080));
+    ///     axum::Server::bind(&addr)
+    ///         .serve(
+    ///             axum::Router::new()
+    ///                 // Register server functions routes with the default handler
+    ///                 .register_server_fns_with_context("", |mut context| {
+    ///                     context.insert(42usize).unwrap()
+    ///                     context
+    ///                 }
+    ///                 .into_make_service(),
+    ///         )
+    ///         .await
+    ///         .unwrap();
+    /// }
+    /// ```
+    fn register_server_fns_with_context<M>(
+        self,
+        server_fn_route: &'static str,
+        context_modifier: M,
+    ) -> Self
+    where
+        M: Fn(DioxusServerContext) -> DioxusServerContext + Send + Sync + Clone + Copy + 'static;
+
     /// Serves the Dioxus application. This will serve a complete server side rendered application.
     /// This will serve static assets, server render the application, register server functions, and intigrate with hot reloading.
     ///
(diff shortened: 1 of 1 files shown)


asdfghmnbvcxser:
Can someone show me the proof?

Calastrophe:
On second thought, it may be better to just pass a custom `DioxusServerContext`, not a closure that generates one.
I'm curious to see how everyone else feels?

ealmloff:
> On second thought, it may be better to just pass a custom `DioxusServerContext`, not a closure that generates one. I'm curious to see how everyone else feels?

I agree, you could insert different data based on the current request but that should be what the original register_server_functions_with_handler does.

I'm a bit worried about an exponential growth of variants here. Instead of having three different ways to register server functions, we could have one builder:

Equivalent of today's register_server_functions
```rust
Router::new()
  .serve_static_assets("dist")
  .connect_hot_reload()
  .register_server_functions("", ServerFunctionHandler::default())
```

Equivalent of today's register_server_functions_with_handler
```rust
Router::new()
  .serve_static_assets("dist")
  .connect_hot_reload()
  .register_server_functions("",
    ServerFunctionHandler::default()
      .map_request(|request| 
        DioxusServerContext::default()
      )
      .map_response(|response: Result<Response<hyper::body::Body>, server_fn::ServerFnError>| response)
  )
```

Adding state to a handler
```rust
Router::new()
  .serve_static_assets("dist")
  .connect_hot_reload()
  .register_server_functions("", ServerFunctionHandler::default().with_state(1234usize))
```

Calastrophe:
For mapping the response [here](https://github.com/DioxusLabs/dioxus/blob/a2dd926f7fec89c0cce84359a045917685fd0830/packages/fullstack/src/adapters/axum_adapter.rs#L265C25-L265C93), do you essentially want to replace the .map() with this custom closure? Also, `ServerFunctionHandler` is similar to `ServerFnHandler` is this meant to be shared across all adapters?

ealmloff:
It would apply after the server function has run [before the response is returned](https://github.com/DioxusLabs/dioxus/pull/1494/files#diff-ac5e88765cd8d13cea09f4edacae036c0dfd830ecdcfd0628dc0e950b71e2e3dR305):

```rust
fn register_server_fns(..., handler: Handler) {
    let request_to_context = handler.map_request;
    let response_to_response = handler.map_response;
    self.register_server_fns_with_handler(server_fn_route, |func| {
        use crate::layer::Service;
        move |req: Request<Body>| async move {
            let (req, body) = req.into_parts();
            let req = Request::from_parts(req, body);
            let server_context = request_to_context(req);
            let mut service = crate::server_fn_service(server_context, func);
            let res = service.run(req);
            match res.await {
                Ok(res) => {
                    let res = response_to_response(res);
                    Ok::<_, std::convert::Infallible>(res.map(|b| b.into()))
                }
                Err(e) => {
                    let mut res = Response::new(Body::from(e.to_string()));
                    *res.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
                    Ok(res)
                }
            }
        }
    })
}
```

> Also, ServerFunctionHandler is similar to ServerFnHandler is this meant to be shared across all adapters?

I think we could share this across adapters because we are mapping a hyper request to a hyper response.

jkelleyrtp:
What are the next steps here? Moving this to draft until we can figure out what's next.

ealmloff:
> What are the next steps here? Moving this to draft until we can figure out what's next.

Implementing this builder API: https://github.com/DioxusLabs/dioxus/pull/1494#issuecomment-1735597415

Or I could pull some server function handler improvements into https://github.com/DioxusLabs/dioxus/pull/1766 which changes a lot of server function internals

PR #1496, opened 2023-09-26
fix(vdom):  try to fix vdom panic

sdwoodbury:
the Uplink project was experiencing a panic on this line. I'm not sure if this is a valid way to fix the issue. The panic happened because `attr_paths` was empty when `dynamic_attrs` wasn't.

### DIFF
--- a/packages/core/src/virtual_dom.rs
+++ b/packages/core/src/virtual_dom.rs
@@ -424,7 +424,10 @@ impl VirtualDom {
                     let target_path = el_ref.path;
 
                     for (idx, attr) in template.dynamic_attrs.iter().enumerate() {
-                        let this_path = node_template.attr_paths[idx];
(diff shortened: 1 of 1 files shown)


ealmloff:
> the Uplink project was experiencing a panic on this line. I'm not sure if this is a valid way to fix the issue. The panic happened because `attr_paths` was empty when `dynamic_attrs` wasn't.

This might fix the panic, but it points to a larger issue: the `template` and `dynamic_attrs` are either pointing to two different rsx calls or one of them is being generated incorrectly in the macro. Do you know what changes in Uplink that triggered this issue? I could take a look to see if there is anything unique that could trigger a bug in Dioxus

ealmloff:
https://github.com/DioxusLabs/dioxus/pull/1402 appears to fix this issue

PR #1497, opened 2023-09-26
Add `use_on_mount` and fix some typos

tigerros:
- Adds a new `use_on_mount` hook. Works like `use_effect` but with no dependencies.
- Adds a little extra documentation to `use_effect`.
- Fixes some typos.

There's some discussion about if `use_on_mount` should instead be `use_on_created` or `use_on_component_mount` because of potential confusion with the `onmounted` event. This is what I prefer, but I guess it's still up for discussion.


ealmloff:
> There's some discussion about if use_on_mount should instead be use_on_created or use_on_component_mount because of potential confusion with the onmounted event. This is what I prefer, but I guess it's still up for discussion.

Copying over some of the discussion from discord for context:

- This hook wouldn't always run after the component is mounted to the dom. It could return None, or be suspended. Because of that some javascript operation that need some element to be mounted would fail unlike the `onmounted` event
- This naming is similar to `use_onunmount` (which mirrors this functionality), but also `onmount` which does something different

packages/hooks/src/use_on_mount.rs
```diff
@@ -0,0 +1,34 @@
...
+        needs_regen: true,
+        task: Cell::new(None),
+    });
+
+    if state.needs_regen {
+        // We don't need regen anymore
+        state.needs_regen = false;
+        let fut = future();
+
+        state.task.set(Some(cx.push_future(async move {
```
ealmloff, line 30: Do we need to store the task here? It looks like it is never used. Tasks will automatically be removed when the scope is dropped

tigerros: Nope. Just a second.

tigerros: Should be good now.

tigerros:
> - This hook wouldn't always run after the component is mounted to the dom. It could return None, or be suspended. Because of that some javascript operation that need some element to be mounted would fail unlike the onmounted event
> - This naming is similar to use_onunmount (which mirrors this functionality), but also onmount which does something different

Yeah, I saw `use_on_unmount` and thought `use_on_mount` makes sense. But the first point is also true, so I really don't know. However, if we're not going to use "mount", then I think we should also rename `use_on_unmount`. Make a suggestion and I'll commit it.

ealmloff:
> Yeah, I saw `use_on_unmount` and thought `use_on_mount` makes sense. But the first point is also true, so I really don't know. However, if we're not going to use "mount", then I think we should also rename `use_on_unmount`. Make a suggestion and I'll commit it.

Completely agree, we should keep the component mount and unmount names similar. I like `use_on_create` and `use_on_destroy` because it feels different from the component being mounted to the DOM. But `use_on_initialized`, and `use_on_removed` or `use_after_first_run` and `use_before_remove` also work. Anything that is clearly different from the element event and is short works. It could also be a good idea to leave this PR open for a bit to see if anyone else has some naming ideas

tigerros:
I updated it to be `use_on_create` and `use_on_destroy`. The problem is that this is now a breaking change.

tigerros:
I added `use_on_unmount` back, but made it deprecated. This might be the best solution, but I'm not sure if it isn't just better to remove it entirely.

ealmloff (approved):

PR #1498, opened 2023-09-26
Rename `dioxus-hooks` modules to use snake_case

tigerros:
Discord transcript:

> tigerros — Today at 23:26
> Why don't `dioxus-hooks` modules use snake_case? For example, there's a `use_effect` function, but a `useeffect` module. It's also not consistent, `use_shared_state` and `use_on_unmount` modules do use snake_case, but others do not. Is there a reason for this? It doesn't really follow naming conventions
> 
> ealmloff — Today at 23:32
> I think the divide is just newer/older hook files. They should all be snake case to match the hook names
> none of the modules are public (except computed) so changing the file names is non-breaking


ealmloff (approved):
Thank you!

Issue #1152, opened 2023-07-04
[RFC] Authentication - Minimal Example and Documentation


stephenandary:
# Authentication in Dioxus

## Purpose

This RFC is to align on a desired authentication example for Dioxus and its relevant documentation as planned in: 
[Documentation Map](https://excalidraw.com/#room=d8e228990924048be951,MBE9NeSUUezwjkfPMyIv_A)

## Proposal

Create an initial example to leverage the [OAuth2 Crate](https://github.com/ramosbugs/oauth2-rs) to authenticate with an external provider. This will allow for social or email login as one can authenticate through [Auth0](https://auth0.com) if email signup is desired. 

This will document the basic authentication flow, and how it works with Dioxus (**with visuals**).

As such, I'd like to avoid sessions for this example; Auth0 can manage sessions. 

If desired I can use Auth0 instead of another OAuth provider for this example. 

Future PRs may involve simplifying the example by abstracting code into components, hooks or something less cumbersome, adding some form of sessions, and moving calls to server functions. 

## Reference

I am currently playing with code in this [Repo](https://github.com/stephenandary/dioxus_playground/tree/main/barebone_app) where I have placed OAuth Library calls in a "services" directory such that if a user wants to move these to a server function instead it should easily be possible. 

I have more previous work [Here](https://github.com/stephenandary/wasm_platform_examples/tree/main/wasmcloud/oauth2).

stephenandary:
There is also this [OpenID Connect Crate](https://github.com/ramosbugs/openidconnect-rs) that can be looked at. But I am using OAuth2 for myself, so this is something I won't be able to look at right now.

Demonthos:
This sounds great! Oauth is a good place to start. We could add extra chapters for other auth providers and sessions in the future.

stephenandary:
After speaking with @jkelleyrtp can also just not use the OAuth2 Crate. I have work archived which does this already. Focus was on Authorization flow with PKCE. And can add other flows in the future. OAuth2 can be somewhat complex to use, and it may be better just to keep it simple.

stephenandary:
I deleted my last comment, as I am sticking with using Oauth2 crate until I can't. Doesn't make sense to not when my blocker before was not being able to compile because of WASI. If I can't compile I will use work I have done to get around this. But I am optimistic wasm-unknown-unknown target shouldn't have this issue.

stephenandary:
Working on this, closing issue.

PR #1500, opened 2023-09-27
Add openidconnect authentication demo

Stygmates:
This is a follow up on https://github.com/DioxusLabs/dioxus/issues/1152, this example shows how to implement authentication using OpenID connect.

I am not an expert on openid/dioxus so I'll require a thorough review if possible!

I also feel like we need a dioxus-auth module for openid/oauth2 to make it easier for anyone to use it, I'm open to help implement it if we have an idea on what the API should look like.
Edit: Happy 1500th pull request!


Stygmates:
Most of the tests fail because the required env variables are not set, the ones that are needed are in the [readme](https://github.com/DioxusLabs/dioxus/blob/e88ac50e55a8b4e901b2f7b3a3fffc087491cfef/examples/openid_connect_demo/README.md).

examples/openid_connect_demo/src/views/header.rs
```diff
@@ -0,0 +1,179 @@
...
+#[component]
+pub fn LogOut(cx: Scope) -> Element {
+    let fermi_client_read = use_read(cx, &FERMI_CLIENT);
+    let fermi_auth_token_write = use_set(cx, &FERMI_AUTH_TOKEN);
+    let auth_token_read = use_read(cx, &FERMI_AUTH_TOKEN);
+    cx.render(match fermi_client_read.oidc_client.clone() {
+        Some(_client) => match auth_token_read.id_token.clone() {
+            Some(id_token) => {
+                let log_out_url_future =
+                    use_future(cx, (), |_| async move { log_out_url(id_token).await });
```
ealmloff, line 24: [You cannot use hooks inside of a conditionals](https://dioxuslabs.com/learn/0.4/reference/hooks#no-hooks-in-conditionals) (the git version of `dx check` should catch this)

examples/openid_connect_demo/build.rs
```diff
@@ -0,0 +1,32 @@
+use dotenvy::dotenv;
+use std::env;
+use std::fs::File;
+use std::io::Write;
+
+// https://dev.to/javiasilis/how-to-pass-environment-variables-to-a-rust-wasm-application-like-yew-dioxus-and-leptos-as-a-typescript-developer-ond
```
ealmloff, line 6: This will work, but it is more complicated than we need for this example. You can pull in specific enviorment varaibles at compile time with the [env](https://doc.rust-lang.org/std/macro.env.html) macro. Because the macro resolves at compile time, it will be included as a string in the WASM binary

ealmloff (changes requested):
Overall this looks amazing! Auth examples are some of the most complex (but very useful) examples we need for Dioxus, thanks for putting this together! There are a few places in the code that could cause issues in the future or could be simplified.

> Most of the tests fail because the required env variables are not set, the ones that are needed are in the [readme](https://github.com/DioxusLabs/dioxus/blob/e88ac50e55a8b4e901b2f7b3a3fffc087491cfef/examples/openid_connect_demo/README.md).

Could you exclude/skip the test in the [CI workflows](https://github.com/DioxusLabs/dioxus/tree/master/.github/workflows)?

Stygmates:
Thanks for the feedback, I'll have a look at it when I can!

Stygmates:
> Overall this looks amazing! Auth examples are some of the most complex (but very useful) examples we need for Dioxus, thanks for putting this together! There are a few places in the code that could cause issues in the future or could be simplified.

You are welcome, but I also needed it for myself and my company, so it's also good for me to get a code review from someone who knows dioxus :).

I also rewrote some of the parts to try to make it simpler/reflect the authentication flow better, so I'm open to to feedback to make it even simpler.
The one thing I was thinking about was to pass fermi atoms like the client as a borrowed prop to get rid of the match pattern in the inner components, would it be a good idea?

> 
> > Most of the tests fail because the required env variables are not set, the ones that are needed are in the [readme](https://github.com/DioxusLabs/dioxus/blob/e88ac50e55a8b4e901b2f7b3a3fffc087491cfef/examples/openid_connect_demo/README.md).
> 
> Could you exclude/skip the test in the [CI workflows](https://github.com/DioxusLabs/dioxus/tree/master/.github/workflows)?

I didn't need to exclude the example from the CI workflow in the end.

Also, I'm still getting this warning when running `dx build` and I'm not sure what to do with it:
```[WARN] failed to parse `name` custom section Invalid name type (at offset 99523911)```

ealmloff:
> Also, I'm still getting this warning when running `dx build` and I'm not sure what to do with it: `` [WARN] failed to parse `name` custom section Invalid name type (at offset 99523911) ``

I think that is an artifact from wasm-bindgen that we currently don't strip. It is safe to ignore for now.

ealmloff:
> The one thing I was thinking about was to pass fermi atoms like the client as a borrowed prop to get rid of the match pattern in the inner components, would it be a good idea?

That sounds like a good idea. There is a lot of duplicate pattern matching between components.

You could try to use borrowed props but you might run into some issues with temporary lifetimes since you would be borrowing from a Ref. If you can't borrow the state, you could clone it before passing it as props. It seems like the objects are relatively small so it shouldn't effect performance too much. (and you save some work matching in child components)

Stygmates:
@ealmloff Could you maybe please chime in in the related [issue](https://github.com/ramosbugs/openidconnect-rs/issues/133) if you have something to add (or if I wasn't clear with something) :pray:

ealmloff:
You can implement PartialEq manually (without the `#[derive]` macro) on the props struct that uses `Client`. If you know that you never change client you could implement PartialEq as `true` or you could get some unique id from the `Client` and compare the props by that id

Stygmates:
> You can implement PartialEq manually (without the `#[derive]` macro) on the props struct that uses `Client`. If you know that you never change client you could implement PartialEq as `true` or you could get some unique id from the `Client` and compare the props by that id

Oh good idea, I think I prefer this solution, cloning the `Client` is probably less expensive than instantiating one, I'll use the `ClientId` to implement `PartialEq`, thanks!

Stygmates:
How would I pass the `Client` as a prop to the login component through the router? Is the only solution to load it before we instantiate the router itself? Maybe the best solution would be to have two different routers depending on whether the client is initalized

ealmloff:
> How would I pass the `Client` as a prop to the login component through the router? Is the only solution to load it before we instantiate the router itself? Maybe the best solution would be to have two different routers depending on whether the client is initalized or not.
> 
> Edit: Maybe the best way would be to keep using fermi for the components that are direct children to the header and use props for the descendents of these children

You need to either use the [context API](https://dioxuslabs.com/learn/0.4/reference/context#using-shared-state) or Fermi. You can use the context API or Fermi to get the router within each route you need it in and then use props to pass it to the children of the route once you get the state you need out of the client

Stygmates:
> > How would I pass the `Client` as a prop to the login component through the router? Is the only solution to load it before we instantiate the router itself? Maybe the best solution would be to have two different routers depending on whether the client is initalized or not.
> > Edit: Maybe the best way would be to keep using fermi for the components that are direct children to the header and use props for the descendents of these children
> 
> You need to either use the [context API](https://dioxuslabs.com/learn/0.4/reference/context#using-shared-state) or Fermi. You can use the context API or Fermi to get the router within each route you need it in and then use props to pass it to the children of the route once you get the state you need out of the client

That's what I thought, thanks! I guess this PR is up for review then. If you see anything else that should be corrected I'm open for feedback, otherwise I think It's good enough to be merged.

ealmloff (approved):
Thank you!

PR #1501, opened 2023-09-27
Fix `examples/todomvc.rs` formatting and warnings

tigerros:
Nothing much I was just copying this to another project and got some warnings so I figured I would fix them.

One Clippy warning is still left: the `App` function is too long. And it is true, but I'm too lazy for that right now. Definitely should be split up in the future, because it's very hard to read. Not a very good showcase if you ask me.

### DIFF
--- a/examples/todomvc.rs
+++ b/examples/todomvc.rs
@@ -146,8 +146,8 @@ pub fn app(cx: Scope<()>) -> Element {
         }
         footer { class: "info",
             p { "Double-click to edit a todo" }
-            p { "Created by ", a { href: "http://github.com/jkelleyrtp/", "jkelleyrtp" }}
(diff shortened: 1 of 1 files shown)


examples/todomvc.rs
```diff
@@ -1,10 +1,13 @@
-#![allow(non_snake_case)]
+#![warn(clippy::pedantic)]
+#![allow(clippy::must_use_candidate)]
```
jkelleyrtp, line 2: Is it worth adding pedantic clippy lints in examples?

IMO I'd rather just keep the examples clean and straightforward. Not worried about warnings.

tigerros: > IMO I'd rather just keep the examples clean and straightforward.

That's why I added extra warnings, to make sure the code is clean. No offense, but I don't think two lines of Clippy make the code any less straightforward.

Issue #1038, opened 2023-05-23
LiveView Router Integration


ealmloff:
## Specific Demand

The LiveView renderer currently uses the desktop router history provider, which does not change the browser URL of the page

## Implement Suggestion

The integration should work similar to the web router feature, but with liveview applications. It can use the [query](https://github.com/DioxusLabs/dioxus/blob/master/packages/liveview/src/query.rs) system introduced in #894 to communicate with the browser

Demonthos:
This issue is blocked on #1026 

In order to implement this, we need some way to communicate when the user presses the forward or backward buttons in the browser (with the [popstate](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) event). There currently isn't a good way to call rust from javascript in liveview.

ealmloff:
The git version of Dioxus now has router support for liveview thanks to DonAlonzo!

HKalbasi:
Maybe this page https://dioxuslabs.com/learn/0.5/reference/liveview needs some update

PR #1505, opened 2023-09-29
Implement LiveView Router Integration

DonAlonzo:
This is in response to #1038.

Included are two bug fixes:

- GoForwardButton used can_go_back to check if it should be disabled, now uses can_go_forward.
- Router context resolve_into_routable resolved all external URLs into an error string instead of the actual URL.


packages/router/src/history/liveview.rs
```diff
@@ -0,0 +1,443 @@
...
+                        WindowEvent::Load {
+                            location,
+                            state,
+                            session,
+                            depth,
+                        } => {
+                            let Ok(route) = R::from_str(&location.to_string()) else {
+                                return;
+                            };
+                            let Ok(state) = serde_json::from_str(&state) else {
```
ealmloff, line 258: Instaed of storing the current history in each history in the browser history, could we just serialize any part of the history with javascript that looks like it could be from the same applicateion (any urls that have the same domain?).

packages/liveview/src/window.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+#[derive(Deserialize, Debug, Clone)]
+#[serde(tag = "type", content = "params")]
+pub enum WindowEvent {
+    #[serde(rename = "load")]
+    Load {
+        location: Location,
+        state: String,
+        session: String,
```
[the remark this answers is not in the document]
DonAlonzo: This is a very good point, and not something I originally considered. I agree with this.

packages/router/src/history/liveview.rs
DonAlonzo: This could prove to be hard, as security restrictions in browsers hinder programmatically getting any other information other than the state of the current page and the total number of pages in the stack. However, I think for this particular problem, with efficiency in mind, we can probably scale down the amount of information being sent drastically.

DonAlonzo: This could prove to be hard, as security restrictions in browsers hinder programmatically getting any other information other than the state of the current page and the total number of pages in the stack. However, I think for this particular problem, with efficiency in mind, we can probably scale down the amount of information being sent drastically. I will investigate this further.

ealmloff (approved):
Much better! Liveview router support has been a big missing feature for several releases, thank you!

This issue closes https://github.com/DioxusLabs/dioxus/issues/1038

packages/router/src/history/liveview.rs
ealmloff: > This could prove to be hard, as security restrictions in browsers hinder programmatically getting any other information other than the state of the current page and the total number of pages in the stack. However, I think for this particular problem, with efficiency in mind, we can probably scale down the amount of information being sent drastically. I will investigate this further.

That makes sense, since it only transfers the history on the initial load, this shouldn't be much of a performance hit

PR #1508, opened 2023-10-02
feat: initial_selected attribute for options

White-Oak:
Adds the `initial_selected` attribute to set an option element's defaultSelected property as described [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement).
Implemented similarly to https://github.com/DioxusLabs/dioxus/pull/1063


ealmloff (approved):
Thank you!

PR #1510, opened 2023-10-03
Update `dioxus-core-macro` README as per #1448

tigerros:
This updates the README with the new changes from #1448. Specifically:
- Adds a description for the `component` attribute macro.
- Removes the description of the deprecated `inline_props` attribute macro.
- Adds dots after each bullet point description.

A bit late to the party, I know. Maybe I searched for `#[inline_props]` (instead of just `inline_props`) when replacing all occurrences of it? I'll see if I find any other occurrences of this tomorrow.


ealmloff (approved):
Thank you

PR #1515, opened 2023-10-03
cli: Log connection errors during `dx serve` in desktop apps

srid:
While debugging the CLI app for a hot-reload issue, I thought this would be useful to have. Let me know if I'm missing something! Previously, the `if let(Ok...` was silently ignoring connection errors.


ealmloff (approved):
Hot reloading will try to connect when you run `cargo run` as well as `dioxus serve --platform desktop`. I added a note about the CLI for users who just use cargo.

Thank you!

PR #1527, opened 2023-10-06
feat: Optional web support for hot-reload crate

marc2332:
### Changes in `hot-reload`
- Enable `dioxus-html` by default but make it optional
- New `custom_file_watcher` feature, it's like the old `file_watcher` but doesn't depend on `dioxus-html`, I will use this one for Freya with `default-features = false` so web is disabled
- `file_watcher` feature now just enables `custom_file_watcher` and `dioxus-html`

No breaking changes and I will be able to not pull `dioxus-html` in freya


ealmloff (approved):
Looks great, thank you!

PR #1535, opened 2023-10-10
CLI: Remove redundant directories

tigerros:
- Removes the following directories from `packages/cli`:
  - `.github` - Workflows from the old CLI repo, never even ran.
  - `docs` - Now lives at [DioxusLabs/docsite/CLI](https://github.com/DioxusLabs/docsite/tree/master/docs-src/0.4/en/CLI).
  - `examples` - Includes an empty README and two Lua plugin examples. We won't be using Lua for plugins though, so this is unnecessary.
- Fix typo


ealmloff (approved):
Thank you!

PR #1537, opened 2023-10-11
Add the ability to map signals

ealmloff:
Adds the ability to display a read only view into a signal with the same subscription characteristics as the original signal

Stacked on top of #1594


srid:
Related question: would it make sense to add a `filter` function as well? Currently I'm writing something to the effect of:

```rust
/// Like `map` but allow dropping certain values. Like `filter`, but allows mapping the filtered out value.
/// This is equivalent to `mapMaybe` in Haskell: https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Maybe.html#v:mapMaybe
 pub fn filter_signal<T, U, F>(cx: Scope, ini: U, sig: Signal<T>, f: F) -> Signal<U>
  where
      F: Fn(T) -> Option<U> + 'static,
  {
      let res: Signal<U> = use_signal(cx, || ini);
      let input = *sig.read();
      use_future(cx, (&input,), |(input,)| async move {
          if f(input) {
              res.set(input);
          }
      });
      res
  }
```

Real-world code:

image

https://github.com/juspay/nix-browser/pull/89/files#diff-fe2ebff4c05d0f2011e0289f1e0cf193e7908cb8685d6a4a72501a204023bc96R40-R53

ealmloff:
> Related question: would it make sense to add a `filter` function as well? Currently I'm writing something to the effect of:
> 
> ```rust
> /// Like `map` but allow dropping certain values. Like `filter`, but allows mapping the filtered out value.
> /// This is equivalent to `mapMaybe` in Haskell: https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Maybe.html#v:mapMaybe
> pub fn filter_signal<T, U, F>(cx: Scope, ini: U, sig: Signal<T>, f: F) -> Signal<U>
>   where
>       F: Fn(T) -> Option<U> + 'static,
>   {
>       let res: Signal<U> = use_signal(cx, || ini);
>       let input = *sig.read();
>       use_future(cx, (&input,), |(input,)| async move {
>           if f(input) {
>               res.set(input);
>           }
>       });
>       res
>   }
> ```
> 
> Real-world code:
> 
> image
> https://github.com/juspay/nix-browser/pull/89/files#diff-fe2ebff4c05d0f2011e0289f1e0cf193e7908cb8685d6a4a72501a204023bc96R40-R53

That is less straight forward. It doesn't map exactly to the iterator version of filter which could be confusing. Feel free to open an issue to discuss it.

You could do it without reruning the component or async like this:
```rust
pub fn filter_signal<T, U, F>(cx: Scope, ini: U, sig: Signal<T>, f: F) -> Signal<U>
  where
      F: Fn(T) -> Option<U> + 'static,
  {
      let res: Signal<U> = use_signal(cx, || ini);
      dioxus_signals::use_effect::(cx, move || {
          let input = *sig.read();
          if f(input) {
              res.set(input);
          }
      });
      res
  }
```

werner291:
`thread 'main' panicked at /home/werner/.cargo/git/checkouts/dioxus-29ab102270016d23/21b1e82/packages/signals/src/impls.rs:325:42:
called `Option::unwrap()` on a `None` value`

Got a crash :3

```
pub fn SessionView(
    cx: Scope,
    session_model: MappedSignal<SessionModel>,
    auto_login: Signal<bool>,
) -> Element {
    to_owned!(session_model);

    // Fan the session state into the various subcomponents.
    let messages = use_selector(cx, move || session_model.read().chats.clone());
```

Looks like you can't read from the inner signal when the outer signal changes to None.

SessionView is supposed to be removed when the outer signal switches, but it seems the selector update fires before that.

werner291:
Yup... Writing `None` to the signal results in a direct stack trace down into the `map` that calls `.unwrap()` on the `None` that was just set.

I see two ways out:
- Always keep the last Some and feed that through the inner signal, so that subscribers to the inner signal never deal with that... They'll have a stale value, though.
- Kill the subscription somehow and never update the lower signal if the outer signal goes to None
- Perhaps consider a pull-based implementation so that unwrap() is never called if it's never needed (because the scope where the value is needed is about to be removed, i.e. the SessionView is about to be removed from the scope tree)

packages/signals/src/impls.rs
```diff
@@ -279,16 +280,51 @@ impl<T: Clone + 'static> IntoIterator for Signal<Vec<T>> {
...
 impl<T: 'static> Signal<Option<T>> {
     /// Returns a reference to an element or `None` if out of bounds.
     pub fn as_mut(&self) -> Option<Write<T, Option<T>>> {
         Write::filter_map(self.write(), |v| v.as_mut())
     }
+
+    /// Try to create a [`MappedSignal`] over the inner value.
+    pub fn as_mapped_ref(&self) -> Option<MappedSignal<T>> {
+        if self.read().is_some() {
+            Some(self.map(|v| v.as_ref().unwrap()))
```
werner291, line 325: So yeah, this line can crash. https://github.com/DioxusLabs/dioxus/pull/1537#issuecomment-1873502974

(I figured I'd thread my replies under here so I top spamming the main PR)

werner291 (changes requested):

packages/signals/src/impls.rs
werner291: Yeah... I think (part of) the problem is that changes get *immediately* pushed over to the receiving signal, rather than making effective use of the tree of scopes you guys have got going on.

Perhaps when a signal is written to, it could be marked "dirty", and downstream subscribers pull the value only once they need it. If the scope is dropped before then, that read never happens, the pull never takes place, and the `.unwrap()` (though it could maybe be `.expect()` at least) never panics.

werner291: Looking more at it, I think you have a bit of a weirdly inconsistent way of implementing push/pull implementation that's leading to the crash I'm facing.

Note how here, when calling `MappedSignal::read()`, you pull the value from the original signal through the mapping into wherever the read occurs.

```
/// Get the current value of the signal. This will subscribe the current scope to the signal.
    pub fn read(&self) -> GenerationalRef<U> {
        (self.mapping.read())()
    }
```

Yet, when an update is available, the information is actively pushed into the subscribing signals, and when they read() it, they just get the value that's just been delivered there.

werner291: It'd probably require a change to how signals work in general, actually... Have the initial update only propagate a "dirty" status in the dependent signals, and have the signal only be actually updated when `.read()` is called.

ealmloff: > It'd probably require a change to how signals work in general, actually... Have the initial update only propagate a "dirty" status in the dependent signals, and have the signal only be actually updated when .read() is called.

For selectors, we need to rerun the selector after any dependency changes because we need to rerun any subscribers that read the selector (like the component) if the value changes

ealmloff: I'm not sure if there is a way to prevent that error. I think fundamentally, you can't do a check once and then assume that check still holds the next time you read the mapped value (especially now that signals can be `Sync+Send`. I removed the iterator and Option helpers that map signals based on some initial property.

ealmloff:
merged into https://github.com/DioxusLabs/dioxus/pull/1594

PR #1540, opened 2023-10-12
feat: initial_checked attribute for inputs

White-Oak:
Adds the `initial_checked` attribute to set an input element's defaultChecked property as described [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement).
Implemented just like previous `initial_selected` attribute in https://github.com/DioxusLabs/dioxus/pull/1508


ealmloff (approved):
Thank you!

PR #1546, opened 2023-10-16
Change the semantics of exprs/for loops allocations strategy

jkelleyrtp:
This commit adjusts how exprs and for loops are handled within rsx. This is a breaking change in terms of codegen, but has slight semantic changes as well.

Now, when exprs/for loops are allocated, they are given a temporary. The temporary is elided to the <'a> lifetime of the bump, to satisfy the borrow checker. This fixes issues with signals where exprs/for loops mapping vecs out of RefCells would be caught up without a temporary lifetime.

Specifically, before, this component could not compile:

```rust
    let todos = use_signal(cx, || vec!["hi"]);

    render! {
        ul {
            for todo in todos.read().iter() {
                li { "{todo}" }
            }

            todos.read().iter().map(|todo| rsx! {
                li { "{todo}" }
            })
        }
    }
}
```

The error we were getting is:

```
error[E0597]: `todos` does not live long enough
  --> src/main.rs:13:25
   |
9  |     let todos = use_signal(cx, || vec!["hi"]);
   |         ----- binding `todos` declared here
...
13 |             for todo in todos.read().iter() {
   |                         ^^^^^^^^^^^^
   |                         |
   |                         borrowed value does not live long enough
   |                         a temporary with access to the borrow is created here ...
...
22 | }
   | -
   | |
   | `todos` dropped here while still borrowed
   | ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `std::cell::Ref<'_, std::vec::Vec<&str>>`
   |
   = note: the temporary is part of an expression at the end of a block;
           consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
   |
11 ~     let x = render! {
12 |         ul {
 ...
20 |         }
21 ~     }; x
   |

For more information about this error, try `rustc --explain E0597`.
error: could not compile `demo-signals-todo` (bin "demo-signals-todo") due to previous error
```

This is fixed now, fixing a major papercut of signals and the work being done in fermi to allow truly global state.

This works because we now do a codegen from

```
make_node(#expr)
```

to

```
{
    let node = #expr.make_node(cx);
    node
}
```

### DIFF
--- a/examples/signals.rs
+++ b/examples/signals.rs
@@ -7,6 +7,7 @@ fn main() {
 
 fn app(cx: Scope) -> Element {
     let mut count = dioxus_signals::use_signal(cx, || 0);
+    let saved_values = dioxus_signals::use_signal(cx, || vec![0]);
 
     use_future!(cx, || async move {
         loop {
@@ -19,9 +20,19 @@ fn app(cx: Scope) -> Element {
         h1 { "High-Five counter: {count}" }
         button { onclick: move |_| count += 1, "Up high!" }
         button { onclick: move |_| count -= 1, "Down low!" }
+        button {
+            onclick: move |_| saved_values.push(count.value()),
+            "Save this value"
+        }
 
+        // We can do boolean operations on the current signal value
         if count.value() > 5 {
             rsx!{ h2 { "High five!" } }
+        }
+
+        // We can cleanly map signals with iterators
+        for value in saved_values.read().iter() {
+            h3 { "Saved value: {value}" }
         }
     })
 }
--- a/examples/signals.rs
+++ b/examples/signals.rs
@@ -8,8 +8,10 @@ fn main() {
 fn app(cx: Scope) -> Element {
     let running = dioxus_signals::use_signal(cx, || true);
     let mut count = dioxus_signals::use_signal(cx, || 0);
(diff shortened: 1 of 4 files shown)


ealmloff (changes requested):
This makes rsx slightly more forgiving and everything gets converted to Vnodes when the component is diffed anyway so it shouldn't have a large performance impact.

A very similar issue can happen with let statements in unterminated if blocks that could be worth addressing in this PR as well:
```rust
    let todos = use_ref(cx, || vec!["hi"]);

    render! {
        ul {
            if let &[first] = todos.read().as_slice() {
                rsx! { li { "single element: {first}" } }
            }
        }
    }
}
```

This isn't specific to signals, it can be reproduces with use_ref (or any Rc<RefCell<T>> like hook) as well:
```rust
    let todos = use_ref(cx, || vec!["hi"]);

    render! {
        ul {
            for todo in todos.read().iter() {
                li { "{todo}" }
            }

            todos.read().iter().map(|todo| rsx! {
                li { "{todo}" }
            })
        }
    }
}
```

jkelleyrtp:
It technically worked before, apparently, but I've updated it to do the temporary coercion for both.

I think there's no performance impact since the temporary should get merged during codegen.

PR #1549, opened 2023-10-16
Fix deref for Write struct in dioxus-signals

esimkowitz:
When mapping a write struct in dioxus-signals, the new Write that is created cannot be dereferenced. This is because the default constraint for the Write struct ties the inner type to the type for the signal. When mapping the inner type to a new type, this constraint is broken. Explicitly separating the type constraints for the inner and the signal allows for dereferencing of the new mapped Write.


jkelleyrtp (approved):
LGTM!

PR #1550, opened 2023-10-16
Make signal callable on stable

ealmloff:
(Thanks to @matthunz for letting me know about [dtolnay](https://github.com/dtolnay)'s amazing article about derefing to closures on stable!)

This restores the previously unsound implementation of `Deref<Target = dyn Fn() -> Ref<'static, T>>` on signals which lets you call the signal like a function instead of reading the signal:

```rust
    let signal = Signal::new("hello world".to_string());
    // You can call signals like functions to get a Ref of their value.
    assert_eq!(&*signal(), "hello world");

    render! {
        "hello world"
    }
}
```

### DIFF
--- a/packages/generational-box/src/lib.rs
+++ b/packages/generational-box/src/lib.rs
@@ -184,7 +184,7 @@ impl<T: 'static> GenerationalBox<T> {
     }
 
     /// Try to read the value. Returns None if the value is no longer valid.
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
What's cool about this is that, with #1546 you can now do very simple iters:

```rust
for todo in todos().iter() { 
    div { "{todo.title}" }
}
```

Looks good

Issue #1502, opened 2023-09-27
Refactor `examples/todomvc.rs` `App` component


tigerros:
## Specific Demand

The `App` component in `examples/todomvc.rs` is long, nested and hard to read. Since TodoMVC is a very popular way of showcasing a certain framework, the implementation should be clean and easy to read.

alexanderjophus:
Can I take a look at this? Recently refactored my first dioxus site, would be good to get feedback/practice on the todomvc refactoring

ealmloff:
> Can I take a look at this? Recently refactored my first dioxus site, would be good to get feedback/practice on the todomvc refactoring

That would be great!

ealmloff:
Fixed by https://github.com/DioxusLabs/dioxus/pull/1556

PR #1556, opened 2023-10-19
refactor todo mvc example

alexanderjophus:
More context in the issue: https://github.com/DioxusLabs/dioxus/issues/1502

Note: I'm learning rust/dioxus, so any/all feedback welcome. I tried to make a component for each of the main UI components. i.e. the header, the list footer, the page footer.

I didn't make a component to wrap the `TodoEntry { }` block, as it seemed unnecessary.


ealmloff (approved):
Thank you! This looks much better

PR #1572, opened 2023-10-23
Feat: add global context functions

jkelleyrtp:
Add functions like window() and router() to allow dynamically grabbing global contexts without have to use the hook variants.

Deprecates the existing hook variants to discourage folks from adding more noise to their codebases.


ealmloff (approved):
Overall, I like this, but there are definitely some caveats. If the API changes to call some function directly instead of using a factory, it can be more difficult to use in threads or tokio futures. For example, the eval hook changes from a function that creates an eval object to a function that directly evaluates some javascript that needs to be within the dioxus runtime. That will make calling the function from a tokio runtime like this impossible:
```rust
tokio::spawn(async move {eval("console.log")});
```

PR #1563, opened 2023-10-21
Append prop documentation to function when using `#[inline_props]` or `#[component]`

tigerros:
From this source code:
```rs
/// Renders a button.
///
/// 
/// 
/// Bunch of lines.
///
/// 
/// # Example
/// ```rs
/// rsx! {
///     Button {
///         label: "Label",
///         action: "Print",
///     }
/// }
/// ```
#[component]
fn Button(
    cx: Scope,
    /// Shows the given label.
    ///
    /// Just two new lines.
    #[deprecated]
    label: Option<&'static str>,
    /// Does this thing when clicked.
    ///
    ///
    ///
    ///
    ///
    ///
    ///
    /// Lot of lines
    action: &'static str
) -> Element {
    let label = label.unwrap_or("defaul");
    render! {
        "Button with label {label} that should do {action} when clicked"
    }
}
```

This documentation is generated:

[image](https://github.com/DioxusLabs/dioxus/assets/76173380/6a698127-6f3a-45db-bd6e-672630aa90be)

*The argument links point to the struct fields.*

There's still some issues. For example, attributes aren't shown, and there's unnecessary spaces in the prop types. I'll mark this as a draft PR so I can work that out. In the meantime, comment any other concerns.

PR #1573, opened 2023-10-23
Fix prop docs line breaks

tigerros:
I'm sorry for creating a third PR for this, but I just noticed that there's one too many line breaks in the generated docs for props (see https://github.com/DioxusLabs/dioxus/pull/1563 and https://github.com/DioxusLabs/dioxus/pull/1565). I changed it to use the Rustdoc approach, which puts text separated by two lines into separate `<p>` elements. Previously, it just replaced `\n` with `<br/>`

Right now, this documentation:
```
/// Doc.
/// Next line.
```

Would generate:

```
Doc.
Next line.
```

But it's actually supposed to be:

```
Doc. Next line.
```

And when you used two new lines, the gap between them would be too big. Example image with the bug:
image
Fixed:
image
Regular Rustdoc:
image

### DIFF
--- a/packages/core-macro/src/component_body_deserializers/inline_props.rs
+++ b/packages/core-macro/src/component_body_deserializers/inline_props.rs
@@ -186,13 +186,13 @@ fn get_props_docs(fn_ident: &Ident, inputs: Vec<&FnArg>) -> Vec<Attribute> {
         let arg_name = arg_name.into_token_stream().to_string();
         let arg_type = crate::utils::format_type_string(arg_type);
 
-        let input_arg_doc =
-            keep_up_to_n_consecutive_chars(input_arg_doc.trim(), 2, '\n').replace('\n', "<br/>");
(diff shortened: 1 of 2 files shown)


tigerros:
I'm confused. `cargo check --all --examples --tests` compiles on my machine. I didn't change any versions either which seems to be the problem...

ealmloff:
> I'm confused why the checks are failing. cargo check --all --examples --tests compiles on my machine. I didn't change any versions either which seems to be the problem...

Some dependency of dioxus may have updated within the last hour since the previous CI run

tigerros:
I ran `cargo update` and `cargo check --all --examples --tests` and still no error. Could you re-run the checks? If not, I can make an empty commit.

jkelleyrtp:
Looks great - thanks @tigerros !

jkelleyrtp (approved):

PR #1575, opened 2023-10-24
Draft: Create hooks that use underlying locks instead of RefCell

DanielAlbl:
I wanted to get some thoughts on:
- If it makes sense to have the separate `use_lock` and `use_async_lock`
- If it would be better to use a `Mutex` instead of an `RwLock`
- If it's necessary to use `Arc` for `value` and `dirty`

### DIFF
--- a/packages/hooks/Cargo.toml
+++ b/packages/hooks/Cargo.toml
@@ -20,6 +20,7 @@ tracing = { workspace = true }
 thiserror = { workspace = true }
 slab = { workspace = true }
 dioxus-debug-cell = "0.1.1"
+async-std = "1.12.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


ealmloff:
There is a use_rw hook in the [dioxus-std crate](https://github.com/DioxusLabs/dioxus-std/blob/master/src/utils/rw/use_rw.rs) that seems similar to use_lock. I think `use_async_lock` could be difficult to work with because it forces every code that interacts with the hook to be in an async block, but if there is some common situation where this hook is necessary, it might fit better in the dioxus-std crate

DanielAlbl:
@ealmloff Wow I didn't even know about dioxus-std. I agree if anything, `use_async_lock` should go there.

PR #1579, opened 2023-10-25
Add Newtype so that Rust-url can be IntoRoutable

stephenandary:
since url::Url is commonly used for many crates, impl IntoRoutable for a rust-url newtype can make it more ergonomic to work with some crates.

### DIFF
--- a/packages/router/src/components/link.rs
+++ b/packages/router/src/components/link.rs
@@ -54,6 +54,8 @@ impl From<&str> for IntoRoutable {
         IntoRoutable::FromStr(value.to_string())
     }
 }
+
+/// Newtype to make is easier to work with rust-url type that other crates commonly use.
 pub struct Url(RustUrl);
 
 impl From<Url> for IntoRoutable {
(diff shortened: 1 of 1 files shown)


packages/router/src/components/link.rs
```diff
@@ -53,6 +55,21 @@ impl From<&str> for IntoRoutable {
     }
 }
 
+/// Newtype to make is easier to work with rust-url type that other crates commonly use.
+pub struct Url(RustUrl);
```
ealmloff, line 59: I think since the type IntoRoutable is defined in this crate, can we implement this directly on url::Url (https://doc.rust-lang.org/std/convert/trait.Into.html#implementing-into-for-conversions-to-external-types-in-old-versions-of-rust)

stephenandary: Sure, I wasn't sure what the best practice was for working with external types. I thought it was to create a newtype and manipulate that.

stephenandary: You beat me to it haha.

ealmloff: > Sure, I wasn't sure what the best practice was for working with external types. I thought it was to create a newtype and manipulate that.

I generally only use newtypes to get around rust's rules about foreign traits on foreign items

stephenandary: Over my head atm, I will trust your judgment. Thanks for updating it.

ealmloff (approved):
Thank you!

PR #1580, opened 2023-10-25
Remove the ignore leaks in MIRI

ealmloff:
This removes the flag that causes MIRI to ignore leaks. We ran into failures with this flag before, but that could have been caused by #1364

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -86,8 +86,7 @@ jobs:
 
         # working-directory: tokio
         env:
-          #  todo: disable memory leaks ignore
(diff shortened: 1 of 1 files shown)


ealmloff:
We might be leaking because of the [Box inside of Component](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L325C40-L325C40) that is bump allocated inside of [dynamic_nodes](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/nodes.rs#L60C15-L60C15)

jkelleyrtp (approved):

PR #1584, opened 2023-10-27
Silence desktop build outputs from cargo

hirschenberger:
Silence additional disturbing output from cargo in desktop builds


ealmloff (approved):
Thank you!

PR #1588, opened 2023-10-29
Make window invisible until the first render

### LINKED ISSUE #1120: Brief flash when opening a dark mode app on a light mode macOS desktop
When opening a dark-mode only macOS Dioxus app on a light mode desktop, there is a very unpleasant flash when it is starting up where the WebView switches from light mode to dark mode.

You can see it in example in a video here:

https://github.com/DioxusLabs/dioxus/assets/132234/b2d274ff-dab5-49ee-8b1a-8768bfd23b4d

I currently maintain a [Dioxus fork](https://github.com/terhechte/dioxus/tree/dioxus_macos_hacks) so that [Ebou](https://github.com/terhechte/Ebou) doesn't have this unpleasant flash.

I'd love to contribute my change / fix but it is quite the hack. At the bottom of this post you can find the source that I'm using for the example app, and I've tried to enforce dark mode in all places.

My current hack to solve this is to delay displaying the web view for a brief amount of time right after window creation. The window itself is properly dark (via the `tao` theme), just the web view needs a couple of frames to parse the initial html / css and then go to dark mode. I also tried setting the background color of the web view to transparent but that lead to other problems.

[Here's my hack](https://github.com/DioxusLabs/dioxus/compare/master...terhechte:dioxus:dioxus_macos_hacks#diff-b0579535f28f7caa593a8664352df35bb34350c47fe2c326d6f4af792bb3a439R98)

``` rs
#[cfg(target_os = "macos")]
    {
        use objc::runtime::Object;
        use objc::runtime::{NO, YES};

        use crate::wry::webview::WebviewExtMacOS;
        let native_webview = build.webview();
        unsafe {
            use cocoa::appkit::NSApp;
            let _: () = msg_send![native_webview, setHidden: YES];

            dispatch::Queue::main().exec_after(Duration::from_millis(250), || {
                let mw: *mut Object = msg_send![NSApp(), mainWindow];
                let mx: *mut Object = msg_send![mw, contentView];
                let uux: *mut Object = msg_send![mx, subviews];
                let first: *mut Object = msg_send![uux, objectAtIndex: 0];
                let _: () = msg_send![first, setHidden: NO];
            });
        }
    }
```

I'm raising this issue to start a discussion around:

- Whether a fix for this is even wanted
- If the hack above would maybe even suffice to include it in Dioxus
- Or if there is a way to allow a Dioxus consumer to hook into window creation so that this hack could live in the corresponding Dioxus apps (such as Ebou).

**Environment:**
 - Dioxus version: `master`
 - OS info: macOS Ventura
 - App platform: `desktop

**Questionnaire**

---

### Example app source

``` toml
[dependencies]
dioxus = { git = "https://github.com/DioxusLabs/dioxus" }
dioxus-desktop = { git = "https://github.com/DioxusLabs/dioxus" }
```

``` rs
#![allow(non_snake_case)]
use dioxus::prelude::*;
use dioxus_desktop::{Config, WindowBuilder};

pub fn default_window() -> WindowBuilder {
    let builder = WindowBuilder::new();

    builder
        .with_title("Darkflash")
        .with_theme(Some(dioxus_desktop::tao::window::Theme::Dark))
}

pub fn main() {
    let style = r#":root { color-scheme: dark; }"#;
    let config = Config::new()
        .with_custom_head(format!(
            r#"
        <style>{style}</style>
        "#
        ))
        .with_window(default_window());

    dioxus_desktop::launch_with_props(RootApp, RootAppProps {}, config);
}

pub struct RootAppProps {}

pub fn RootApp(cx: Scope<'_, RootAppProps>) -> Element<'_> {
    render! {
        div {
            "Hello World"
        }
    }
}

```

### LINKED ISSUE #1585: Blinking window at startup
**Problem**

At the loading of window, at first an empty window is created, then it becomes filled with content.

**Steps To Reproduce**

- Create a hello world desktop project and run it.

**Expected behavior**

The content should immediately appear on the window. If it is not feasible, the window should be hidden until the content is loaded.

**Screenshots**

https://github.com/DioxusLabs/dioxus/assets/45197576/3e99ed95-50c4-4f18-b40e-87bd3b15aad7

Sorry if it is duplicate, and thanks for the great project!

HKalbasi:
fix #1585 
fix #1120

### DIFF
--- a/packages/desktop/src/lib.rs
+++ b/packages/desktop/src/lib.rs
@@ -161,6 +161,7 @@ pub fn launch_with_props<P: 'static>(root: Component<P>, props: P, cfg: Config) 
     // iOS panics if we create a window before the event loop is started
     let props = Rc::new(Cell::new(Some(props)));
     let cfg = Rc::new(Cell::new(Some(cfg)));
+    let mut is_visible_before_start = true;
 
     event_loop.run(move |window_event, event_loop, control_flow| {
         *control_flow = ControlFlow::Wait;
@@ -209,6 +210,8 @@ pub fn launch_with_props<P: 'static>(root: Component<P>, props: P, cfg: Config) 
 
                 // Create a dom
                 let dom = VirtualDom::new_with_props(root, props);
+
+                is_visible_before_start = cfg.window.window.visible;
 
                 let handler = create_new_window(
                     cfg,
@@ -323,6 +326,10 @@ pub fn launch_with_props<P: 'static>(root: Component<P>, props: P, cfg: Config) 
                 EventData::Ipc(msg) if msg.method() == "initialize" => {
                     let view = webviews.get_mut(&event.1).unwrap();
                     send_edits(view.dom.rebuild(), &view.desktop_context.webview);
+                    view.desktop_context
+                        .webview
+                        .window()
+                        .set_visible(is_visible_before_start);
                 }
 
                 EventData::Ipc(msg) if msg.method() == "browser_open" => {
--- a/packages/desktop/src/webview.rs
+++ b/packages/desktop/src/webview.rs
@@ -13,7 +13,7 @@ pub fn build(
     proxy: EventLoopProxy<UserWindowEvent>,
 ) -> (WebView, WebContext) {
     let builder = cfg.window.clone();
-    let window = builder.build(event_loop).unwrap();
+    let window = builder.with_visible(false).build(event_loop).unwrap();
     let file_handler = cfg.file_drop_handler.take();
     let custom_head = cfg.custom_head.clone();
     let index_file = cfg.custom_index.clone();


packages/desktop/src/lib.rs
```diff
@@ -323,6 +323,7 @@ pub fn launch_with_props<P: 'static>(root: Component<P>, props: P, cfg: Config)
                 EventData::Ipc(msg) if msg.method() == "initialize" => {
                     let view = webviews.get_mut(&event.1).unwrap();
                     send_edits(view.dom.rebuild(), &view.desktop_context.webview);
+                    view.desktop_context.webview.window().set_visible(true);
```
ealmloff, line 326: We should probably respect the visible setting the user sets in the [with_window](https://github.com/DioxusLabs/dioxus/blob/d4e6876ff648159c78ad008ade4e3b1b469bb996/packages/desktop/src/cfg.rs#L98) function in the builder. We could store the visible setting that is set after that function call and only make the window visible again if the visible setting is true

(Invisible windows are useful for headless testing)

ealmloff (changes requested):

ealmloff (approved):
Thank you!

PR #1590, opened 2023-10-29
Add DesktopElement::eval method and export it

matthunz:
Adds a new method `eval` to the `DesktopElement` struct and makes the struct public.

This allows calling javascript from dioxus-desktop using an html element. The difference with `use_eval` is the javascript being executed here has access to `window.currentNode` which is the currently mounted HTML node.

It's still pretty rough code but I'd love to know what you guys think!

### DIFF
--- a/packages/desktop/src/element.rs
+++ b/packages/desktop/src/element.rs
@@ -1,7 +1,8 @@
+use crate::{desktop_context::DesktopContext, query::QueryEngine};
 use dioxus_core::ElementId;
 use dioxus_html::{geometry::euclid::Rect, MountedResult, RenderedElementBacking};
-
-use crate::{desktop_context::DesktopContext, query::QueryEngine};
+use serde::de::DeserializeOwned;
+use std::borrow::Cow;
 
 /// A mounted element passed to onmounted events
 pub struct DesktopElement {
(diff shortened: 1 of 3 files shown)


ealmloff:
Overall, I think an API like this would be very useful, but I'm not sure about this API. Ideally we should expand some API to all platforms that can support something like this (web, desktop, liveview). If we enable an API like this on liveview, it will make interacting with multiple elements at once difficult. Each time you want to interact with an element you need to use that element's eval function instead of the normal eval. For example, imagine you create a list of elements with onmounted on each one and you want to find the combined area of those elements. With this API, you need to create a new script for each element which would result in more network requests on liveview.

We could expose a get_element function on the element that returns either a function call with all the parameters filled in as a string. For example, if you are on element id 1 on desktop: get_element would return `"window.get_dioxus_element(1)"`

jkelleyrtp:
> Overall, I think an API like this would be very useful, but I'm not sure about this API. Ideally we should expand some API to all platforms that can support something like this eventually (web, desktop, liveview). If we enable an API like this on liveview, it will make interacting with multiple elements at once difficult. Each time you want to interact with an element you need to use that element's eval function instead of the normal eval. For example, imagine you create a list of elements with onmounted on each one and you want to find the combined area of those elements. With this API, you need to create a new script for each element which would result in more network requests on liveview.
> 
> We could expose a get_element function on the element that returns either a function call with all the parameters filled in as a string. For example, if you are on element id 1 on desktop: get_element would return `"window.get_dioxus_element(1)"`

Agreed, I think tweaking this PR or a new PR that allows you to query our element slab would probably be better. The way around this right now is to use onmounted or set an ID on the element and then query the element by its ID. Both of which aren't perfect but should allow developers to do most things.

PR #1593, opened 2023-10-30
Add read untracked to signals

ealmloff:
Adds a read untracked method to Signals. This method is like read but it doesn't subscribe to the current context. This can be useful for signals you do not want to track inside of a `dioxus_signals::effect`


jkelleyrtp:
Can we rename this to peek and then mention it in the docs for regular reads?

jkelleyrtp:
Just FYI peak and peek are two different spellings. We want peek (peak is like for mountains)

jkelleyrtp (approved):

Issue #1568, opened 2023-10-23
Send + Sync Signals


ealmloff:
## Specific Demand

Signals can currently only be used in the thread they were created in. This works great within the virtual dom, but it makes using signals within threads or tokio futures difficult.

## Implement Suggestion

We can add a generic to signals to make them automatically sync + send when you use them in a context like a new thread that requires them to be sync + send. Something like [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6681e7810eac2ed853ed6f26eb255720)

jkelleyrtp:
Solved by #1791

PR #1594, opened 2023-10-30
Make signals optionally Send+Sync

ealmloff:
Adds a generic on signals to mark the signal as sync + send.

# Design

Why is this not a separate type? (generics can be confusing)

If signal was the only thing that needs to be sync, I think a separate type would be simpler. The issue is if we switch signal to be the preferred state primitive, every hook would need two different versions: one that is Send + Sync with the Send + Sync signal type and one version that is not Send + Sync. Currently, state uses a completely separate type for Send + Sync state: [use_rw](https://github.com/DioxusLabs/dioxus-std/blob/master/src/utils/rw/use_rw.rs). This works for simple state, but it makes Send + Sync versions of hooks like `use_memo`, `use_shared_state`, and `use_future` difficult

closes https://github.com/DioxusLabs/dioxus/issues/1568


jkelleyrtp:
This got merged into #1791 so I'm closing it here

PR #1597, opened 2023-11-01
Use indentation settings in dx fmt and vscode extension

### LINKED ISSUE #1595: dx fmt does not respect file's indentation style
**Problem**

`dx fmt` forces an indentation style of four spaces per indent level, even if the file uses a different amount of spaces or tabs.

**Steps To Reproduce**

- Create a new project with `dx create`
- Create a `rustfmt.toml` file at the project root with either `hard_tabs = true` or `tab_spaces = 2` (or some other number other than 4)
- Run `cargo fmt`
- Run `dx fm`
- View the `main.rs` file and enable visible whitespaces in your editor.

**Expected behavior**

The indentation produced by `dx fmt` matches the indentation of the rest of the file.

**Screenshots**

20231031_09h42m45s_grim

**Environment:**
 - Dioxus version: dioxus 0.4.1
 - Rust version: 1.72.1
 - OS info: Arch Linux
 - App platform: Not Applicable

**Questionnaire**

ColonelThirtyTwo:
Adds the ability to specify an indent string to public autofmt methods - either a sequence of spaces or a tab character.

Get the indentation style and size from rustfmt for dx fmt, or from the editor settings for the vscode extension.

Closes #1595


packages/autofmt/src/element.rs
```diff
@@ -86,7 +86,7 @@ impl Writer<'_> {
 
         // if we have few children and few attributes, make it a one-liner
         if is_short_attr_list && is_small_children {
-            if children_len.unwrap() + attr_len + self.out.indent * 4 < 100 {
+            if children_len.unwrap() + attr_len + self.out.indent_level * 4 < 100 {
```
ColonelThirtyTwo, line 89: This should be adjusted

packages/autofmt/src/lib.rs
```diff
@@ -76,7 +77,7 @@ pub fn fmt_file(contents: &str) -> Vec<FormattedBlock> {
 
         let rsx_start = macro_path.span().start();
 
-        writer.out.indent = leading_whitespaces(writer.src[rsx_start.line - 1]) / 4;
+        writer.out.indent_level = leading_whitespaces(writer.src[rsx_start.line - 1]) / 4;
```
ColonelThirtyTwo, line 80: This should be adjusted

ColonelThirtyTwo:
I don't have the setup to test changes to the VS Code extension, so testing that would be appreciated.

ealmloff:
> I don't have the setup to test changes to the VS Code extension, so testing that would be appreciated.

It appears `editor.options.indentSize` does not exist
```
ERROR in /Users/evanalmloff/Desktop/Github/dioxus/packages/extension/src/main.ts
./src/main.ts 28:64-74
[tsl] ERROR in /Users/evanalmloff/Desktop/Github/dioxus/packages/extension/src/main.ts(28,65)
      TS2339: Property 'indentSize' does not exist on type 'TextEditorOptions'.
```

ColonelThirtyTwo:
> > I don't have the setup to test changes to the VS Code extension, so testing that would be appreciated.
> 
> It appears `editor.options.indentSize` does not exist
> 
> ```
> ERROR in /Users/evanalmloff/Desktop/Github/dioxus/packages/extension/src/main.ts
> ./src/main.ts 28:64-74
> [tsl] ERROR in /Users/evanalmloff/Desktop/Github/dioxus/packages/extension/src/main.ts(28,65)
>       TS2339: Property 'indentSize' does not exist on type 'TextEditorOptions'.
> ```

Hmm, it's documented here: https://code.visualstudio.com/api/references/vscode-api#TextEditorOptions

I can switch to `tabSize` for now.

ealmloff (approved):
I confirmed the VSCode extension is working now, thank you!

(waiting for checks before merging this)

ColonelThirtyTwo:
Looks like the x86_64-apple-darwin image is missing `cargo fmt`, which this patch uses to get the indentation settings for `dx fmt`.

packages/extension/src/main.ts
```diff
@@ -90,7 +90,13 @@ function fmtDocument(document: vscode.TextDocument) {
 		if (!editor) return; // Need an editor to apply text edits.
 
 		const contents = editor.document.getText();
-		const formatted = dioxus.format_file(contents, !editor.options.insertSpaces, editor.options.tabSize);
+		let tabSize: number;
+		if (typeof editor.options.tabSize === 'number') {
```
ColonelThirtyTwo, line 94: The docs say this is always a number when being read.

ealmloff: I'm not sure about the docs, but when I tried to compile this the type of `editor.options.insertSpaces` is `number | string`. It could be something with my setup. I might have an older version of the type definitions. If you can get the extension compiling on your machine without this check, let me know.

If you are interested in testing this, you can follow the build settings [here](https://code.visualstudio.com/api/working-with-extensions/bundling-extension). Dioxus has the webpack config already setup

ColonelThirtyTwo: Yeah, it definitely needs changing to make the types line up. I'm saying at runtime,  you'll always get back a number, so the if branch will always match (though it won't hurt).

It's typed as `number | string` because you can assign `"auto"` to it to have it to `indentSize`, but when you read it, it will always be a `number`.

ealmloff:
> Looks like the x86_64-apple-darwin image is missing cargo fmt, which this patch uses to get the indentation settings for dx fmt.

You can modify the CI to add the rustfmt component like we do [here](https://github.com/DioxusLabs/dioxus/blob/master/.github/workflows/main.yml#L69)

jkelleyrtp:
@ColonelThirtyTwo LGTM - fixed the CI. Thanks!!

PR #1598, opened 2023-11-01
Update root README to use `#[component]` macro

tigerros:
- Renames the `app` component in the root README to `App`.
- Adds the `#[component]` macro to mentioned component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@
 Dioxus is a portable, performant, and ergonomic framework for building cross-platform user interfaces in Rust.
 
 ```rust
-fn app(cx: Scope) -> Element {
+#[component]
+fn App(cx: Scope) -> Element {
     let mut count = use_state(cx, || 0);
 
     cx.render(rsx! {
@@ -155,7 +156,7 @@ Dioxus is also a very extensible platform.
 So... Dioxus is great, but why won't it work for me?
 - It's not fully mature yet. APIs are still shifting, things might break (though we try to avoid it)
 - You need to run in a no-std environment.
-- You don't like the React-hooks model of building UIs
+- You don't like the React-hooks model of building UIs.
 
 
 ## Contributing


ealmloff:
Not every component needs the component attribute. If the component has no props, or a manual props struct it doesn't need the automatic props struct generation. The app component should never need the component attribute because it never accepts props and it is never used as a component so it doesn't need to be upper camel case. If it doesn't need the component attribute, I think the example is more clear without extra macros

tigerros:
> Not every component needs the component attribute. If the component has no props, or a manual props struct it doesn't need the automatic props struct generation.

I think it's better in the long term to make a standard component macro. It's functionality is not necessarily limited to generate the props, and it doesn't generate them if there's only one argument, which also works for a manual props struct. This way, it's immediately clear what functions are components, and there is a clear convention at play.

> The app component doesn't need the component attribute because it doesn't accept props and it is never used as a component in the rsx macro so it doesn't need to be upper camel case.

By doing this, we basically break our own naming conventions. React's `App` component also uses upper camel case.

> I think the example is more clear without extra macros because the component macro is unneeded and it can be unclear what the component macro is doing if you are new to the framework.

I don't feel it's necessary to know what the macro is doing, as long it's clear what it's declaring, seeing as Dioxus is an explicitly declarative framework. Also see my first point.

ealmloff:
> > Not every component needs the component attribute. If the component has no props, or a manual props struct it doesn't need the automatic props struct generation.
> 
> I think it's better in the long term to make a standard component macro. It's functionality is not necessarily limited to generate the props, and it doesn't generate them if there's only one argument, which also works for a manual props struct. This way, it's immediately clear what functions are components, and there is a clear convention at play.
> 
> > The app component doesn't need the component attribute because it doesn't accept props and it is never used as a component in the rsx macro so it doesn't need to be upper camel case.
> 
> By doing this, we basically break our own naming conventions. React's `App` component also uses upper camel case.
> 
> > I think the example is more clear without extra macros because the component macro is unneeded and it can be unclear what the component macro is doing if you are new to the framework.
> 
> True, but see my first point.

That is fair. I think changing app to be App makes sense to differentiate app as a component, but until the `#[component]` does something to components that take no props, I think we can leave off the macro (at least for the first example anyone new to the framework sees)

jkelleyrtp:
Yeah, agree with @ealmloff  here - it's not necessary for the readme example and IMO takes away from the message of simplicity we're trying to convey.

PR #1599, opened 2023-11-01
Fix `#[component]` expansion Clippy warning

tigerros:
Adds a `#[allow(clippy::inline_always)]` attribute to the generated `__dx_inner_comp` function.


ealmloff (approved):
Thank you!

PR #1602, opened 2023-11-03
Fix the scroll event on the web renderer

ealmloff:
Fixes a type mismatch that broke the scroll event on the web renderer


packages/web/src/dom.rs
```diff
@@ -294,7 +294,7 @@ pub fn virtual_event_from_websys_event(event: web_sys::Event, target: Element) -
         "select" => Rc::new(SelectionData {}),
         "touchcancel" | "touchend" | "touchmove" | "touchstart" => Rc::new(TouchData::from(event)),
 
-        "scroll" => Rc::new(()),
+        "scroll" => Rc::new(ScrollData {}),
```
Lz020316, line 297: How to use onscorll, I want to get scollheight, scrolltop, etc. Is there a simple example?

PR #1603, opened 2023-11-03
Bump tauri bundler for using proxy to fetch resources

divinerapier:


packages/cli/Cargo.toml
```diff
@@ -72,8 +72,8 @@ cargo-generate = "0.18"
 toml_edit = "0.19.11"
 
 # bundling
-tauri-bundler = { version = "=1.3.0", features = ["native-tls-vendored"] }
-tauri-utils = "=1.4.*"
+tauri-bundler = { version = "=1.4.*", features = ["native-tls-vendored"] }
```
ealmloff, line 75: We can't allow minor version bumps of tauri-bundler because the crate doesn't follow semvar versioning. The version just matches the tauri version regardless of breaking changes

PR #1607, opened 2023-11-04
fix syntax error at `#[derive(Props)]` using const generics

syrflover:
# Reproduce problem
https://github.com/syrflover/derive-props-with-const-generics

To trigger an error, change the version of dioxus to 0.4 in `Cargo.toml`

# What is problem

image

https://github.com/DioxusLabs/dioxus/blob/a3e6d0adcaee49495db5c324dbf9f20d5ec529ed/packages/core-macro/src/props/mod.rs#L554-L566

https://github.com/DioxusLabs/dioxus/blob/a3e6d0adcaee49495db5c324dbf9f20d5ec529ed/packages/core-macro/src/props/mod.rs#L645-L651

When it is `GenericParam::Const`, it doesn't return anything. However, if that item is included in an iterator, a syntax error occurs because it generates commas as many times as the length of the iterator when creating tokens.

# How to fixed it

https://github.com/syrflover/dioxus/blob/4f8b60834af43fef879c7eb0ff46b74d23a6eabe/packages/core-macro/src/props/mod.rs#L554-L564

Changed to returns `None`, when it is `GenericParam::Const` using `filter_map`.

### DIFF
--- a/packages/core-macro/src/props/mod.rs
+++ b/packages/core-macro/src/props/mod.rs
@@ -551,18 +551,16 @@ mod struct_info {
             let generics_with_empty = modify_types_generics_hack(&ty_generics, |args| {
                 args.insert(0, syn::GenericArgument::Type(empties_tuple.clone().into()));
             });
-            let phantom_generics = self.generics.params.iter().map(|param| match param {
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Good catch and very clear explanation, thank you!

Issue #1608, opened 2023-11-04
`AtomState::current()` doesn't return the current value


marc2332:
**Problem**

Fermi's `AtomState::current()` doesn't actually return the current value, it does the same thing as `AtomState::get()`
 but it clons the inner value
**Steps To Reproduce**

1. Create an AtomState
2. Clone it and hold it in some event handler / async block, and do regular println!() on the value returned by .current()
3. Modify it
4. The value printed in 2. will still be the old one

**Expected behavior**

Print the new value

**Environment:**
 - Dioxus version: 0.4
 - Rust version: 1.73
 - OS info: Windows 11
 - App platform: freya

**Questionnaire**
- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

PR #1609, opened 2023-11-04
fix: Read value from root when calling `AtomState::current()`

marc2332:
Fixes https://github.com/DioxusLabs/dioxus/issues/1608

This will make `AtomState::current()` read from the Root value instead of the AtomState inner value, this makes it work properly on async contexts / event handlers


ealmloff (approved):
Thank you!

PR #1610, opened 2023-11-04
Serializing FormData while form 'onsubmit'.

bunnyBites:
current ForData values:
`{ username: ["rust"], password: ["dioxus"]}`

 what we need it to be:
`{ username: "rust", password: "dioxus"}`

Reason:
I faced this while I was trying to get the form values that are to be sent to an API. I thought it would be better if the framework gave us an option to get the ready-to-use values object.


ealmloff (commented):
I have a few questions about this API, but overall I think adding an easy way to deserialize form data into a struct is a great addition to Dioxus!


If `get_parsed_values` should only be used to deserialize form values, can we directly expose a function that deserializes form values and make `get_parsed_values` private?

```rust
fn parse_json<T>(&self) -> Result<T, serde_json::Error>
where
    T: serde::de::DeserializeOwned,
{
    let formatted = self.get_parsed_values();
    let as_string = serde_json::to_string(&formatted)?;
    serde_json::from_str(&as_string)
}
```


This will not handle deserializing multiple select elements correctly

If we have a form like this:

```rust
form {
    onsubmit: move |evt| {
        // How should we deserialize the form data?
        println!("{:#?}", evt.values);
    },
    select {
        multiple: "true",
        name: "options",

        option {
            value: "option1",
            "The first option"
        }
        option {
            value: "option2",
            "The second option"
        }
        option {
            value: "option3",
            "The third option"
        }
    }
}
```

I would create this struct to deserialize the form data:

```rust
#[derive(Deserialize)]
enum MyOptions {
    #[serde(rename = "option1")]
    Option1,
    #[serde(rename = "option2")]
    Option2,
    #[serde(rename = "option3")]
    Option3,
}

#[derive(Deserialize)]
struct FormData {
    options: Vec<MyOptions>,
}
```

But options will not deserialize correctly because the form data has items separated by commas. We could either add something to the form data to indicate that a value was a list or we could modify the deserializer to try to deserialize a single value list as either a single value or a list.

bunnyBites:
> You modified the terminal renderer html value. The main [html](https://github.com/DioxusLabs/dioxus/blob/a3e6d0adcaee49495db5c324dbf9f20d5ec529ed/packages/html/src/events/form.rs#L10) is used by all renderers. This function should probably be added to that crate instead of the terminal renderer crate

Sorry, my bad. I would move the code to the appropriate place 🫡.

bunnyBites:
> I have a few questions about this API, but overall I think adding an easy way to deserialize form data into a struct is a great addition to Dioxus!
> 
> If `get_parsed_values` should only be used to deserialize form values, can we directly expose a function that deserializes form values and make `get_parsed_values` private?
> 
> ```rust
> fn parse_json<T>(&self) -> Result<T, serde_json::Error>
> where
>     T: serde::de::DeserializeOwned,
> {
>     let formatted = self.get_parsed_values();
>     let as_string = serde_json::to_string(&formatted)?;
>     serde_json::from_str(&as_string)
> }
> ```
> 
> This will not handle deserializing multiple select elements correctly
> 
> If we have a form like this:
> 
> ```rust
> form {
>     onsubmit: move |evt| {
>         // How should we deserialize the form data?
>         println!("{:#?}", evt.values);
>     },
>     select {
>         multiple: "true",
>         name: "options",
> 
>         option {
>             value: "option1",
>             "The first option"
>         }
>         option {
>             value: "option2",
>             "The second option"
>         }
>         option {
>             value: "option3",
>             "The third option"
>         }
>     }
> }
> ```
> 
> I would create this struct to deserialize the form data:
> 
> ```rust
> #[derive(Deserialize)]
> enum MyOptions {
>     #[serde(rename = "option1")]
>     Option1,
>     #[serde(rename = "option2")]
>     Option2,
>     #[serde(rename = "option3")]
>     Option3,
> }
> 
> #[derive(Deserialize)]
> struct FormData {
>     options: Vec<MyOptions>,
> }
> ```
> 
> But options will not deserialize correctly because the form data has items separated by commas. We could either add something to the form data to indicate that a value was a list or we could modify the deserializer to try to deserialize a single value list as either a single value or a list.

In that case, I think for multi-valued elements like select, we would expect the same result as the 'values' (vector/array of values), which we can get to know based on the length of values.

For example:
 ``if values = { anime: ["DBZ", "Naruto", "Demon slayer"] } // length of anime > 1, no need to deserialize. ``
 
Also, we would expect the same structure from our submit form. Please correct me if I am wrong. In the meantime, let me first see how the app works with select elements with enum variants.

bunnyBites:
> I have a few questions about this API, but overall I think adding an easy way to deserialize form data into a struct is a great addition to Dioxus!
> 
> If `get_parsed_values` should only be used to deserialize form values, can we directly expose a function that deserializes form values and make `get_parsed_values` private?
> 
> ```rust
> fn parse_json<T>(&self) -> Result<T, serde_json::Error>
> where
>     T: serde::de::DeserializeOwned,
> {
>     let formatted = self.get_parsed_values();
>     let as_string = serde_json::to_string(&formatted)?;
>     serde_json::from_str(&as_string)
> }
> ```
> 
> This will not handle deserializing multiple select elements correctly
> 
> If we have a form like this:
> 
> ```rust
> form {
>     onsubmit: move |evt| {
>         // How should we deserialize the form data?
>         println!("{:#?}", evt.values);
>     },
>     select {
>         multiple: "true",
>         name: "options",
> 
>         option {
>             value: "option1",
>             "The first option"
>         }
>         option {
>             value: "option2",
>             "The second option"
>         }
>         option {
>             value: "option3",
>             "The third option"
>         }
>     }
> }
> ```
> 
> I would create this struct to deserialize the form data:
> 
> ```rust
> #[derive(Deserialize)]
> enum MyOptions {
>     #[serde(rename = "option1")]
>     Option1,
>     #[serde(rename = "option2")]
>     Option2,
>     #[serde(rename = "option3")]
>     Option3,
> }
> 
> #[derive(Deserialize)]
> struct FormData {
>     options: Vec<MyOptions>,
> }
> ```
> 
> But options will not deserialize correctly because the form data has items separated by commas. We could either add something to the form data to indicate that a value was a list or we could modify the deserializer to try to deserialize a single value list as either a single value or a list.

I think i have made some progress with my latest commit along with the feedbacks provided. Please bare with me if i made silly mistakes 😔. Please provide feedbacks if i got deviated from the expectations and how to improve upon it 🤓.

packages/html/src/events/form.rs
```diff
@@ -23,6 +30,52 @@ pub struct FormData {
...
+            /*
+            case 1 - multiple values, example { driving_types: ["manual", "automatic"] }
+                In this case we want to return the values as it is, NO point in making
+                driving_types: "manual, automatic"
+
+            case 2 - single value, example { favourite_language: ["rust"] }
+                In this case we would want to deserialize the value as follows
+                favourite_language: "rust"
+            */
+            if values.len() > 1 {
```
ealmloff, line 56: Much better! This still leaves one situation where this will deserialize incorrectly:
If I have rsx like this and I only select the first option:
```rust
form {
    onsubmit: move |evt| {
        // How should we deserialize the form data?
        println!("{:#?}", evt.values);
    },
    select {
        multiple: "true",
        name: "options",

        option {
            value: "option1",
            "The first option"
        }
        option {
            value: "option2",
            "The second option"
        }
        option {
            value: "option3",
            "The third option"
        }
    }
}
```
And I want to deserialize it into this struct:
```rust
#[derive(Deserialize)]
enum MyOptions {
    #[serde(rename = "option1")]
    Option1,
    #[serde(rename = "option2")]
    Option2,
    #[serde(rename = "option3")]
    Option3,
}

#[derive(Deserialize)]
struct FormData {
    options: Vec<MyOptions>,
}
```

We can't know from the data we currently have in the FormData struct if the data is from a single value or from many values.

This is the issue I was talking about here:
> We could either add something to the form data to indicate that a value was a list or we could modify the deserializer to try to deserialize a single value list as either a single value or a list.

We could [write a custom data format](https://serde.rs/data-format.html) or just change the type of values to `Hashmap<String, ValueType>` and serialize the data correctly when we send the event from javascript

bunnyBites: Yes, that will be a problem. It would be better if I get the type of values and use it in the if condition. I'll check if I can get a grab on that.

packages/html/src/events/form.rs
```diff
@@ -12,6 +20,8 @@ pub struct FormData {
 
     pub values: HashMap<String, Vec<String>>,
 
+    pub value_types: HashMap<String, String>,
```
ealmloff, line 23: Instead of having two different maps for the values and the value types, can we use one hashmap of `HashMap<String, ValueType>`? That will let different renderers decide what events have multiple values and it insures that every value has a value_type

bunnyBites: That will be neat. But in that case, we might need to check the effect. I'll make the changes and let you know 👾

bunnyBites: I also wanted to ask if this code can be moved as a common util function:

```rust
fn convert_hashmap_to_json<K, V>(hashmap: &HashMap<K, V>) -> serde_json::Result<String>
where
    K: Serialize + std::hash::Hash + Eq,
    V: Serialize,
{
    serde_json::to_string(hashmap)
}

```

ealmloff: I don't feel like that will come up again. We don't have a common crate that is shared between all of the Dioxus crates. If we need to deserialize from a hashmap into another structure we can just copy-paste the function or reconsider and pull out the function

bunnyBites: > Instead of having two different maps for the values and the value types, can we use one hashmap of `HashMap<String, ValueType>`? That will let different renderers decide what events have multiple values and it insures that every value has a value_type

Do you think we can make it a separate PR? Because it seems like a new feature. I would gladly work on that as well 😇.

ealmloff: You can open a separate PR if you would prefer that. It is just a different way to structure the data, not necessarily a new feature. I'm also happy to take over implementing that change if it isn't something you are interested in.

We are preparing for a 0.5 release soon. Because this changes the public API of `FormData`, I don't think we should merge this yet if we know the API will change soon so that we can release 0.5 before merging the followup PR.

bunnyBites: Sure, I understand. I have already started working on the requirements, We will work this out together. 

Really excited for 0.5 🧬🧬🧬.  I'm just glad that I'm part of it (I am right ? 🫣)

packages/html/src/events/form.rs
```diff
@@ -1,16 +1,24 @@
 use std::{any::Any, collections::HashMap, fmt::Debug};
 
 use dioxus_core::Event;
+use serde::{Deserialize, Serialize};
 
 pub type FormEvent = Event<FormData>;
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(untagged)] // this will serialize Text(String) -> String and VecText(Vec<String>) to Vec<String>
+pub enum ValueType {
```
ealmloff, line 10: Serde is an optional dependancy, this will fail if serde is disabled. We need to make serde always enabled to fix this

packages/html/src/events/form.rs
```diff
@@ -23,6 +31,26 @@ pub struct FormData {
...
+    serde_json::to_string(hashmap)
+}
+
+impl FormData {
+    pub fn parse_json<T>(&self) -> Result<T, serde_json::Error>
+    where
+        T: serde::de::DeserializeOwned,
+    {
+        let parsed_json =
+            convert_hashmap_to_json(&self.values.clone()).expect("Failed to parse values to JSON");
```
ealmloff, line 48: I don't think we need to clone here because you take a reference to the value directly after you clone.
```suggestion
            convert_hashmap_to_json(&self.values).expect("Failed to parse values to JSON");
```

ealmloff (changes requested):
Overall looks pretty good! The API looks great. There is one feature issue and I need to test this with a few examples, but after that this great

packages/html/src/events/form.rs
bunnyBites: I think that currently, this will enable serde as default for ValueType. Are you saying to include #[cfg(feature = "serialize")] for each enum's variants ?

packages/html/src/events/form.rs
bunnyBites: That is a blunder. ill remove it.

packages/html/src/events/form.rs
bunnyBites: Did you mean something like this ?

```rust
#[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
#[derive(Debug, Clone, PartialEq)]
#[serde(untagged)] // this will serialize Text(String) -> String and VecText(Vec<String>) to Vec<String>
pub enum ValueType {
    Text(String),
    VecText(Vec<String>),
}
```

ealmloff: I meant removing `optional = true` in this line https://github.com/DioxusLabs/dioxus/blob/master/packages/html/Cargo.toml#L15. The serialize feature enables serialization of events for renderers, but we can enable deserialization of form values automatically.

bunnyBites: ohh, my bad i saw the wrong file, that's why I misunderstood your expectation.
I checked here by mistake: https://github.com/DioxusLabs/dioxus/blob/master/Cargo.toml

Ill change it.

bunnyBites: > I meant removing `optional = true` in this line https://github.com/DioxusLabs/dioxus/blob/master/packages/html/Cargo.toml#L15. The serialize feature enables serialization of events for renderers to reduce the binary size because of how many events Dioxus supports, but we can enable deserialization of form values automatically. I don't think it would effect the binary size very much

This change is causing issues, need to figure out why ... do you know what might be the cause?

ealmloff: You also need to remove it from the serialize feature now that it is not optional

ealmloff:
It looks like desktop form serialization is broken with this example:
```rust
use dioxus::prelude::*;
use serde::{Deserialize, Serialize};

fn main() {
    dioxus_desktop::launch(app);
}

    cx.render(rsx! {
        div {
            h1 { "Form" }
            form {
                onsubmit: move |ev| println!("Submitted {:?}", ev.values),
                oninput: move |ev| {
                    println!("Input {:?}", ev.values);
                    println!("Submitted {:?}", ev.parsed_values::<MyFormData>());
                },
                input { r#type: "text", name: "username" }
                input { r#type: "text", name: "full-name" }
                input { r#type: "password", name: "password" }
                input { r#type: "radio", name: "color", value: "red" }
                input { r#type: "radio", name: "color", value: "blue" }
                button { r#type: "submit", value: "Submit", "Submit the form" }
            }
        }
    })
}

#[derive(Debug, Serialize, Deserialize)]
struct MyFormData {
    username: String,
    #[serde(rename = "full-name")]
    full_name: String,
    password: String,
    color: String,
}
```

bunnyBites:
> It looks like desktop form serialization is broken with this example:
> 
> ```rust
> use dioxus::prelude::*;
> use serde::{Deserialize, Serialize};
> 
> fn main() {
>     dioxus_desktop::launch(app);
> }
> 
> fn app(cx: Scope) -> Element {
>     cx.render(rsx! {
>         div {
>             h1 { "Form" }
>             form {
>                 onsubmit: move |ev| println!("Submitted {:?}", ev.values),
>                 oninput: move |ev| {
>                     println!("Input {:?}", ev.values);
>                     println!("Submitted {:?}", ev.parsed_values::<MyFormData>());
>                 },
>                 input { r#type: "text", name: "username" }
>                 input { r#type: "text", name: "full-name" }
>                 input { r#type: "password", name: "password" }
>                 input { r#type: "radio", name: "color", value: "red" }
>                 input { r#type: "radio", name: "color", value: "blue" }
>                 button { r#type: "submit", value: "Submit", "Submit the form" }
>             }
>         }
>     })
> }
> 
> #[derive(Debug, Serialize, Deserialize)]
> struct MyFormData {
>     username: String,
>     #[serde(rename = "full-name")]
>     full_name: String,
>     password: String,
>     color: String,
> }
> ```

Yeah, i figured out the issue. The issue is that we have used "form" to get the value type instead, we should have used "target". After changing it now the example you provided will not break anymore.

ealmloff (approved):
Thank you, the example now works as expected on all platforms

(Waiting for a second review before merging this)

bunnyBites:
> Thank you, the example now works as expected on all platforms
> 
> (Waiting for a second review before merging this)

That's awesome. Please let me know if there are any changes I have to make. Super excited right now :)!!!

PR #1612, opened 2023-11-04
Add a noop evaluator to the ssr renderer

ealmloff:
Adds a noop evaluator to the SSR renderer that ignores any incoming evals. This makes it easier to use Eval with the fullstack crate without adding `#[cfg(feature = "web")]` for every eval

### DIFF
--- a/packages/fullstack/examples/axum-hello-world/src/main.rs
+++ b/packages/fullstack/examples/axum-hello-world/src/main.rs
@@ -24,6 +24,7 @@ fn app(cx: Scope<AppProps>) -> Element {
 
     let mut count = use_state(cx, || 0);
     let text = use_state(cx, || "...".to_string());
+    let eval = use_eval(cx);
 
     cx.render(rsx! {
         div {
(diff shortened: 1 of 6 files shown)


anvlkv:
Hi! Thank you for picking up the issue so quickly. 

I [tried with your branch](https://github.com/anvlkv/red-siren/commit/e4d6ff2d83bd5adac47d77387a83ec7588d8d58c) but I'm still experiencing the issue. 

```
[INFO] dioxus_fullstack::render - Rebuilding vdom
thread 'tokio-runtime-worker' panicked at /Users/anvlkv/.cargo/git/checkouts/dioxus-5df58c476196bcfb/160c809/packages/html/src/eval.rs:41:14:
evaluator not provided
[ERROR] dioxus_core::any_props - Error while rendering component `app`: Any { .. }
```

I came around [this issue](https://discord.com/channels/899851952891002890/943190605067079712/1170459845040799804) mainly looking at mobile firefox on iOS which appears to be slow with wasm, and would benefit from some prerendering. 


I've also tried using [`#[cfg(feature=web)]`](https://github.com/anvlkv/red-siren/commit/bfdcee5d818c3766203cf92dab3aa5c2e71f7299#diff-e526094b440f01ccfaa8fe24f7ecf7f09e7ed25f207ec91f71123bd0bcb9611e) this does solve the error, but then shared state is not updated in child component, nor I see any logs from my loop, I do see that eval is running from browser logs.

ealmloff:
I forgot to apply the changes to fullstack. It should be fixed now

jkelleyrtp (approved):

PR #1614, opened 2023-11-06
fix: wasm-bindgen version mismatch

xTeKc:
This PR fixes the issue of `wasm-bindgen` version mismatch by ensuring that all instances of wasm-bindgen in the project are using the same version (`0.2.88`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,7 +88,7 @@ slab = "0.4.2"
 futures-channel = "0.3.21"
 futures-util = { version = "0.3", default-features = false }
 rustc-hash = "1.1.0"
-wasm-bindgen = "0.2.87"
+wasm-bindgen = "0.2.88"
 html_parser = "0.7.0"
 thiserror = "1.0.40"
 prettyplease = { package = "prettier-please", version = "0.2", features = [


swpecht:
Should we instead change to: `wasm-bindgen = "=0.2.88"` (note the equal sign in the version string).

I may misunderstand the root cause, but I think this should keep users from experiencing this issue with future wasm-bingen updates until we manually bump the version.

ealmloff:
Minor versions will update automatically if you don't pass the `--locked` flag when you install the CLI. Adding an `=` would make the version not update automatically when you reinstall the CLI. The important thing here is to make the version of `dioxus-web`'s wasm-bindgen and the CLI's version of [wasm-bindgen-cli-support](https://github.com/DioxusLabs/dioxus/blob/c18f91142937937a32708d9f9c9294cf249e9d76/packages/cli/Cargo.toml#L15) match. If we don't lock the version of `wasm-bindgen` with an `=`, we need to not recommend using the locked flag when installing the CLI. If we do lock the `wasm-bindgen` with an `=`, we need to also lock the `wasm-bindgen-cli-support` version with either an `=` or the `--locked` flag

Issue #85, opened 2022-01-07
Need for synchronous PreventDefault across Tao/Wry boundary


jkelleyrtp:
Unfortunately, there's no way today to conditionally call PreventDefault across the Tao/Wry boundary.

Developers should be able to disable default in a way that's not just hardcoded fields - IE allowing default behavior when right-clicking links in the browser.

The FFI boundary is particularly challenging because we cannot synchronously handle the event.

I *think* the solution is to *always* trap the event and prevent default.

Then, ff 1) the event has default behavior and 2) the user didn't prevent that default behavior in their Dioxus handler, then we should craft the event again and push it back in the real dom.

valyagolev:
> we should craft the event again and push it back in the real dom.

from my (maybe flawed) testing in the chrome console, synthetic events still manage to have a different behaviour:

```
(function(a) {
    document.body.appendChild(a);
    //a.setAttribute('href', 'https://github.com');
    a.setAttribute('target', '_blank');
    a.dispatchEvent((function(e) {
        e.initMouseEvent("click", true, true, window, 0, 0, 0, 0, 0, 
                         true, false, false, false, 0, null);
        return e
    }(document.createEvent('MouseEvents'))))
}(document.querySelector('a[href^=http]')))
```

This opens a new tab and *focuses* it, as opposed to opening it in the background when the cmd-click is done using the real mouse

ealmloff:
We cannot create an event that is identical to the event triggered by the browser. Events that are triggered by a user are [trusted](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted). Events that are modified by javascript will not be trusted

bishoyroufael:
Is there any current workaround for calling prevent default on an event conditionally? I am particularly interested for desktop platform.

nicoburns:
Relevant Wry issue: https://github.com/tauri-apps/wry/issues/454. Looks like there are potential APIs for this, but Wry has no plans to expose them. So possibilities are:
- Synchronous HTTP requests (not nice, but would probably work decently so long as the webviews don't remove support)
- Waiting until webviews implement better APIs which Wry then picks up

ealmloff:
From the mdn docs:
> [Notes](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault#notes)
Calling preventDefault() during any stage of event flow cancels the event, meaning that any default action normally taken by the implementation as a result of the event will not occur.

[...]

PR #1617, opened 2023-11-07
rename onclick_only to provided_onclick_only in dioxus_router's Link

valyagolev:
otherwise it doesn't work :)

this is needed to allow cmd-clicking link. working example:

```rust
#[component]
fn Link<'a>(cx: Scoped<'a>, class: Option<&'a str>, to: Route, children: Element<'a>) -> Element {
    let navigator = use_navigator(&cx);

    let eval_provider = use_eval(cx);

    let to_url = to.to_string();

    cx.render(rsx! {
        dioxus_router::prelude::Link {
            class: class.unwrap_or(""),
            to: to.clone(),
            onclick: move |e: Event<MouseData>| {
                let ms = e.data.modifiers();
                
                if ms.meta() || ms.ctrl() {
                    if let Err(e) = eval_provider(&format!("
                        window.open('{to_url}', '_blank');
                    ")) {
                        eprintln!("Error evaluating JS: {e:?}");
                    };
                } else {
                    navigator.push(to.clone());
                }
            },
            provided_onclick_only: true,
            children
        }
    })
}
```

~~btw, what would you say if I add this example code to Link as well? (in a separate PR). I think it's important to allow cmd-click for opening in new tabs. I only ever worked with Liveview though, so may be missing something~~

actually, `window.open` doesn't allow the focus to be retained in the original tab anymore. so without some `prevent_default` hackery, I'll have to resort to the normal `a` here. anyway, this particular PR might still be of value, if this attribute is to be around

### DIFF
--- a/packages/router/src/components/link.rs
+++ b/packages/router/src/components/link.rs
@@ -91,11 +91,12 @@ pub struct LinkProps<'a> {
     #[props(default)]
     /// Whether the default behavior should be executed if an `onclick` handler is provided.
     ///
-    /// 1. When `onclick` is [`None`] (default if not specified), `onclick_only` has no effect.
-    /// 2. If `onclick_only` is [`false`] (default if not specified), the provided `onclick` handler
-    ///    will be executed after the links regular functionality.
-    /// 3. If `onclick_only` is [`true`], only the provided `onclick` handler will be executed.
-    pub onclick_only: bool,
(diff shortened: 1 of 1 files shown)


ealmloff:
Once https://github.com/DioxusLabs/dioxus/issues/85 is resolved we can fix ctrl-clicking links. We need to conditionally enable prevent default if the control key was pressed which is currently not possible

Issue #1564, opened 2023-10-21
Integrate base_url with the router


ealmloff:
## Specific Demand

The router component currently ignores the base_url set in the `dioxus.toml`. It should read this value and use it to parse the route without needing to duplicate the base url in the router

aqrun:
Static files url also need support base_url

I want site deploy at subpath, e.g: abc.com/site-a  abc.com/site-b

PR #1622, opened 2023-11-08
Pull out CLI configs into a separate library

ealmloff:
This PR pulls out the CLI config into a separate library so that we can more easily integrate the config into internal libraries that need to read the config at runtime during dev mode. In the future this will let us read something like the assets path that is supplied in the config inside of the fullstack crate to serve the correct assets in both web and fullstack mode. Overall this will result in less duplication between configuration in code and configuration in the `Dioxus.toml` format

closes https://github.com/DioxusLabs/dioxus/issues/1564


ealmloff:
We need to expose const access to the cli config so that the parts of the config we need are baked into the binary at compile time. I am marking this as a draft again until that is added

ealmloff:
That makes the `base_url` from `Dioxus.toml` work in the router and the `app title` from `Dioxus.toml` work in desktop apps automatically

Issue #1623, opened 2023-11-08
bug: Props macro swallows any allowed rule from clippy


marc2332:
**Problem**

When a struct that derives the `Props` macro and is big enough for clippy to trigger the `type_complexity` warning will still throw the warning even if you bypass it with `#[allow(clippy::type_complexity)]`

**Steps To Reproduce**

- Create a large struct enough to trigger the warning
- Add `#[derive(Props)`
- Add `#[allow(clippy::type_complexity)]`

**Expected behavior**

Don't throw the warning.

**Screenshots**

![image](https://github.com/DioxusLabs/dioxus/assets/38158676/eefc433b-f805-48f8-8426-70421495d44d)

**Environment:**
 - Dioxus version: master
 - Rust version: 1.73
 - OS info: Windows 11
 - App platform: `freya`
 - 
**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later
- [x] I have no idea :sob:

tigerros:
Fixed in #1626

PR #1626, opened 2023-11-09
Fix 1623

tigerros:
See https://github.com/DioxusLabs/dioxus/issues/1623. This just allows the warning where it should be allowed.


ealmloff (approved):
Thank you!

Issue #1539, opened 2023-10-11
Configure automated release versioning and publishing


esimkowitz:
## Specific Demand

As Dioxus grows in size and scope, it's time for the project to evolve its release management process. Ideally, a GitHub workflow can be configured to remove the complex manual steps required to tag, version, and publish new releases. This can also help with security as fewer Crates.io tokens need to be floating around between maintainers. Instead, a single copy can be uploaded to the organization and used by the pipeline for automated publishing.

## Implement Suggestion

There are many tools available for automating versioning and publishing, the following seem the most promising. I'm still investigating this and will update this list with benefits and tradeoffs as I learn more.

### [release-plz](https://release-plz.ieni.dev/)

This is a rust-specific CI tool that automates the creation of release-PRs, which manage versioning and changelog generation for all changes between the last successful release and the current one. It is capable of automating versioning using either [conventional commits schema](conventionalcommits.org) or [cargo-semver-checks](https://docs.rs/crate/cargo-semver-checks/0.5.0), making it flexible to projects that don't have a well-established PR hygiene.

### [release-please](https://github.com/googleapis/release-please)

This is a Google-maintained project that was the original inspiration for release-plz, though it is compatible with many languages (Rust among them). This project enforces stronger PR hygiene by requiring all PRs to use the [conventional commits schema](conventionalcommits.org). This project also seems to support additional workspace-specific tooling, though I am still looking into this.

### [cargo-workspaces](https://github.com/pksunkara/cargo-workspaces)

This is a workspaces-optimized CLI for managing crate versioning and publishing. It doesn't appear to have the best support for CI automation, though I'm still looking into this.

### [cargo-smart-release](https://github.com/Byron/cargo-smart-release)

This is a very opinionated toolset for managing and publishing releases, though the opinions appear to be pretty well-aligned to the current Dioxus workflow. Still looking into this and how it can be CI-automated.

esimkowitz:
[...]

PR #1632, opened 2023-11-12
adds pipeline to bump semver and release to crates.io automatically

alexanderjophus:
I can't figure out how to mark this as a draft myself, but I want to check a few things before this is actually ready.

This change introduces a new pipeline for the maintainers to release dioxus in a simple manner.

Things to note:

- The following diff is necessary, this defaults to the code in the workspace enabling maintainers/contributors to easily and quickly write code across different crates. The version should be maintained by a bot and automatically bumped as/when a new version is released, it is required for `cargo workspaces` to work (as it depends on crates.io's requirements of there being a version number)
```diff
+ dioxus-rsx = { workspace = true, version = "0.4.2" }
- dioxus-rsx = { workspace = true }
```

- I'm still trying to figure out how to solve for unreleased packages such as signals and generational-box. I _believe_ `cargo workspaces` allows you to specify which packages to release but this is not implemented yet. As such, the PR as it stands would bump everything to `0.4.3` on the next patch regardless of if it's `0.4.2` now or `0.0.0`.

Fixes part of https://github.com/DioxusLabs/dioxus/issues/1539

Example commit in the much simpler [slides project](https://github.com/alexanderjophus/dioxus-slides/commit/3c52b02bc2e3fd0a6b0e732cfee80881565d8632) I'm working on. Note how it maintains the version in all the packages.


alexanderjophus:
After a discussion with Marc on discord, we're going to try pinning to `0.4` to see what happens, otherwise it may publish unnecessary versions a lot.

alexanderjophus:
Looks like pinning to 0.4 will also work and is possibly preferable.

Proof: https://github.com/alexanderjophus/dioxus-slides/actions/runs/6840221582/job/18599245617

alexanderjophus:
Noticed there's a possible hiccup with this when running some examples.

My terminal gets littered with these - exploring options now:
```
warning: /Users/ajt/projects/tmp/dioxus/packages/dioxus-tui/Cargo.toml: unused manifest key: dependencies.dioxus-core.version
```

ealmloff:
> Noticed there's a possible hiccup with this when running some examples.
> 
> My terminal gets littered with these - exploring options now:
> 
> ```
> warning: /Users/ajt/projects/tmp/dioxus/packages/dioxus-tui/Cargo.toml: unused manifest key: dependencies.dioxus-core.version
> ```

dioxus-tui might not use dioxus-core anymore (instead it just uses dioxus). You could try entirely removing the dependency

alexanderjophus:
The only fix I've found is the following in the package cargo.toml files. 
```diff
+ dioxus = { path = "../dioxus", version = "0.4" }
- dioxus = { workspace = true, version = "0.4" }
```
I tried various permutations in the root Cargo.toml file, most googling suggests .cargo/config.toml, which showed no luck either.

ealmloff:
> The only fix I've found is the following in the package cargo.toml files.
> 
> ```diff
> + dioxus = { path = "../dioxus", version = "0.4" }
> - dioxus = { workspace = true, version = "0.4" }
> ```
> 
> I tried various permutations in the root Cargo.toml file, most googling suggests .cargo/config.toml, which showed no luck either.
> 
> Question for yourself is; is that ok?

That wouldn't be great. using workspace dependancies makes it much easier to move around, rename and add new crates. What exactly caused the error? Did adding the versions cause the error? Is it an issue in cargo workspace? It seems like this might be related to https://github.com/pksunkara/cargo-workspaces/pull/56. In the discussion, there is a fork linked that fixed this issue. Maybe we could try that fork

alexanderjophus:
Not sure I have the bandwidth to finish this. I do agree that we should stick with the `workspace = true` setting.

As far as I know that ticket you linked is not related (could be wrong though), but I do think the issue is in cargo workspace, I've just not dug through to find out more.

jkelleyrtp:
Hey! Thanks for getting this off the ground. This is exactly what we need. I'll work on bringing it up to speed.

Issue #1547, opened 2023-10-16
Allow to use `dioxus-cli` within a cargo workspace


ryo33:
I'd like to work on this on my own if this feature is welcomed.

## Specific Demand

I'm using dioxus in a cargo workspace and want to run the `dx` command within the workspace root instead of `cd` to the dioxus project before run `dx`. Also, although I did not do it yet, a workspace could include multiple dioxus projects in the monorepo context.

## Implement Suggestion

Add `-p, --package [<SPEC>]        Package with the target to run` option to several `dx`'s commands.

ealmloff:
This would be a great addition to the CLI! This would be useful for the main Dioxus repo which uses workspace.

jkelleyrtp:
Cargo build, serve, bundle, and fmt now all apply based on the package arg.

Only `check` is not yet supported on a per-package basis.

PR #1642, opened 2023-11-15
Related to #1547: use `dioxus-cli` within a workspace (wildcard-members, real package names)

RGafiyatullin:
(Related to #1547)

Although the `dx` CLI allows to specify a package name to chose from workspace members, it does not support workspace members specified as glob-wildcards. Neither it respects the effective package name, specified in the crate's `Cargo.toml`.

This PR addresses that issue:
- upon `dx build ...`, if the `--bin` CLI-argument is provided, treat the current dir as a workspace;
- search through the workspace's `members`: resolve each of them with `glob`;
- assume that any workspace member has a `Cargo.toml` in it (cargo does it, so it's okay);
- read said manifest, and check the package name in it;
- if found — there we have our sought package.

### DIFF
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -11,7 +11,6 @@ keywords = ["react", "gui", "cli", "dioxus", "wasm"]
 [dependencies]
 # cli core
 clap = { version = "4.2", features = ["derive"] }
-glob = "^0.3"
 thiserror = { workspace = true }
 wasm-bindgen-cli-support = "0.2"
 colored = "2.0.0"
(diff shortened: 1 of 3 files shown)


ealmloff:
What is your use case for glob patterns for the bin field?

I realize this PR doesn't introduce this error, but as a note, we don't actually need to assume that the current directory is the root of the package. We can use `cargo-metadata` to get the members of the workspace associated with the current path (even if we are in a different workspace member) https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#method.workspace_packages

ealmloff:
Interesting, I didn't know that the workspace field had glob support!

RGafiyatullin:
I had a look at how the rest of the CLI works: a lot of it could use moving to cargo-metadata...

ealmloff (approved):
This looks much better, thank you!

PR #1648, opened 2023-11-19
Fix outdated lazynode documentation

### LINKED ISSUE #1647: `Lazynode` documentation not up to date
**Problem**

In the example of the docs, there is a `let f = LazyNodes::new(move |f| f.element("div", [], [], [] None));`

But `f.element` is no longer a valid method.

Here is the link:
https://docs.rs/dioxus/latest/dioxus/prelude/struct.LazyNodes.html#implementations

That would be nice to fix it.

**Questionnaire**
- [ x] I don't have time to fix this right now, but maybe later

Exotik850:
Fix error in Lazynode documentation that uses `ScopeState::element` which doesn't exist. Change to `ScopeState::render`

Fixes #1647

### DIFF
--- a/packages/core/src/lazynodes.rs
+++ b/packages/core/src/lazynodes.rs
@@ -23,8 +23,37 @@ use crate::{innerlude::VNode, ScopeState};
 ///
 ///
 /// ```rust, ignore
-/// LazyNodes::new(|f| f.render(rsx!(div {})).unwrap())
+/// LazyNodes::new(|f| {
+///        static TEMPLATE: dioxus::core::Template = dioxus::core::Template {
+///         name: "main.rs:5:5:20", // Source location of the template for hot reloading
+///         roots: &[
+///             dioxus::core::TemplateNode::Element {
+///                 tag: dioxus_elements::div::TAG_NAME,
+///                 namespace: dioxus_elements::div::NAME_SPACE,
+///                 attrs: &[],
+///                 children: &[],
+///             },
+///         ],
+///         node_paths: &[],
+///         attr_paths: &[],
+///     };
+///     dioxus::core::VNode {
+///         parent: None,
+///         key: None,
+///         template: std::cell::Cell::new(TEMPLATE),
+///         root_ids: dioxus::core::exports::bumpalo::collections::Vec::with_capacity_in(
+///                 1usize,
+///                 f.bump(),
+///             )
+///             .into(),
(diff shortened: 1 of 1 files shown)


packages/core/src/lazynodes.rs
```diff
@@ -23,7 +23,7 @@ use crate::{innerlude::VNode, ScopeState};
 ///
 ///
 /// ```rust, ignore
-/// LazyNodes::new(|f| f.element("div", [], [], [] None))
+/// LazyNodes::new(|f| f.render(rsx!(div {})).unwrap())
```
ealmloff, line 26: If you want to use rsx, you shouldn't wrap it in LazyNodes::new(). rsx!(div{}) expands to something like LazyNodes::new(|f| ...create div):

```rust
::dioxus::core::LazyNodes::new(move |
        __cx: &::dioxus::core::ScopeState,
    | -> ::dioxus::core::VNode {
        static TEMPLATE: ::dioxus::core::Template = ::dioxus::core::Template {
            name: "src/main.rs:5:5:83",
            roots: &[
                ::dioxus::core::TemplateNode::Element {
                    tag: dioxus_elements::div::TAG_NAME,
                    namespace: dioxus_elements::div::NAME_SPACE,
                    attrs: &[],
                    children: &[],
                },
            ],
            node_paths: &[],
            attr_paths: &[],
        };
        ::dioxus::core::VNode {
            parent: None,
            key: None,
            template: std::cell::Cell::new(TEMPLATE),
            root_ids: dioxus::core::exports::bumpalo::collections::Vec::with_capacity_in(
                    1usize,
                    __cx.bump(),
                )
                .into(),
            dynamic_nodes: __cx.bump().alloc([]),
            dynamic_attrs: __cx.bump().alloc([]),
        }
    })
```

That makes the expanded version of this code something like:

```rust
LazyNodes::new(|f| **resolve lazy nodes to the elements they create** LazyNodes::new(|f| **actually create the elements**))
```

We could show the expanded version of `rsx!{div{}}` instead and link to https://dioxuslabs.com/learn/0.4/contributing/walkthrough_readme#the-rsx-macro.

ealmloff (changes requested):

packages/core/src/lazynodes.rs
Exotik850: Alright, changed it to a closure that makes an empty div element

ealmloff (approved):
Thank you!

PR #1649, opened 2023-11-19
Fix tailwind.css 404 in Cli web platform build

hjin-me:
```
[application]
default_platform = "web"
[application.tools]
tailwindcss = { input = "input.css", config = "tailwind.config.js" }
```

when enable tailwindcss, CLI will auto insert `<link rel="stylesheet" href="tailwind.css">` into `index.html`.

this will cause 404 when route enabled.

this PR change the path to absolute path to fix this.


ealmloff (approved):
Thank you!

PR #1650, opened 2023-11-19
Add an optional cleanup closure to the use effect hook

ealmloff:
Allows users to return an optional cleanup closure from the use effect hook like react allows. This makes it easier to clean up any effects that effect dom elements.

### DIFF
--- a/packages/hooks/src/use_effect.rs
+++ b/packages/hooks/src/use_effect.rs
@@ -1,5 +1,10 @@
 use dioxus_core::{ScopeState, TaskId};
-use std::{any::Any, cell::Cell, future::Future};
+use std::{
+    any::Any,
+    cell::{Cell, RefCell},
+    future::Future,
+    rc::Rc,
+};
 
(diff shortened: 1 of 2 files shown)


packages/hooks/src/use_effect.rs
```diff
@@ -33,6 +38,16 @@ use crate::UseFutureDep;
 ///         }
 ///     });
 ///
+///     // Only fetch the user data when the id changes.
+///     use_effect(cx, (id,), |(id,)| {
+///         to_owned![name];
+///         async move {
+///             let user = fetch_user(id).await;
+///             name.set(user.name);
+///             move || println!("Cleaning up from {}", id);
```
marc2332, line 47:
```suggestion
///             move || println!("Cleaning up from {}", id)
```

marc2332 (changes requested):

jkelleyrtp (approved):

PR #1651, opened 2023-11-19
Add into attribute value impl for String

ealmloff:
This makes it possible to pass string directly into an attribute. String is a very common type and this is something I have seen people hit several times when they first use dioxus.


rambip:
I do agree, I had to find very ellaborate workarounds to solve this issue.

And tha fact that `rsx!{div {style: "{1}"}}` does work but `rsx! {div {style: format!("{1}")}}` doesn't is not user-friendly at all

jkelleyrtp (approved):

Issue #1661, opened 2023-11-23
Documentation of `EventHandler` not exposed


rambip:
**Problem**

https://docs.rs/dioxus/latest/dioxus/prelude/struct.ScopeState.html#method.event_handler

If you look at the documentation of `event_handler`, you can see a function that returns a `EventHandler`.
With an lsp, I can click on it to see its documentation, but on the docs.rs, I cannot.

I  think it is just a broken link issue, or something similar

**Questionnaire**
- [ ] I may contribute to the project later, but for now I don't know the framework well enough

PR #1665, opened 2023-11-27
remove #[doc(hidden)] from EventHandler

Exotik850:
fixes [1661](https://github.com/DioxusLabs/dioxus/issues/1661)


ealmloff (approved):
Oops, it looks like that attribute was originally applied to another bit of code that was moved. Thank you!

Issue #1562, opened 2023-10-21
use_future is not clone


bbigras:
**Problem**

use_future is not clone.

**Steps To Reproduce**

```rust
#![allow(non_snake_case)]
use dioxus::prelude::*;

async fn fetch_data() -> String {
    "ok".to_string()
}

fn main() {
    dioxus_desktop::launch(App);
}

fn App(cx: Scope) -> Element {
    let data = use_future(cx, (), |_| fetch_data());

    let log_in = move |_| {
        let data = data.clone();
        cx.spawn({
            async move {
                data.restart();
            }
        });
    };

    let data2 = format!("{:?}", data.value());

    render! {
        div {
            button {
                onclick: log_in,
                "test"
            },
            "data: {data2}"
        }
    }
}
```

```
error[E0521]: borrowed data escapes outside of function
  --> src/main.rs:13:16
   |
12 | fn App(cx: Scope) -> Element {
   |        --
   |        |
   |        `cx` is a reference that is only valid in the function body
   |        has type `&'1 Scoped<'1>`
13 |     let data = use_future(cx, (), |_| fetch_data());
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                |
   |                `cx` escapes the function body here
   |                argument requires that `'1` must outlive `'static`

For more information about this error, try `rustc --explain E0521`.
error: could not compile `test-dioxus` (bin "test-dioxus") due to previous error
```

**Expected behavior**

**Screenshots**

**Environment:**
 - Dioxus version: v0.4
 - Rust version: [e.g. 1.43.0, `nightly`]
 - OS info: nixos
 - App platform: `desktop`

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

ealmloff:
If you are interested in fixing this, we need to change [this code](https://github.com/DioxusLabs/dioxus/blob/master/packages/hooks/src/use_future.rs#L72) to derive clone. There are a few fields we need to be careful with:
```rust
needs_regen: Cell<bool>,
task: Rc<Cell<Option<TaskId>>>,
```
needs to change to this to make it `Clone`:
```rust
needs_regen: Rc<Cell<bool>>,
task: Rc<Cell<Option<TaskId>>>,
```

[...]

PR #1666, opened 2023-11-27
Make UseFuture Clone, factor out dependencies field

Exotik850:
Fixes [1562](https://github.com/DioxusLabs/dioxus/issues/1562)


ealmloff (approved):
Thank you!

Issue #571, opened 2022-09-30
`use_const` hooks/init only once hooks


DrSloth:
## Specific Demand
Its sometimes necessary to share state between components, which is only initialised once and later only read. One case where this arised is after fetching something and displaying this content in multiple components. The data is shown on the page but not really changed.

This can be done relatively simply with borrowing, but these can not be memoized. One way to do this easily would be using something like an Rc.

I don't exactly know wether this is would be considered unidiomatic and should be solved using borrowing, utilizing the parent components as a "barrier". But that seems definitely harder...

## Implement Suggestion
An easy way would be providing a `use_const` hook which gives some `UseConst` value, that just contains an `Rc` which implements `PartialEq` using [`Rc::ptr_eq`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.ptr_eq)

Thanks for your time :)

jkelleyrtp:
We basically have this now with use_signal() which can decompose into ReadOnlySIgnal via an Into conversion.

Closed by #1791

PR #1667, opened 2023-11-27
Add use_const hook

Exotik850:
Adds `use_const` hook for [571](https://github.com/DioxusLabs/dioxus/issues/571) so data that doesn't need to be mutated can be easily memoized and used across components

### DIFF
--- a/packages/hooks/src/use_const.rs
+++ b/packages/hooks/src/use_const.rs
@@ -47,8 +47,7 @@ impl<T: core::fmt::Display> core::fmt::Display for UseConst<T> {
 }
 
 impl<T> UseConst<T> {
-    #[allow(dead_code)]
-    fn get_rc(&self) -> &Rc<T> {
+    pub fn get_rc(&self) -> &Rc<T> {
         &self.value
     }
 }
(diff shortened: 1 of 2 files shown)


packages/hooks/src/use_const.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl<T: core::fmt::Display> core::fmt::Display for UseConst<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.value.fmt(f)
+    }
+}
+
+impl<T> UseConst<T> {
+    #[allow(dead_code)]
+    fn get_rc(&self) -> &Rc<T> {
```
ealmloff, line 51: If you make this public it will get rid of the dead code warning

Exotik850: ah didn't even cross my mind

ealmloff (approved):
Thank you! I think this a good addition to the built in hooks. When hooks are switched to signals, this could be changed to a read only signal that is copy

Issue #1630, opened 2023-11-11
Spawning a task with cx.spawn() should poll the task at least once


jkelleyrtp:
## Specific Demand

With some coroutine code, I tend to do some setup work and *then* wait for some async work to happen. Sometimes the async code can exit immediately without waiting for the future to actually progress. This would be the type of futures that can resolve instantly.

In these cases, I would expect `cx.spawn` to poll the future at least once. This makes sense for suspended components where the future resolves immediately during hydration - why come back to a component to render it if the future was already ready immediately?

## Implement Suggestion

During insertion of the task during `push_future`, we have the waker ready. We should directly poll the task right away before committing it to the task queue. This is a really easy change which is technically breaking.

PR #1669, opened 2023-11-28
Make cx.spawn() return an Option<TaskId>, returns None if finished immediately

Exotik850:
This should close [1630](https://github.com/DioxusLabs/dioxus/issues/1630)

Code passes all the tests besides the last Miri stress test, but it doesn't look like it's ~~coming from my changes~~ I don't know how my changes would be causing them

Surprisingly few changes needed to be made

### DIFF
--- a/packages/core/src/scheduler/task.rs
+++ b/packages/core/src/scheduler/task.rs
@@ -19,7 +19,7 @@ pub struct TaskId(pub usize);
 /// the task itself is the waker
 pub(crate) struct LocalTask {
     pub scope: ScopeId,
-    pub(super) task: RefCell<Pin<Box<dyn Future<Output = ()> + 'static>>>,
+    pub task: RefCell<Pin<Box<dyn Future<Output = ()> + 'static>>>,
     pub waker: Waker,
 }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/scheduler/task.rs
```diff
@@ -48,20 +52,25 @@ impl Scheduler {
...
+        Some(task_id)
     }
 
     /// Drop the future with the given TaskId
     ///
     /// This does not abort the task, so you'll want to wrap it in an aborthandle if that's important to you
-    pub fn remove(&self, id: TaskId) {
-        self.tasks.borrow_mut().try_remove(id.0);
+    pub fn remove(&self, id: TaskId) -> Option<LocalTask> {
+        self.tasks.try_borrow_mut().ok()?.try_remove(id.0)
```
Exotik850, line 73: Forgot to change this back when getting tests to pass, but doesn't effect stress tests

packages/fullstack/src/hooks/server_future.rs
```diff
@@ -60,7 +58,7 @@ where
         }
     }
 
-    if dependencies.clone().apply(&mut state.dependencies) || state.needs_regen.get() {
+    if dependencies.clone().apply(&mut Vec::new()) || state.needs_regen.get() {
```
ealmloff, line 61: This will break memorization because the vec will always be empty which will cause apply to return true and always rerun. You could pull it out into a hook like #1666

ealmloff (changes requested):

packages/fullstack/src/hooks/server_future.rs
Exotik850: trying to factor it out but didn't remember exactly what the steps were, thank you, will fix now

ealmloff:
@Exotik850 Was this closed intentionally? I still think this PR would be a useful addition to Dioxus, we are just waiting to merge any breaking PRs for 0.5 until 0.4.3 is published.

Exotik850:
I didn't even realize it was closed, I'll try to reopen it

Exotik850:
Ah, I was cleaning out the repo because I messed up a branch in my fork, I'll fix it in a bit

Exotik850:
I didn't know it would close the PR if I deleted and forked the main repo again, I'm getting used to git and reverting changes is a hassle 😅

Issue #1472, opened 2023-09-18
Create a way to run all checks on local machine before contributing


tigerros:
## Specific Demand

Right now, contributors have to run every CI check separately, so it would good to streamline the process and combine all checks into one command that would execute all of them.

## Implement Suggestion

I'm pretty sure cargo-make can do this.

PR #1670, opened 2023-11-28
Add cargo make tidy/miri

Exotik850:
Adds `cargo make tidy` and `cargo +nightly make miri` and changes `cargo make tests` to include `tidy`

Closes [1472](https://github.com/DioxusLabs/dioxus/issues/1472)

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -24,10 +24,52 @@ script = [
 ]
 script_runner = "@duckscript"
 
+[tasks.format]
+command = "cargo"
+args = ["fmt", "--all"]
+
+[tasks.check] 
+command = "cargo"
+args = ["check", "--workspace", "--examples", "--tests"]
+
+[tasks.clippy]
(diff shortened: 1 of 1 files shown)


ealmloff (changes requested):
Cargo check, fmt, and clippy already exist in the github actions workflow here: https://github.com/DioxusLabs/dioxus/blob/master/.github/workflows/main.yml#L62-L84. Because one of the actions runs `cargo make test`, this will cause those parts of the CI to always run twice. Can you change the CI to only run the tests part of the tests. I think we should keep the different steps separate in CI because it is useful to see what part is failing in a PR at a glance. We could either duplicate the commands or run different `cargo make *` in different steps for the CI

Exotik850:
Didn't notice the github actions, will fix now!

ealmloff (approved):
Looks good. It is nice do have an option with everything centralized. Thank you

Issue #1311, opened 2023-08-06
Rink: Use ratatui instead of tui-rs


undersquire:
**Problem**

`rink` currently depends on [tui-rs](https://github.com/fdehau/tui-rs), which is [no longer maintained](https://github.com/fdehau/tui-rs/issues/654).

**Solution**

Switch to [ratatui](https://github.com/ratatui-org/ratatui), which has been named the [official successor to tui-rs](https://github.com/fdehau/tui-rs/commit/335f5a4563342f9a4ee19e2462059e1159dcbf25).

It should just be a drop-in replacement, simply changing the Cargo.toml entry to:
```toml
tui = { package = "ratatui", version = "0.22", features = ["all-widgets"]}
```

Esales301080:
Nice

Demonthos:
We don't really use the TUI crate for anything except rendering. Wrap their widget trait and events. We could switch to a more raw terminal rendering crate like [termwiz](https://docs.rs/termwiz/latest/termwiz/index.html). Termwiz supports images, and raw pixel graphics on some terminals unlike TUI or ratatui

undersquire:
> We could switch to a more raw terminal rendering crate like [termwiz](https://docs.rs/termwiz/latest/termwiz/index.html). Termwiz supports images, and raw pixel graphics on some terminals unlike TUI or ratatui

That does sound like a good idea. I have looked into termwiz and it seems like a better option if Rink isn't using any widgets from tui-rs. (also ratatui does have a termwiz backend if dioxus-tui ever does find use in the widgets).

PR #1671, opened 2023-11-28
Switch outdated tui crate for ratatui

Exotik850:
It would probably be pretty simple to take it one step lower and use [`termwiz`](https://docs.rs/termwiz/latest/termwiz/index.html) directly, but for now `ratatui` is a drop in replacement for an outdated crate.
Fixes [1311](https://github.com/DioxusLabs/dioxus/issues/1311)


ealmloff (approved):
Thank you!

PR #1672, opened 2023-11-29
Merge Old https://github.com/DioxusLabs/cli/pull/124 PR to main repo

Exotik850:
Basically just a straight copy, besides a few spots where there were some gaps in the old and new code.
Hastily put together so please review, any suggestions are welcome, all credit to [mrxiaozhuox](https://github.com/mrxiaozhuox)

Edit: Link [124](https://github.com/DioxusLabs/cli/pull/124)

Issue #1192, opened 2023-07-14
Tailwind Support for CLI


itsezc:
## Specific Demand

For DX purposes, it'd be great if we can have Tailwind to auto-generate CSS files through the CLI similar to how trunk approaches this: 

![image](https://github.com/DioxusLabs/dioxus/assets/33750251/89e3a95b-7e5f-4940-bb95-8fec92ce4bd4)

## Implement Suggestion

N/A

Exotik850:
I think it might be better to have commands that are run whenever `dx serve` is run to be set in `Dioxus.toml` similar to how tauri uses `npm run dev` which launches both the backend and the frontend with separate commands, just to make the config more flexible to on-file-change systems like `tailwindcss --watch`

Could also be a good solution to [1458](https://github.com/DioxusLabs/dioxus/issues/1458)

ealmloff:
> I think it might be better to have commands that are run whenever `dx serve` is run to be set in `Dioxus.toml` similar to how tauri uses `npm run dev` which launches both the backend and the frontend with separate commands, just to make the config more flexible to on-file-change systems like `tailwindcss --watch`

+1 I don't think the configuration should live in the html file. A [plugin system](https://github.com/DioxusLabs/cli/pull/124) would allow us to build different composable pieces that hook into different parts of the build process. Tailwind integration could be a plugin

itsezc:
Some additional context to keep in mind is Tailwind CLI will be moving to Oxide (https://github.com/tailwindlabs/tailwindcss/tree/master/oxide) in a few minor versions (3.5 - 3.7) which isn't far away. I would also love it if we didn't have a "tailwind.css" in our public folder that we had to commit, I'm aware we can just ignore it with `.gitignore` but while running with `dx serve` I'd like to not have generated there if possible (nice to have).

ealmloff:
> I would also love it if we didn't have a "tailwind.css" in our public folder that we had to commit, I'm aware we can just ignore it with .gitignore but while running with dx serve I'd like to not have generated there if possible (nice to have).

You don't need to have it in your public folder, but the dist folder should have everything required for the application to run, including the CSS files

peanutbother:
For my current project I am using a build.rs script where I run `tailwindcss` if a source file changes:

```rust
use std::process::Command;

fn main() {
    Command::new("tailwindcss")
        .args(&["-i", "./src/tailwind.css", "-o", "./assets/tailwind.css"])
        .status()
        .unwrap();
[...]

PR #1673, opened 2023-11-29
Create CLI plugin system

ealmloff:
Adds a WASI plugin to the CLI that will allow users to write plugins in Rust (instead of the Lua plugin system added in https://github.com/DioxusLabs/dioxus/pull/1672)


packages/cli-plugin/src/export_plugin.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+",
+            world: "plugin-world",
+            exports: {
+                world: $name,
+                "plugins:main/definitions": $name
+            },
+        });
+    };
+}
```
Andrew15-5, line 185: Missing last LF.

packages/cli/src/assets/dioxus.toml
```diff
@@ -14,16 +14,17 @@ out_dir = "dist"
 # resource (static) file folder
 asset_dir = "public"
 
+[watcher]
+
+index_on_404 = true
```
ealmloff, line 19: Renaming web.watcher to watcher probably makes sense, but it could be confusing that index_on_404 doesn't effect liveview or fullstack if it isn't marked as a web property

packages/cli/src/cli/build.rs
```diff
@@ -55,6 +57,8 @@ impl Build {
             }
         }
 
+        plugins_after_command(BuildEvent).await;
```
ealmloff, line 60: It might make sense to move this after the `index.html` file is generated so that plugins can modify that file. I'm thinking of asset plugins like the tailwind plugin automatically reaching in and checking that you have a tailwind file imported somewhere in your `<head>`

packages/cli/src/lock.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    pub fn load() -> crate::error::Result<Self> {
+        let crate_dir = crate::cargo::crate_root()?;
+
+        let crate_dir = crate_dir.as_path();
+
+        let Some(dioxus_conf_file) = acquire_dioxus_lock(crate_dir) else {
+            return Ok(Self::new(crate_dir.join("Dioxus.lock")));
```
ealmloff, line 32: It might be better to output Dioxus.lock inside the target directory or the new `.dioxus` directory so that people don't have to add it to their `.gitignore`. It shouldn't be something that users need to know about

packages/cli/src/server/mod.rs
```diff
@@ -37,95 +50,108 @@ async fn setup_file_watcher<F: Fn() -> Result<BuildResult> + Send + 'static>(
...
+            return;
+        };
+
+        if chrono::Local::now().timestamp() <= last_update_time {
+            return;
+        }
+
+        let mut needs_full_rebuild = false;
+        let change =
+            futures::executor::block_on(plugins_watched_paths_changed(&e.paths, &config.crate_dir));
```
ealmloff, line 64: We can use a channel and a seperate task to make this block on shorter: https://github.com/notify-rs/notify/blob/c3929ed114fbb0bc7457a9a498260461596b00ca/examples/async_monitor.rs#L39

packages/cli/src/builder.rs
```diff
@@ -582,135 +583,128 @@ fn build_assets(config: &CrateConfig) -> Result<Vec<PathBuf>> {
...
-        let sass_conf = dioxus_tools.get("sass").unwrap();
-        if let Some(tab) = sass_conf.as_table() {
-            let source_map = tab.contains_key("source_map");
-            let source_map = if source_map && tab.get("source_map").unwrap().is_bool() {
-                if tab.get("source_map").unwrap().as_bool().unwrap_or_default() {
-                    "--source-map"
-                } else {
-                    "--no-source-map"
+
+    if !sass.is_installed() || !dioxus_tools.contains_key("sass") {
```
ealmloff, line 587: We should remove all of the remnents of dioxus_tools now that we have plugins

packages/cli-plugin/examples/dx-parcel-minifier/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+use plugins::main::{
+    imports::{get_project_info, log},
+    toml::{Toml, TomlValue},
+    types::{CommandEvent, PluginInfo, ResponseEvent, RuntimeEvent},
+};
+
+struct CSSMinifer;
+
+fn minify_css() -> Result<ResponseEvent, ()> {
+    for entry in fs::read_dir("/dist").unwrap() {
```
ealmloff, line 14: This could be different if the user changes the output directory

Exotik850: That part isn't completely finished, but the "/dist" directory would be sandboxed for all the plugins and would be just set to whatever the host project has in the asset directory and output directory configs, so any plugin would just read and write from "/dist" and "/assets" and it would pull from where the configs are set to. At the moment there is a `has_output_dir` and `has_asset_dir` variables in the `get_project_info` function that tells the plugin if the host project has these defined, which would guarantee the existence of the "asset" and "dist" directories to be available to plugins since the plugin sandboxes these when they are loaded [here](https://github.com/ealmloff/dioxus/blob/b6e83d03dc49ad75f58033542b5a031dd23b0074/packages/cli/src/plugin/mod.rs#L291), but maybe a `get_asset_dir -> option<path>` function would be better?

Exotik850: We also just create the dist directory [here](https://github.com/ealmloff/dioxus/blob/b6e83d03dc49ad75f58033542b5a031dd23b0074/packages/cli/src/cli/build.rs#L73C24-L73C24) anyway, so just defaulting to "/dist" if it's not available in the config shouldn't be too much of an issue

ealmloff: Oh, interesting! I think `get_asset_dir` would be a bit more clear, but mapping the /dist and /assets directory also works

packages/cli/src/server/mod.rs
Exotik850: just importing `block_on` from futures separately cleans it up quite a bit, and it's not as trivial as it should be to just throw the function into a task because lifetimes and futures mix like rocks and glass in a blender

ealmloff:
I had another thought on the double drop issue. Here we do own the resource returned from call_get_default_config:
```rust
pub async fn get_default_config(&mut self) -> wasmtime::Result<toml::Value> {
        let default_config = self
            .bindings
            .plugins_main_definitions()
            .call_get_default_config(&mut self.store)
            .await?;
        let t = self
            .store
            .data_mut()
            .get_toml(default_config)
            .convert_with_state(self.store.data_mut())
            .await;
        Ok(t)
}
```

Can we do something like take_toml that removes the toml from the store and returns it (maybe with mem::forget to prevent the default_config we pass into take_config so it doesn't get dropped after it is removed :grimacing:).

FragrantArmpit:
All CSS utility feature requests are blocked by this:

1. #1192 
2. #2813 
3. #2425 

API-wise we probably want a way for plugins to add assets to the manganis manifest, i.e. run tailwind/whatever in `on_watched_paths_change` and add the tailwind/whatever output a CSS asset.

PR #1676, opened 2023-11-29
add signals for existing hooks

bunnyBites:


bunnyBites:
> We can keep the use_signal hook in the dioxus-signals crate. In discord, I was talking about changing the existing hooks in `dioxus-hooks` to use signals under the hood so that they can be `Copy`. For example, we could change [use_shared_state](https://github.com/DioxusLabs/dioxus/blob/master/packages/hooks/src/use_shared_state.rs) to return a `Option<Signal<T>>` instead of `Option<&UseSharedState<T>>` and change [use_future](https://github.com/DioxusLabs/dioxus/blob/master/packages/hooks/src/use_future.rs#L78) to use `Signal<Option<T>>` instead of `UseState<Option<T>>` for the current future state

bunnyBites:
@ealmloff I should have raised the PR a little sooner. I totally misunderstood the requirement. Now, I think I got it.
Thanks

jkelleyrtp:
Closing since it seems like the PR is a lot thinner now than it was when it was opened.

PR #1678, opened 2023-12-01
docs(desktop): fix wrong example code

zhangzhonglai:
The docs for `dioxus_desktop::launch_cfg` uses code from a long time ago, and that is now unavailable, which can cause problems for developer.


ealmloff (approved):
Thank you! I removed the ignore tag from the doc tests so these examples will be checked by CI in the future

PR #1681, opened 2023-12-01
Add a warning about wry gnu support to Dioxus desktop

ealmloff:
Dioxus desktop uses wry internally. There are some [limitations](https://github.com/wravery/webview2-rs#cross-compilation) that make compiling to GNU more difficult. This PR adds a warning about the limitations and suggests that most users switch to the MSVC, but it also provides information for the users that choose not to switch.

(This is an issue that has come up several times in the discord especially when cross compiling to windows)

### DIFF
--- a/packages/desktop/src/lib.rs
+++ b/packages/desktop/src/lib.rs
@@ -19,17 +19,6 @@ mod webview;
 #[cfg(any(target_os = "ios", target_os = "android"))]
 mod mobile_shortcut;
 
-// WARN about wry support on windows gnu targets. GNU windows targets don't work well in wry currently
-#[cfg(all(windows, target_env = "gnu", not(feature = "gnu")))]
-mod wry_gnu_warning {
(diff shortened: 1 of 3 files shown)


packages/desktop/src/lib.rs
```diff
@@ -19,6 +19,17 @@ mod webview;
 #[cfg(any(target_os = "ios", target_os = "android"))]
 mod mobile_shortcut;
 
+// WARN about wry support on windows gnu targets. GNU windows targets don't work well in wry currently
```
jkelleyrtp, line 22: Don't people usually stick this check in build.rs?

jkelleyrtp (approved):

packages/desktop/src/lib.rs
ealmloff: That is true, build scripts let us create a much nicer looking warning message. I have been writing too many macros 😆

PR #1686, opened 2023-12-04
Fixed conflict of  "--verbose" and "--quiet" flags

raulcesar:
When running `dx serve --verbose` an error appears on console stating: "cannot set both --verbose and --quiet"
I noted that in the file "packages/cli/src/builder.rs", the "quiet" flag appeared to be set in default command, and the "verbose" flag was being appended, which caused both flags to be put into the command, so I made them mutually exclusive instead.


ealmloff (approved):
Thank you!

PR #1687, opened 2023-12-04
feat(fullstack): add `render_handler_with_state`

sd2k:
When using server functions, the current pattern to access state such as database connections is to use `register_server_fns_with_handler` on an Axum router and 'inject' the state into the context provided to the server function. However, this only affects function calls which go via the Axum router; SSR renders bypass this, and therefore don't have access to any state.

This commit adds an alternative `render_handler` which accepts some additional state. That state is injected into the context in a similar manner to `register_server_fns_with_handler`. SSR renders can then proceed to run in the same way as HTTP calls.

### DIFF
--- a/packages/fullstack/src/adapters/axum_adapter.rs
+++ b/packages/fullstack/src/adapters/axum_adapter.rs
@@ -419,7 +419,7 @@ fn apply_request_parts_to_response<B>(
 /// ```
 pub async fn render_handler_with_context<
     P: Clone + serde::Serialize + Send + Sync + 'static,
-    F: for<'a> FnMut(&'a mut DioxusServerContext),
+    F: FnMut(&mut DioxusServerContext),
 >(
     State((mut inject_context, cfg, ssr_state)): State<(F, ServeConfig<P>, SSRState)>,
     request: Request<Body>,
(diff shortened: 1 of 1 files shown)


packages/fullstack/src/adapters/axum_adapter.rs
```diff
@@ -369,15 +369,19 @@ fn apply_request_parts_to_response<B>(
     }
 }
 
-/// SSR renderer handler for Axum
-pub async fn render_handler<P: Clone + serde::Serialize + Send + Sync + 'static>(
-    State((cfg, ssr_state)): State<(ServeConfig<P>, SSRState)>,
+/// SSR renderer handler for Axum with added state.
+pub async fn render_handler_with_state<
+    P: Clone + serde::Serialize + Send + Sync + 'static,
+    S: Send + Sync + 'static,
```
ealmloff, line 375: This limits the context that is injected to one state. Can you change this to a callback that modifies the `DioxusServerContext` to make it possible to insert as many states as you want?

sd2k: Sure, done. I think the APIs here might be nicer if we used a similar pattern to axum where there's only one state allowed, which can be composed of substates. In particular getting `register_server_fns_with_handler` to work is pretty hairy right now!

That can always happen in a future refactor though.

ealmloff (approved):
Much better, thank you!

PR #1696, opened 2023-12-06
add default menu bar in dioxus-desktop to resolve #1691

### LINKED ISSUE #1691: Some macOS keybindings do not work (Copy, Paste, Select all)
**Problem**
Some keybindings like Cmd+A for select all, Cmd+c for copy or Cmd+v for paste, Cmd+x, Cmd+Z (and potentially others? im no macOS expert) do not work/do nothing. Other macOS specific keybindings like Cmd+backspace and Cmd+shift+left do work

**Steps To Reproduce**

Try to use the following code: 
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app);
}

pub fn app(cx: Scope) -> Element {
    render! {
      input {
        style: "width:100%;",
        value: "A lot of text to remove. Maybe you wish to select all to remove it?"
      }
    }
}
```

**Expected behavior**

Standard keybindings work the same as in Safari,

**Environment:**
 - Dioxus version: 0.4.0/git main
 - Rust version: 1.76.0-nightly
 - OS info: macOS 14.1
 - App platform: desktop

**Questionnaire**

pascalbehmenburg:
## Feature: add default menu bar to dioxus-desktop
It looks like this on macOS (please ignore firefox underneath it):
image image

This also resolves #1691 since, as mentioned there, macOS requires menu bar items with keybindings assigned to them to enable system default text editing and window shortcut behaviors.

### Solution
The feature was implemented by adding a function `fn create_default_menu_bar() -> MenuBar` in [webview.rs](packages/desktop/src/webview.rs) which constructs a platform specific menu bar that may be passed into `window_builder` if `disable_default_menu_bar` is `false` which is defined in [config.rs](packages/desktop/src/cfg.rs) to enable the option to opt out of having a menu bar at all. `diasable_default_menu_bar` is `false` by default, and therefore the `window_builder` will use the default menu bar if the user does not opt out, as implemented in [webview](packages/desktop/src/webview.rs).

I think this should be exactly what @ealmloff described in his [comment](https://github.com/DioxusLabs/dioxus/issues/1691#issuecomment-1842846024).

`clippy` and all tests including `playwright` passed on `master`.

**Note**: due to `tao` not supporting a lot of native menu bar items on Linux, there is no edit menu on Linux yet.

### Example usage
`Cargo.toml`:
```toml
dioxus = { path = "../dioxus/packages/dioxus/" }
dioxus-desktop = { path = "../dioxus/packages/desktop/" }
```

`main.rs`:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch_cfg(
        App,
        dioxus_desktop::Config::new()
            .with_disable_default_menu_bar(false)
    );
}

    render! {
        input {
            r#type: "text",
            placeholder: "Enter smth. and use Ctrl + A / Cmd + A...",
        }
    }
}
```


packages/desktop/src/cfg.rs
```diff
@@ -65,9 +66,16 @@ impl Config {
             root_name: "main".to_string(),
             background_color: None,
             last_window_close_behaviour: WindowCloseBehaviour::LastWindowExitsApp,
+            disable_default_menu_bar: false,
         }
     }
 
+    /// Set whether the default menu bar should be disabled.
+    pub fn with_disable_default_menu_bar(mut self, disable: bool) -> Self {
```
ealmloff, line 74: I'm not sure about this name. It could leads to double negatives which can be confusing: `with_disable_default_menu_bar(false)` means the default menu bar is enabled. This is already a pattern that we have introduced into the desktop config builder, so I completely understand why you copied it, but I think `with_default_menu_bar` with a comment about the default being true might be more clear

ealmloff (commented):
I have one small question about a function name, but overall looks great! This will help make desktop apps work more like I would expect out of box. (I also really appreciate the detailed PR description 🙂)

jkelleyrtp (approved):

esimkowitz:
Do we want this to be the default behavior on all platforms or only on Mac? I haven't been targeting Windows or Linux recently but if I remember correctly they weren't impacted by the keybindings issue

packages/desktop/src/cfg.rs
pascalbehmenburg: I agree, just tried to be consistent but actually found the naming kind of bad too in this case, will definitely adjust that to what you've suggested. I will probably make the changes later today.

pascalbehmenburg:
> Do we want this to be the default behavior on all platforms or only on Mac? I haven't been targeting Windows or Linux recently but if I remember correctly they weren't impacted by the keybindings issue

Actually, I don't think it is an issue on Linux or Windows, though I haven't been able to test it either. Otherwise, I do think it won't be bad to have those very basic menus included on these platforms too. It is more consistent that way, otherwise one might wonder where the menus went when cross-compiling or similar.

> Agreed, looks great, definitely something we want to support. It might be worth to make the function that builds the menubar public and then change the naming scheme to what @ealmloff mentioned. Other than that, LGTM!

So you basically want to expose the standard menu bar that was built so that one could adjust it further instead of having to re-create a new Menubar when intending to customize the existing one? I think that would be nice, but I am not too sure where to put it. I thought about exposing it and accepting a MenuBar instead of a bool, but MenuBar does not implement copy / clone. Also, I think no one would expect it to be found in webview.rs when using it from the outside. If you have an idea on that I would be really happy about adjusting it, otherwise I'd suggest keeping it the way it is with the incoming naming changes and opening an issue about exposing it where we can discuss this further.

ealmloff (approved):
Thank you!

ealmloff:
It might be better to wait to merge this until the last 0.4 version goes out. If you already have the system menu added manually, it wouldn't be great to add a second menu even if the API isn't technically breaking

Issue #1616, opened 2023-11-06
Use a tempdir to hide the `dioxusin` pipe for hotreloading


jkelleyrtp:
## Specific Demand

The `dioxusin` file gets created in the top-level directory which shows up in ls/vscode. 

## Implement Suggestion

We should put the pipe in the target folder or a tempdir instead to hide the pipe.

PR #1697, opened 2023-12-06
Use a temp directory for the hot reloading pipe

ealmloff:
This hides the hot reloading file by placing it in the temp directory.

This reverses #842. If the issue is just with `@` in a path, this shouldn't revert that fix.

Closes https://github.com/DioxusLabs/dioxus/issues/1616

### DIFF
--- a/packages/cli/src/server/desktop/mod.rs
+++ b/packages/cli/src/server/desktop/mod.rs
@@ -109,7 +109,8 @@ pub async fn serve(config: CrateConfig, hot_reload_state: Option<HotReloadState>
 }
 
 async fn start_desktop_hot_reload(hot_reload_state: HotReloadState) -> Result<()> {
-    match LocalSocketListener::bind("@dioxusin") {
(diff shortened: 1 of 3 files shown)


jkelleyrtp (changes requested):
This will cause issues if two projects are being hotreloaded simultaneously, right? We should try one of three things:
- put the pipe in the target directory
- put the pipe in a folder in the temp dir
- multiplex the pipe itself

Eventually we probably want 3 but either 1 or 2 would be fine.

ealmloff:
> This will cause issues if two projects are being hotreloaded simultaneously, right? We should try one of three things:
> 
> * put the pipe in the target directory
> * put the pipe in a folder in the temp dir
> * multiplex the pipe itself
> 
> Eventually we probably want 3 but either 1 or 2 would be fine.

It now puts the socket in the target folder

jkelleyrtp (approved):

PR #1698, opened 2023-12-07
Fix disconnects that happen while a server function is being resolved

ealmloff:
When a server function is called, but then the client disconnect, the oneshot channel send panics. This PR fixes that issue by joining the task instead


jkelleyrtp (approved):
At some point it might be nice to set up test harnesses for stuff like this.

PR #1701, opened 2023-12-09
feat: `--split-line-attributes` parameter for CLI `fmt`

marc2332:
## To Do
- [ ] Tests
- [ ] Docs


marc2332:
Without inlined attributes: 
image

With inlined attributes:
image

paulo-granthon:
I'm having trouble with this flag, for some reason it adds a bunch of empty lines, which keep being added with consecutive executions

2026_02_14_20_23_16_sshot

Issue #1586, opened 2023-10-29
tab char ( `\t` ) is not working in RSX


felixf4xu:
**Problem**

tab key ( `\t` ) is not working in RSX:
1) there is a error message when adding '\t' char during hot reloading. 
  the error message is:
>thread '<unnamed>' panicked at D:\github\dioxus\packages\hot-reload\src\lib.rs:34:83:
called `Result::unwrap()` on an `Err` value: Error("invalid type: string \"hello\\tgui\", expected a borrowed string", line: 1, column: 166)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

2) '\t' char is not displayed correctly 

**Steps To Reproduce**

- Create RSX like this:
```rust
fn App(ctx: Scope) -> Element{
    ctx.render(rsx!{
        div { 
            "hellogui"
        }
    })
}
```
- then run it with `dx serve --hot-reload --platform desktop`. 
 During hot reloading, add a tab char '\t' inside the string of "hellogui". In this step, make sure your editor **keeps** tab char, not converting it into spaces. This is the first bug, error messages will be displayed.
- Then re-run `dx serve` command
 the GUI can be displayed (so it's still valid RSX) but tab key is ignored. (I can verify this by copy the string to another editor and check if `\t` is copied. This is the second bug, tab char is not displayed.

**Expected behavior**

Hot reloading works, and tab char is displayed correctly.

**Environment:**
 - Dioxus version: 0.5/master, locally compiled.
 - Rust version: 1.73
 - OS info: Windows 11
 - App platform: desktop

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

ealmloff:
This seems related to https://github.com/serde-rs/json/issues/907. Serde json will not accept unescaped `\t`s in the json. https://crates.io/crates/serde-json-fmt might be able to fix this

i123iu:
I've looked into this and it is more complicated than just differently handling the serialization (as suggested with the crate `serde_json_fmt`). 

The problematic characters (`\n`, `\t`, …) need to be escaped in serialization by the JSON spec ([here](https://www.ietf.org/rfc/rfc4627.txt), 2.5)

[...]

PR #1703, opened 2023-12-10
Fix escaping ascii codes on serializing reload messages

Exotik850:
Fixes [1586](https://github.com/DioxusLabs/dioxus/issues/1586)


ealmloff:
I still get the same error when I change the rsx like this:
```
div {
   "hello world"
}
```

to
```
div {
   "hello\tworld"
}
```

The error

```
thread '<unnamed>' panicked at /Users/evanalmloff/Desktop/Github/dioxus/packages/hot-reload/src/lib.rs:38:83:
called `Result::unwrap()` on an `Err` value: Error("invalid type: string \"Hello\\tWorld!\", expected a borrowed string", line: 1, column: 167)
```

We might need to manually modify the formatted json if `serde-json-fmt` isn't working

jkelleyrtp:
Moved this back to draft until the right fixes are applied.

Issue #1520, opened 2023-10-04
`dx fmt` should skip `gitignore`d files


srid:
## Specific Demand

What I see right now:
<img width="961" alt="image" src="https://github.com/DioxusLabs/dioxus/assets/3998/85eb8902-b6d6-4772-bf87-52be7b23ba69">

## Implement Suggestion

Some options:

- Look for `.gitignore` and ignore those files automatically.
- Implement a whitelist to look sources
- Accept directory in command line and format the sources under that recursively, so I can just run `dx fmt ./src`

PR #1704, opened 2023-12-10
Autoformat ignores files in `.gitignore`

Exotik850:
Fixes [1520](https://github.com/DioxusLabs/dioxus/issues/1520)

### DIFF
--- a/packages/cli/src/cli/autoformat.rs
+++ b/packages/cli/src/cli/autoformat.rs
@@ -97,7 +97,9 @@ async fn autoformat_project(check: bool) -> Result<()> {
     let gitignore_path = crate_config.crate_dir.join(".gitignore");
     if gitignore_path.is_file() {
         let gitigno = gitignore::File::new(gitignore_path.as_path()).unwrap();
-        files_to_format.extend(gitigno.included_files().unwrap());
+        if let Some(files) = gitigno.included_files() {
+            files_to_format.extend(files)
+        };
     } else {
         collect_rs_files(&crate_config.crate_dir, &mut files_to_format);
     }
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/autoformat.rs
```diff
@@ -207,24 +206,20 @@ fn indentation_for(file_or_dir: impl AsRef<Path>) -> Result<IndentOptions> {
...
     for entry in folder {
         let Ok(entry) = entry else {
             continue;
         };
-
         let path = entry.path();
-
         if path.is_dir() {
             collect_rs_files(&path, files);
+            return;
```
ealmloff, line 221: If we return here, I think it will only read the first directory in this structure:
- Folder 1 (collect_rs_files is called from this path)
  - Folder 2 (It reaches this folder)
  - Folder 3 (collect_rs_files so it never reaches this folder)

packages/cli/src/cli/autoformat.rs
```diff
@@ -97,7 +93,16 @@ async fn autoformat_project(check: bool) -> Result<()> {
     let crate_config = crate::CrateConfig::new(None)?;
 
     let mut files_to_format = vec![];
-    collect_rs_files(&crate_config.crate_dir, &mut files_to_format);
+
+    let gitignore_path = crate_config.crate_dir.join(".gitignore");
+    if gitignore_path.is_file() {
+        let gitigno = gitignore::File::new(gitignore_path.as_path()).unwrap();
+        if let Ok(files) = gitigno.included_files() {
+            files_to_format.extend(files)
```
ealmloff, line 101: We should filter out any `.rs` files even if they are allowed in the gitignore. You might want to track a file in git that is not a rust file. Currently, I think this would try to format the `cargo.toml` file

Exotik850: I felt like I was forgetting something but it completely slipped my mind, nice catch! gotta love reviews

ealmloff (approved):
I changed is_target_dir to not ignore paths like `/src/target/windows`. Other than that it looks good. Thank you!

PR #1706, opened 2023-12-10
Fix doc tests by adding formatting to CLI README

benmanns:
The README for cli gets included in lib.rs via `include_str!`. The code snippets then get assumed to be Rust, which fails if running `cargo test --doc`. This is not a huge problem, but was brought to my attention because Nix's `buildRustPackage` runs `cargo test` including doc tests (unlike what get run for this project on CI). This means I can't update the Nix package to 0.4.3 without changing the test flags (which I think I'll see about doing anyways, as that's how Dioxus runs CI).

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -10,7 +10,7 @@ It handles building, bundling, development and publishing to simplify developmen
 
 ### Install the stable version (recommended)
 
-```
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thank you!

PR #1709, opened 2023-12-11
fix: dx create not work

mrxiaozhuox:
Because the `get_bin` function will use `cargo_metadata`, but for some command we don't have it.


ealmloff (approved):
Thank you!

The code for the plugin system here is kind of messy. All of it should be configged out, but it isn't related to this change. I can fix it in the plugins PR.

PR #1710, opened 2023-12-11
Make cx.spawn poll the task before returning `TaskId`

Exotik850:
Reopen of [1669](https://github.com/DioxusLabs/dioxus/pull/1669) (oops)

### DIFF
--- a/packages/core/src/scheduler/task.rs
+++ b/packages/core/src/scheduler/task.rs
@@ -19,7 +19,7 @@ pub struct TaskId(pub usize);
 /// the task itself is the waker
 pub(crate) struct LocalTask {
     pub scope: ScopeId,
-    pub(super) task: RefCell<Pin<Box<dyn Future<Output = ()> + 'static>>>,
+    pub task: RefCell<Pin<Box<dyn Future<Output = ()> + 'static>>>,
     pub waker: Waker,
 }
 
@@ -33,7 +33,11 @@ impl Scheduler {
     ///
     /// Spawning a future onto the root scope will cause it to be dropped when the root component is dropped - which
     /// will only occur when the VirtuaalDom itself has been dropped.
-    pub fn spawn(&self, scope: ScopeId, task: impl Future<Output = ()> + 'static) -> TaskId {
+    pub fn spawn(
+        &self,
+        scope: ScopeId,
+        task: impl Future<Output = ()> + 'static,
+    ) -> Option<TaskId> {
         let mut tasks = self.tasks.borrow_mut();
 
         let entry = tasks.vacant_entry();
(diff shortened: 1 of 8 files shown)


packages/core/src/scheduler/task.rs
```diff
@@ -33,7 +33,11 @@ impl Scheduler {
     ///
     /// Spawning a future onto the root scope will cause it to be dropped when the root component is dropped - which
     /// will only occur when the VirtuaalDom itself has been dropped.
-    pub fn spawn(&self, scope: ScopeId, task: impl Future<Output = ()> + 'static) -> TaskId {
+    pub fn spawn(
+        &self,
+        scope: ScopeId,
+        task: impl Future<Output = ()> + 'static,
+    ) -> Option<TaskId> {
```
jkelleyrtp, line 40: Would prefer to keep returning `TaskId` so we can move it to a `JoinHandle` -esque API in the future.

This would involve creating the waker as you've already done on line 57, but not returning None. 

If you wanted to take this is a step further, adding some sort of JoinHandle methods onto TaskId would be useful by checking with the current Dioxus runtime.

jkelleyrtp (changes requested):

packages/hooks/src/use_coroutine.rs
```diff
@@ -69,11 +69,9 @@ where
     G: FnOnce(UnboundedReceiver<M>) -> F,
     F: Future<Output = ()> + 'static,
 {
-    cx.use_hook(|| {
-        let (tx, rx) = futures_channel::mpsc::unbounded();
-        let task = cx.push_future(init(rx));
-        cx.provide_context(Coroutine { tx, task })
-    })
+    let (tx, rx) = futures_channel::mpsc::unbounded();
+    let task = cx.push_future(init(rx));
```
ealmloff, line 73: I think this should still be in the use_hook. We only want to create and spawn the coroutine future when the component is first created, not every time the component is rerun

Exotik850: Forgot to change it back, thanks!

Exotik850:
Even after moving the check to [here](https://github.com/DioxusLabs/dioxus/blob/695bf009d11be0a8b717257ab2d56adf8873c489/packages/core/src/scope_context.rs#L217) so that it won't add it to the spawned_tasks if the task is ready, it's still failing, so i'm not sure what's causing the `"async fn" resumed after completion` error

packages/core/src/scheduler/task.rs
```diff
@@ -48,6 +48,9 @@ impl Scheduler {
             })),
         };
 
+        let mut cx = std::task::Context::from_waker(&task.waker);
+        let _ = task.task.borrow_mut().as_mut().poll(&mut cx);
+
         entry.insert(task);
 
         self.sender
```
ealmloff, line 56: The error you are running into might be related to the future being woken up after it is finished because you notify the task immediately here

Exotik850: you're absolutely right, just added the check if it's ready before it notifies the task

ealmloff (approved):

jkelleyrtp:
Thanks! Should be going out in a pre-release shortly...

Issue #1718, opened 2023-12-13
Support dynamically-generated assets


willcrichton:
## Specific Demand

A webview may have dynamically generated content. My use case is an EPUB reader where the user provides a ZIP file containing web files with assets. I can dynamically load the web file (e.g. using `dangerous_inner_html`), but if that HTML contains an `<img src="custom.jpg" />`, then I have no easy way to intercept that request and redirect it to the ZIP file.

## Implement Suggestion

Something around `WebViewBuilder::with_custom_protocol` is probably the way to go. I'll mess around with this and post here if I make progress.

willcrichton:
One possible solution is to have the user provide a custom asset handler in the `Config` structure. I prototyped one such interface here: https://github.com/willcrichton/dioxus/commit/deded8063da08244df661a0886f046af1f009311

ealmloff:
closed by https://github.com/DioxusLabs/dioxus/pull/1719

PR #1719, opened 2023-12-13
Add custom asset handler to desktop config

willcrichton:
This addresses issue #1718 by adding a user-configurable callback to the asset handler. The callback is a function `&Path -> Option<Vec<u8>>`. 

Some considerations:
- Should the user be able to return an `Option` (handle or fallback), an `Error` (handle or fail), or a ternary thing (handle, fail, or fallback)?
- Should the user return a buffer only, or should they be expected to construct the entire `Response`?


ealmloff:
> Should the user be able to return an Option (handle or fallback), an Error (handle or fail), or a ternary thing (handle, fail, or fallback)?

I think returning an option is fine. I think most of the time if a handler reached an error, it should just log it and continue trying to find the asset which can happen within the callback

> Should the user return a buffer only, or should they be expected to construct the entire Response?

Being able to read the rest of the body and return a request would be nice for streaming assets like [this tauri example](https://github.com/tauri-apps/tauri/blob/ef21b681e237a80592c9118b9c023c1d57231bac/examples/streaming/main.rs#L65)

Adding an option to the config is a good starting point, but I think adding hooks to dynamically create an asset handler might be more flexible (especially for any libraries that use custom assets). We could maintain a list of handlers wrapped in a `Arc<RwLock<T>>` like we currently do for [shortcuts](https://github.com/DioxusLabs/dioxus/blob/master/packages/desktop/src/shortcut.rs#L31). That would let us support something like this API:

```rust
// Add an asset handler dynamically
let asset = use_asset_handler(cx, |request| todo!());

// Or just handle a specific asset path
let asset = use_asset(cx, "my-asset-path", || todo!());
```

willcrichton:
@ealmloff I rebased off your `bump-wry` branch and added a `use_asset_handler` API: https://github.com/DioxusLabs/dioxus/pull/1719/files#diff-d5641e90df32efe5aa1817711d1139fc1e699abfd33229af54bc692c3d7041f6

Does this look better?

packages/desktop/src/protocol.rs
```diff
@@ -51,12 +63,132 @@ fn module_loader(root_name: &str) -> String {
...
+/// This handler is stashed indefinitely in a context object, so it must be `'static`.
+pub trait AssetHandler<F: AssetFuture>: Fn(&Path) -> F + Send + Sync + 'static {}
+impl<F: AssetFuture, T: Fn(&Path) -> F + Send + Sync + 'static> AssetHandler<F> for T {}
+
+/// An identifier for a registered asset handler, returned by [`AssetHandlerRegistry::register_handler`].
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
+pub struct AssetHandlerId(usize);
+
+struct AssetHandlerRegistryInner {
+    handlers: HashMap<
```
ealmloff, line 84: This looks a lot like a [slab](https://docs.rs/slab/latest/slab/). If we don't need the AssetHandlerId to be unique throughout time, it might be simpler to use a slab directly

ealmloff:
> @ealmloff I rebased off your `bump-wry` branch and added a `use_asset_handler` API: https://github.com/DioxusLabs/dioxus/pull/1719/files#diff-d5641e90df32efe5aa1817711d1139fc1e699abfd33229af54bc692c3d7041f6
> 
> Does this look better?

That looks a lot better, thank you! I might add a way to read the rest of the request so that we can implement streaming video without breaking this API (unless you had any concerns). Then this API looks pretty good!

willcrichton:
Simplified it with slab. Feel free to extend the API for streaming video.

ealmloff:
I pushed a few changes to make it possible to read the request. Streaming video is not implemented, but it should now be possible to implement it without breaking this API in another PR

ealmloff (approved):
Looks good, thank you! (waiting a second review before merging this because it is breaking)

PR #1723, opened 2023-12-15
dx bundle  support generate package with signature like tauri

x-falcon:
Then we can easily update our app by  https://docs.rs/cargo-packager-updater/latest/cargo_packager_updater


Abrahamh08:
Why did you close this? Did you find an alternative

x-falcon:
> Why did you close this? Did you find an alternative

It's been so long that I can't even remember why anymore.

PR #1724, opened 2023-12-15
add error log when root element isn't found

### LINKED ISSUE #1231: Web: Error when Dioxus fails to mount to the real dom
## Specific Demand

On the web renderer, Dioxus should error to the console when it fails to find any elements with an id of `main` to mount to.

w3irdrobot:
closes #1231

### DIFF
--- a/packages/web/Cargo.toml
+++ b/packages/web/Cargo.toml
@@ -42,6 +42,7 @@ features = [
     "HtmlFormElement",
     "Text",
     "Window",
+    "console",
 ]
 
 [features]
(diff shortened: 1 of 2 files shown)


ealmloff:
It looks like we only enable the console feature of web-sys [here](https://github.com/DioxusLabs/dioxus/blob/1ec230ae292ecb011d71675c983f12757eff6772/packages/web/Cargo.toml#L52) if the hydrate feature is enabled. The workspace has that feature enabled, so I don't think the issue will show up in CI, but we need to add it to the list of web-sys features that are always enabled [here](https://github.com/DioxusLabs/dioxus/blob/1ec230ae292ecb011d71675c983f12757eff6772/packages/web/Cargo.toml#L44)

ealmloff (approved):
This should make a missing root id much easier to debug. Thank you!

PR #1764, opened 2023-12-30
Send mount event when hydrating

### LINKED ISSUE #1763: `mounted` event is not triggered in fullstack renderer
**Problem**

I am trying to listen to the mounted event in a component to get it's size after it has been rendered. The problem is that I can't get it to work in the fullstack renderer (dioxus-fullstack). It does work in the web renderer (dioxus-web) however.

**Steps To Reproduce**

A small example of how I am using it:

```rust
use dioxus::html::geometry::euclid::Rect;
use dioxus_fullstack::prelude::*;

fn main() {
    dioxus_logger::init(log::LevelFilter::Debug).expect("failed to initialize logger");
    
    /* Works woth dioxus-web. */
    dioxus_web::launch(App);

    /* Doesn't work with dioxus-fullstack. */
    // LaunchBuilder::new(App).launch();
}

pub fn App(cx: Scope) -> Element {
    let size = use_state(cx, Rect::zero);

    log::debug!("App");

    render! {
        div {
            onmounted: |event|
            {
                log::debug!("Mounted");
                to_owned![size];
                async move {
                    if let Ok(rect) = event.inner().get_client_rect().await {
                        size.set(rect);
                    }
                }
            },
            "{size:?}"
        }
    }
}
```

The problem is also reproducible in the following `main()` setup:

```rust
fn main() {
    dioxus_logger::init(log::LevelFilter::Debug).expect("failed to initialize logger");

    #[cfg(feature = "web")]
    dioxus_web::launch_with_props(
        App,
        get_root_props_from_document().unwrap_or_default(),
        dioxus_web::Config::new().hydrate(true),
    );

    #[cfg(feature = "ssr")]
    {
        tokio::runtime::Runtime::new()
            .unwrap()
            .block_on(async move {
                let addr = std::net::SocketAddr::from(([127, 0, 0, 1], 8080));
                axum::Server::bind(&addr)
                    .serve(
                        axum::Router::new()
                            .serve_dioxus_application("", ServeConfigBuilder::new(App, ()))
                            .into_make_service(),
                    )
                    .await
                    .unwrap();
            });
    }
}
```

**Expected behavior**

Using the web renderer will yield the output "App\nMounted\nApp" with the inner HTML "Rect(860.0x18.0 at (0.0, 0.0))". Using the fullstack renderer will yield the output "App" with the inner HTML "Rect(0.0x0.0 at (0.0, 0.0))". Expected behavior is that both methods should have the same output.

**Environment:**
 - Dioxus version: 0.4.3
 - Rust version: 1.74.1
 - OS info: macOS Monteray
 - App platform: fullstack

**Questionnaire**

hampuslidin:
Resolves #1763.

### DIFF
--- a/packages/web/src/dom.rs
+++ b/packages/web/src/dom.rs
@@ -256,17 +256,21 @@ impl WebsysDom {
         i.flush();
 
         for id in to_mount {
-            let node = get_node(id.0 as u32);
-            if let Some(element) = node.dyn_ref::<Element>() {
-                let data: MountedData = element.into();
-                let data = Rc::new(data);
-                let _ = self.event_channel.unbounded_send(UiEvent {
-                    name: "mounted".to_string(),
-                    bubbles: false,
-                    element: id,
-                    data,
-                });
-            }
+            self.send_mount_event(id);
+        }
+    }
+
+    pub(crate) fn send_mount_event(&self, id: ElementId) {
+        let node = get_node(id.0 as u32);
+        if let Some(element) = node.dyn_ref::<Element>() {
+            let data: MountedData = element.into();
+            let data = Rc::new(data);
+            let _ = self.event_channel.unbounded_send(UiEvent {
+                name: "mounted".to_string(),
+                bubbles: false,
+                element: id,
+                data,
+            });
         }
     }
 }
--- a/packages/web/src/rehydrate.rs
+++ b/packages/web/src/rehydrate.rs
@@ -125,6 +125,7 @@ impl WebsysDom {
                 children, attrs, ..
             } => {
                 let mut mounted_id = None;
+                let mut should_send_mount_event = true;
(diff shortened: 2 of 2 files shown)


hampuslidin:
This will conflict with #1732, so this code needs to adapt to that PR. Most notably, the `set_node` JS function binding has been replaced with `hydrate`, which handles all the hydration in one go, as well as creating event listeners. I assume it will incorrectly create an event listener for the "mount" event, instead of firing it, which is what this PR solves.

ealmloff (approved):
This looks good for now, I can fix the conflicts in #1732. Thank you!

Fixed in https://github.com/DioxusLabs/dioxus/pull/1732/commits/0a7873fcd0dbde76e9e4dc2869f49b97dc641bb7

PR #1766, opened 2023-12-30
Server function improvements

ealmloff:
This PR integrates https://github.com/gbj/server_fns which is a rewrite of server functions that contains middleware that can be shared between dioxus and leptos. It also includes more options for codecs including streaming server functions


ealmloff:
folded into https://github.com/DioxusLabs/dioxus/pull/1944

PR #1772, opened 2024-01-01
Implemented From<Signal<T>> for ReadOnlySignal<T>

werner291:
I noticed I couldn't convert a Signal into a ReadOnlySignal with .into()

This implements the trait that enables that.


ealmloff (approved):
Thank you!

PR #1775, opened 2024-01-02
Add doc comment on generated inline properties struct

hampuslidin:
Adds a doc comment to the derived properties struct generated from the inline arguments of the component.

This prevents the `missing_docs` lint to trigger if that is setup for a project (for instance, a CI might prevent progress by denying all lint warnings). This could probably be extended to provide custom doc comments to the struct, or use the doc comment of the component function, but I didn't see the need for that, as this PR will include a link to the original component function in the docs for further info.

### DIFF
--- a/packages/core-macro/src/component_body_deserializers/component.rs
+++ b/packages/core-macro/src/component_body_deserializers/component.rs
@@ -58,8 +58,11 @@ impl ToTokens for ComponentDeserializerOutput {
     fn to_tokens(&self, tokens: &mut TokenStream2) {
         let comp_fn = &self.comp_fn;
         let props_struct = &self.props_struct;
+        let fn_ident = &comp_fn.sig.ident;
 
+        let doc = format!("Properties for the [`{fn_ident}`] component.");
         tokens.append_all(quote! {
+            #[doc = #doc]
             #props_struct
             #[allow(non_snake_case)]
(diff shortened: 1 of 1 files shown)


packages/core-macro/src/component_body_deserializers/inline_props.rs
```diff
@@ -102,7 +102,9 @@ fn get_props_struct(component_body: &ComponentBody) -> ItemStruct {
         quote! { #generics }
     };
 
+    let doc = format!("Properties for the [`{fn_ident}`] component.");
```
ealmloff, line 105: Could you apply the same change to the component macro [here](https://github.com/DioxusLabs/dioxus/blob/a0b95a82a0fa16d9101897554444878da7096f74/packages/core-macro/src/component_body_deserializers/component.rs#L63C20-L63C20)?

hampuslidin: In that case, in order to avoid having the doc comment appearing twice, I would have to revert the change in the `inline_props.rs` file, since `component.rs` is indirectly referencing the `get_props_struct()` function via the `InlinePropsDeserializerOutput` type. But I guess that is fine if that macro is deprecated anyway!

ealmloff (approved):
Thank you!

PR #1777, opened 2024-01-02
Add Cargo Arguments and Cross-Architecture Support

### LINKED ISSUE #1347: `dx build`: add `no-default-features`
## Specific Demand

I have a full-stack application that is primarily used as a server, and should build as a server by default. To do this, I have `default = ["server"]` specified (which is responsible for bringing in all of the server dependencies and calling `serve_dioxus_application`).

However, `dx build` for the web doesn't let you turn off the default features, which means that `dx build --features web` will build with both `server` and `web`, leading to the expected shenanigans.

Removing `server` from the `default` works fine, but it means both the DX and UX are slightly worse as a result.

## Implement Suggestion

Add `no-default-features` as a flag, similar to Cargo, and pass it through while building.

### LINKED ISSUE #1693: Support cross compilation in dioxus-CLI
## Specific Demand

The ability to pass Cargo's [`--target`](https://doc.rust-lang.org/cargo/commands/cargo-build.html#compilation-options) parameter to `dx build`, to support cross compilation for Desktop and Fullstack applications.

## Implement Suggestion

Add specific support for the `--target` parameter, or a more general way to passthrough arguments to `cargo build`.

Bennett-Petzold:
Changes apply for the CLI actions `build`, `bundle`, and `serve`. Adds `-- [CARGO_FLAGS]` and `--platform-triple`.

Cargo arguments can now be passed through after `--` (e.g. `dx build -- --timings`)  via escaped conditionals. This does lock out that space for any later usage without a breaking change to the CLI, but I can't think of any other necessary use for post-argument parsing on this tool.

Since the `--target` flag is currently taken, a `--platform-triple` flag allows alternative values for `cargo build --target`. This can't be done directly in passthrough, since dioxus-cli needs to know the output folder. If reasonable, the existing `--target` flag should probably be renamed so we can be 1:1 with cargo naming.

If the codebase has any other instances where bundled anyhow errors are flattened by "{}" they should probably be replaced with "{:?}" or "{:#?}" -- this caused a lot of confusion trying to bundle my first Dioxus project.

Closes #1693, closes #1347

### DIFF
--- a/packages/cli/src/builder.rs
+++ b/packages/cli/src/builder.rs
@@ -93,6 +93,8 @@ pub fn build(config: &CrateConfig, quiet: bool) -> Result<BuildResult> {
         cmd
     };
 
+    let cmd = cmd.args(&config.cargo_args);
+
     let cmd = match executable {
         ExecutableType::Binary(name) => cmd.arg("--bin").arg(name),
         ExecutableType::Lib(name) => cmd.arg("--lib").arg(name),
@@ -285,6 +287,8 @@ pub fn build_desktop(config: &CrateConfig, _is_serve: bool) -> Result<BuildResul
         let features_str = config.features.as_ref().unwrap().join(" ");
         cmd = cmd.arg("--features").arg(features_str);
     }
+
+    let cmd = cmd.args(&config.cargo_args);
 
     let cmd = match &config.executable {
         crate::ExecutableType::Binary(name) => cmd.arg("--bin").arg(name),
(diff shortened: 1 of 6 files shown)


ealmloff:
> Since the --target flag is currently taken, a --platform-triple flag allows alternative values for cargo build --target. This can't be done directly in passthrough, since dioxus-cli needs to know the output folder. If reasonable, the existing --target flag should probably be renamed so we can be 1:1 with cargo naming.

We can use the target flag for the target triple for parity with the cargo commands. It is breaking, but it would make it a lot easier to use coming from cargo. `--target` is also not a very clear name for the index path argument and it looks like it is currently unused. If we expose a different flag to use a different `index.html` path in the future, it can have a different name

Bennett-Petzold:
I've removed `--platform-triple` and given it's behavior to `--target`, so it works the same as cargo now.

ealmloff (approved):
Looks good, this changes the `--target` arg, but because that argument did nothing, I think it is OK to make that change in a minor release.

Thank you!

PR #1782, opened 2024-01-03
Fix rsx rosetta element and attribute mapping

ealmloff:
This PR changes rsx-rosetta to use the mapping of html->rsx names for elements and attributes defined in dioxus-html instead of converting the case of the attributes and elements.

This fixes the following cases:

- translating h1-h6

old:
```
h_1 {}
```
new:
```
h1 {}
```

- translating web components

old:
```
my_component {}
```
new:
```
my-component {}
```

- translating unsupported attributes

old:
```
div { unrecognizedattribute: "asd", "hello world" }
```
new:
```
div { "unrecognizedattribute": "asd", "hello world" }
```


jkelleyrtp:
Sweet, thanks!

Issue #1753, opened 2023-12-27
UseState::get is easy to misuse


werner291:
**Problem**

Hi!

So, the `UseState<T>::get()` method gets the value of the state hook when it was *initially* set.

While this works exactly as documented, it's easy to mistake for a way to simply get the value of the state; it's led to bugs from at least two separate devs (including me) on a project using Dioxus.

May I suggest renaming it to something like `get_original` to make mistakes less likely?

ealmloff:
It is easy to misuse `UseState` because it uses the value of the state when the hook was rerun, but that is central to the design of `UseState`. `UseState` [implements](https://docs.rs/dioxus-hooks/latest/src/dioxus_hooks/use_state.rs.html#390-396) `Deref` and [other traits](https://github.com/DioxusLabs/dioxus/issues/619) with the same issue.

also relevant: https://github.com/DioxusLabs/dioxus/issues/619#issuecomment-1343715579

UseState is used in a lot of Dioxus apps, so changing the name of `get` would be very breaking. It would also not fix the other issues with the hook.

In the short term: you can continue to use `UseState` without any breaking changes. It may be depreciated in a future release
In the long term: Dioxus is switching to [signals](https://github.com/DioxusLabs/dioxus/tree/master/packages/signals) as the main state management solution. Signals will never return an out of state value, but they are almost as easy to read as `UseState`'s `Deref` because you can call a signal as a function to read it

Issue #1405, opened 2023-08-29
Signals: with_mock_scope


ealmloff:
## Specific Demand

You can create a signal with an owner scope, but you need to explicitly pass that scope id everywhere. This can be problematic for deserialization and can result in duplicates of a lot of functions. (`State::new()`, `State::new_in_scope()`)

## Implement Suggestion

Create a new with_owner_scope function. If you create a signal inside of the callback you pass to that function it will assign the owner to the scope you set:

```rust
// This will create a new signal owned by the root scope
with_mock_scope(ScopeId::ROOT, || {Signal::new(0});
```

This would be implemented by pushing and popping a scope from the [scope_stack](https://github.com/DioxusLabs/dioxus/blob/master/packages/core/src/runtime.rs#L49)

Issue #1374, opened 2023-08-18
Static Event Handlers for use in futures


ealmloff:
## Specific Demand

To call an event handler in a future, the event handler must be static, or have a `'static` handle.

## Implement Suggestion

If we could implement this with a channel so that the event handler can still take a `'bump` lifetime that should be preferred. That would make normal event handlers usable in a future. Otherwise we could create a new `StaticEventHandler` struct like this:
```rust
use std::{rc::Rc, future::Future, pin::Pin};

use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

fn app(cx: Scope) -> Element {
    cx.render(rsx! {
        Test {
            handlefile: |_| async {
                println!("Hello world");
            },
            other: &(),
        }
    })
}

#[derive(Clone)]
struct StaticCallback<T>{
    inner: Rc<RefCell<Box<dyn FnMut(T) -> Pin<Box<dyn Future<Output = ()>>>>>>,
}

impl<F, Fut, T> From<F> for StaticCallback<T> where F: FnMut(T)-> Fut + 'static, Fut: Future<Output = ()> + 'static {
    fn from(mut f: F) -> Self {
        Self {
            inner: Rc::new(RefCell::new(Box::new(move |input| Box::pin(f(input)))))
        }
    }
}

#[derive(Props)]
struct TestProps<'a> {
    other: &'a (),
    #[props(into)]
    handlefile: Option<StaticCallback<Vec<u8>>>
}

fn Test<'a>(cx: Scope<'a, TestProps<'a>>) -> Element<'a> {
    render! {
        div {
            onclick: move |_| {
                let handle_file = cx.props.handlefile.clone();

                async move {
                    let file = vec![1, 2, 3];
                    if let Some(handler) = &handle_file {
                        (handler.inner.borrow_mut())(file).await;
                    }
                }
            }
        }
    }
}
```

ealmloff:
> Does this require the handler to be a future?
> 
> Also, if I have a state inside App and try to read/write it in the handler, I get a lifetime error since the state is an `'a` and it needs to be a `'static`.

This does, but if this was merged into the framework, we could make it work by returning a trait. We do this today with event handlers here: https://github.com/DioxusLabs/dioxus/blob/master/packages/html/src/events.rs#L12

> Also, if I have a state inside App and try to read/write it in the handler, I get a lifetime error since the state is an 'a and it needs to be a 'static.

[...]

Issue #1032, opened 2023-05-18
Element components cannot be reused


Niedzwiedzw:
**Problem**
```rust
use super::*;

#[derive(Props)]
pub struct DropdownMenuProps<'children> {
    dropdown_content: Element<'children>,
    controller_open: Element<'children>,
    controller_closed: Element<'children>,
}

pub fn DropdownMenu<'children>(cx: Scope<'children, DropdownMenuProps<'children>>) -> Element {
    let open = use_state(cx, || false);
    let onclick = |_| open.set(!open.get());

    cx.render(rsx!(
        div { class: "DropdownMenu",
            div { class: "controller-content", onclick: onclick,
                if *open.get() {
                    &cx.props.controller_open
                } else {
                    &cx.props.controller_closed
                }
            }
            div { class: "dropdown-content",
                open.get().then_some(()).and_then(|_| cx.props.dropdown_content.as_ref())
            }
        }
    ))
}

```
```rust
panicked at 'called `Option::unwrap()` on a `None` value', /home/niedzwiedz/.cargo/git/checkouts/dioxus-1e619ce595d3799d/6acec6f/packages/core/src/diff.rs:213:57
```
above component crashes after I toggle it 3 times.

**Steps To Reproduce**
dioxus = { git = "https://github.com/DioxusLabs/dioxus", rev = "c97f43146d17abca049f77a616ae6b116d90ca34" }
dioxus-web = { git = "https://github.com/DioxusLabs/dioxus", rev = "c97f43146d17abca049f77a616ae6b116d90ca34" }

Niedzwiedzw:
discord thread:
https://discord.com/channels/899851952891002890/899851952891002893/1108757735522177147

Demonthos:
Could you post the props you pass into the element in the example where it crashes?

I tested this code and could not reproduce the issue:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

fn app(cx: Scope) -> Element {
    let dynamic = String::from("testing");
    cx.render(rsx!(DropdownMenu {
        dropdown_content: render! {
            div { "Hello, world! {dynamic}" }
        },
        controller_open: render! {
            div { "Close {dynamic}" }
        },
        controller_closed: render! {
            div { "Open {dynamic}" }
        },
    }))
}

#[derive(Props)]
pub struct DropdownMenuProps<'children> {
    dropdown_content: Element<'children>,
    controller_open: Element<'children>,
[...]

PR #1791, opened 2024-01-06
Remove scope, use_state, use_ref, bump allocator and make everything 'static

ealmloff:
This PR is a massively breaking change to the entirety of Dioxus, reworking state management almost entirely. It moves Dioxus away from the classic React useState towards a signal-based paradigm introduced in 0.4.3.

The inspiration for this work comes from a number of places like SolidJS with an entirely 'static lifetime system borrowed from Leptos. The manifestation of this work will look very similar to the final API of the upcoming React Forget compiler where dependencies on hooks like use_effect are managed automatically by the compiler.

This has a number of advantages
- No more dependency arrays
- No more requirement to call `clone` to bring state into closures and async blocks
- Precise re-renders based on selectors and (eventually) incremental fine-grained reactivity
- No loss of reactivity thanks to diffing and re-rendering architecture
- Ability to call event handlers from async functions
- Much simpler and more easily composable global APIs
- Removal of the scope cx object
- Removal of all lifetimes
- Ability to cache vnodes for things like virtualized lists

Dioxus will still retain:
- hooks
- global context
- components 
- diffing and re-rendering

The impetus for this work is finally acknowledging the difficulty of managing a lifetime managed state system while simultaneously looking forward toward's React's future. This state system is novel and does not exist in any other one particular framework.

Closes https://github.com/DioxusLabs/dioxus/issues/1374
Closes https://github.com/DioxusLabs/dioxus/issues/1032
Closes https://github.com/DioxusLabs/dioxus/issues/619
Closes https://github.com/DioxusLabs/dioxus/issues/1753
Closes https://github.com/DioxusLabs/dioxus/issues/1793
Closes https://github.com/DioxusLabs/dioxus/issues/1405
Closes https://github.com/DioxusLabs/docsite/pull/192

Todo:
- [x] Fix core tests
- [x] Change effect order so use_effect/use_selectors are deferred and refreshed
- [x] Signals as props should be made in child scope
- [ ] Introduce boxed variants of `Readable<T>` and `Writable<T>`?
- [ ] Overhaul examples
- [ ] Overhaul tests
- [x] flush_sync for consistent effect order
- [ ] Overhaul documentation
- [x] Overhaul desktop APIs
- [x] Overhaul router APIs
- [ ] ...more


ealmloff:
Here is the JS framework benchmark result after just removing the bump allocator:

Screenshot 2024-01-07 at 4 45 36 PM

Still the same ranking. It might be a bit slower, but there is a lot more we can do to get some of the performance benefits back without the unsafe like using a slotmap of Elements and using [diffable arguments](https://github.com/ealmloff/diffable-arguments)

jkelleyrtp:
Are the use_effect/use_future/use_memo all updated in this PR? Curious about test driving it with the new primitives.

As far as performance... It looks like we moved, right? We were at 1.12 before? I'd like to stay closer to that if we can. I think some obvious wins are reusing buffers (like strings) and merging buffers (again, like strings, and element vecs). Event handlers are a bit more complex and box is *probably* fine for those, but that's where stuff like the bump really shined.

ealmloff:
> Are the use_effect/use_future/use_memo all updated in this PR? Curious about test driving it with the new primitives.

Not yet, dioxus-hooks is currently just commented out. `dioxus-signals` works, but hasn't been changed

> As far as performance... It looks like we moved, right? We were at 1.12 before? I'd like to stay closer to that if we can. I think some obvious wins are reusing buffers (like strings) and merging buffers (again, like strings, and element vecs). Event handlers are a bit more complex and box is _probably_ fine for those, but that's where stuff like the bump really shined.

Yeah, it is a bit slower. We are at 1.12 on the official benchmark. I just wanted to test what I naive implementation would look like before I spent too much time on optimizations, but I have been working on some optimizations today. One pretty easy win is keeping the mounted information across renders instead of allocating a new buffer and copying them every time

ealmloff:
Here is what the js framework benchmark looks like:
```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;
use dioxus_signals::*;
use js_sys::Math;

fn random(max: usize) -> usize {
    (Math::random() * 1000.0) as usize % max
}

fn main() {
    dioxus_web::launch(app);
}

#[derive(Copy, Clone, PartialEq)]
struct Label {
    key: usize,
    label: Signal<String>,
}

impl Label {
    fn new(num: usize, label: String) -> Self {
        Label {
            key: num,
            label: Signal::new(label),
        }
    }

    fn new_list(num: usize, key_from: usize) -> Vec<Self> {
        let mut labels = Vec::with_capacity(num);
        append(&mut labels, num, key_from);
        labels
    }
}

fn append(list: &mut Vec<Label>, num: usize, key_from: usize) {
    list.reserve_exact(num);
    for x in 0..num {
        let adjective = ADJECTIVES[random(ADJECTIVES.len())];
        let colour = COLOURS[random(COLOURS.len())];
        let noun = NOUNS[random(NOUNS.len())];
        let mut label = String::with_capacity(adjective.len() + colour.len() + noun.len() + 2);
        label.push_str(adjective);
        label.push(' ');
        label.push_str(colour);
        label.push(' ');
        label.push_str(noun);
        list.push(Label::new(x + key_from, label));
    }
}

#[derive(Clone, PartialEq)]
struct LabelsContainer {
    last_key: usize,
    labels: Vec<Label>,
}

impl LabelsContainer {
    fn new(num: usize, last_key: usize) -> LabelsContainer {
        let labels = Label::new_list(num, last_key + 1);
        LabelsContainer {
            labels,
            last_key: last_key + num,
        }
    }

    fn append(&mut self, num: usize) {
        self.labels.reserve(num);
        append(&mut self.labels, num, self.last_key + 1);
        self.last_key += num;
    }

    fn overwrite(&mut self, num: usize) {
        self.labels.clear();
        append(&mut self.labels, num, self.last_key + 1);
        self.last_key += num;
    }

    fn swap(&mut self, a: usize, b: usize) {
        if self.labels.len() > a + 1 && self.labels.len() > b {
            self.labels.swap(a, b);
        }
    }

    fn remove(&mut self, key: usize) {
        if let Some(to_remove) = self.labels.iter().position(|x| x.key == key) {
            self.labels.remove(to_remove);
        }
    }
}

fn app(_: ()) -> Element {
    let labels_container = use_signal(|| LabelsContainer::new(0, 0));
    let selected: Signal<Option<usize>> = use_signal(|| None);
    let prev_selected: Signal<Option<usize>> = use_signal(|| None);
    let selected_selector: Signal<rustc_hash::FxHashMap<usize, Signal<bool>>> = use_signal(Default::default);
    dioxus_signals::use_effect(move || {
        let currently_selected = selected.value();
        let selected_selector = selected_selector.read();
        let mut prev_selected = prev_selected.write();
        {
            let prev_selected = *prev_selected;
            if let Some(prev_selected) = prev_selected {
                if let Some(is_selected) = selected_selector.get(&prev_selected) {
                    is_selected.set(false);
                }
            }
        }
        if let Some(currently_selected) = currently_selected {
            if let Some(is_selected) = selected_selector.get(&currently_selected) {
                is_selected.set(true);
            }
        }
        *prev_selected = currently_selected;
    });

    render! {
        div { class: "container",
            div { class: "jumbotron",
                div { class: "row",
                    div { class: "col-md-6", h1 { "Dioxus" } }
                    div { class: "col-md-6",
                        div { class: "row",
                            ActionButton { name: "Create 1,000 rows", id: "run",
                                onclick: move |_| labels_container.write().overwrite(1_000),
                            }
                            ActionButton { name: "Create 10,000 rows", id: "runlots",
                                onclick: move |_| labels_container.write().overwrite(10_000),
                            }
                            ActionButton { name: "Append 1,000 rows", id: "add",
                                onclick: move |_| labels_container.write().append(1_000),
                            }
                            ActionButton { name: "Update every 10th row", id: "update",
                                onclick: move |_| {
                                    let labels = labels_container();
                                    for i in 0..(labels.labels.len()/10) {
                                        *labels.labels[i*10].label.write() += " !!!";
                                    }
                                },
                            }
                            ActionButton { name: "Clear", id: "clear",
                                onclick: move |_| labels_container.write().overwrite(0),
                            }
                            ActionButton { name: "Swap rows", id: "swaprows",
                                onclick: move |_| labels_container.write().swap(1, 998),
                            }
                        }
                    }
                }
            }

            table { class: "table table-hover table-striped test-data",
                tbody { id: "tbody",
                    labels_container().labels.iter().map(|item| {
                        render! {
                            Row {
                                label: item.clone(),
                                labels: labels_container.clone(),
                                selected_row: selected.clone(),
                                is_in_danger: {
                                    let read_selected_selector = selected_selector.read();
                                    match read_selected_selector.get(&item.key) {
                                        Some(is_selected) => *is_selected,
                                        None => {
                                            drop(read_selected_selector);
                                            let mut selected_selector = selected_selector.write();
                                            let is_selected = Signal::new(false);
                                            selected_selector.insert(item.key, is_selected);
                                            is_selected
                                        }
                                    }
                                },
                                key: "{item.key}"
                            }
                        }
                    })
                }
            }

            span { class: "preloadicon glyphicon glyphicon-remove", aria_hidden: "true" }
        }
    }
}

#[derive(Copy, Clone, Props)]
struct RowProps {
    label: Label,
    labels: Signal<LabelsContainer>,
    selected_row: Signal<Option<usize>>,
    is_in_danger: Signal<bool>
}

impl PartialEq for RowProps {
    fn eq(&self, other: &Self) -> bool {
        self.label == other.label
    }
}

fn Row(props: RowProps) -> Element {
    let RowProps {
        label,
        labels,
        selected_row,
        is_in_danger
    } = props;
    let is_in_danger = if is_in_danger.value() {
        "danger"
    } else {
        ""
    };

    render! {
        tr { class: "{is_in_danger}",
            td { class:"col-md-1", "{label.key}" }
            td { class:"col-md-4", onclick: move |_| {
                    selected_row.set(Some(label.key))
                },
                a { class: "lbl", "{label.label}" }
            }
            td { class: "col-md-1",
                a { class: "remove", onclick: move |_| labels.write().remove(label.key),
                    span { class: "glyphicon glyphicon-remove remove", aria_hidden: "true" }
                }
            }
            td { class: "col-md-6" }
        }
    }
}


#[derive(Clone, Props, PartialEq)]
struct ActionButtonProps {
    name: &'static str,
    id: &'static str,
    onclick: EventHandler,
}

fn ActionButton(
    ActionButtonProps {
        name,
        id,
        onclick,
    }: ActionButtonProps,
) -> Element {
    render! {
        div {
            class: "col-sm-6 smallpad",
            button {
                class:"btn btn-primary btn-block",
                r#type: "button",
                id: id.to_string(),
                onclick: move |_| onclick.call(()),
                *name
            }
        }
    }
}

static ADJECTIVES: &[&str] = &[
    "pretty",
    "large",
    "big",
    "small",
    "tall",
    "short",
    "long",
    "handsome",
    "plain",
    "quaint",
    "clean",
    "elegant",
    "easy",
    "angry",
    "crazy",
    "helpful",
    "mushy",
    "odd",
    "unsightly",
    "adorable",
    "important",
    "inexpensive",
    "cheap",
    "expensive",
    "fancy",
];

static COLOURS: &[&str] = &[
    "red", "yellow", "blue", "green", "pink", "brown", "purple", "brown", "white", "black",
    "orange",
];

static NOUNS: &[&str] = &[
    "table", "chair", "house", "bbq", "desk", "car", "pony", "cookie", "sandwich", "burger",
    "pizza", "mouse", "keyboard",
];
```

jkelleyrtp:
Status update:

We're getting really really close. Most examples compile now, but there are bugs we're finding that are getting stomped out bit by bit.

Currently the main challenges involve:
- Signals made at prop boundaries need to be properly scoped
- A number of APIs rely on clone-behavior with drop semantics. These are being transitioned to `Copy` types.
- Scoping per-runtime might have some issues - the tests here aren't great.
- For desktop, we don't always set the runtime guard, causing some callbacks to not work properly. This needs fixing.
- We need an opt-in to manual dependencies for a few hooks
- We need to accumulate dependencies on the future spawn by use_resource. This is somewhat complex.
- Lints will be critical 
- We should have infinite loop detection - either at compile time or runtime
- Documentation hasn't been updated yet for the new model
- Some code is just ugly/undocumented/highly generic or produces weird errors. Dogfooding fixes this.
- Sometimes effects run incorrectly, leading to panics
- The lock types on signal need to properly consume the borrowed lifetime so read/write in the same scope don't panic.
- The launch function has been tricky to get right with Send/Sync bounds but I think it's done now.

Generally, it's a balance between clean code, simple code, and code that gets the job done.

packages/hooks/src/use_on_destroy.rs
```diff
@@ -1,85 +1,10 @@
...
-struct LifeCycle<D: FnOnce()> {
-    ondestroy: Option<D>,
-}
-
-impl<D: FnOnce()> Drop for LifeCycle<D> {
-    fn drop(&mut self) {
-        let f = self.ondestroy.take().unwrap();
-        f();
-    }
+pub fn use_on_unmount<D: FnOnce() + 'static>(destroy: D) {
```
marc2332, line 8: What's the recommended way now? `use_drop` or `use_on_unmount` ? If you ask me, I like `use_on_unmount` more as it is more "semantic"

marc2332: or perhaps `use_on_unmount` could just be renamed to `use_on_destroy`

packages/signals/src/signal.rs
```diff
@@ -104,44 +166,49 @@ pub(crate) struct SignalData<T> {
...
+pub struct Signal<T: 'static, S: Storage<SignalData<T>> = UnsyncStorage> {
+    pub(crate) inner: CopyValue<SignalData<T>, S>,
 }
 
+/// A signal that can safely shared between threads.
+pub type SyncSignal<T> = Signal<T, SyncStorage>;
+
 #[cfg(feature = "serde")]
-impl<T: serde::Serialize + 'static> serde::Serialize for Signal<T> {
+impl<T: serde::Serialize + 'static, S: Storage<SignalData<T>>> serde::Serialize for Signal<T, S> {
```
atty303, line 202: This will generate the following error.

```
error[E0403]: the name `S` is already used for a generic parameter in this item's generic parameters
   --> C:\Users\atty\.cargo\git\checkouts\dioxus-1e619ce595d3799d\87dc409\packages\signals\src\signal.rs:203:18
    |
202 | impl<T: serde::Serialize + 'static, S: Storage<SignalData<T>>> serde::Serialize for Signal<T, S> {
    |                                     - first use of `S`
203 |     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
    |                  ^ already used
```

ealmloff: Fixed by https://github.com/DioxusLabs/dioxus/pull/1791/commits/55b37cc381f832ee842e3d0a5c177db9b2e19498

marc2332:
Amazing work @ealmloff @jkelleyrtp ! 👏 🧬

PR #1792, opened 2024-01-06
Add use_stream hook

### LINKED ISSUE #433: Add `use_stream` hook
## Specific Demand

Like `use_future`, except instead of taking a `Future<Output = T>`, takes a `futures::Stream<Item = T>`

itytophile:
closes #433 

The code is from `use_future` but it has been tuned for `Stream`. The provided example works great. I am not too sure about the macros but they work fine.

The following example shows that the stream is reset correctly after a dependency change:
```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;
use futures_util::{future, stream, StreamExt};
use std::time::Duration;

fn main() {
    dioxus_web::launch(App);
}

#[component]
fn Clock(cx: Scope, prefix: &'static str) -> Element {
    let count = use_state(cx, || 0);
    let make_stream = |_| {
        stream::once(future::ready(0)).chain(stream::iter(1..).then(|second| async move {
            gloo_timers::future::sleep(Duration::from_secs(1)).await;
            second
        }))
    };
    let on_item = {
        let count = count.clone();
        move |second| count.set(second)
    };
    use_stream(cx, prefix, make_stream, on_item);
    cx.render(rsx! {"{prefix} seconds: {count}"})
}

    let prefix = use_state(cx, || false);
    cx.render(rsx! {
        button { onclick: |_| { prefix.modify(|prefix| !*prefix) }, "Change prefix" }
        div { Clock { prefix: if *prefix.get() { "wouf" } else { "miaou" } } }
    })
}
```

[Capture vidéo du 2024-01-06 01-58-25.webm](https://github.com/DioxusLabs/dioxus/assets/45497473/e1cbf7a0-e37c-4078-ae48-d712e46356a2)

### DIFF
--- a/packages/hooks/Cargo.toml
+++ b/packages/hooks/Cargo.toml
@@ -20,9 +20,9 @@ tracing = { workspace = true }
 thiserror = { workspace = true }
 slab = { workspace = true }
 dioxus-debug-cell = "0.1.1"
+futures-util = { workspace = true, default-features = false }
 
 [dev-dependencies]
-futures-util = { workspace = true, default-features = false }
 dioxus-core = { workspace = true }
 dioxus = { workspace = true }
 web-sys = { version = "0.3.64", features = ["Document", "Window", "Element"] }
--- a/packages/hooks/src/lib.rs
+++ b/packages/hooks/src/lib.rs
@@ -81,6 +81,9 @@ pub use use_coroutine::*;
 mod use_future;
 pub use use_future::*;
 
+mod use_stream;
+pub use use_stream::*;
+
 mod use_effect;
 pub use use_effect::*;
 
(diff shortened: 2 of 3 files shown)


jkelleyrtp:
Hey! Thanks for the PR.

However, I'm not sure if we need a use_stream hook.

The design space for working with async is rather large and I don't think we could do a good job of covering all the various primitives.

Just playing with the pieces here, I think it's easy enough to work with streams using use_future/use_coroutine.

Screenshot 2024-01-08 at 11 15 40 AM

itytophile:
Well, I agree with you. I think I got confused while working with stream and dioxus and I stumbled upon #433.
In my opinion your solution covers my `use_stream` implementation. I don't know how I haven't thought of this.

Anyway, thanks for your work on Dioxus!

PR #1797, opened 2024-01-08
Clean up examples, documentation, simplify a variety of structs/methods, and more general polish

jkelleyrtp:
Just going through a number of APIs and simplifying them where possible. We've accumulated quite a bit of complexity in certain places, so this PR is taken an opinionated approach to simplifying and polishing these APIs. Will be breaking in quite a few ways, but should result in easier to navigate APIs.


jkelleyrtp:
Gonna close this since the conflicts with #1791 will be too big to justify this in its current state

PR #1810, opened 2024-01-11
Disambiguate if expressions in rsx by requiring curlies, allow shorthand component/element initialization

jkelleyrtp:
This PR changes how rsx interprets expressions, requiring them to be wrapped in curly braces.

This is a hugely breaking PR but allows a number of goodies like letting us keep consistency of if/for special cases and shorthand struct initialization.

It could also lead the way to improved autocomplete via partial expansion, however this is not yet implemented.

In it's current form it provides *way* better autocomplete when working inside curly braces and better autocomplete of attributes and props.

So, this works now:
```rust
if cond {
    div {}
} else {
   h1 {}
}
```

And so does this:
```rust
let a = 123;
let b = 456;
let c = 789;
let children = render!{ "hi" };

rsx! {
    Component { a, b, c, children }
}
```

This also supports elements:

```rust
let class = "123";

rsx! {
    div { class } 
}
```

However we don't special-case children and currently panic if an attribute named "children" is present.

In theory we could accept the `children` argument in elements, but I just didn't get around to it here.

### DIFF
--- a/packages/rsx/src/element.rs
+++ b/packages/rsx/src/element.rs
@@ -8,6 +8,7 @@ use syn::{
     parse::{Parse, ParseBuffer, ParseStream},
     punctuated::Punctuated,
     spanned::Spanned,
+    token::Brace,
     Expr, Ident, LitStr, Result, Token,
 };
 
@@ -59,6 +60,7 @@ impl Parse for Element {
             }
 
             // Parse the raw literal fields
+            // "def": 456,
             if content.peek(LitStr) && content.peek2(Token![:]) && !content.peek3(Token![:]) {
                 let name = content.parse::<LitStr>()?;
                 let ident = name.clone();
@@ -84,6 +86,8 @@ impl Parse for Element {
                 continue;
             }
 
+            // Parse
+            // abc: 123,
             if content.peek(Ident) && content.peek2(Token![:]) && !content.peek3(Token![:]) {
                 let name = content.parse::<Ident>()?;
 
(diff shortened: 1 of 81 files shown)


packages/rsx/src/element.rs
```diff
@@ -130,38 +129,61 @@ impl Parse for Element {
...
+                && !content.peek2(Brace)
+                && !content.peek2(Token![:])
+                && !content.peek2(Token![-])
+            {
+                let name = content.parse::<Ident>()?;
+                let name_ = name.clone();
+
+                // If the shorthand field is children, these are actually children!
+                if name == "children" {
+                    panic!("children must be wrapped in braces");
```
ealmloff, line 143: This should probably return a syn error instead of panicing so that it highlights the right part of the rsx

packages/rsx/src/node.rs
```diff
@@ -112,7 +112,27 @@ impl Parse for BodyNode {
...
         }
 
-        Ok(BodyNode::RawExpr(stream.parse::<Expr>()?))
+        // Match statements are special but have no special arm syntax
+        // we could allow arm syntax if we wanted
+        //
+        // ```
+        // match {
+        //  val => div {}
+        //  other_val => div {}
```
ealmloff, line 121: I think allowing you to do normal rsx in match arms would be more consistant with if statements and loops in the future, but it doesn't need to be in this PR

ealmloff (approved):
Looks good! Excited for better autocomplete

PR #1819, opened 2024-01-11
Fix muda segfault

jkelleyrtp:
Fixes the issue that https://github.com/DioxusLabs/dioxus/pull/1816 papers over.

We still technically need a way of allowing menubars to be modified, etc.

### DIFF
--- a/packages/desktop/src/webview.rs
+++ b/packages/desktop/src/webview.rs
@@ -11,7 +11,7 @@ use crate::{
 use dioxus_core::VirtualDom;
 use dioxus_html::prelude::EvalProvider;
 use futures_util::{pin_mut, FutureExt};
-use std::{rc::Rc, task::Waker};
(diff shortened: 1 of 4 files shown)


packages/desktop/src/webview.rs
```diff
@@ -22,16 +22,19 @@ pub struct WebviewInstance {
     // Wry assumes the webcontext is alive for the lifetime of the webview.
     // We need to keep the webcontext alive, otherwise the webview will crash
     _web_context: WebContext,
+
+    // Same with the menu.
+    // Currently it's a box<dyn any> because 1) we don't touch it and 2) we support a number of platforms
```
ealmloff, line 27: We could change this to a type alias exported from `impl_` to avoid the any once we need access to the menu to support custom menus defined in components/hooks

ealmloff (approved):

Issue #1822, opened 2024-01-12
Form submit catching stopped working


serzhiio:
Here is a part of my code generating web form:
```
removed
```
i marked the part of code which generates the error, if i comment it out everything is fine, this code was working for many months :(
On actual submit i receive error: 
`wasm-bindgen: imported JS function that was not marked as ``catch`` threw an error: target is not defined`

serzhiio:
Any specific reason for this?

ealmloff:
> Any specific reason for this?

Can you create an isolated example that reproduces this issue (without `DNS_CFG`, or a coroutine from the context)? There have been several recent changes to the event system including changing how events bubble and changing events from serialized objects to a trait.

serzhiio:
main.rs
```
use dioxus::prelude::*;
use log::LevelFilter;

fn main() {
    // Init debug
    dioxus_logger::init(LevelFilter::Info).expect("failed to init logger");
    console_error_panic_hook::set_once();

    log::info!("starting app");
    dioxus_web::launch(app);
}

fn app(cx: Scope) -> Element {

    let fields = (0..10).into_iter().map(|id| rsx!(
        input{
            r#type: "input",
            checked: false,
            name: "input-{id}",
            value: "dyn"
        }
    ));

    cx.render(rsx! {
        div {
            h1 { "Form" }
            form {
                onsubmit: move |ev| log::info!("Submitted {:?}", ev.data.values()),
                oninput: move |ev| log::info!("Input {:?}", ev.data.values()),
                {fields}
                button { r#type: "submit", value: "Submit", "Submit the form" }
            }
        }
    })
}
```
Cargo.toml:
```
[package]
name = "submit-issue"
version = "0.1.0"
edition = "2021"

[dependencies]

dioxus-web = { git="https://github.com/DioxusLabs/dioxus" }
dioxus = { git="https://github.com/DioxusLabs/dioxus" }

# Debug
log = "0.4.19"
dioxus-logger = "0.4.1"

console_error_panic_hook = "0.1.7"
```
P.S.: Current crates.io version 0.4.3 is perfectly fine

serzhiio:
If to change `fields` iter like this, values does not aggregate several input values into vec (i mean `input.0=[Text("txt0"),Text("s0")]`):
```
    let fields = (0..10).into_iter().map(|id| rsx!(
        div{
            input{
                r#type: "input",
                checked: false,
                name: "input.{id}",
                value: "txt{id}"
            }
[...]

PR #1833, opened 2024-01-14
Fix the web form event

ealmloff:
There were some merged conflicts with the new event system that broke form events. This PR fixes that issue.

Closes https://github.com/DioxusLabs/dioxus/issues/1822#issuecomment-1890994140


serzhiio:
https://github.com/DioxusLabs/dioxus/issues/1822#issuecomment-1891641856

PR #1836, opened 2024-01-15
Aggregating input values into arrays if required

serzhiio:


packages/web/src/event.rs
```diff
@@ -398,24 +416,18 @@ impl HasFormData for WebFormData {
...
                         if let Ok(item_values) = array.get(1).dyn_into::<Array>() {
-                            let item_values: Vec<String> =
-                                item_values.iter().filter_map(|v| v.as_string()).collect();
-
-                            values.insert(name, FormValue::VecText(item_values));
+                            item_values
+                                .iter()
+                                .filter_map(|v| v.as_string())
+                                .for_each(|v| insert_value(&mut values, name.clone(), v));
                         } else if let Ok(item_value) = array.get(1).dyn_into::<JsValue>() {
```
serzhiio, line 423: Not sure this branch is required, since all values come as an array

serzhiio:
In this PR we are able to aggregate several values of different inputs under same `name` attribute.
Also `FormValue` struct now implemets `.as_slice()` ans `.to_vec()` methods wich are useful if multiple select input has only one value and still it could be used as slice or Vec.

packages/web/src/event.rs
```diff
@@ -398,24 +416,18 @@ impl HasFormData for WebFormData {
...
-                            values.insert(name, FormValue::Text(item_value.as_string().unwrap()));
+                            insert_value(&mut values, name, item_value.as_string().unwrap());
                         }
                     }
                 }
             }
         }
 
-        // try to fill in select element values
```
ealmloff, line 413: This branch makes on input for select elements work.

serzhiio: In my case everything works without it

ealmloff: Select multiple elements is broken. Try this example:

```rust
use dioxus::prelude::*;

fn main() {
    dioxus_web::launch(app);
}

    let selected = use_state(cx,String::new);

    render! {
        select {
            multiple: true,
            oninput: |e| {
                selected.set(format!("{e:#?}"));
            },
            option { "Hello1" }
            option { "Hello2" }
            option { "Hello3" }
            option { "Hello4" }
        }

        pre { "{selected}" }
    }
}
```

https://github.com/DioxusLabs/dioxus/assets/66571940/5f8dd4ac-52e1-4d48-a477-f14e2e6283ca

ealmloff (approved):
Thank you! Accessing the values as a slice is much nicer

PR #1837, opened 2024-01-15
Minor fixes in cli

hem1t:
- restructured match block to not load bin when not required (at `main.rs`)
- removed an unnecessary line? although it might have been there readability (at `logging.rs`)
- there was difference in pattern of if block at same kind of code, so changed it all to `if let` pattern (at `serve.rs`)

If these are ok changes, then I may make more in this same PR?

### DIFF
--- a/packages/cli/src/main.rs
+++ b/packages/cli/src/main.rs
@@ -79,39 +79,37 @@ async fn main() -> anyhow::Result<()> {
             Ok(())
         }
         action => {
-            let bin = get_bin(args.bin);
-
-            if let Ok(bin) = &bin {
-                let _dioxus_config = DioxusConfig::load(Some(bin.clone()))
+            let bin = get_bin(args.bin)?;
+            let _dioxus_config = DioxusConfig::load(Some(bin.clone()))
                        .map_err(|e| anyhow!("Failed to load Dioxus config because: {e}"))?
                        .unwrap_or_else(|| {
                            log::info!("You appear to be creating a Dioxus project from scratch; we will use the default config");
                            DioxusConfig::default()
                     });
 
-                #[cfg(feature = "plugin")]
-                PluginManager::init(_dioxus_config.plugin)
-                    .map_err(|e| anyhow!("🚫 Plugin system initialization failed: {e}"))?;
-            }
+            #[cfg(feature = "plugin")]
+            PluginManager::init(_dioxus_config.plugin)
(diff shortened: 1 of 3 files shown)


packages/cli/src/main.rs
```diff
@@ -110,6 +78,41 @@ async fn main() -> anyhow::Result<()> {
 
             Ok(())
         }
-        _ => Err(anyhow::anyhow!(bin.unwrap_err())),
+        action => {
+            let bin = get_bin(args.bin);
```
ealmloff, line 82: We can use `get_bin(args.bin)?` now that all of the commands in this branch need bin

hem1t:
```rust
        action => {
            let bin = get_bin(args.bin)?;
            let _dioxus_config = DioxusConfig::load(Some(bin.clone()))
                       .map_err(|e| anyhow!("Failed to load Dioxus config because: {e}"))?
                       .unwrap_or_else(|| {
                           log::info!("You appear to be creating a Dioxus project from scratch; we will use the default config");
                           DioxusConfig::default()
                    });

            #[cfg(feature = "plugin")]
            PluginManager::init(_dioxus_config.plugin)
                .map_err(|e| anyhow!("🚫 Plugin system initialization failed: {e}"))?;

            match action {
                Build(opts) => opts
                    .build(Some(bin.clone()), None)
                    .map_err(|e| anyhow!("🚫 Building project failed: {}", e)),

                Clean(opts) => opts
                    .clean(Some(bin.clone()))
                    .map_err(|e| anyhow!("🚫 Cleaning project failed: {}", e)),

                Serve(opts) => opts
                    .serve(Some(bin.clone()))
                    .await
                    .map_err(|e| anyhow!("🚫 Serving project failed: {}", e)),

                Bundle(opts) => opts
                    .bundle(Some(bin.clone()))
                    .map_err(|e| anyhow!("🚫 Bundling project failed: {}", e)),

                _ => Err(anyhow::anyhow!(bin.unwrap_err())),
            }
```

Like this? although Idk what last arm will return, since there may exist no error till there, So, should I make it `Ok(())`

ealmloff: Yes, that looks much nicer! The `_ =>` branch can be `unreachable!()`

hem1t:
There is no `init` option on dx should I create one?

ealmloff:
> There is no `init` option on dx should I create one?

That would be nice!

ealmloff (approved):
Thank you, this looks much cleaner!

Issue #1778, opened 2024-01-02
MathML support


stefnotch:
## Specific Demand

[MathML Core](https://caniuse.com/mathml) is implemented in every major browser, and it adds a HTML element, and child elements. (Similar to SVG)

https://developer.mozilla.org/en-US/docs/Web/MathML/Authoring#mathml_in_html_pages

However, Dioxus does not seem to expose it

```rust
fn MathElement(cx: Scope) -> Element {
    cx.render(rsx! {
        math {
            "Hello, world!"
        }
    })
}
```

## Implement Suggestion

Would it be possible to get support for it?

Jah-On:
I second this feature request.

ealmloff:
Adding support for MathML elements would be great! It is a pretty easy part of the codebase to get started with. You just need to add each element and attribute in this format to the end of [this list](https://github.com/DioxusLabs/dioxus/blob/master/packages/html/src/elements.rs#L1684):
```rust
// elementname (with no underscores) namespace
math "http://www.w3.org/1998/Math/MathML" {
    // rust name of the attribute | unit (this is currently ignored, you can use String as a default) | the HTML attribute name (this can be DEFAULT for a stringified name or a custom string like "display2")
    display: String DEFAULT,
};
```

Ameyanagi:
I am interested in adding this feature.

stefnotch:
@Ameyanagi That'd be awesome! If you do find the time to make a pull request which adds this feature, please do ping me. I'll be more than happy to review it. (Even if I know very little about the dioxus code base)

Jah-On:
> @Ameyanagi That'd be awesome! If you do find the time to make a pull request which adds this feature, please do ping me. I'll be more than happy to review it. (Even if I know very little about the dioxus code base)

Same for me!

Ameyanagi:
@stefnotch @Jah-On 
I will do!

Ameyanagi:
@ealmloff 
Hi, I am working on this, and I implemented most of the MathML core tags, except for <annotation-xml>
https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics

In my understanding, we cannot use "-" or "_" for the rsx tag names, and are there any naming conventions that should be applied here?

I was initially thinking something like the following, but I saw in your comment that underscore is not allowed.

```
/// Build a
/// [`<annotation-xml>`](https://w3c.github.io/mathml-core/#dfn-annotation-xml)
/// element.
[...]

PR #1846, opened 2024-01-18
Adding MathML Core support to rsx

Ameyanagi:
This PR adds a [MathML Core](https://w3c.github.io/mathml-core/#mathml-elements-and-attributes) support to the rsx as discussed in issue https://github.com/DioxusLabs/dioxus/issues/1778.

Deprecated tags and attributes are not included in the code base.
annotation-xml is aliased to annotationXml due to the naming restrictions of rsx macro.

@stefnotch @Jah-On

### DIFF
--- a/packages/html/src/elements.rs
+++ b/packages/html/src/elements.rs
@@ -1744,4 +1744,196 @@ builder_constructors! {
     r#use ["use", "http://www.w3.org/2000/svg"] {
         href: String DEFAULT,
     };
+
+    // MathML elements
+
+    /// Build a
+    /// [`<annotation>`](https://w3c.github.io/mathml-core/#dfn-annotation)
+    /// element.
+    annotation "http://www.w3.org/1998/Math/MathML" {};
+
+    /// Build a
+    /// [`<annotation-xml>`](https://w3c.github.io/mathml-core/#dfn-annotation-xml)
+    /// element.
+    annotation_xml ["annotation-xml", "http://www.w3.org/1998/Math/MathML"] {};
+
+    /// Build a
+    /// [`<merror>`](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/merror)
+    /// element.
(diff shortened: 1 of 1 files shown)


Jah-On (approved):
Looks good!

ealmloff (approved):
Thank you!

packages/html/src/elements.rs
```diff
@@ -1744,4 +1744,199 @@ builder_constructors! {
...
+    munderover "http://www.w3.org/1998/Math/MathML" {
+        accent: Bool DEFAULT,
+        accentunder: Bool DEFAULT,
+    };
+
+    /// Build a
+    /// [`<semantics>`](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics)
+    /// element.
+    semantics "http://www.w3.org/1998/Math/MathML" {
+        encoding: String DEFAULT,
```
stefnotch, line 1940: The semantics element doesn't accept an encoding

stefnotch:
Will we need a MathML equivalent of https://github.com/DioxusLabs/dioxus/blob/53727b344de9af5bcd160fb99703a09ec0da2da4/packages/html/src/global_attributes.rs#L1643 ? The MathML core spec specifies a few global attributes.

packages/html/src/elements.rs
ealmloff: It looks like it does:
- https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics#encoding
Screenshot 2024-01-19 at 8 01 12 AM

Ameyanagi:
@stefnotch @ealmloff 
I will try to add the global attributes.

packages/html/src/elements.rs
stefnotch: Huh, interesting. The MathML Core spec seems to not mention it https://w3c.github.io/mathml-core/#semantics-and-presentation

PR #1851, opened 2024-01-20
dioxus-web: Add `Config::rootelement`

### LINKED ISSUE #1848: web renderer: Allow passing a `web_sys::Element` to `launch_cfg`
## Specific Demand

To use the web renderer, we have to use `dioxus_web::launch` or `dioxus_web::launch_cfg`. While it is possible to specify the root element by its html id, it is not possible to directly specify a `web_sys::Element`. Dioxus will always use `window.document.get_element_by_id` to find the root element.

This makes it very hard to use more complex roots, i.e. roots in a popup window or in a shadow-root.

It should be possible to directly specify an element if we have one

## Implement Suggestion
`dioxus_web::Config` could be easily modified to support both rootnames and root elements. Since the implementation details for `Config` are private, it's no breaking change.

Only a new method would be added: `Config::rootelement()`.

I'd be happy to create a quick PR for this if there are no objections

mirkootter:
This PR affects dioxus-web and allows to specify a direct element as dom-root. Before, it was necessary to specify the element by ID. This is annoying, for example if the element was created dynamically.

It is now possible to directly specify the element:
```rust
let doc = web_sys::window().unwrap().document().unwrap();
let elem = doc.get_element_by_id("main").unwrap();

use dioxus_web::Config;
dioxus_web::launch_cfg(App, Config::new().rootelement(elem));
```
This is equivalent to
```rust
use dioxus_web::Config;
dioxus_web::launch_cfg(App, Config::new().rootname("main"));
```

The old approaches still work. This PR just adds the new method `Config::rootelement`. It is necessary for dynamically created elements, for example in shadow roots.

# Open problems
Unfortunately, it still does not work with iframes and popup windows, because of a nice JavaScript detail: JavaScript types like `Object`, `Element`, `Node`, ... are tied to a window. Therefore, JavaScript code like the following will print "false":

```javascript
let iframe = /* ... */;
let elem = iframe.content_document.get_element_by_id("main");
console.log(elem instanceOf Element); // false :(
```

This behaviour obviously transfers to WASM, i.e. the following code panics:

```rust
let iframe: web_sys::HtmlIFrameElement = /* ... */;
let document = iframe.content_document().unwrap(); // ok, but tied to another window
let body = document.body().unwrap(); // ok, HtmlElement
let body: web_sys::Element = body.dyn_into().unwrap(); // panics :(
```

The last line is necessary because we need to cast `HtmlElement` to `Element`. Unfortunately, this does not work, because rust does not support "Runtime types", for example types which are tied to specific windows which are created at runtime.

Note that the problem is only caused by the fact that `dyn_into` uses the `instanceOf` operator to check if the cast is ok. Just using `body.into()` works fine. I adjusted all places in `dioxus-web`, and it worked fine with both iframes and popup windows 🚀🚀🚀

However, this may be a little too risky for this PR. Maybe I'll add another after this one.

# Issues
Closes #1848

### DIFF
--- a/packages/web/src/cfg.rs
+++ b/packages/web/src/cfg.rs
@@ -10,7 +10,7 @@
 pub struct Config {
     #[cfg(feature = "hydrate")]
     pub(crate) hydrate: bool,
-    pub(crate) rootname: String,
+    pub(crate) root: ConfigRoot,
     pub(crate) cached_strings: Vec<String>,
     pub(crate) default_panic_hook: bool,
 }
@@ -20,7 +20,7 @@ impl Default for Config {
         Self {
             #[cfg(feature = "hydrate")]
             hydrate: false,
-            rootname: "main".to_string(),
+            root: ConfigRoot::RootName("main".to_string()),
             cached_strings: Vec::new(),
             default_panic_hook: true,
         }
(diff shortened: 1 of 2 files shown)


ealmloff:
> Note that the problem is only caused by the fact that dyn_into uses the instanceOf operator to check if the cast is ok. Just using body.into() works fine. I adjusted all places in dioxus-web, and it worked fine with both iframes and popup windows 🚀🚀🚀

That seems like a good solution, feel free to include it in this PR or another PR

ealmloff (approved):
Looks great, thank you!

PR #1855, opened 2024-01-20
Mark hot-reload flag incompatible with release

### LINKED ISSUE #1757: Add a note about incompatibility between `--hot-reload` and `--release` options
If `--hot-reload` doesn't work with `--release` option (according to @ealmloff, it doesn't work), then there is probably should be a note in either one of the options addressing this issue.

```
      --release
          Build in release mode [default: false]
      --hot-reload
          Build with hot reloading rsx (doesn't work in release mode) [default: false]
```

@ealmloff suggested a warning log instead (which I can agree with).

egegungordu:
- added a new group to `ConfigOptsServe` that conflicts with `release`
- added `hot_reload` to the group
- changed help comment of `hot_reload`

Using `--hot-reload` option with `--release` now shows:

```
error: the argument '--release' cannot be used with '--hot-reload'

Usage: dx.exe serve --release [-- <CARGO_ARGS>...]

For more information, try '--help'.
```

Also reflected on help:
```
--hot-reload
  Build with hot reloading rsx. Will not work with release builds. [default: false]
```

Instead of my approach, the `"release-incompatible"` group could also be refactored into a struct using the '#[group]' macro, moving the incompatible arguments inside and flattening them to the `ConfigOptsServe` struct.

closes #1757

### DIFF
--- a/packages/cli/src/cli/cfg.rs
+++ b/packages/cli/src/cli/cfg.rs
@@ -78,6 +78,7 @@ impl From<ConfigOptsServe> for ConfigOptsBuild {
 }
 
 #[derive(Clone, Debug, Default, Deserialize, Parser)]
+#[command(group = clap::ArgGroup::new("release-incompatible").multiple(true).conflicts_with("release"))]
 pub struct ConfigOptsServe {
     /// Port of dev server
     #[clap(long)]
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Looks great, thank you!

PR #1858, opened 2024-01-21
Fix attribute diff infinite loop

### LINKED ISSUE #1843: Props spreading causes an infinite loop
**Problem**

https://github.com/DioxusLabs/dioxus/blob/5b6f9d6b80ef06c83a55a5399d3ddcd7e29479df/packages/core/src/diff.rs#L144-L163

It appears that the iterator is not consumed when it enters this case.
I tried calling left_iter.next() right_iter.next() and the infinite loop was resolved.
Is this the right way to fix it?

**Environment:**
 - Dioxus version: master
 - Rust version: 1.75.0]
 - OS info: Windows
 - App platform: web

**Questionnaire**

- [x] I would like to fix and I have a solution, but I don't know how to test it.

egegungordu:
- attribute diff check now consumes both iterators when both exist
- added a test for attribute diffing

fixes  #1843

### DIFF
--- a/packages/core/src/diff.rs
+++ b/packages/core/src/diff.rs
@@ -145,25 +145,29 @@ impl<'b> VirtualDom {
                                 (Some(left), Some(right)) => {
                                     // check which name is greater
                                     match left.name.cmp(right.name) {
-                                        std::cmp::Ordering::Less => self.remove_attribute(
-                                            left.name,
-                                            left.namespace,
-                                            mounted_id,
-                                        ),
-                                        std::cmp::Ordering::Greater => self.write_attribute(
-                                            right_template,
-                                            right,
-                                            idx,
-                                            mounted_id,
-                                        ),
+                                        std::cmp::Ordering::Less => {
+                                            self.remove_attribute(
+                                                left.name,
+                                                left.namespace,
+                                                mounted_id,
+                                            );
+                                            left_iter.next();
+                                        }
+                                        std::cmp::Ordering::Greater => {
+                                            self.write_attribute(
+                                                right_template,
+                                                right,
+                                                idx,
+                                                mounted_id,
+                                            );
+                                            right_iter.next();
+                                        }
                                         std::cmp::Ordering::Equal => {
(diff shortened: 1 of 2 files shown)


packages/core/src/diff.rs
```diff
@@ -160,6 +160,10 @@ impl<'b> VirtualDom {
                                             self.diff_attribute(left, right, mounted_id)
                                         }
                                     }
+
+                                    // consume both
+                                    left_iter.next();
```
ealmloff, line 165: I don't think consuming both items is the right behavior for the greater or less branches. We should only consume the attribute that is less than the other attribute.

If we have the attributes:
old_attributes: [`aaaa: "world"`, `aaab: "world"`]
new_attributes: [`aaab: "world"`]

first iteration:
left = `aaaa: "world"`
right `aaab: "world"`
left < right, so we remove the attribute "aaaa: world"

If we move both iterators forward one item, then the remaining iterators would have the items:
old_attributes: [`aaab: "world"`]
new_attributes: []

In the second iteration, it would incorrectly try to create the attribute `aaab: "world"` instead of diffing it

egegungordu: Yep, overlooked that bit.

However, the approach of consuming the smaller value assumes the elements are sorted by name, right? Also, the Miri test fails due to a memory leak from `cx.bump().alloc()` (i think?).

Any suggestions for an alternative approach to allocate the attributes for the test?

ealmloff: > Yep, overlooked that bit.
> 
> However, the approach of consuming the smaller value assumes the elements are sorted by name, right? 

Yes, they should be sorted by name in the rsx macro

> Also, the Miri test fails due to a memory leak from `cx.bump().alloc()` (i think?).
> 
> Any suggestions for an alternative approach to allocate the attributes for the test?

If you allocate anything in the bump allocator, it will not be dropped. That is fine for values like numbers and simple structs, but boxes and vecs rely on the drop implementation so we need to avoid them. You can allocate individual slides with `cx.bump().allow([your slice])` instead of allocating a vec for each branch to fix the issue

PR #1863, opened 2024-01-23
Add `isComposing` support

### LINKED ISSUE #1857: Please support `isComposing` of `KeybaordEvent`.
## Specific Demand

I could not find this in doc, and it is useful for CJK users.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

egegungordu:
Added isComposing to the js interpreter and updated necessary implementations of keyboard logic.

Since crossterm doesn't seem to have a similar feature to isComposing from the KeyboardEvent web API, I've set the value to be false when translating a crossterm key event.

Also, looks like the code snippet in the docs [here](https://dioxuslabs.com/learn/0.4/cookbook/custom_renderer#event-loop) will need to reflect the changes. I can create a pr for that once this pr is resolved 👍 

closes #1857

### DIFF
--- a/packages/desktop/headless_tests/events.rs
+++ b/packages/desktop/headless_tests/events.rs
@@ -154,7 +154,7 @@ fn app(cx: Scope) -> Element {
         ctrlKey: false,
         metaKey: false,
         shiftKey: false,
-        isComposing: false,
+        isComposing: true,
         which: 65,
         bubbles: true,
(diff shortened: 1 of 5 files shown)


ealmloff (approved):
This looks good, thank you!

(waiting for a second review before merging because this is a breaking change)

PR #1868, opened 2024-01-25
fix: initialize linux wry webview

tirithen:
In the wry 0.32 examples the linux webviews are initialized in a different way to render properly using
tao::platform::unix::WindowExtUnix. Tested with Gnome 45.3 and Wayland.

The changes was taken from the wry example files like https://github.com/tauri-apps/wry/blob/7234b2b99a8efc59484c96b767f3be8b52127816/examples/simple.rs#L16

### DIFF
--- a/packages/desktop/src/webview.rs
+++ b/packages/desktop/src/webview.rs
@@ -91,7 +91,28 @@ impl WebviewInstance {
             }
         };
 
-        let mut webview = WebViewBuilder::new(&window)
+        #[cfg(any(
+            target_os = "windows",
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
We have had a few other people who have had issues on linux in the past few days so it is nice to get this fixed. Thank you!

tirithen:
Nice! Glad to have it working I have now started playing around with a small set of components now that it works, it's I nice project and I look forward to exploring its possibilities more both for Linux and Android. :-)

PR #1872, opened 2024-01-27
Added `dx create` options to skip user interaction

### LINKED ISSUE #1871: Add `dx create` options to skip user interaction
## Specific Demand

Right now, when creating a Dioxus project with `dx create` you can only specify the template that will be used or use the default one if the option is omitted. For more convenience and easier reproduction examples, a way to skip the user interaction would be great (using `echo | dx create` doesn't work, and it's more of a hack). So something like `--yes` to "answer yes" for all questions when creating a cargo project (to use the default values for all questions/variables).

## Implement Suggestion

The specifics of the implementation are discussed here: https://discord.com/channels/899851952891002890/1200469217884000288.

Andrew15-5:
This PR is being made with the help from @ealmloff in the https://discord.com/channels/899851952891002890/1200469217884000288.

Closes #1871.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1024,15 +1024,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "btoi"
-version = "0.4.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dd6407f73a9b8b6162d8a2ef999fe6afd7cc15902ebf42c5cd296addf17e0ad"
-dependencies = [
- "num-traits",
-]
-
-[[package]]
 name = "built"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Andrew15-5:
The CLI command that works is `dx create -n name -y -o platform=web -o router=true -o styling=Vanilla`.

Andrew15-5:
This is a blocking issue: #23. It needs to be resolved first to avoid breaking changes. Since you only can create cargo project interactively, the changes will not break anything, IMO.

Andrew15-5:
Now that default options are working, sub templates ware added to the template..._aaand_ there default sub template is not working: https://github.com/cargo-generate/cargo-generate/issues/1181. :/

But you can still manually select it. When that issue will be resolved, ideally, just updating the `cargo-generate` version should fix it in `dx`. So, not really a blocker.

ealmloff (approved):
Love the small diff, thank you!

PR #1873, opened 2024-01-27
Removed 1 CLI dependency by updating `cargo-generate`

Andrew15-5:
When building the `dx` CLI there is 1 less dependency (from 710 to 709). I thought this is pretty nice, because it's a double win.


ealmloff (approved):
Thank you!

PR #1874, opened 2024-01-27
feat: Reduce cli deps

marc2332:
Just reduces the deps from 740 to 738 on Windows


ealmloff (approved):
Thank you! It looked like the versions already matched, but I switched the future-utils version to the workspace so that it is easier to update in the future

PR #1882, opened 2024-01-30
Converted `out_dir` and `asset_dir` fields to methods in `CrateConfig` (cli-config refactor)

Andrew15-5:
`out_dir` and `asset_dir` are now methods, because they derive from `crate_dir` and `dioxus_config`. As a result a few other files had to be updated as well.


ealmloff (approved):
Nice to see the CLI getting refactored, Thank you!

PR #1898, opened 2024-02-04
Now fullstack client uses correct config via `dx build`

Andrew15-5:
The server already uses the correct config. Now the `./dist/binary` works as expected — all the buttons work.


ealmloff (approved):
good catch, thank you!

PR #1917, opened 2024-02-09
fix `HasFileData` for `WebDragData`

spookyvision:


spookyvision:
clippy isn't happy because now the `element` field of `WebDragData` looks unused. Shall I remove it?

ealmloff:
> clippy isn't happy because now the `element` field of `WebDragData` looks unused. Shall I remove it?

Yes, I don't think we use it outside of that implementation

ealmloff (approved):
Thank you! This fixes the [file_upload example](https://github.com/DioxusLabs/dioxus/blob/master/examples/file_upload.rs) on the web platform

PR #1923, opened 2024-02-11
feat(fullstack): Added Cloudflare adapter for Cloudflare Pages + Functions deployment

### LINKED ISSUE #1829: Cloudflare Pages / Workers integration
## Specific Demand

It's easy to build static sites and CSR with Dioxus and have it built ready to go on multiple platforms including Cloudflare Pages, however this is not easily possible with SSR, as Cloudflare uses workerd, a custom JS runtime, which also supports Rust. 

It'd be great if fullstack applications with Dioxus can run on CF pages / workers runtimes. Cloudflare is currently the largest runtime platform behind AWS Lambda, with it continuously growing thanks to its amazing pricing.

## Implement Suggestion

Perhaps the internal macros for SSR could also generate the necessary code to turn it with workers-rs.

Reference: https://github.com/cloudflare/workers-rs

atty303:
closes #1829

Demo: https://dioxus-fullstack-hello-world.pages.dev/

In order to run a server with Cloudflare Pages (Workers), the server must be built with WASM.
This PR makes changes around threads to build the server with WASM.

I can confirm that Server functions and SSR are working in cloudflare-workers-hello-world.

I am not familiar with fullstack as I have just started touching it. I think I have the following TODOs, is that correct?

- [x] SSR/Hydration support
- [x] Regression test for platforms other than WASM
   - Only `axum-hello-world` was confirmed to work. Others did not pass the compilation even with `master`.
- [x] README in example

## Future work

- Make IncrementalRenderer cache pluggable
   - Cloudflare Workers can uses KV
- ~~Hot Reload cannot be supported on Cloudflare Workers~~
   - Workers had no file system access, but supported WebSocket client/server and TCP client. So it seems to me that dx CLI could support Hot Reload by setting up a Hot Reload server and Workers could relay between the Web and the Hot Reload server.
   - However, this is a lot of work and we would like to try another time.
- Documentation/Guide. English is not my first language so writing documentation is tough. (This text also relies on DeepL)


packages/fullstack/examples/cloudflare-workers-hello-world/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+    console_error_panic_hook::set_once();
+
+    let fmt_layer = tracing_subscriber::fmt::layer()
+        .with_ansi(true)
+        .without_time()
+        .with_writer(MakeConsoleWriter);
+    tracing_subscriber::registry().with(fmt_layer).init();
+
+    GetServerData::register_explicit().unwrap();
+    PostServerData::register_explicit().unwrap()
```
atty303, line 17: You must explicitly register Server Functions.

packages/fullstack/examples/cloudflare-workers-hello-world/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+    env: worker::Env,
+    _ctx: worker::Context,
+) -> worker::Result<worker::Response> {
+    handle_dioxus_application("/api/", req, env).await
+}
+
+pub fn app() -> Element {
+    let mut count = use_signal(|| 0);
+    let text = use_signal(|| "...".to_string());
+    // let server_future = use_server_future(get_server_data)?;
```
atty303, line 33: If this is commented out, the following loop will enter an infinite loop. I could not figure out how to resolve this.

https://github.com/DioxusLabs/dioxus/blob/a3f8d1cd2cfe086373108416886d9967aa4fde9d/packages/web/src/lib.rs#L155

packages/fullstack/src/render.rs
```diff
@@ -152,6 +153,122 @@ impl SsrRendererPool {
                 });
                 let (freshness, html) = rx.await.unwrap()?;
 
+                Ok((freshness, html))
+            }
+        }
+        #[cfg(target_arch = "wasm32")]
```
atty303, line 159: This code has not been tested with changes to pass compilation. Is there a better way to write it?

packages/fullstack/examples/cloudflare-workers-hello-world/src/lib.rs
ealmloff: #1926 fixes that issue

packages/ssr/src/incremental.rs
```diff
@@ -152,6 +152,7 @@ impl IncrementalRenderer {
             }
         }
         // check the file cache
+        #[cfg(not(target_arch = "wasm32"))]
```
ealmloff, line 155: It looks like cloudflare workers supports wasi, tokio has [experimental support](https://docs.rs/tokio/latest/tokio/index.html#unstable-wasm-support) for wasi, and wasi has experimental support for fs operations. Does cloudflare not support fs, or does tokio not support wasi fs?

packages/fullstack/src/render.rs
ealmloff: It looks like there is a lot of duplication between the wasm and not wasm implementations. Is it possible to pull out our own version of spawn_local that tries to use `get_local_pool().spawn_pinned` on most platforms but uses `spawn_local` directly on wasm?

packages/ssr/src/incremental.rs
atty303: Cloudflare appears to support WASI.

https://github.com/cloudflare/workers-wasi

However, I don't think actual Workers supports storage.

https://community.cloudflare.com/t/about-cloudflare-workers-filesystem/481642/2

packages/fullstack/examples/cloudflare-workers-hello-world/src/lib.rs
atty303: Thanks, I'm going to try to make it work.

atty303:
Snag_64cb8c0

SSR and hydration now works. However, local development and build workflows still need to be adjusted.

packages/cli/src/server/web/mod.rs
```diff
@@ -369,8 +369,8 @@ async fn start_server(
     #[cfg(feature = "plugin")]
     PluginManager::on_serve_start(_config)?;
 
-    // Parse address
-    let addr = format!("0.0.0.0:{}", port).parse().unwrap();
+    // Bind the server to `[::]` and it will LISTEN for both IPv4 and IPv6. (required IPv6 dual stack)
+    let addr = format!("[::]:{}", port).parse().unwrap();
```
atty303, line 373: Wrangler uses `localhost` when proxying static files to `dx serve`. `localhost` resolves to an IPv6 Lookback (`[::1]`) on Windows, but dx serve only LISTENS to IPv4, so the proxy fails.
So I would like to change it so that dx serve LISTENs to IPv6 dual stack.

atty303:
As for myself, I have solved all TODOs and am ready to merge. Please review.

packages/fullstack/src/adapters/mod.rs
```diff
@@ -65,6 +67,21 @@ impl ServerFnHandler {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+macro_rules! spawn_platform {
```
ealmloff, line 71: Can this be a function instead of a macro? Macros are a bit harder to follow and end up generating more code

packages/fullstack/src/render.rs
```diff
@@ -1,18 +1,38 @@
...
-use tokio::task::spawn_blocking;
 
 use crate::prelude::*;
-use dioxus_lib::prelude::*;
+use crate::render::dioxus_core::NoOpMutations;
+use crate::server_context::SERVER_CONTEXT;
+
+#[cfg(not(target_arch = "wasm32"))]
+macro_rules! spawn_platform {
+    ($f:expr) => {
```
ealmloff, line 21: I think this could be a function as well

packages/fullstack/src/adapters/mod.rs
atty303: I tried, but I could not describe it in a function, not in a macro. If you can rewrite it for me, you are welcome to do so.

packages/fullstack/src/render.rs
atty303: This one, too, could only be solved with macros to my knowledge.

atty303:
```rust
fn spawn_platform<Fut>(f: Fut) -> JoinHandle<Fut::Output>
where
    Fut: Future + 'static,
    Fut::Output: Send + 'static,
{
    spawn_blocking(move || {
        tokio::runtime::Runtime::new()
            .expect("couldn't spawn runtime")
            .block_on(f)
    })
    // spawn_local(f)
}
```

This implementation will result in the following error.

```
error[E0277]: `Fut` cannot be sent between threads safely
   --> packages\fullstack\src\render.rs:25:20
    |
25  |       spawn_blocking(move || {
    |       -------------- ^------
    |       |              |
    |  _____|______________within this `{closure@packages\fullstack\src\render.rs:25:20: 25:27}`
    | |     |
    | |     required by a bound introduced by this call
26  | |         tokio::runtime::Runtime::new()
27  | |             .expect("couldn't spawn runtime")
28  | |             .block_on(f)
29  | |     })
    | |_____^ `Fut` cannot be sent between threads safely
    |
note: required because it's used within this closure
   --> packages\fullstack\src\render.rs:25:20
    |
25  |     spawn_blocking(move || {
    |                    ^^^^^^^
note: required by a bound in `spawn_blocking`
   --> C:\Users\atty\.cargo\registry\src\index.crates.io-6f17d22bba15001f\tokio-1.36.0\src\task\blocking.rs:206:28
    |
204 |     pub fn spawn_blocking<F, R>(f: F) -> JoinHandle<R>
    |            -------------- required by a bound in this function
205 |     where
206 |         F: FnOnce() -> R + Send + 'static,
    |                            ^^^^ required by this bound in `spawn_blocking`
help: consider further restricting this bound
    |
22  |     Fut: Future + 'static + std::marker::Send,
    |                           +++++++++++++++++++
```

It seems to me that `f` must be written inline with the `block_on` argument since it is a `!Send`. I think a macro would be necessary.

packages/fullstack/src/adapters/mod.rs
atty303: This was solved without the use of macros.

packages/fullstack/src/render.rs
atty303: @ealmloff I understand that you were able to solve the problem with a closure. Thank you very much for the learning opportunity!

atty303:
Conflict is resolved.

atty303:
Hmmm, does this PR conflict with #1944? Personally, I think there is great value in having Dioxus Fullstack work with Cloudflare. But there may be other ways, such as integrating with Cloudflare via Axum. Either way, I would be happy to have Fullstack Server remain working with WASM.

ealmloff:
> Hmmm, does this PR conflict with #1944? Personally, I think there is great value in having Dioxus Fullstack work with Cloudflare. But there may be other ways, such as integrating with Cloudflare via Axum. Either way, I would be happy to have Fullstack Server remain working with WASM.

Yes, this will conflict with #1944. I think supporting Cloudflare via Axum is a better solution at least until fullstack is more stable. The WASM fixes in this PR are still very useful, but the cloudflare adapter might be better as a separate crate or in https://github.com/DioxusLabs/web-adapters.

atty303:
Okay, I'll create a PR that only supports Fullstack's WASM support, and I'll figure out a way to keep Cloudflare support out of this repository.

atty303:
This PR is closed as we have split the Cloudflare-independent fix to #1948.

PR #1933, opened 2024-02-14
Only subscribe scopes to signals when rendering

### LINKED ISSUE #1930: bug: Scopes get subscribed when signals are read inside event handlers
**Problem**

Scopes get subscribed when signals are read inside event handlers

```rs
fn app() -> Element {
    let mut show = use_signal(|| false);

    println!("Why am I running?");

    rsx!(
        rect {
            onclick: move |_| {
                let x = show.read();
                drop(x);
                *show.write() = true;
            },
            label { "Toggle" }
        }
    )
}
```

**Steps To Reproduce**
- Run it

**Expected behavior**

Should not print `Why am I running?` every time I click because there is no reader in component function, only in the event handler.

**Screenshots**

**Environment:**
 - Dioxus version: `master` 
 - Rust version: 1.75
 - OS info: Windows
 - App platform: any

ealmloff:
Closes #1930


ealmloff:
I restored the signal tests which are failing because of #1935

PR #1940, opened 2024-02-14
calling `eval` should not return an error - eval() should always be available

jkelleyrtp:
Change the behavior of `eval` to not require unwrapping the result. If no EvalProvider is found, this will inject a dummy evaluator that always returns NotSupported.

@ealmloff should we keep the Poll::pending eval implementation for SSR or just not include one?


ealmloff:
> @ealmloff should we keep the Poll::pending eval implementation for SSR or just not include one?

I think we can just use the NotSupported evaluator on SSR. Non-suspense futures shouldn't be running on the server and if you do try to use eval in a suspense future, then returning an error is better than waiting forever

ealmloff (approved):

jkelleyrtp:
Okay, going to hijack this PR to fix some cfg issues

PR #1948, opened 2024-02-16
feat(fullstack): support wasm target

atty303:
This is a PR from #1923 with only Fullstack's WASM support extracted.

In order to run a server with Cloudflare Pages Functions (Workers), the server must be built with WASM.
This PR makes changes around threads to build the server with WASM.


ealmloff (approved):
Looks great, thank you!

I swapped out web-time for chrono which also supports wasm because we use chrono in other crates in the workspace

Issue #1957, opened 2024-02-21
SVG global attribute "type" is wrong


fdgStilla:
**Problem**
Try to create a svg filter:
```rust
svg {
    filter {
        id: "gamma_filter",
        feComponentTransfer {
            feFuncR {
                r#type: "linear",
                slope: "0.5",
                intercept: "0.5"
            }
        }
    }
}
```
=> the resulting dom is 
```html
<svg><filter id="gamma_filter"><feComponentTransfer><feFuncR _type="linear" slope="0.5" intercept="0.5"></feFuncR></feComponentTransfer></filter></svg>
```
The filter does not work because `_type` attribute does not exist, it should be `type`

**Environment:**
 - Dioxus version: [0.4.3]
 - Rust version: [1.76.0]
 - OS info: [Windows 10]
 - App platform: [`desktop]

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

PR #1958, opened 2024-02-21
Fix name for SVG attribute "type"

fdgStilla:
Fix https://github.com/DioxusLabs/dioxus/issues/1957


ealmloff (approved):
Good catch, thanks!

PR #1960, opened 2024-02-21
autorelease workflow

jkelleyrtp:
Get the first prerelease of 0.5 out and set up an autorelease workflow that will enable us to ship faster.

- Adds backporting workflow
- Adds autorelease workflow for each branch
- Releases CLI, extension, and docs in one go
- Enables prereleases for cli and extension

### backporting

We now have the ability to backport fixes. This is done automatically unless the PR is marked "breaking" and thus will be excluded. This fits our normal workflow. We also do cargo-semverchecks to ensure non-breaking PRs are infact nonbreaking.

### auto-publish

We now automatically publish new versions of crates to crates.io using cargo-workspaces. This works for stable using stable bumps and nightly using the pre-release choice.

### releasing vscode extension

We now automatically release the vscode extension and bump its version number so its kept in sync with dioxus code. I have yet to implement prereleases here.


.github/workflows/merge.yml
```diff
@@ -0,0 +1,92 @@
...
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+
+jobs:
+  # Attempt to backport a merged pull request to the latest stable release
```
korthout, line 23: @jkelleyrtp It's great to see you're trying out backport-action 🙌 

By default, backport-action only backports pull requests with a label matching the regex `^backport ([^ ]+)$`. If you want to always backport to a specific branch instead of having to label pull requests, you can use the [`target_branches`](https://github.com/korthout/backport-action?tab=readme-ov-file#target_branches) input.

For example, if you want to always backport to the latest stable release branch, and this branch is called `latest-stable`, you can use:

```yml
uses: korthout/backport-action@v2
with:
  target_branches: 'latest-stable'
```

Issue #1972, opened 2024-02-24
webview should never automatically navigate


jkelleyrtp:
**Problem**

In a number of instances:
- [] form submission
- [] dropping image
- [] forward/backward keyboard shortucts

the webview will attempt to navigate.

We should block these - the webview has no reason to ever navigate.

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Issue #1963, opened 2024-02-22
futures get polled on desktop before window is loaded


jkelleyrtp:
**Problem**

We have occasional CI failures around sending events into the webviews because the window is taking a long time to load. This is noticeable more on the windows runners than anything else, but it does mean async code can hit a race condition with the load state of the webview.

We should wait for the webview to be loaded completely before polling any futures.

**Environment:**
 - Dioxus version: [e.g. v0.17, `master`]
 - Rust version: [e.g. 1.43.0, `nightly`]
 - OS info: [e.g. MacOS]
 - App platform: [e.g. `web`, `desktop`]

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [x] I don't have time to fix this right now, but maybe later

Issue #1950, opened 2024-02-17
use_server_future is not properly reactive


Nnamdi-sys:
**Problem**

I'm struggling with re-rendering  HTML using the dangerous_inner_html attribute.

The HTML String is being generated from a server function and called with use_server_future.

On the first launch the HTML renders correctly, but after that the HTML doesn't get rendered.

I have checked the console logs and I can see the HTML string is generated correctly.

**Expected behavior**

HTML Should be re-rendered on submit of the form.

**Environment:**
 - Dioxus version: 0.4.3
 - Rust version: 1.72.0
 - OS info: MacOS
 - App platform: fullstack
 
main.rs
 ```
#![allow(non_snake_case, unused)]

use dioxus::html::{button, input, label};
use dioxus::prelude::*;
use dioxus_fullstack::prelude::*;
use serde::{Deserialize, Serialize};

#[cfg(feature = "ssr")]
use finalytics::charts::portfolio::PortfolioCharts;
#[cfg(feature = "ssr")]
use finalytics::data::ticker::Interval;
#[cfg(feature = "ssr")]
use finalytics::analytics::optimization::ObjectiveFunction;
use log::log;


fn main() {
    #[cfg(feature = "web")]
    dioxus_web::launch_cfg(app, dioxus_web::Config::new().hydrate(true));
    #[cfg(feature = "web")]
    wasm_logger::init(wasm_logger::Config::default());
    #[cfg(feature = "ssr")]
    {
        tokio::runtime::Runtime::new()
            .unwrap()
            .block_on(async move {
                let addr = std::net::SocketAddr::from(([127, 0, 0, 1], 8080));
                axum::Server::bind(&addr)
                    .serve(
                        axum::Router::new()
                            .serve_dioxus_application("", ServeConfigBuilder::new(app, ()))
                            .into_make_service(),
                    )
                    .await
                    .unwrap();
            });
    }
}

fn app(cx: Scope) -> Element {
    cx.render(rsx! {
        h1 { "Finalytics" }
        div {
            PortfolioComponent {}
        }
    })
}


#[component]
pub fn PortfolioComponent(cx: Scope) -> Element {
    let symbols = use_state(cx, || "AAPL,MSFT,NVDA,BTC-USD".to_string());
    let benchmark_symbol = use_state(cx, || "^GSPC".to_string());
    let start_date = use_state(cx, || "2020-01-01".to_string());
    let end_date = use_state(cx, || "2023-12-31".to_string());
[...]

Issue #1922, opened 2024-02-09
oninput for forms isn't working (desktop)


jkelleyrtp:
**Problem**

onsubmit works as expected, but oninput is not working as expected (not returning all the values with the .values() method)

ealmloff:
This is also broken on the web renderer if the event originates from an element within the form instead of the form:
```rust
//! Forms
//!
//! Dioxus forms deviate slightly from html, automatically returning all named inputs
//! in the "values" field

use dioxus::prelude::*;
use std::collections::HashMap;

fn main() {
    launch(app);
}

fn app() -> Element {
    let mut values = use_signal(|| HashMap::new());
    rsx! {
        div {
            "{values:#?}"
            h1 { "Form" }
            form {
                oninput: move |ev| values.set(ev.values()),
                input {
                    r#type: "text",
                    name: "username",
                    oninput: move |ev| values.set(ev.values())
                }
                input { r#type: "text", name: "full-name" }
                input { r#type: "password", name: "password" }
                input { r#type: "radio", name: "color", value: "red" }
                input { r#type: "radio", name: "color", value: "blue" }
                button { r#type: "submit", value: "Submit", "Submit the form" }
            }
        }
    }
}
```

PR #1974, opened 2024-02-24
Fix form inputs, form submits navigating, file drop, multiple root elements

### LINKED ISSUE #1921: FileDrop is currently busted on master
**Problem**

Dragging an image and dropping it into the desktop renderer overwrites the renderer. We should just intercept all drag-and-drop events and forward them to the right element.
**Steps To Reproduce**

- Drag a file onto the desktop webview
- Webview is replaced with the image

**Expected behavior**
- We intercept the drag-drop

**Environment:**
 - Dioxus version: [`master`]
 - Rust version: [`1.76.0`]
 - OS info: [e.g. MacOS]
 - App platform: [`desktop`]

**Questionnaire**

jkelleyrtp:
Fix
#1921
#1922
#1972
#1963
#1907
#1826
#1950

- [x] fix RFD
- [x] Migrate to typescript + bun
- [x] FormInputs
- [x] Navigating
- [x] tests for all

Add test cases for each of these.
Adds typescript support for the interpreter.
Unifies binary protocol with its base host

This should clean up our interpreter code and fix a number of longstanding issues around eventhandling across the boundary.

We still don't have a great solution for liveview.

Wry let's us intercept navigations now, too, which should make accidental navigates impossible.


jkelleyrtp:
Going to leave synchronous events for a separate PR as it will need to touch core for event dispatching.

PR #1975, opened 2024-02-24
Fix Clippy warning in `Routable` macro

tigerros:
`Routable` creates types which derive `PartialEq`, but can also derive `Eq`. Clippy complains about this. This PR makes them derive `Eq` as well. If deriving only `PartialEq` is intentional let me know and I'll add an `#[allow]` instead.

### DIFF
--- a/packages/router-macro/src/lib.rs
+++ b/packages/router-macro/src/lib.rs
@@ -559,7 +559,8 @@ impl RouteEnum {
             #(#type_defs)*
 
             #[allow(non_camel_case_types)]
-            #[derive(Debug, PartialEq, Eq)]
+            #[allow(clippy::derive_partial_eq_without_eq)]
+            #[derive(Debug, PartialEq)]
             pub enum #match_error_name {
                 #(#error_variants),*
             }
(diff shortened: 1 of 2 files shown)


ealmloff (commented):
Part of the error types that the routable macro generates comes from the error type in different router traits including FromRouterSegments. Deriving Eq on the error enum requires each of those error types to implement Eq which causes this code to not compile:
```rust
#[rustfmt::skip]
#[derive(Routable, Clone, PartialEq)]
enum Route {
    #[route("/:id")]
    Blog { id: MyCustomType },
}

#[derive(Clone, Debug, PartialEq)]
struct MyCustomType;

impl std::fmt::Display for MyCustomType {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "My Custom Type")
    }
}

impl dioxus_router::routable::FromRouteSegment for MyCustomType {
    /// The error that can occur when parsing a route segment.
    type Err = f32;

    /// Create an instance of `Self` from a route segment.
    fn from_route_segment(route: &str) -> Result<Self, Self::Err> {
        if route == "my-custom-type" {
            Ok(Self)
        } else {
            Err(0.0)
        }
    }
}

#[component]
fn Blog(id: MyCustomType) -> Element {
    rsx! {
        h1 { "How to make: " }
        p { "{id}" }
    }
}
```

ealmloff (approved):
Thank you!

PR #1984, opened 2024-02-27
Add better documentation to use_future, use_resource, use_effect and use_context

agreyyy:
# better documentation for 0.5
I feel like the documentation for **use_future, use_resource, use_effect and use_context** was pretty lackluster, so I made a few changes. I added a little concise code snippet showing how to use each of these hooks in an app, so that the developer experience can be better when using them. Since these hooks are probably the most important/used, along with **use_signal** and **use_memo**, I think they deserve to be documented better with code examples and descriptions, like **use_signal**.


packages/hooks/src/use_future.rs
```diff
@@ -8,10 +8,34 @@ use dioxus_signals::*;
 use dioxus_signals::{Readable, Writable};
 use std::future::Future;
 
-/// A hook that allows you to spawn a future
-///
+/// A hook that allows you to spawn a future.
+/// This future will **not** run on the server
 /// The future is spawned on the next call to `flush_sync` which means that it will not run on the server.
 /// To run a future on the server, you should use `spawn` directly.
+/// `use_future` assumes your future will never complete - **it won't return a value**.
```
ealmloff, line 15: Use future can be used with both futures that stop or run forever. It doesn't require the future to run forever

packages/hooks/src/use_context.rs
```diff
@@ -22,6 +34,24 @@ pub fn use_context<T: 'static + Clone>() -> T {
...
 /// Once the context has been provided, it is immutable. Mutations should be done via interior mutability.
+/// Context can be read by any child components of the context provider, and is a solution to prop
+/// drilling, using a context provider with a Signal inside is a good way to provide global/shared
+/// state in your app:
+/// ```rust
+///fn app() -> Element {
+///    use_context_provider(|| Signal::new(0));
+///    rsx! { Child {} }
+///}
+// This component does read from the signal, so when the signal changes it will rerun
```
ealmloff, line 45: Comments are a bit odd within doc comments. I think this needs to be `/// // This...` if you want the comment to show up in the docs

ealmloff (commented):
Thanks for adding all these code examples! The reference includes a lot more detail about some hooks, but having some of that information inline in the docs is also very useful

Once we get the 0.5 docs merged, I would love to link different sections of the dioxus reference in the rust docs. (For example, we could link the [use_resource](https://ealmloff.github.io/docsite/learn/0.4/reference/use_resource/) guide at the end of the rust docs for the hook

packages/hooks/src/use_future.rs
agreyyy: Most of the text for these snippets was paraphrased from the examples folder, this one specifically from future.rs at  https://github.com/DioxusLabs/dioxus/blob/main/examples/eval.rs. I will change the comments inside that file too, so that they can both be correct.

ealmloff (approved):

Issue #1907, opened 2024-02-06
dioxus_web::launch_cfg cannot be called multiple times on different root elements.


JJJHolscher:
**Problem**

When trying to have multiple dioxus widgets on different pre-existing elements, the browser console will tell about the fact that `null has no property X`, where X can be `listeners` or `nodeType` when calling `dioxus_web::launch_cfg` for the second time using either a `Config::rootname` or `Config::rootelement` on an element that is not a child of first root element.

**Steps To Reproduce**

- Clone my [repo](https://github.com/JJJHolscher/jupyter-and-dioxus) and cd into it.
- Checkout to 3f5abc55ba1981a8fcb0e2d13915834f238ba0bb
- Make sure you have wasm-pack.
- `cmd/build`
- `python -m http.server 8080`
- `$BROWSER http://localhost:8080`

My repo works with custom elements, but my exposed `change` function can also be used to launch dioxus on an element given its id, and you will again notice that this function errors when dioxus was already launched, either because there was already a rendered custom element or because the function was called before on a different id.

**Expected behavior**

When pointing the browser to index.html, you only see a single group of 2 buttons + 1 header. I expect there to be two groups, since there are two custom elements in index.html. Instead, there is the aforementioned error in the console.

**Screenshots**

![image](https://github.com/DioxusLabs/dioxus/assets/25668596/8b1e50de-1764-4922-8aa5-d587ad2deb51)

**Environment:**
 - Dioxus version: commit dc5e4e3738ac80cb3e333a10340674b575c95abd (<2 weeks old)
 - Rust version: 1.76.0-nightly
 - OS info: ArchLinux
 - App platform: web

**Willingness to solve this myself**
I'm interested in fixing this myself but don't know where to start.
If there is a way to solve this without using dioxus_web, that's also fine by me. I just want multiple dioxus widgets without all those widgets necessarily having to share the same root element, since I'm using dioxus inside a pre-existing html that sometimes adds new elements itself (I want dioxus widgets inside a jupyter notebook's cell output). If my goal is doable with a single dioxus root, then that's _also_ fine (I did not try this yet), but my gut tells me dioxus won't play nice with external sources adding (complex) elements.

**Bug or Feature**
[...]

PR #1986, opened 2024-02-27
Bump the sledgehammer bindgen version

ealmloff:
The new version of sledgehammer bindgen uses a class instead of globals to work towards fixing https://github.com/DioxusLabs/dioxus/issues/1907

### DIFF
--- a/packages/interpreter/src/interpreter.js
+++ b/packages/interpreter/src/interpreter.js
@@ -1,12 +1,6 @@
-class InterpreterConfig {
-  constructor(intercept_link_redirects) {
-    this.intercept_link_redirects = intercept_link_redirects;
-  }
-}
-
 // this handler is only provided on the desktop and liveview implementations since this
 // method is not used by the web implementation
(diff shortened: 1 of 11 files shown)


jkelleyrtp:
This conflicts with #1974 so I'm going cherry pick some changes out of it.... Thank you for doing this!!

packages/interpreter/src/interpreter.js
```diff
@@ -35,7 +29,7 @@ async function handler(event, name, bubbles, config) {
             const href = a_element.getAttribute("href");
             if (href !== "" && href !== null && href !== undefined) {
               window.ipc.postMessage(
-                window.interpreter.serializeIpcMessage("browser_open", { href })
+                this.serializeIpcMessage("browser_open", { href })
```
jkelleyrtp, line 32: will `this` be bound to the right `this` when this function is called?

jkelleyrtp (approved):

PR #1993, opened 2024-02-29
Fix effect ordering and futures being run after scopes are dropped

ealmloff:
This PR fixes futures that are rerun after scopes are dropped. The following example is broken on the main branch, but working on this branch:

```rust
use dioxus::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {
    let mut vec = use_signal(|| vec![1, 2, 3]);

    let len = vec.len();

    rsx! {
        button {
            onclick: move |_| {
                let mut vec = vec.write();
                vec.push(len);
            },
            "Add"
        }
        button {
            onclick: move |_| {
                vec.pop();
            },
            "Remove"
        }
        for i in 0..len {
            Child {
                index: i,
                vec,
            }
        }
    }
}

#[component]
fn Child(index: usize, vec: Signal<Vec<usize>>) -> Element {
    let item = use_memo(move || vec.read()[index]);
    rsx! {
        div { "Item: {item}" }
    }
}
```


jkelleyrtp:
Note a huge huge fan of the complexity of combining tasks and scopes when pulling out work but it's something we can revisit - it's much more important that this functionality just works.

Thanks for taking this on. LGTM

PR #2002, opened 2024-03-04
Fix signals leaking memory

### LINKED ISSUE #1999: bug: Signals are leaking because their inner values are not being dropped.
**Problem**

When a signal's owner scope is dropped, the signal inner value isn't dropped, therefore it's leaking memory.

**Steps To Reproduce**

```rs

#[derive(Default)]
struct VecWrapper(Vec<f64>);

impl Drop for VecWrapper {
    fn drop(&mut self) {
        println!("Dropped VecWrapper");
    }
}

#[derive(Default, Clone)]
struct OkayWrapper(std::rc::Rc<Okay>);

#[derive(Default, Clone)]
struct Okay;

impl Drop for Okay {
    fn drop(&mut self) {
        println!("Dropped Okay");
    }
}

#[allow(non_snake_case)]
fn Something() -> Element {
    // Dropped
    let blabla = use_hook(OkayWrapper::default);

    // Not dropped
    let mut data = use_signal(VecWrapper::default);

    let increase = move |_| {
        data.write().0.extend((0..99999).map(|_| 1f64));
    };

    rsx!(
        button {
            onclick: increase,
            p { "Increase {data.read().0.len()}" }
        }
    )
}


fn app() -> Element {
    let mut show = use_signal(|| true);

    rsx!(
        button {
              onclick: move |_| show.toggle(),
              p { "Toggle" }
        }
        if *show.read() {
               Something { }
        }
    )
}

```

**Expected behavior**

Drop their inner values.

**Screenshots**

https://discord.com/channels/899851952891002890/928812591126569000/1213503960229486634

**Environment:**
 - Dioxus version: `9ae3d14e676c90c73a99f5cfb41ff85c7cbf4bc2` - latest commit
 - Rust version: 1.75
 - OS info: Windows
 - App platform: Freya

ealmloff:
Signals should be owned by the scope they were created in. They use an owner which lives in the context API. Dioxus signals had a bug that caused child scopes to use the parent scope's owner and only drop when that parent is dropped. This PR fixes the leak by only looking for the signal owner in the current scope instead of looking through all parent scopes (including the root scope). 

Closes #1999

### DIFF
--- a/packages/signals/README.md
+++ b/packages/signals/README.md
@@ -36,18 +36,18 @@ use dioxus_signals::*;
 #[component]
 fn App() -> Element {
     // Because signal is never read in this component, this component will not rerun when the signal changes
-    let signal = use_signal(|| 0);
+    let mut signal = use_signal(|| 0);
 
     rsx! {
         button {
             onclick: move |_| {
-                *signal.write() += 1;
+                signal += 1;
             },
             "Increase"
         }
         for id in 0..10 {
             Child {
-                signal: signal,
+                signal,
             }
         }
     }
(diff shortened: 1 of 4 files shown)


jkelleyrtp (approved):
Looks great, figured it was simple. Thanks!

Issue #1804, opened 2024-01-10
sync server functions in a use future fail to hydrate


ealmloff:
**Problem**

Running this program fails to hydrate:
```rust
#![allow(non_snake_case, unused)]
use dioxus::prelude::*;
use dioxus_fullstack::{
    launch::{self, LaunchBuilder},
    prelude::*,
};

fn app(cx: Scope) -> Element {
    let future = use_future(cx, (), |_| async move {
        let data = get_server_data().await.unwrap();
        println!("Client received: {}", data);
    });

    match future.value() {
        Some(_) => render! { div {} },
        None => render! {"Loading..."},
    }
}

#[server]
async fn get_server_data() -> Result<String, ServerFnError> {
    Ok("Hello from the server!".to_string())
}

fn main() {
    LaunchBuilder::new(app).launch()
}
```

**Screenshots**

<img width="1512" alt="Screenshot 2024-01-10 at 9 05 42 AM" src="https://github.com/DioxusLabs/dioxus/assets/66571940/f4eb460a-fccb-4adb-9789-a01dc77f5b82">

**Environment:**
 - Dioxus version: `master`
 - Rust version: `nightly`
 - OS info: MacOS
 - App platform: `fullstack`

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

ealmloff:
This is caused by https://github.com/DioxusLabs/dioxus/pull/1710:
- If a user tries to use a use_future with a sync server function, on the server it will poll once and resolve to a value. On the client, it will not resolve immediately. This can cause hydration to fail, because the server can renderer something different than the client.

In reality, users should probably be using use_server_future instead of use_future. We could:
- Warn the user about the issue and force them to change it to a use_server_future
- Always wait at least one poll before resolving server functions

Nnamdi-sys:
Encountering similar problem on the use_server_future as well in the issue here:
https://github.com/DioxusLabs/dioxus/issues/1950

ealmloff:
Closed by #2005

PR #2005, opened 2024-03-06
Only poll suspended futures, lazy memos

### LINKED ISSUE #1935: Memos can give out of date values
**Problem**

You can read an out of date value from a memo if you update a dependency and then immediately read the memo

**Steps To Reproduce**

1) Create a memo
2) Update a state the memo depends on
3) Read an out of date value from the memo

```rust
use dioxus::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {

    let mut signal = use_signal(|| 0);
    let memo = use_memo({
        move || {
            println!("Signal: {:?}", signal);
            signal()
        }
    });
    assert_eq!(memo(), 0);
    signal += 1;
    assert_eq!(memo(), 1);

    None
}
```

**Expected behavior**

Memos should always have up to date values

**Screenshots**

**Environment:**
 - Dioxus version: `master`
 - Rust version: `nightly`
 - OS info: MacOS
 - App platform: `hooks`

**Questionnaire**

### LINKED ISSUE #2003: suspend should take an explicit future
## Specific Demand

Right now to suspend a component you call `suspend()?`. This makes dioxus wait for wakeups in that component, requiring us to run every future, even if that future doesn't contribute to suspense. This is particularly annoying on the server where we need some futures running on the server and some only on the client.

If we tweaked suspend to take a future, we would know exactly *which* futures are causing the component to be suspended, and then only poll those futures when running `wait_for_suspense`.

This is particularly handy when it comes to data loading - we don't want to run any non-dataloading futures while waiting for a route/component to be loading its data. This would ensure a suspense tree is completely inert, making queuing suspense futures side-effect free, which is essential on the server.

ealmloff:
This PR fixes suspense polling futures that are unrelated to suspense. This will fix some hydration errors including https://github.com/DioxusLabs/dioxus/issues/1804.

Currently during suspense, all futures that have been spawned will be polled until suspense is finished. This causes hydration issues if a future finishes on the server but not the client and renders different content.

Fixes #2003
Fixes #1935

### DIFF
--- a/packages/signals/src/signal.rs
+++ b/packages/signals/src/signal.rs
@@ -150,8 +150,10 @@ impl<T: 'static, S: Storage<SignalData<T>>> Signal<T, S> {
         {
             let inner = self.inner.read();
 
-            let mut subscribers = inner.subscribers.lock().unwrap();
-            subscribers.retain(|reactive_context| reactive_context.mark_dirty())
+            // We cannot hold the subscribers lock while calling mark_dirty, because mark_dirty can run user code which may cause a new subscriber to be added. If we hold the lock, we will deadlock.
+            let mut subscribers = std::mem::take(&mut *inner.subscribers.lock().unwrap());
+            subscribers.retain(|reactive_context| reactive_context.mark_dirty());
+            *inner.subscribers.lock().unwrap() = subscribers;
         }
     }
 
--- a/packages/signals/src/signal.rs
+++ b/packages/signals/src/signal.rs
@@ -1,11 +1,9 @@
+use crate::Memo;
 use crate::{
     read::Readable, write::Writable, CopyValue, GlobalMemo, GlobalSignal, ReactiveContext,
-    ReadOnlySignal, ReadableRef,
+    ReadableRef,
 };
-use dioxus_core::{
-    prelude::{spawn, IntoAttributeValue},
-    ScopeId,
-};
+use dioxus_core::{prelude::IntoAttributeValue, ScopeId};
 use generational_box::{AnyStorage, Storage, SyncStorage, UnsyncStorage};
 use std::{
     any::Any,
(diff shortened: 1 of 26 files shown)


packages/signals/src/signal.rs
```diff
@@ -237,7 +208,7 @@ impl<T: 'static, S: Storage<SignalData<T>>> Writable for Signal<T, S> {
     }
 
     #[track_caller]
-    fn try_write(&self) -> Result<Self::Mut<T>, generational_box::BorrowMutError> {
+    fn try_write(&mut self) -> Result<Self::Mut<T>, generational_box::BorrowMutError> {
```
marc2332, line 211: I was using this method some times as a way to get a writable value without the need to making mut the value, is there a way to avoid adding &mut? Or maybe is it possible to have another method that takes &self?

ealmloff: The idea with signals is by default borrow checking within the same scope works like normal. This should give you a compile time error just like normal state:
```rust
let mut signal = Signal::new(0);
let data = signal.read();
signal.set(1);
println!("{data}");
```
But once you copy it into a different scope, it resets any borrows:
```rust
let mut signal = Signal::new(0);
let data_callback = move || signal.read();
let data = data_callback();
signal.set(1);
println!("{data}");
```

Two issues currently exist:
1) We don't thread the lifetimes through the generational borrows, so both of the above work
2) The easy way to opt out of this checking was removed during some other updates. We used to support `write_unchecked` which took `&self` instead of `&self`. Right now, you can make a copy of the signal and write to that, but it isn't as easy:
let signal = Signal::new(0);
let signal_borrow = &signal;
let mut copied = signal;
// You can write to the copy while you hold a reference to signal
copied.set(1);
println!("{signal_borrow}");
```

marc2332: Regarding the .2, couldn't there be something like `with_mut` that actually takes `&self` but gives you `&mut T` by giving you a mutable reference to a copy of the signal? or something like that, maybe calling it `with_unchecked`

ealmloff: Yes, either `with_unchecked` or `write_unchecked` would work

jkelleyrtp:
Looks great :) Clean diff, straightforward code. I like that we collapsed the DirtyScopes into VirtualDom. Fantastic work, thanks!

jkelleyrtp (approved):

packages/signals/src/signal.rs
marc2332: > Yes, either `with_unchecked` or `write_unchecked` would work

Maybe I can implement this myself

ealmloff: `with_unchecked`? Yes with https://github.com/DioxusLabs/dioxus/pull/2022 you can implement that pretty easily

jkelleyrtp: @marc2332 in the short term you can re-ref which will give you a copy and relax the borrow checker.

PR #2008, opened 2024-03-06
fix sledgehammer bindings on web

jkelleyrtp:
We were missing a `let` for `els` in the JS bindings.

It's weird that this wasn't getting caught by the fullstack tests. Not sure what happened - maybe we bypassed merge or something?


ealmloff (approved):
I don't think we test that part of diffing in the headless tests currently. We should test more cases or try to do headless fuzzing

Issue #1973, opened 2024-02-24
Auto detect the platform from the features and serve that one


jkelleyrtp:
## Specific Demand

`dx serve` should autodetect which platform you're targeting and serve that platform instead of defaulting to web. This would make using it with desktop much nicer.

This should be easy enough with a few simple heuristics - we might be able to ship it for 0.5

PR #2010, opened 2024-03-06
added automatic platform detection to cli if none is passed

rtretter:
Should close https://github.com/DioxusLabs/dioxus/issues/1973

### DIFF
--- a/packages/cli/src/cli/serve.rs
+++ b/packages/cli/src/cli/serve.rs
@@ -3,6 +3,7 @@ use manganis_cli_support::AssetManifest;
 
 use super::*;
 use std::{fs::create_dir_all, io::Write, path::PathBuf};
+use cargo_toml::Dependency::{Simple, Inherited, Detailed};
 
 /// Run the WASM project on dev-server
 #[derive(Clone, Debug, Parser)]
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli/serve.rs
```diff
@@ -41,11 +42,28 @@ impl Serve {
...
+                let features = match dependency {
+                    Inherited(detail) => detail.features.to_vec(),
+                    Detailed(detail) => detail.features.to_vec(),
+                    Simple(_) => vec![]
+                };
+
+                platform = features
+                    .iter()
+                    .next()
+                    .and_then(|first_feature| serde_json::from_str(&format!(r#""{}""#, first_feature)).ok());
```
jkelleyrtp, line 60: I think we should choose the first feature that matches a valid platform:

```
fullstack
desktop
mobile
web
ssr
liveview
axum
tui
```

Either that or complain if we see two platforms enabled at once.

Otherwise, fantastic work, thank you!

jkelleyrtp (changes requested):

packages/cli/src/cli/serve.rs
rtretter: Oh, I only found the Platform-Enum in cli-config containing
* web
* desktop
* fullstack
Where are the others located?

Currently the first entry is always picked, I will adjust it to search for the first valid platform.

jkelleyrtp: Oh, we might only support those platforms in the CLI then. That list is coming from the valid features on `dioxus/packages/cargo.toml`.

rtretter:
Excuse my force pushes I didn't think they would appear as such 😅 
I'll be sure to avoid them in future PRs.

jkelleyrtp:
Looks great, thank you!

jkelleyrtp (approved):

PR #2033, opened 2024-03-09
Kick server futures on the client to track reactivity

jkelleyrtp:
Fix: #1950

This kicks the future until its pending, fixing the use cases where you use a future like:

```rust
use_sever_future(|| async move { sf(reactive_argument()).await });
```

We still need a better way of sending reactive dependencies to the client, but this should fix that particular issue.

### DIFF
--- a/packages/fullstack/src/hooks/server_future.rs
+++ b/packages/fullstack/src/hooks/server_future.rs
@@ -14,8 +14,6 @@ where
 
     let resource = use_resource(move || {
         async move {
-            // this is going to subscribe this resource to any reactivity given to use in the callback
-            // We're doing this regardless so inputs get tracked, even if we drop the future before polling it
             let user_fut = cb.call();
 
             let currently_in_first_run = first_run.cloned();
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Running code before the first await point may break some futures and this will also not subscribe to any reads after the first Pending result:

```rust
let mut state = use_signal(|| 0);
use_server_future(|| async move {
    // I wouldn't expect this to run on the client after hydration
    state.set(0);
    sf().await;
    // use_server_future will not subscribe to state because it is after the first pending result
    other_future(state()).await
});
```

Longer term (0.7?) we should find a better way to do this that doesn't have side effects, but this is fine for now. We might be able to serialize the subscriptions by the index read and send that along with the server function result to the client

jkelleyrtp:
> Running code before the first await point may break some futures and this will also not subscribe to any reads after the first Pending result:
> 
> ```rust
> let mut state = use_signal(|| 0);
> use_server_future(|| async move {
>     // I wouldn't expect this to run on the client after hydration
>     state.set(0);
>     sf().await;
>     // use_server_future will not subscribe to state because it is after the first pending result
>     other_future(state()).await
> });
> ```
> 
> Longer term (0.7?) we should find a better way to do this that doesn't have side effects, but this is fine for now. We might be able to serialize the subscriptions by the index read and send that along with the server function result to the client

Yeah it's a sucky problem to deal with, but I think subscription serialization is the right way to handle it, and we should have enough metadata plumbed throughout the system to make it possible.

PR #2049, opened 2024-03-10
feat: Add `Resource::clear()` and also updated `Resource` docs

marc2332:
Add `Resource::clear()` and also updated `Resource` docs

### DIFF
--- a/packages/hooks/src/use_resource.rs
+++ b/packages/hooks/src/use_resource.rs
@@ -12,7 +12,7 @@ use std::ops::Deref;
 use std::{cell::Cell, future::Future, rc::Rc};
 
 /// A memo that resolves to a value asynchronously.
-/// Similar to `use_memo` but `use_resource` runs asynchronously.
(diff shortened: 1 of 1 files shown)


packages/hooks/src/use_resource.rs
```diff
@@ -11,8 +11,8 @@ use futures_util::{future, pin_mut, FutureExt, StreamExt};
 use std::ops::Deref;
 use std::{cell::Cell, future::Future, rc::Rc};
 
-/// A memo that resolve to a value asynchronously.
-/// Unlike `use_future`, `use_resource` runs on the **server**
+/// A memo that resolves to a value asynchronously.
+/// Similar to `use_memo` but `use_resource` runs asynchronously.
```
ealmloff, line 15: `use_resource` will not memorize the value returned from the resource unlike `use_memo`. It is pretty similar to `use_future` with a return value

ealmloff (approved):
Looks good, thank you!

PR #2050, opened 2024-03-10
docs: Add Turkish Translation

Tahinli:


ealmloff (approved):
Thanks for putting this together! (I do not know Turkish, but the technical information through google translate looks correct)

PR #2069, opened 2024-03-12
Manual non-reactive dependancies for use_memo, use_effect, and use_resource

ealmloff:
Adds a way to create a memo, effect, and resource with a manual non-reactive dependency tuple:

```rust
use dioxus::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {
    let mut signal = use_signal(|| 0);

    use_future(move || async move {
        loop {
            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
            signal += 1;
        }
    });

    rsx! {
        "Parent count: {signal}"
        Child {
            non_reactive_prop: signal()
        }
    }
}

#[component]
fn Child(non_reactive_prop: i32) -> Element {
    let mut signal = use_signal(|| 0);

    // You can manually specify the dependencies with `use_reactive` for values that are not reactive like props
    let computed = use_memo(use_reactive!(
        |(non_reactive_prop,)| non_reactive_prop + signal()
    ));
    use_effect(use_reactive!(|(non_reactive_prop,)| println!(
        "{}",
        non_reactive_prop + signal()
    )));
    let fut = use_resource(use_reactive!(|(non_reactive_prop,)| async move {
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
        non_reactive_prop + signal()
    }));

    rsx! {
        button {
            onclick: move |_| signal += 1,
            "Child count: {signal}"
        }

        "Sum: {computed}"

        "{fut():?}"
    }
}
```


ealmloff:
> I think `with_dependencies` would be better than `use_dependencies`

`use_dependencies` is a hook and must be called following the rules of hooks which is why it starts with the `use_` prefix

ealmloff:
I think the test failures here are unrelated. It looks like the same failures are happening in the `main` branch

PR #2086, opened 2024-03-15
Fix: Add dioxus-desktop To Mobile Feature

DogeDark:
Mobile relies on dioxus-desktop and without it, cargo complains about undeclared crates or modules named ``dioxus-desktop``


jkelleyrtp:
Perfect, thank you!

PR #2103, opened 2024-03-18
Fix #2043: use formvalue instead of String for forms

### LINKED ISSUE #2043: Form values should return a Vec of values
Here i suggested more classic browser behavior #1834 , which was implemented in #1836.
Why was it changed again?

Let's assume we have group of fields with same name, in browser they are aggregated under same name in Array, like this:
```field_name = ["value1","value2","value 3"]```
even if one of values is empty it will persist in the array as empty value.
`field_name = ["","","value 3"]`

Current behavior is pushing all values into String with comma as separator:
`field_name = "value1,value2,value3"`

If one of values is empty it is simply excluded, but this is very tricky bc looks like there are no other fields:
`field_name = "value1"` - here looks like we have no more fields

And even this is possible:
`field_name = ",value2,value3"` - here first value is excluded, but the value exists but epmty

Снимок экрана от 2024-03-09 19-27-39

jkelleyrtp:
- [x] Fix #2043

Reverts/simplifies the changes I made when fixing forms. Instead of an enum of Val/VecVal, just uses a VecVel and an accessor to get the "value" if there's only one option.

Now uses a vec of strings instead of just a string to represent the fields of the form.

@serzhiio can you verify this works for you before we ship it?


ealmloff (approved):

serzhiio:
Yeah, i'm fine with current implementation

PR #2125, opened 2024-03-21
Check type of launch config

ealmloff:
I ran into this issue a couple times while updating apps to the new launch builder API. The launch builder currently accepts any type and silently fails when you pass it the wrong type.

This can lead to situations where you pass in a config and see no errors, but your config is not applied.

This PR changes the launch builder to check the type of the config at compile type instead of using a boxed any type.


srid:
Aside: I was wondering if this PR helped address [this compile error](https://discord.com/channels/899851952891002890/1221445356668194907) with `LaunchBuilder`, but it didn't.

jkelleyrtp:
Played around with it locally but not super happy with it since we still have the platform methods here that don't work with the config they specify.  Will likely punt on it unless there's a quick fix to allow Builder::desktop().with_cfg()

ealmloff:
> Played around with it locally but not super happy with it since we still have the platform methods here that don't work with the config they specify. Will likely punt on it unless there's a quick fix to allow Builder::desktop().with_cfg()

```rust
Builder::desktop().with_cfg()
```
now works even if you have the `web` feature enabled as well.

PR #2133, opened 2024-03-24
Fix nix support, by upgrading Rust compiler version

srid:
This PR mainly updates `nixpkgs` so that we use the newer Rust compiler.

Without this, the devShell won't be able to build local crates, viz.:

```sh
❯ nix develop -c cargo build -p dioxus-cli
error: package `normpath v1.2.0` cannot be built because it requires rustc 1.74.0 or newer, while the currently active rustc version is 1.73.0
Either upgrade to rustc 1.74.0 or newer, or use
cargo update -p normpath@1.2.0 --precise ver
where `ver` is the latest version of `normpath` supporting rustc 1.73.0
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,54 +1,15 @@
 {
   "nodes": {
-    "crane": {
-      "inputs": {
-        "flake-compat": "flake-compat",
-        "flake-utils": "flake-utils",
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
Cool! Thanks!

PR #2134, opened 2024-03-24
Allow <input type="file"> without event handlers to open web view file dialog on Desktop

matteron:
This add no real benefit other than hopefully prevent confusion in the future.

Currently, clicking an input with type file without any attached event handlers will cause the event to just get swallowed.  This just lets the event propagate like normal in those cases.

This seems intentionally disabled at some point, but seems to behave normally now, so I figured might as well.


jkelleyrtp:
Thanks for the contribution :) looks good

jkelleyrtp (approved):

PR #2139, opened 2024-03-25
Fix fullstack render server context

DonAlonzo:
Calls to `extract` during initial render would behave differently from calls to `extract` in server functions called through the API. This fixes it.


ealmloff (approved):
Good catch! Because servers may be running multiple server functions at once, we need to only set the server context while we are polling the future instead of the entire time the future is running. I pushed a change that fixes that issue and verified that extracting a header map works both during the initial render and in server functions

Thank you!

PR #2157, opened 2024-03-27
Revision: Change Default Logging In CLI

DogeDark:
Makes the default logs be ``warn`` for the CLI and manganis and ``error`` for other deps unless the ``DIOXUS_LOG`` environment variable is set in which it will use what is specified there.


jkelleyrtp (approved):
Great PR, thanks!

PR #2159, opened 2024-03-27
Require every field in router variants to be present in the route in the web history

ealmloff:
In the web history if a field is not present in the URL, it will not be passed to the route (even without reloading the page). This PR adds a compiler error when the web feature is enabled and a field included in the router is not present in the URL.

### DIFF
--- a/packages/router-macro/Cargo.toml
+++ b/packages/router-macro/Cargo.toml
@@ -22,3 +22,4 @@ slab = { workspace = true }
 
 [features]
 default = []
+web = []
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
What's the plan to associate metadata with a route? Maybe later in the release we can add a metadata tag to the router that exempts fields from this rule?

Also are you trying to get this into the release?

ealmloff:
> Also are you trying to get this into the release?

Ideally, because this is a breaking change but it could also be pushed to 0.6

jkelleyrtp:
Cool, when tests pass I'll merge it

PR #2198, opened 2024-03-30
Implement the "Toggle Developer Tools" menu item in desktop

mullr:


jkelleyrtp:
Thanks! I think we expected this to be automatic (some menubar commands are actually - like copy, paste, cut, hide, hide-all, etc) but I guess not this one.

PR #2209, opened 2024-04-01
Fix for Counters sample code in README

ekanna:
Fix for Counters sample code in README. It was throwing syntax errors.


ealmloff (approved):
Thank you!

PR #2213, opened 2024-04-01
Fix for Counters sample code in Readme

ekanna:
For leptos code, there is no need for Memorization here. As memorization of the values of the text nodes in the renderer already happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ fn Counters() -> impl IntoView {
         >
             <li>
                 <button on:click=move |_| counters.update(|n| n[idx] += 1)>
-                    {move |_| counters.with(|n| n[idx])}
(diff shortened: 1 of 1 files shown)


ealmloff:
> For leptos code, there is no need for Memorization here. As memorization of the values of the text nodes in the renderer already happens.

This doesn't appear to be the case in the latest commit of leptos `main`. As I understand it, Leptos doesn't use a virtual dom so nodes are not automatically memorized. Here is the behavior when running your modification. Each time I change any of the counter values, each text node is rerendered:


https://github.com/DioxusLabs/dioxus/assets/66571940/79eb06df-0245-4fd7-ab81-72ed71e48085

ealmloff (approved):

Issue #2227, opened 2024-04-02
EventHandlers on elements are owned by parent, causing long-running memory usage


jkelleyrtp:
**Problem**

In https://github.com/DioxusLabs/dioxus/discussions/2195 we received a report that global signals might be causing large memory usage. This shouldn't be the case but it's worth investigating and writing tests just in case.

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

marc2332:
Small repro:

```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;
use dioxus::prelude::dioxus_core::NoOpMutations;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    fn app() -> Element {
        let mut count = use_signal(|| 0);

        use_hook(|| spawn(async move {
            loop {
                tokio::time::sleep(std::time::Duration::from_millis(2)).await;
                let val = *count.peek_unchecked();
                if val == 50 {
                    count.set(0);
                } else {
                    count.set(val + 1);
                }
            }
        }));

        rsx! {
            div { 
                color: "red",
                for el in 0..*count.read() {
                    rect {
                        key: "{el}",
                        onclick: move |_| { },
                        label {
                            "{el}"
                        }
                    }
                }
            }
        }
    }

    // create the vdom, the real_dom, and the binding layer between them
    let mut vdom = VirtualDom::new(app);

    vdom.rebuild(&mut NoOpMutations);

    // we need to run the vdom in a async runtime
    tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()?
        .block_on(async {
            loop {
                // wait for the vdom to update
                vdom.wait_for_work().await;

                // get the mutations from the vdom
                 vdom.render_immediate(&mut NoOpMutations);
            }
        })
}
```

marc2332:
Although the leaking is now much smaller, it still seems to leak.

New repro:

```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;
use dioxus::prelude::dioxus_core::NoOpMutations;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    fn app() -> Element {
        let mut count = use_signal(|| 0);

        use_hook(|| spawn(async move {
            loop {
                tokio::time::sleep(std::time::Duration::from_micros(50)).await;
                let val = *count.peek_unchecked();
[...]

PR #2244, opened 2024-04-04
Fix: memory leak on listeners for elements

jkelleyrtp:
We switched over to Copy types for EventHandlers on elements which is equivalent to calling Signal::new(). This is not necessary for elements, just for child components, and thus caused a memory leak for long-running components.

This PR adds a drop impl to VNode to manually drop listeners when the last VNodeInner is dropped. We can't add the drop impl anywhere else since VNode is the only struct in the chain that can't be destructured, making it semver compatible.

Fix: #2227

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2453,6 +2453,7 @@ dependencies = [
  "serde_json",
  "serde_repr",
  "tokio",
+ "tracing",
  "wasm-bindgen",
  "web-sys",
 ]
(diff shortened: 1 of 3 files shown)


ealmloff (approved):
Looks good

I think we should publish this fix in a minor release, but just noting that could break some code using 0.5 if anyone is reading the eventhandler we now drop properly. There isn't any way to fix the leak without making a breaking change like this

PR #2248, opened 2024-04-05
fix the relative paths issue in the fullstack crates.io build

ealmloff:
This just copies the dioxus-cli's script back into fullstack inline. In the future, we should try to pull out the hot reloading code into a separate package that both the CLI and fullstack can use to serve the websocket


ealmloff:
Tracking the duplication in https://github.com/DioxusLabs/dioxus/issues/2249

PR #2258, opened 2024-04-05
Hotreloading of `for/if/body`, formatted strings, literals, component props, nested rsx, light CLI rewrite, cli TUI

jkelleyrtp:
Hotreload the contents of for loops, if chains, component bodies, props, attributes, and any literals discovered in rsx!

This is breaking since we need to open a *bunch* of APIs making this no longer compatible with all the support crates in 0.5. Anyways, plan is to ship by end of april so it's fine that we keep this PR out here for a while.

Also brings in binary patching for components for *nix targets under an experimental flag.

Specifics/todos:
- [x] rewrite hotreload logic to allow collecting nested templates
- [x] hotreload for loop bodies
- [x] hotreload if chains bodies
- [x] Hotreload logic for ifmt
- [x] enable hotreload for textnodes
- [x] enable hotreload for component props
- [x] A more stable way of assigning IDs to literals/ifmts
- [x] enable hotreload for generic literals (numbers, &str, etc) 
- [x] partial expand expressions where possible (autocomplete in expressions)
- [x] PartialExpr
- [x] PartialClosure
- [x] merge element and component parsing for consistency
- [x] merge element and component autoformatting for consistency
- [ ] Give all components .children() methods even if they're not used (some sort of warning) if not defined

Tests:
- [x] Parsing attribute
- [x] Parsing block
- [x] Validation of element
- [x] Validation of component

CLI Rewrite (prep for plugins):
- [x] merge all hotreloading behind a single websocket port
- [x] merge all file watchers behind a single watcher
- [x] use a tokio select / tx/rx pattern
- [x] Drop interprocess
- [x] async builders
- [x] TUI for serve

Final todo:
- [x] some merge conflicts broke template registration
- [x] need to wire signal reloading into web
- [x] the `..` syntax is not actually generating a spread for elements/components
- [x] Bug around setting a HR literal back to its original
- [x] Remove `krates` from rsx?

Demo:

https://github.com/DioxusLabs/dioxus/assets/10237910/0c685dbc-537d-4676-a7dc-7730fedf161a

https://github.com/DioxusLabs/dioxus/assets/10237910/1eff75a3-3bba-4e28-93b3-9dbd76537ce8


jkelleyrtp:
This is done, actually, but since it requires breaking changes to core, I'm gonna keep it open until we're ready to start merging break PRs. In the meantime I'll try adding more hot reloading features like formatted strings and prop literals to this pr.

DogeDark:
Will this include hot reloading for workspaces? Would massively speed up components in combo with a manganis issue fixed.

jkelleyrtp:
Gonna punt binary patching for another PR. We will still ship the binary patching under an experimental flag for unix-based platforms in 0.6 and then WASM based patching later.

packages/cli/src/server/watcher.rs
```diff
@@ -0,0 +1,259 @@
...
+
+/// This struct stores the file watcher and the filemap for the project.
+///
+/// This is where we do workspace discovery and recursively listen for changes in Rust files and asset
+/// directories.
+pub struct Watcher {
+    tx: UnboundedSender<notify::Event>,
+    rx: UnboundedReceiver<notify::Event>,
+    last_update_time: i64,
+    watcher: FsEventWatcher,
```
LiamKarlMitchell, line 22: This should use RecommendedWatcher not FsEventWatcher so that it can build on different platforms e.g. Windows/Linux.

packages/cli/src/server/watcher.rs
```diff
@@ -0,0 +1,259 @@
...
+
+use dioxus_cli_config::CrateConfig;
+use dioxus_hot_reload::HotReloadMsg;
+use dioxus_html::HtmlCtx;
+use dioxus_rsx::hot_reload::FileMap;
+use futures_channel::mpsc::{UnboundedReceiver, UnboundedSender};
+use futures_util::StreamExt;
+use notify::{
+    event::{DataChange, ModifyKind},
+    EventKind, FsEventWatcher,
```
LiamKarlMitchell, line 11: Use RecommendedWatcher

packages/cli/src/server/watcher.rs
LiamKarlMitchell: Was seeing error compiling on windows.
```
error[E0432]: unresolved import `notify::FsEventWatcher`                                                                                                                                                          
   --> packages\cli\src\server\watcher.rs:11:16
    |
11  |     EventKind, FsEventWatcher,
    |                ^^^^^^^^^^^^^^ no `FsEventWatcher` in the root

```

Made a PR if helpful.
https://github.com/DioxusLabs/dioxus/pull/2596

packages/cli/src/server/output.rs
```diff
@@ -1,157 +1,518 @@
...
+            if let KeyCode::Char('c') = key.code {
+                if key.modifiers.contains(KeyModifiers::CONTROL) {
+                    return Err(io::Error::new(io::ErrorKind::Interrupted, "Ctrl-C"));
+                }
+            }
+        }
+
+        if let Event::Key(key) = input {
+            if let KeyCode::Char('/') = key.code {
+                self.fly_modal_open = !self.fly_modal_open;
```
LiamKarlMitchell, line 197: May need to debounce user inputs like this or track key state over time.

As when I press/hold it to toggle visible then not visible straight away whilst I am pressing the key sometimes I see the modal flicker but rare to see if visible.

Perhaps there is a key down / key up event?
Or can add one with checking previous key state.
I'm running from RustRover.

packages/desktop/src/app.rs
```diff
@@ -148,14 +149,32 @@ impl App {
...
     ))]
     pub fn connect_hotreload(&self) {
-        let Ok(cfg) = dioxus_cli_config::CURRENT_CONFIG.as_ref() else {
+        let Some(cli_args) = dioxus_cli_config::ServeArguments::from_cli() else {
             return;
         };
+        let port = cli_args.port;
+        let addr = cli_args
+            .addr
+            .unwrap_or(std::net::IpAddr::V4(std::net::Ipv4Addr::new(0, 0, 0, 0)));
```
LiamKarlMitchell, line 158: When running on Windows 10, it seems `0.0.0.0` does not connect to hot reload server.
```Failed to start hotreload server```
However if I use `127.0.0.1` or perhaps specify the IP address exactly it connects and when I edit my files I get hot

If I recall before this was changed it was giving me an error like below when `0.0.0.0` was used.
`HotReload failed to start IO error: The requested address is not valid in its context. (os error 10049)`

```rust
.unwrap_or(if cfg!(target_os = "windows") {
    std::net::IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1))
} else {
    std::net::IpAddr::V4(std::net::Ipv4Addr::new(0, 0, 0, 0))
});
```

Not sure why it couldn't always use `127.0.0.1` as default ipv4 as all zeros is normally used for binding not connecting?

packages/cli/src/server/output.rs
```diff
@@ -1,157 +1,518 @@
...
+                .children
+                .iter_mut()
+                .find_map(|(platform, child)| {
+                    if platform == &result.platform {
+                        Some((child.stdout.take().unwrap(), child.stderr.take().unwrap()))
+                    } else {
+                        None
+                    }
+                })
+                .unwrap();
```
LiamKarlMitchell, line 276: When starting up with `--platform web` this line panicks.
thread 'main' panicked at packages\cli\src\server\output.rs:276:18:
called `Option::unwrap()` on a `None` value

Worked around it with this.
```rust
match build_engine
    .children
    .iter_mut()
    .find_map(|(platform, child)| {
        if platform == &result.platform {
            Some((child.stdout.take(), child.stderr.take()))
        } else {
            None
        }
    }) {
    Some((Some(stdout), Some(stderr))) => {
        let platform = result.platform.clone();

        let app = RunningApp {
            result,
            stdout: BufReader::new(stdout).lines(),
            stderr: BufReader::new(stderr).lines(),
            stdout_line: String::new(),
            stderr_line: String::new(),
        };

        self.running_apps.insert(platform, app);
    },
    _ => {
        // Handle the None case or continue the loop
        continue;
    }
}
```

packages/rsx/src/partial_closure.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.as_expr_closure().map(|closure| Expr::Closure(closure))
+    }
+
+    /// Convert this partial closure into a full closure if it is valid
+    /// Returns err if the internal tokens can't be parsed as a closure
+    pub fn as_expr_closure(&self) -> Result<ExprClosure> {
+        // If there's a brace token, we need to parse the body as a block
+        // Otherwise we can parse it as an expression
+        let body = match self.brace_token.as_ref() {
+            Some(brace) => {
```
jkelleyrtp, line 162: Expr::Block(ExprBlock {
                    attrs: Vec::new(),
                    label: None,
                    block: Block {
                        brace_token: brace.clone(),
                        stmts: Block::parse_within.parse2(self.body.clone())?,
                    },
                })

ealmloff (changes requested):

packages/desktop/src/app.rs
jkelleyrtp: Localhost should generally work but reloading over the network might require the 0.0.0.0 binding

LiamKarlMitchell: Sorry maybe I have misunderstood?

Server side:
Dev servers I've seen in the past listen by default on localhost perhaps as a measure of secure by default?

They leave it to the user to configure if wanting to listen on a specific network or bind all.

For the server, binding all addresses by default may be convenient but it may be better to only listen on localhost unless otherwise specified? If worried about exposing service to other machines on network etc. (Of course users could have firewalls and configure there too). E.g. a worm developer finds a vulnerability in your service and exploits it to connect to all machines on a network because they listen on 0.0.0.0 by default. Bad press could come from that right.

Binding localhost can cause a minor confusion if developers are not knowing to set the bind ip however that could be noted in docs.

Client side:
Connecting for TCP over network needs the exact ip.
On Windows a connect to 0.0.0.0 does not loopback.

In a scenario with different subnets
1. Host machine
   Wifi: Network 1
   Ethernet: Network 2
   Internet: Network 3 via Network 2.
   Can connect to self on localhost
   Also runs Websocket Server which does bind all via 0.0.0.0 "good to be configurable"
2. Mobile connect via wifi
3. Laptop connected via ethernet
4. Another desktop connected via internet through portforward+nat.

Each connecting client would have a different IP for the Host.

0.0.0.0 could be used for starting a server to listen on all devices but if this code is for connecting to the server then must use a exact IP for its route to the host.

packages/web/Cargo.toml
```diff
@@ -12,6 +12,8 @@ keywords = ["dom", "ui", "gui", "react", "wasm"]
 [dependencies]
 dioxus-core = { workspace = true }
 dioxus-html = { workspace = true, features = ["wasm-bind"] }
+dioxus-hot-reload = { workspace = true, features = ["client"] }
```
LiamKarlMitchell, line 15: Trying to run web platform gave me this error.
Need to enable feature for CloseEvent for HotReload.

Cargo.toml for web package.
Under features section for hot_reload add `"web-sys/CloseEvent"`

```
error[E0432]: unresolved import `web_sys::CloseEvent`                             
  --> packages\web\src\hot_reload.rs:15:23                                        
   |                                                                              
15 | use web_sys::{window, CloseEvent, MessageEvent, WebSocket};                  
   |                       ^^^^^^^^^^                                             
   |                       |                                                      
   |                       no `CloseEvent` in the root                            
   |                       help: a similar name exists in the module: `MouseEvent`
```

packages/cli/src/server/output.rs
jkelleyrtp: Good catch! Stdout is not guaranteed for web builds - we want to actually pipe the output to the terminal. I've updated the code to do that.

packages/signals/src/global/mod.rs
```diff
@@ -1,18 +1,54 @@
...
+/// The context for global signals
 #[derive(Clone)]
-pub(crate) struct GlobalSignalContext {
-    signal: Rc<RefCell<HashMap<*const (), Box<dyn Any>>>>,
+pub struct GlobalSignalContext {
+    signal: Rc<RefCell<HashMap<GlobalKey, Box<dyn Any>>>>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum GlobalKey {
```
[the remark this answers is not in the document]
ealmloff: This doesn't work because you need to allocate the buffer somewhere at compile time. You could do it with a macro, but we don't want the entry point for global signals to be a macro

```rust
const BUFFER_SIZE: usize = 2usize.pow(16);
const fn location_to_str(location: &'static Location<'static>) -> [u8; BUFFER_SIZE] {
    let mut buffer = [0u8; BUFFER_SIZE];
    let mut fill_index = 0;
    // The first part of the location is the file name
    let file_name = location.file().as_bytes();
    while fill_index < file_name.len() {
        buffer[fill_index] = file_name[fill_index];
        fill_index += 1;
    }

    // Then the colon
    buffer[fill_index] = b':';
    fill_index += 1;
    // Then the line number
    let mut line_number = location.line();
    while line_number != 0 {
        let digit = line_number % 10;
        buffer[fill_index] = b'0' + digit as u8;
        fill_index += 1;
        line_number /= 10;
    }

    // Then the colon
    buffer[fill_index] = b':';
    fill_index += 1;
    // Then the file name
    let file_name = location.file().as_bytes();
    while fill_index < file_name.len() {
        buffer[fill_index] = file_name[fill_index];
        fill_index += 1;
    }

    let (filled, _) = buffer.split_at(fill_index);

    match std::str::from_utf8(filled) {
        Ok(s) => s,
        Err(_) => panic!("formatting the location created invalid utf-8"),
    }
}
```

jkelleyrtp:
- [ ]  Send HR changes down the pipe on initial connect
- [ ]  Dont send every template on every update
- [ ]  Change the ordering of TUI logs
- [ ]  Filtering system for logs (warn/error/tracing/location) - not required but something simple would be nice
- [ ]  WASM ws-log coloring bad - all colors are lost, and some html is sent down
    - [ ]  use dioxus-logger where possible
- [ ]  Add console logs to tui when a hotreload occurs / file watcher surfaces changes
    - [ ]  easier debugging and some insight on what is causing the full rebuild
    - [ ]  log this as a console log
- [ ]  The TUI perf seems bad - we should test in release mode
    - [ ]  Scrolling has way too much momentum sometimes
- [ ]  don’t necessarily wait for an entire line - maybe pull a line or poll for a character
- [ ]  TUI help screen should not just say “hello world” 10 times.
    - [ ]  Extra tabs for tutorials, server address, uptime, etc
    - [ ]  Tabs should start with 1 (keyboard order)
    - [ ]  Autoswitching on errors should bring you to the build tab
    - [ ]  Maybe add a settings configurator for settings we can change on the fly?
    - [ ]  add in network address, https, compiler info, dx verison info
    - [ ]  maybe move the platform span to the sidebar
    - [ ]  fix the sidebar collapse responsiveness
- [ ]  Wire up shortcuts that say they work but definitely don’t work
    - [ ]  open the app
    - [ ]  maybe float the desktop app / float on top / surface?
    - [ ]  open devtools could be cool!
    - [ ]  clear should clear the terminal
- [ ]  The web loading screen is pretty bad
    - [ ]  No percentages for optimizing assets, etc
    - [ ]  Style is pretty rough too
- [ ]  A toast would be pretty nice if the build failed (autoclear perhaps?) - +/- some way of accessing our own devtools
- [ ]  Toast for disconnected devserver
- [ ]  Add more logging in the cli
    - [ ]  Virtualdom handles events - do a log
    - [ ]  ServerFns - inspect payload
    - [ ]  Expand logs
    - [ ]  Spanning of events → link to file in editor
- [ ]  rsx things that weren’t resolved in the way I wanted
    - [ ]  special key type instead of stringly typed keys
    - [ ]  would be nice to rethink or remove owners with a vcomp factor type
    - [ ]  pass-thru of tests again in case we missed things that should be asserts
    - [ ]  PrettyUnparse should be an extra dev-dependency crate in the workspace?
    - [ ]  cargo-fmt would be nice too instead of pretty unparse
    - [ ]  Look at scoring again and see if we can simplify/remove some things
- [ ]  components take .children() method
- [ ]  Binary size analysis of core - simple things we can do to improve this?
- [ ]  windows builds fail - crypto libraries
- [ ]  playwright fails - loading screens are loaded instead
    - [ ]  how to fix this? disable loading screens?
- [ ]  merged attributes with combined if-chains might still not be hotreloadable?
- [ ]  the changed-lits problem but for files (need a third file or a dynamic pool)
    - [ ]  remove template hotreload logic in core? move all to a dynamic pool
- [ ]  Tests for each crate that test feature combinations
    - [ ]  Some things randomly failing when tested on their own
    - [ ]  Tests for release mode feature combinations too - some things fail in release with certain features
    - [ ]  Maybe we can have fewer features in some places?
- [ ]  How many places is track_caller wrong?
- [ ]  Duplicate deps / old deps
    - [ ]  Speed up compile times?
    - [ ]  CLI compile times have gotten really slow
- [ ]  More test harness, more speedy tests
- [ ]  Can we optimize the repo itself for iterating?
    - [ ]  Changes in Rust-analyzer necessary?
- [ ]  Caching missing in CI?
- [ ]  Nightly CLI builds
- [ ]  Clear old build logs when new build logs come through
    - [ ]  Currently just accumulating all the build logs
- [ ]  html crate just be absolutely wilding now
- [ ]  the gawd damn prelude is sooooo big
    - [ ]  stuff macros use should not in the prelude if we can help it

Things to test:

- suspense
- rsx outputs
- keys
- diffing
- hotreloading in weird cases
- hotreload scoring
- TUI - this should be bulletproof!!!!!
    - can’t just ship a worse terminal.
- streaming
- feature combinations
- end-to-end
- hydration
- generational-box (sync/non-sync)
- global signals/memos etc
- more async signal stuff
- passing things via cli args and what to do with version mismatches
- liveview
- the launch file had a lot of churn and is suuuuuper annoying

chungwong:
Does it PR respect
```bash
❯ cat .cargo/config.toml 
[build]
rustflags = ["--cfg", "tokio_unstable"]
```
For some reason it didn't work
```diff
-dioxus = { git = "https://github.com/DioxusLabs/dioxus.git", rev = "742721b", features = ["fullstack", "router"] }
+dioxus = { git = "https://github.com/DioxusLabs/dioxus.git", rev = "443b9a4", features = ["fullstack", "router"] }
```
```bash
thread 'main' panicked at /home/chung/.cargo/registry/src/index.crates.io-6f17d22bba15001f/console-subscriber-0.3.0/src/lib.rs:265:9:
task tracing requires Tokio to be built with RUSTFLAGS="--cfg tokio_unstable"!
```
Of course if I run `RUSTFLAGS="--cfg tokio_unstable" dx serve` then everything works.

Ref: https://github.com/tokio-rs/console?tab=readme-ov-file#using-it

PR #2268, opened 2024-04-07
docs: :pencil2: typo -> unnecessary need to in global_context.rs

Tahinli:


Tahinli:
Why test is failed, is it normal to fail ? I just changed 2 words in document 😶‍🌫️

ealmloff (approved):

PR #2271, opened 2024-04-08
Fix features not passed from dx bundle to cargo

apuddy:
Playing around with [Publishing](https://dioxuslabs.com/learn/0.5/cookbook/publishing), the docs recommended to use a feature to disable the console on Windows when bundling. That worked with `dx build` but not `dx bundle`.

This PR fixes that by passing features to `cargo` when using `dx bundle`, which is the same behaviour as `dx build`


ealmloff (approved):
Good catch, thanks!

PR #2275, opened 2024-04-08
Fix empty for loop, if statements and children

### LINKED ISSUE #2270: Rsx panics on an empty conditional block
**Problem**

This fails with a panic:
```
#[component]
fn Home() -> Element {
    rsx! {
        if true {
        }
    }
}
```

But this works:
```
#[component]
fn Home() -> Element {
    rsx! {
    }
}
```

The failure case was extra confusing to me because I was using `if signal() {...}` -- where the insides of those braces was empty while I was developing / debugging the signal itself. So I kept thinking I had done something wrong with the signal, rather than simply having an empty block after an rsx conditional.

**Steps To Reproduce**

It is trivial to recreate this, just `dx new` and change the `Home` Element as above.

```
dx new
✔ 🤷   Which sub-template should be expanded? · Web
🤷   Project Name: empty-conditional
✔ 🤷   How do you want to create CSS? · Vanilla
✔ 🤷   Should the application use the Dioxus router? · true

nano src/main.rs
dx serve
```

**Expected behavior**

Do not panic in the browser.

**Screenshots**

The panic message was:

```
panicked at /home/chris/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-core-0.5.1/src/diff/node.rs:642:16:
index out of bounds: the len is 0 but the index is 0

Stack:
[very long stack]
```

**Environment:**
 - Dioxus version: `0.5.1`
 - Rust version: `rustc 1.74.1 (a28077b28 2023-12-04)`
 - OS info: `Ubuntu 22.04.3 LTS` (in WSL2)
 - App platform: `web`

**Questionnaire**

ealmloff:
We expand empty rsx statements to `None` instead of an empty template so that there is a root node for the template for diffing. This PR expands that behavior to the body of if statements, for loops and children

Fixes #2270

### DIFF
--- a/packages/rsx/src/attribute.rs
+++ b/packages/rsx/src/attribute.rs
@@ -3,7 +3,7 @@ use std::fmt::{Display, Formatter};
 use super::*;
 
 use proc_macro2::{Span, TokenStream as TokenStream2};
-use quote::{quote, ToTokens, TokenStreamExt};
+use quote::quote;
 use syn::{parse_quote, spanned::Spanned, Expr, ExprIf, Ident, LitStr};
 
 #[derive(PartialEq, Eq, Clone, Debug, Hash)]
--- a/packages/rsx/src/component.rs
+++ b/packages/rsx/src/component.rs
@@ -11,18 +11,15 @@
 //! - [ ] Keys
 //! - [ ] Properties spreading with with `..` syntax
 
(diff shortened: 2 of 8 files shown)


jkelleyrtp:
This is technically a breaking change but since it only fixes cases it's fine!

Issue #2286, opened 2024-04-09
Event handlers are dropped twice


matthunz:
**Problem**

In certain situations event handlers can be dropped twice and panic

Looks to come from
```rust
 #[doc(hidden)]
/// This should only be used by the `rsx!` macro.
pub fn __take(&self) -> ExternalListenerCallback<T> {
    self.callback
        .manually_drop()
        .expect("Signal has already been dropped")
        .expect("EventHandler was manually dropped")
}
```

**Steps To Reproduce**

- Create an event handler in `rsx!`
- TODO, still tracking this down
- 

**Expected behavior**

Make sure event handlers don't drop twice.

**Screenshots**

**Environment:**
 - Dioxus version: v0.5.1
 - Rust version: `nightly`
 - OS info: MacOS
 - App platform: Custom

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [*] I would like to fix and I (maybe) have a solution
- [ ] I don't have time to fix this right now, but maybe later

Issue #2272, opened 2024-04-08
bug(memory leak): Event handlers passed as props are not properly dropped


marc2332:
**Problem**

```rs
#![allow(non_snake_case)]
use dioxus::prelude::dioxus_core::NoOpMutations;
use dioxus::prelude::*;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    fn app() -> Element {
        let mut count = use_signal(|| 0);

        use_hook(|| spawn(async move {
            loop {
                tokio::time::sleep(std::time::Duration::from_nanos(5)).await;
                let val = *count.peek_unchecked();
                if val == 70 {
                    count.set(0);
                } else {
                    count.set(val + 1);
                }
            }
        }));

        rsx! {
            div { 
                color: "red",
                onwheel: move |_| { println!("wheel"); },
                for el in 0..*count.read() {
                    div {
                        key: "{el}",
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        Comp {
                            onclick: move |_| { println!("click"); },
                        }
                        
                    }
                }
            }
        }
    }

    #[component]
    fn Comp(onclick: EventHandler<()>) -> Element {
        rsx! {
            div { 
                onclick: move |_| onclick.call(()),
            }
        }
    }

    // create the vdom, the real_dom, and the binding layer between them
    let mut vdom = VirtualDom::new(app);

    vdom.rebuild(&mut NoOpMutations);

    // we need to run the vdom in a async runtime
    tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()?
        .block_on(async {
            loop {
                // wait for the vdom to update
                vdom.wait_for_work().await;

                // get the mutations from the vdom
                 vdom.render_immediate(&mut NoOpMutations);
            }
        })
}
``` 

**Expected behavior**

Not leak

**Screenshots**

after running for some seconds

![image](https://github.com/DioxusLabs/dioxus/assets/38158676/3e076f78-9d6a-416d-82a7-692408743f3b)

**Environment:**
 - Dioxus version: latest stable 0.5
 - Rust version: stable
 - OS info: Windows
 - App platform: Any

PR #2288, opened 2024-04-09
Fix event handler drops

matthunz:
#2286 Event handlers used as component props may currently be dropped twice or not at all due to a bug in props.

This fixes the error for `Option<EventHandler>` by updating `looks_like_event_handler_type`


marc2332:
> not at all

Unfortunately, my [issue](https://github.com/DioxusLabs/dioxus/issues/2272) persists 🥲

ealmloff (approved):

PR #2289, opened 2024-04-10
Simplify and fix some issues with `#[component]` macro

tigerros:
- Gets rid of a ton of code (almost half of the `component.rs` file + the 130 line long creatively named `utils.rs` file), which wasn't even being used!! This is the doc building that I introduced in [1563](https://github.com/DioxusLabs/dioxus/pull/1563), and I'm sorry for introducing it. It's completely unnecessary because people can simply view the props struct through the `props` argument, and I don't know how it even got merged. Anyway, since then there have been some changes, and the original features made in [1563](https://github.com/DioxusLabs/dioxus/pull/1563) just didn't work. I tested it out and the macro simply would not build any docs. So there was a bunch of useless code, technically being used at compilation, but not being used at runtime. Or maybe it was being used somehow, slowing down compilation but not producing any results.
- Proper casing. I've included a little trick which will actually make the compiler prefer camel case names for the function identifiers. Previous implementations just applied `#[allow(non_snake_case)]` for the entire function, including the body, which would silence other stuff like variables (and also wouldn't make the compiler prefer camel case). This PR makes it so that the macro only changes case lints for the function identifier.
- Allow struct pattern parameter parsing (e.g. `fn Navbar(NavbarProps { title }: NavbarProps)`). This was previously being parsed incorrectly and the macro just wouldn't compile. I noticed this was used a lot in Freya which is probably why Freya doesn't use the `#[component]` macro, instead opting for putting `#[allow(non_snake_case)]` everywhere (which, as mentioned, is incorrect because it silences warnings of other identifiers).
- Update and generally improve the outdated docs.
- When "inlining" props, changes the expansion to be a little cleaner. The change is very minor but it does make generated docs quite a bit cleaner (IMHO).
    ```rust
    #[component]
    fn Foo(bar: usize) ...

    // old expansion:
    fn Foo(mut __props: FooProps) ...
    // rustdoc:
    fn Foo(__props: FooProps) ...

    // new expansion:
    fn Foo(FooProps { bar }: FooProps) ...
    // rustdoc:
    fn Foo(_: FooProps) ...
    ```


ealmloff:
Expanding to a struct will cause lints on any components that use the `component_name` syntax with an underscore instead of upper camel case

tigerros:
> Expanding the component name to a struct will cause lints on any components that use the `component_name` syntax with an underscore instead of upper camel case

That's the point, since the convention is upper camel case and not snake case.

jkelleyrtp:
> > Expanding the component name to a struct will cause lints on any components that use the `component_name` syntax with an underscore instead of upper camel case
> 
> That's the point, since the convention is upper camel case and not snake case.

Love the fact that we're trimming the fat in this module - I was very confused by the complexity of it all last time I was digging around here.

Can we keep support for lowercase components? They're basically our unofficial syntax for wrapped web components. IE you'd wrap a stringly typed `kebab-case-element {}` with a `strongly_typed_element {}`.

We don't do anything differently with codegen there, per se, but it's a decent heuristic when glancing at code as to why those components are different.

jkelleyrtp (changes requested):
Love the fact that we're trimming the fat in this module - I was very confused by the complexity of it all last time I was digging around here.

Can we keep support for lowercase components? They're basically our unofficial syntax for wrapped web components. IE you'd wrap a stringly typed kebab-case-element {} with a strongly_typed_element {}.

We don't do anything differently with codegen there, per se, but it's a decent heuristic when glancing at code as to why those components are different.

tigerros:
> Can we keep support for lowercase components? They're basically our unofficial syntax for wrapped web components. IE you'd wrap a stringly typed kebab-case-element {} with a strongly_typed_element {}.

Done. I just want to point out that they're definitely "supported", they will just trigger a warning (which you can easily suppress with `#[allow(non_camel_case_types)]`), not an error. I thought it made sense because camel case is the convention, but I guess I didn't consider web components.

packages/core-macro/src/component.rs
```diff
@@ -22,12 +22,13 @@ impl ToTokens for ComponentBody {
...
+        // Components can also use the struct pattern to "inline" their props.
+        // Freya uses this a bunch (because it's clean),
+        // e.g. `fn Navbar(NavbarProps { title }: NavbarProps)` was previously being incorrectly parsed
+        if self.is_explicit_props_ident() || self.has_struct_parameter_pattern() {
             let comp_fn = &self.item_fn;
-            tokens.append_all(quote! {
-                #[allow(non_snake_case)]
-                #comp_fn
-            });
+            tokens.append_all(allow_camel_case_for_fn_ident(comp_fn).into_token_stream());
```
jkelleyrtp, line 31: Let's keep this.

packages/core-macro/src/component.rs
```diff
@@ -75,37 +74,32 @@ impl ComponentBody {
...
-            true => quote! {},
-            false => quote! { mut __props: #struct_ident #ty_generics },
-        };
-        let expanded_struct = match inputs.is_empty() {
-            true => quote! {},
-            false => quote! { let #struct_ident { #(#struct_field_names),* } = __props; },
+        let inlined_props_argument = if inputs.is_empty() {
+            quote! {}
+        } else {
+            quote! { #struct_ident { #(#struct_field_names),* }: #struct_ident #impl_generics }
```
jkelleyrtp, line 92: and this as well.

examples/rsx_usage.rs
```diff
@@ -197,12 +197,6 @@ fn app() -> Element {
             Label { text: "hello geneirc world!" }
             Label { text: 99.9 }
 
-            // Lowercase components work too, as long as they are access using a path
-            baller::lowercase_component {}
-
-            // For in-scope lowercase components, use the `self` keyword
-            self::lowercase_helper {}
-
```
jkelleyrtp, line 205: we want to keep lower_case component support without warnings.

jkelleyrtp (changes requested):
Reviewing this now, sorry for the wait.

We want to keep lowercase component support, but also allow the new deconstructed props syntax.

Also I think keeping the extra props doc is nice - it helps with discovery on optional fields.

ealmloff (approved):

jkelleyrtp (approved):

PR #2298, opened 2024-04-12
Fix event handler memory leak

### LINKED ISSUE #2272: bug(memory leak): Event handlers passed as props are not properly dropped

### LINKED ISSUE #2286: Event handlers are dropped twice

### LINKED ISSUE #2342: `ValueDroppedError when redrawing` in `dioxus-core-0.5.1/src/events.rs`
Dropped(ValueDroppedError { created_at: Location { file: "/home/niedzwiedz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-core-0.5.1/src/events.rs", line: 177, col: 9 } })

I spent entire day trying to reproduce it in a separate repo (it's a closed source project) but I was unable to

ealmloff:
This PR fixes two bugs with event handlers:
1) EventHandler::from was changed to leak memory by default in #2245 in a minor release which makes traits like the `From` implementation leak memory. This PR reverts that change and moves the leaky version to a `leak` method. We might want to remove the normal new method in the next major release, but currently it is very unintuitive that calling `.into` leaks memory
2) EventHandler in props took the new value of the event handler which breaks when you clone elements like this:
```rust
use dioxus::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {
    let mut count = use_signal(|| 10);

    let x = rsx! {
        TakesEventHandler {
            onclick: move |_| {},
        }
    };

    rsx! {
        button { onclick: move |_| count -= 1, "Down!" }
        for _ in 0..count() {
            // When diffing onclick is moved out of the TakesEventHandler multiple times which panics
            {x.clone()}
        }
    }
}

#[component]
fn TakesEventHandler(onclick: EventHandler<MouseEvent>) -> Element {
    None
}
```
This PR changes the behavior to clone event handler out of the props instead of taking the data

Fixes #2286
Fixes #2272
Fixes #2342


ealmloff:
The same bug exists with `ReadOnlySignal`, but I don't think we can fix that issue without breaking changes because we don't own the type inside ReadOnlySignal

Niedzwiedzw:
this fixes an app I've put quite a lot of work into, so I vote let's merge this :D

jkelleyrtp:
> this fixes an app I've put quite a lot of work into, so I vote let's merge this :D

Will get it in and do another patch release today!

Niedzwiedzw:
I'd like to point out that your response times are super impressive guys, kudos!

ealmloff:
There is something wrong with generational box recycling boxes. Working on a fix

PR #2300, opened 2024-04-12
Fix: CLI Compilation Spam

### LINKED ISSUE #2221: bug: Bloated CLI rendering in certain terminal emulators
Wezterm:

image

every update causes a new line break

DogeDark:
Closes #2221 

**Cause**
https://github.com/console-rs/indicatif/issues/638

**Solution**
Switch the cog emoji to a similar one.


ealmloff (approved):

Silvea12:
I want to make sure this is relayed here: https://github.com/console-rs/indicatif/issues/638#issuecomment-2080314224
`unicode-width` has just had an update that fixes this issue, so indicatif may end up with it fixed upstream shortly!

PR #2303, opened 2024-04-13
fix: Update `use_resource` docs

marc2332:


onweru:
`current_weather()` shorthand also works. I like it because it better as it's less verbose. Could have been what the one who edited that bit was going for 🤞🏼 

It's documented here https://github.com/DioxusLabs/dioxus/blob/05b662ee2d2a9b17c4b094df5e4e81c9a05c37c8/packages/hooks/src/use_resource.rs#L269

marc2332:
> `current_weather()` shorthand also works. I like it better because it's less verbose. Could have been what the one who edited that bit was going for 🤞🏼
> 
> It's documented here
> 
> https://github.com/DioxusLabs/dioxus/blob/05b662ee2d2a9b17c4b094df5e4e81c9a05c37c8/packages/hooks/src/use_resource.rs#L269

Calling a signal will clone it's inner value, which might not be desirable in all the cases, hence I think it's better to use `read()` for this kind of examples 😄

onweru:
> > `current_weather()` shorthand also works. I like it better because it's less verbose. Could have been what the one who edited that bit was going for 🤞🏼
> > It's documented here
> > https://github.com/DioxusLabs/dioxus/blob/05b662ee2d2a9b17c4b094df5e4e81c9a05c37c8/packages/hooks/src/use_resource.rs#L269
> 
> Calling a signal will clone it's inner value, which might not be desirable in all the cases, hence I think it's better to use `read()` for this kind of examples 😄

Hadn't considered that. Fair point, thanks!

ealmloff (approved):

PR #2308, opened 2024-04-14
Replace `AttributeValue::Text` with `Cow<'static, str>`

matthunz:
```rust
pub enum AttributeValue {
    Text(Cow<'static, str>),
    ...
}
```

This would allow for less clones when using text nodes.

However this does introduce the tradeoff of not automatically converting `&str` to `String`, only `&'static str` is permitted to reduce cloning.

Definitely open to changing any of this! 😄

### DIFF
--- a/packages/core/src/nodes.rs
+++ b/packages/core/src/nodes.rs
@@ -953,6 +953,12 @@ impl IntoAttributeValue for AttributeValue {
     }
 }
 
+impl IntoAttributeValue for Cow<'static, str> {
+    fn into_value(self) -> AttributeValue {
+        AttributeValue::Text(self)
+    }
+}
+
 impl IntoAttributeValue for &'static str {
     fn into_value(self) -> AttributeValue {
         AttributeValue::Text(self.into())
--- a/packages/core/src/nodes.rs
+++ b/packages/core/src/nodes.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use crate::{Properties, VirtualDom};
 use core::panic;
+use std::borrow::Cow;
 use std::ops::Deref;
 use std::rc::Rc;
 use std::vec;
@@ -711,7 +712,7 @@ impl Attribute {
 /// variant.
 pub enum AttributeValue {
     /// Text attribute
-    Text(String),
+    Text(Cow<'static, str>),
 
     /// A float
     Float(f64),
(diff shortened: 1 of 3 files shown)


onweru:
> However this does introduce the tradeoff of not automatically converting &str to String, only &'static str is permitted to reduce cloning.

Does this mean I won't be able to do this

```
#[component]
fn Text(class: String) {
  rsx! { 
     div { class, ... }
  }
}

Text {  class: "some-class" }
```
Hated having to

```
Text {  class: "some-class".into() } // or the to_string() 
```

matthunz:
> > However this does introduce the tradeoff of not automatically converting &str to String, only &'static str is permitted to reduce cloning.
> 
> Does this mean I won't be able to do this
> 
> ```
> #[component]
> fn Text(class: String) {
>   rsx! { 
>      div { class, ... }
>   }
> }
> 
> Text {  class: "some-class" }
> ```
> 
> Hated having to
> 
> ```
> Text {  class: "some-class".into() } // or the to_string() 
> ```

Yes unfortunately this change would require that. I'd saw the optional solution is `#[props(into) class: Cow<'static, str>` to avoid having to allocate the `String`

ealmloff:
I don't think the added complexity is worth it here, but I would be interested to see benchmarks if this has a large performance impact.

Dioxus will currently not allocate with static strings inline in the rsx:

```rust
rsx! {
   div {
      // "hello world" is included in the template which only supports &'static str attributes
      width: "hello world",
   }
}
```

This PR would enable less allocation for the following code:

```rust
const STATIC: &str = "hello world";
const STATICS: &[&str] = &["hello world"];

let index = use_signal(|| 0);
rsx! {
   div {
      // Theoretically this could be part of the template as well, but we are currently not smart enough to know that it is static
      width: STATIC,
      // This shouldn't be part of the template because it is dynamic
      height: STATICS[index()],
   }
}
```

There is also some overlap with https://github.com/DioxusLabs/dioxus/pull/2258/files#diff-b96a80b4b2e2f0da84fe76143d9f213d159383fd322e41068eba85fcd7fe93f3 which introduces string type with static and dynamic segments (currently only used in debug mode for hot reloading)

matthunz:
TODO: Greg Johnson's new [oco_ref](https://docs.rs/oco_ref/0.1.1/oco_ref/l) crate might fit great here 👀

PR #2314, opened 2024-04-16
fix path of set_server_url

chungwong:
`set_server_url` is no longer directly under the [prelude](https://github.com/DioxusLabs/dioxus/blob/main/packages/fullstack/src/lib.rs#L40)


ealmloff (approved):
Thank you!

Issue #1333, opened 2023-08-09
Streaming File Uploads


ealmloff:
## Specific Demand

The current [FileEngine](https://docs.rs/dioxus/latest/dioxus/events/trait.FileEngine.html) exposes two ways to read a file: Reading the file to a string or reading the file to a byte array. Both of these methods perform poorly with large files. In addition to poor performance, you cannot provide updates to the user based on file upload progress.

## Implement Suggestion

We should expose a third API that returns the size of the file (depending on the platform) and a [stream](https://docs.rs/futures-util-preview/latest/futures_util/stream/index.html) with the file data

Related to https://github.com/DioxusLabs/dioxus/pull/1181

rogusdev:
Reminder on file size visibility: https://github.com/DioxusLabs/dioxus/pull/2323

And for the stream part, in my own web only project with file uploads, I use: `web_file: web_sys::File` and with https://github.com/MattiasBuelens/wasm-streams :
```
    let mut stream = ReadableStream::from_raw(web_file.stream());
    let reader = stream.get_byob_reader();
    let mut reader = ChunkedReader::new(reader, MAX_ALLOWED_SIZE_CHUNK_PADDED);

    while let Some(res) = reader.read_chunk().await {
```

PR #2323, opened 2024-04-17
add file size to FileEngine

rogusdev:
This allows file sizes to be checked *before* loading the entire file into memory (e.g. with `read_file`) when it might be too big. (And without requesting the native file to do this check manually.)

### DIFF
--- a/packages/html/src/file_data.rs
+++ b/packages/html/src/file_data.rs
@@ -20,6 +20,11 @@ impl FileEngine for SerializedFileEngine {
         self.files.keys().cloned().collect()
     }
 
+    async fn file_size(&self, file: &str) -> Option<u64> {
+        let file = self.files.get(file)?;
+        Some(file.len() as u64)
+    }
+
     async fn read_file(&self, file: &str) -> Option<Vec<u8>> {
         self.files.get(file).cloned()
     }
@@ -41,6 +46,9 @@ impl FileEngine for SerializedFileEngine {
 pub trait FileEngine {
     // get a list of file names
     fn files(&self) -> Vec<String>;
+
+    // get the size of a file
+    async fn file_size(&self, file: &str) -> Option<u64>;
 
     // read a file to bytes
     async fn read_file(&self, file: &str) -> Option<Vec<u8>>;
--- a/packages/html/src/native_bind/native_file_engine.rs
+++ b/packages/html/src/native_bind/native_file_engine.rs
@@ -25,6 +25,11 @@ impl FileEngine for NativeFileEngine {
             .collect()
     }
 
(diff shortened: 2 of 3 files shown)


rogusdev:
Note: there were no automated tests for any of the affected structs that I could see, so I simply tested with:
```
cd packages/web && cargo build && cd -
cd packages/html && cargo build -F native-bind && cd -
```

Which seemed to catch any compiler errors in the affected code, at least. And, ofc, it works in my own application depending on the new feature :) (manually tested web only tho)

rogusdev:
I don't know how I affected the failing workflow test.

As for the labels: this is a new function that should be indiscernible to any users who don't reach for it. How is it "breaking"? Which is to say, how does this project define "breaking" changes?

Thanks for starting to review it!

jkelleyrtp:
We have a flakey windows test that needs to be resolved - it's not a blocker for your PR!

We define breaking changes as changes that are not semver compatible and I believe adding a new non-default method to a trait is breaking.

https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default

If the trait provided a default (IE read the file and then measure it) then this could get merged as part of 0.5 release cycle. We haven't started doing it yet, but our flow involves keeping a stable branch (0.5) and a nightly branch (main) and then backporting fixes onto the stable branch. However we're currently still in the release phase where everything is fixes and all the new features are being held for a bit.

rogusdev:
I'm in no rush personally -- I'm using my local version now.

I suspect that part of semver is referring to libraries where others are implementing that trait. Certainly then it would be a breaking changes, but that seems out of place in the context of dioxus. That said, I defer to your process -- and certainly bug fixes are more important than this right now!

Thanks :)

ealmloff:
Flakey windows tests are fixed in https://github.com/DioxusLabs/dioxus/pull/2332 🙂

ealmloff (commented):
I completely agree an API like this needs to exist, but I'm not sure this is what the API should look like long term.

There is some overlap between a file size API, and a file streaming API (https://github.com/DioxusLabs/dioxus/issues/1333). Once file streaming is added, I think it would be better to include everything in a single file object that supports reading the size, reading the stream, name, etc.

```rust
// Getting a file object should be very cheap
for file in file_event.files() {
    println!("name: {}", file.name());
    println!("size: {}", file.size());
    println!("contents: {}", file.read().await);
}
```

That said, a nicer object based file API would require other breaking changes to the current API, so if this were non-breaking it could be fine for `0.5`

rogusdev:
For sure longer term this should get incorporated into something better, definitely agree. I think including this sooner rather than later allows the need for those bigger changes to be clearer, as well as what they should look like.

jkelleyrtp (approved):

PR #2332, opened 2024-04-17
Fix flakey windows tests

ealmloff:
We have had some tests that are flakey on windows for a while. This has caused some actual failures to slip by in the past and makes windows CI useless

It looks like we were not running doc tests as well. This fixes that issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2489,6 +2489,7 @@ dependencies = [
 name = "dioxus-lib"
 version = "0.5.2"
 dependencies = [
+ "dioxus",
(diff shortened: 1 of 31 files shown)


jkelleyrtp:
I think your rsx autofmt might be eating keys and duplicating entries, here's a build with the updated extension with all the fixes

[dioxus-0.5.0.vsix.zip](https://github.com/DioxusLabs/dioxus/files/15017167/dioxus-0.5.0.vsix.zip)

jkelleyrtp (approved):

PR #2337, opened 2024-04-18
nonreactive examples for hooks were not compilable as written

rogusdev:
My code was out of date, so I missed https://github.com/DioxusLabs/dioxus/commit/b6d3da2b31c336fdf38845a790969edda9f8e0e7 -- but I still caught one more, and added an example. Also tweaked the comments.


rogusdev:
Notably: for the example, I did not call `&*res.read_unchecked()` as the docs suggest on https://dioxuslabs.com/learn/0.5/reference/use_resource -- instead choosing to simply do `res()`, which definitely works. If the longer syntax is preferred / recommended, I would like to update my example to keep best practices in all places, but I'd also like to hear why, thanks.

examples/router_resource.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[component]
+fn App() -> Element {
+    rsx! {
+        Router::<Route> {}
+    }
+}
+
+#[component]
+fn Blog(id: i32) -> Element {
```
ealmloff, line 29: This is a really good use case for `ReadOnlySignal<i32>`. We automatically convert `T` to `ReadOnlySignal<T>` in the props so it will work out of the box with the router and it will be reactive without the extra `use_reactive` hook

examples/router_resource.rs
```diff
@@ -0,0 +1,65 @@
...
+        n
+    }
+
+    // if you use the naive approach, the "Blog post {id}" below will never update when clicking links!
+    // let res = use_resource(move || future(id));
+
+    // the use_reactive hook is required to properly update when clicking links to this component, from this component
+    let res = use_resource(use_reactive(&id, |id| future(id)));
+
+    match res() {
```
ealmloff, line 40: Using `res()` here is recommended over `res.read_unchecked()` because the result of the resource is very cheap to clone. `read_unchecked` should be preferred when you don't want to clone the data

examples/router_resource.rs
ealmloff: use_reactive is good for non-reactive state that is created in the middle of a component like this:
```rust
let state = use_signal(|| 0);
let doubled = state() * 2;
use_effect(use_reactive((&doubled,), |(doubled,)| println!("Manually manipulate the dom")));
```

rogusdev: Ok, that makes sense, however 2 questions then:
1. Is that distinction documented somewhere? After this PR gets merged (in some form), I already intend to update https://dioxuslabs.com/learn/0.5/reference/use_resource with a link to the example and an explanation
2. If it is already done with props, so should it happen with the `#[component]` approach as well? As in, should this just work without the `use_reactive` or explicit `ReadOnlySignal` at all?

examples/router_resource.rs
rogusdev: That sounds very much like something worth documenting! Would https://dioxuslabs.com/learn/0.5/reference/use_resource be the right place, or is there somewhere else better, in your mind?

examples/router_resource.rs
ealmloff: > 1. Is that distinction documented somewhere? After this PR gets merged (in some form), I already intend to update https://dioxuslabs.com/learn/0.5/reference/use_resource with a link to the example and an explanation

Not yet

> 2. If it is already done with props, so should it happen with the `#[component]` approach as well? As in, should this just work without the `use_reactive` or explicit `ReadOnlySignal` at all?

The conversion happens when you use the component if you explicitly use `ReadOnlySignal<T>` in your props or `#[component]`.

```rust
// Because value is a signal, it is reactive and any reads will automatically add the signal as a dependency to use_effect, use_resource, etc
#[component]
fn Comp(value: ReadOnlySignal<i32>) -> Element {None}

Comp {
   // A new signal is created here
   value: 0
}
```

rogusdev: I will update the example, and also add documentation around this.

ealmloff (approved):
Thank you! Non reactive props are also documented better after #2460

examples/router_resource.rs
```diff
@@ -0,0 +1,65 @@
...
+#[component]
+fn Blog(id: ReadOnlySignal<i32>) -> Element {
+    async fn future(n: i32) -> i32 {
+        n
+    }
+
+    // if you use the naive approach, the "Blog post {id}" below will never update when clicking links!
+    // let res = use_resource(move || future(id));
+
+    // the use_reactive hook is required to properly update when clicking links to this component, from this component
```
rogusdev, line 37: You merged this with comments that no longer apply because you switched to readonlysignal

ealmloff: Oops, fixed in 77b08f2330ebf6ad750adbcfd9f28967509765d4

rogusdev: Thanks! The header also mentions use_reactive, and honestly I wonder if using both use_reactive and readonlysignal in this example would help guide people to when to use one vs the other. Or is there a different example we should point at in the header for use_reactive, for comparison?

ealmloff: The [use_reactive docs](https://github.com/DioxusLabs/dioxus/blob/77b08f2330ebf6ad750adbcfd9f28967509765d4/packages/hooks/docs/use_resource.md?plain=1#L112-L141) also show how to use both ReadOnlySignals and `use_reactive` with `use_resource`. Any time you can turn your value into a reactive value with readonlysignal, I would recommend that approach. If you have a raw rust value that you can't make reactive, you can use `use_reactive` instead

PR #2338, opened 2024-04-18
Add access to the Element attributes related to scrolling

ASR-ASU:
Add the capability to access to the following Element attributes related to the scrolling:

- https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight
- https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft
- https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop
- https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth


ealmloff:
Thanks for adding support for these methods! These would be great to support in dioxus.

Instead of exposing the browser methods directly, we could merge some of the methods into more rusty versions like we do with other event methods today. Maybe something like this:

- `get_scroll_offset` that returns [Vector2D](https://docs.rs/euclid/0.22.9/euclid/struct.Vector2D.html)
- `get_scroll_size` that returns [Size2D](https://docs.rs/euclid/0.22.9/euclid/struct.Size2D.html)

ASR-ASU:
Should I rather use (or add new ones)  the primitives defined by the https://github.com/DioxusLabs/dioxus/blob/main/packages/html/src/geometry.rs instead of euclid types directly?

ealmloff:
> Should I rather use (or add new ones) the primitives defined by the https://github.com/DioxusLabs/dioxus/blob/main/packages/html/src/geometry.rs instead of euclid types directly?

Reusing the unit type definitions from geometry would be great

ealmloff (approved):
Thank you, this looks great!

PR #2351, opened 2024-04-20
enables ability to turn off hot reloading

### LINKED ISSUE #2233: `dx`'s help message shows that the hot-reloading is disabled by default, which is false (no pun intended)
**Problem**

```
      --hot-reload
          Build with hot reloading rsx. Will not work with release builds. [default: false]
```

It should say `true`. But how would you disable it? Can you pass in true/false? Maybe there should be a `--no-hot-reload` option instead.

**Steps To Reproduce**

- `dx -h`
- see the thing

**Expected behavior**

The message should say `true`.

**Environment:**
 - Dioxus version: v0.5.4
 - Rust version: 1.79.0, nightly
 - OS info: Pop_!OS
 - App platform: fullstack

**Questionnaire**

alexanderjophus:
Resolves #2233 

Changes the description to accurately reflect the default.

Adds ability to set the `hot_reload` to either `true` or `false`, requiring an `=`

### DIFF
--- a/packages/cli/src/cli/cfg.rs
+++ b/packages/cli/src/cli/cfg.rs
@@ -117,8 +117,14 @@ pub struct ConfigOptsServe {
     #[clap(long, value_enum)]
     pub platform: Option<Platform>,
 
-    /// Build with hot reloading rsx. Will not work with release builds. [default: false]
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thank you!

PR #2381, opened 2024-05-02
Improve the base_path story

pyrrho:
This PR updates how the the axum router constructed for `dx serve`, the `--open` option, and the dioxus-router `WebHistory` interact with the Dioxus.toml `base_path` field.

Prior to this change, running `dx serve` with a `base_path` set always resulted in my browser displaying the "Outside of the base path: blah" message, even when I was navigating to routes that were very definitely inside of the base path.

Three major changes are included in this PR;
- Setting a `base_path` now causes the `dx serve` router to be nested under the base path, rather than treating the base path as a route that is handled with a `axum::routing::any_service`. 
- The `--open` flag now opens to `localhost:8080/<base_path>` rather than to `localhost:8080`.
- The dioxus router's default web history initializes to a URL that includes the base_path, rather than to the root.  
  This last change is actually what I consider to be the most important, as it impacts dioxus web applications hosted outside of the `dx serve` axum server. Without this change, navigating to e.g. `localhost:8080/base/path/` would cause the browser address bar to resolve to `localhost:8080`, even though following a `Link { to: Route::Home {}, .. }` would push `localhost:8080/base/path/` onto the history stack.

### DIFF
--- a/packages/cli/src/server/web/server.rs
+++ b/packages/cli/src/server/web/server.rs
@@ -62,8 +62,8 @@ pub async fn setup_router(
         .and_then(move |response| async move { Ok(no_cache(file_service_config, response)) })
         .service(ServeDir::new(config.out_dir()));
 
-    // Setup websocket
-    let mut router = Router::new().route("/_dioxus/ws", get(ws_handler));
+    // Setup router
+    let mut router = Router::new();
 
     // Setup proxy
     for proxy_config in config.dioxus_config.web.proxy {
@@ -81,9 +81,16 @@ pub async fn setup_router(
     ));
 
     router = if let Some(base_path) = config.dioxus_config.web.app.base_path.clone() {
-        let base_path = format!("/{}", base_path.trim_matches('/'));
+        // NB. The trailing `/` is required for Dioxus Routers to match on root paths. Without it,
+        // navigating to e.g. `/base/path/` results in the below fallback being hit, and users
+        // receiving a message that `/base/path/` is "Outside of the base path: /base/path".
+        // Navigating to e.g. `/base/path` results in the below fallback  _not_ being hit, but a
+        // panic getting thrown from `WebHistory::route_from_location`. Interestingly, the
+        // `RouteParseError` passed into the panic reports that attempts were made to match on no
+        // routes at all.
+        let base_path = format!("/{}/", base_path.trim_matches('/'));
         Router::new()
-            .route(&base_path, axum::routing::any_service(router))
+            .nest(&base_path, router)
             .fallback(get(move || {
                 let base_path = base_path.clone();
                 async move { format!("Outside of the base path: {}", base_path) }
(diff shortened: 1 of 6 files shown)


packages/cli/src/server/web/server.rs
```diff
@@ -92,6 +99,9 @@ pub async fn setup_router(
         router
     };
 
+    // Setup websocket
+    router = router.route("/_dioxus/ws", get(ws_handler));
```
pyrrho, line 103: Calling out this change -- related to the change on :65-66 -- as I'm not 100% sure how this will interact with the proxy configuration on :69-71. I know the order of some methods called on axum routers matter, but I'm not terribly familiar with the framework. I'm also not sure how to correctly test this interaction.

packages/cli/src/server/web/server.rs
```diff
@@ -81,9 +81,16 @@ pub async fn setup_router(
...
     router = if let Some(base_path) = config.dioxus_config.web.app.base_path.clone() {
-        let base_path = format!("/{}", base_path.trim_matches('/'));
+        // NB. The trailing `/` is required for Dioxus Routers to match on root paths. Without it,
+        // navigating to e.g. `/base/path/` results in the below fallback being hit, and users
+        // receiving a message that `/base/path/` is "Outside of the base path: /base/path".
+        // Navigating to e.g. `/base/path` results in the below fallback  _not_ being hit, but a
+        // panic getting thrown from `WebHistory::route_from_location`. Interestingly, the
+        // `RouteParseError` passed into the panic reports that attempts were made to match on no
+        // routes at all.
+        let base_path = format!("/{}/", base_path.trim_matches('/'));
```
pyrrho, line 91: After looking at the router-macro source a bit more, I realize this is (probably?) intentional behavior. A [check](https://github.com/ealmloff/dioxus/blob/58304b7a69c123f842e05ab5814ef89a5d469619/packages/router-macro/src/lib.rs#L491) introduced in #1272 explicitly rejects all routes without a leading slash.

I'm not sure if there should be a special case in the generated `from_str` method for the `s.trim().len() == 0` case, or if I'm misunderstanding something about canonical URI construction?

ealmloff: If I remember correctly, the check was introduced so that the route `href=bla` is not parsed as `href=/bla` in a link because browsers do not interpret it that way. Parsing an empty route we get from the browser as the base route seems unproblematic

packages/cli/src/server/web/server.rs
ealmloff: I believe axum matches exact routes first. I tested that hot reloading still works if the base_path is `_dioxus` and the proxy ends in `_dioxus`. The proxy just adds new routes, not layers so I don't think this will effect behavior

ealmloff (approved):
This works much better. Thank you!

pyrrho:
Thank you for finalizing this PR, and buttoning up the big ugly comment!

Issue #2285, opened 2024-04-09
base_path needed for deploy, but breaks local serve


hkBst:
**Problem**

I need to set [web.app]base_path in Dioxus.toml to get deployment on Github Pages working. But then locally the app refuses to work as it is replaced by an error message ("Outside of the base path: /your-path"). Is it possible to get both to work?

**Environment:**
 - Dioxus version: 0.5.1
 - Rust version: 1.76.0
 - OS info: Linux
 - App platform: web

s1gtrap:
I had some success using the latest CLI instead (thanks to #2381):
```bash
cargo install --git https://github.com/DioxusLabs/dioxus.git dioxus-cli
```
With that, fetching `http://localhost:8080/<base_path>` seems to respond correctly, although `http://localhost:8080/<base_path>/` still yields `Outside of the base path: /<base_path>` (is this intended?)

This wasn't all the way there in my case as assets weren't loading properly. Naturally, as there was no trailing `/` in the URL, all relatively addressed assets were being loaded from the root instead of after the `base_path`, despite them (correctly) only being served after the `base_path`.

So I tried adding a trailing `/` in `packages/cli/src/server/web/server.rs` and it's actually working as intended now. In fact, it seems to be working much better, as even `http://localhost:8080/<base_path>` is redirected to `http://localhost:8080/<base_path>/` ensuring all assets are available regardless of how the address is entered.

It's available here: https://github.com/s1gtrap/dioxus/tree/fix-relative-assets-w-base-path but I'll create a PR in a sec.

pyrrho:
@s1gtrap I can share some details from my digging for #2381 and #2395

> ... fetching `http://localhost:8080/<base_path>` seems to respond correctly, although `http://localhost:8080/<base_path>/` still yields `Outside of the base path: /<base_path>` (is this intended?)

[...]

PR #2403, opened 2024-05-10
Fetch relative assets from `base_path` instead of root when `base_path` is set

s1gtrap:
I also ran into #2285, and went ahead and had a look. Turns out #2381 fixed most of it, but I believe it needs a trailing `/` in order to serve assets correctly. Without a trailing `/`, assets that are relatively addressed will search in the root path instead of relative to the `base_path`, causing 404s as assets are (correctly) served from the `base_path`.

This seems as if it was the intended use of `Router::nest` as requests to `/<base_path>` are redirected to `/<base_path>/` automatically.

The alternative would be to hardcode the `base_path` in the assets but I don't think this is a very intuitive approach.

### DIFF
--- a/packages/cli/src/server/web/server.rs
+++ b/packages/cli/src/server/web/server.rs
@@ -81,7 +81,7 @@ pub async fn setup_router(
     ));
 
     router = if let Some(base_path) = config.dioxus_config.web.app.base_path.clone() {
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
This unfortunately doesn't work since base_path *needs* to be set as part of the absolute path very every asset.

PR #2415, opened 2024-05-15
Add test for using `use_memo` with `use_effect`

pablosichert:
This PR is a reproduction test case for #2416.

Currently, when starting with `(0, 0)` and writing

```
Write(Some(0), None)
Write(None, Some(0))
Write(Some(0), Some(0))
Write(Some(1), None)
Write(None, Some(1))
Write(Some(1), Some(1))
Write(Some(2), Some(2))
Write(Some(0), Some(0))
```

this test results in the following reads:

```
Read(0, 0)
Read(1, 1)
Read(1, 1)
Read(0, 0)
Read(0, 0)
```

However, the expected result is

```
Read(0, 0)
Read(1, 0)
Read(1, 1)
Read(2, 2)
Read(0, 0)
```

### DIFF
--- a/packages/hooks/tests/memo.rs
+++ b/packages/hooks/tests/memo.rs
@@ -68,3 +68,111 @@ async fn memo_updates() {
         println!("Signal: {signal:?}");
     }
 }
+
+#[tokio::test]
+async fn use_memo_with_use_effect() {
+    use dioxus::prelude::*;
+    use futures_util::future::select;
+    use std::{cell::RefCell, collections::VecDeque, pin::pin, rc::Rc};
+    use tokio::sync::mpsc;
+
+    let (action_sender, mut action_receiver) = mpsc::unbounded_channel::<Action>();
+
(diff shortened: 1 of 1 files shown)


packages/hooks/tests/memo.rs
```diff
@@ -68,3 +68,111 @@ async fn memo_updates() {
...
+        use_effect(move || {
+            let (a, b) = tuple();
+            action_sender
+                .0
+                .borrow_mut()
+                .send(Action::Read(a, b))
+                .unwrap();
+        });
+
+        None
```
pablosichert, line 134: Ok, interesting, if you change this line to:

```suggestion
        rsx! {"{tuple:?}"}
```

The test passes.

ealmloff:
The issue with memos over subscribing is fixed in #2506, but this test will still fail because it assumes that effects rerun once every time a dependency updates. They may drop updates freely as long as the latest value is consistent (in this test `(0, 0)`). I added a test to #2506 that tests for regressions in synchronous updates with memo

Issue #2239, opened 2024-04-04
Bundle error on Linux: dx bundle copy the dist/bundle into itself endlessly


SadCl0wn:
**Problem**

If i try to bundle a project, I get the following error:

```
Adding assets from dist to bundle
thread 'main' panicked at packages/cli/src/cli/bundle.rs:193:13:
Failed to bundle project: BundlerError(
    Error {
        context: "Failed to build data folders and files",
        source: BundlerError(
            Error {
                context: "Failed to copy resource files",
                source: IoError(
                    Os {
                        code: 36,
                        kind: InvalidFilename,
                        message: "File name too long",
                    },
                ),
            },
        ),
    },
)
```

in Dioxius the out_dir is "dist" and looking into the content of `dist/bundle` there is a recursive patern:
`packages/cli/dist/bundle/deb/dx_0.5.4_amd64/data/usr/lib/dx/dist/bundle/deb/dx_0.5.4_amd64/data/usr/lib/dx/dist/bundle/deb/dx_0.5.4_amd64/data/usr/lib/dx/dist/bundle/deb/dx_0.5.4_amd64/data/usr/lib/dx/dist/bundle`
it seems that during the bundle process, bundle try to copy the `out_dir` content into `out_dir/bundle` but doesn't ignore the content of `out_dir/bundle` itself

**Steps To Reproduce**

- cd into `packages/cli`
- run `cargo run -- bundle`

for the context, I first tried to bundle a project of my own

**Expected behavior**

AppImage and other bundle files being generated

**Environment:**
 - Dioxus version: origin/main and 5.4.0
 - Rust version:    stable-x86_64-unknown-linux-gnu - rustc 1.79.0-nightly  and 1.77.1
 - OS info: Fedora 40 (linux)
 - App platform: `desktop`

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Beiri22:
the same happens here on Arch

tkr-sh:
Yeah, it don't really make sens for it to be specific to fedora if the pkg manager is not involved.

SadCl0wn:
For the fedora part, I suspected that it wasn't linked to SELinux being activated or other tweaks, but I did not try on other distros.

[...]

PR #2419, opened 2024-05-16
Fix recursive copy while bundling

ealmloff:
This PR changes dx bundle to copy individual assets instead of the whole dist directory to avoid recursively copying into itself

Fixes https://github.com/DioxusLabs/dioxus/issues/2239

### DIFF
--- a/packages/cli/src/cli/bundle.rs
+++ b/packages/cli/src/cli/bundle.rs
@@ -1,6 +1,6 @@
 use core::panic;
 use dioxus_cli_config::ExecutableType;
-use std::{fs::create_dir_all, str::FromStr};
+use std::{env::current_dir, fs::create_dir_all, str::FromStr};
 
 use tauri_bundler::{BundleSettings, PackageSettings, SettingsBuilder};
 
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
> I tested this PR, and it does resolve the infinite copy. But I now get this error [tauri-apps/tauri#8929](https://github.com/tauri-apps/tauri/issues/8929) when building the appimage and I need to manually delete the folder dist/bundle between tries as the script get:
> 
> ```
> thread 'main' panicked at packages/cli/src/cli/bundle.rs:218:13:
> Failed to bundle project: IoError(
>     Os {
>         code: 13,
>         kind: PermissionDenied,
>         message: "Permission denied",
>     },
> )
> ```
> 
> even thought the permission looked alright `drwxr-xr-x@ - sadclown 21 mai 09:35 bundle` (and the subfolders are the same). I think the issue is that the bundle is in `out_dir` and there is no filter at packages/cli/src/cli/bundle.rs:163:31 tauri would delete the bundle folder content if it weren't in the bundle_settings.resources_map. I got stuck on this when I tried to fix this issue myself ^^', that is why I didn't open a PR.

I'm going to merge this PR and then open this as an issue - it could be an issue with the app image bundler that we need to do some upgrades to fix.

jkelleyrtp (approved):

PR #2436, opened 2024-05-21
Improved fullstack reload_upon_connect timeout

luveti:
The default 8 second timeout is a bit too long for projects that compile faster than that. With this change, we start with a 250 millisecond timeout that doubles with each failed connection attempt, up to a max of 4 seconds. This results in a much faster page reload than before.

Allowing this to be customized by the user would be even better, as they could tune things to fit their project and compilation speed.


ealmloff (approved):
This looks great! Faster reloads while actively developing fullstack will be very nice.

The hot reloading code for fullstack and web was merged in https://github.com/DioxusLabs/dioxus/pull/2226. I'm going to move this over to the new location so it fixes web rebuilds as well

PR #2444, opened 2024-05-23
Fix task dead cancelation race condition

### LINKED ISSUE #2440: Resource does not restart from async event handler
```rust
use dioxus::prelude::*;

pub fn main() {
    launch(app)
}

pub fn app() -> Element {
    let mut num = use_server_future(get_next_number);

    rsx! {
        div {
            // This does work
            // onclick: move |e| {
            //     if let Some(num) = &mut num {
            //         num.restart();
            //     }
            // },
            // This does not work
            onclick: move |e| async move {
                if let Some(num) = &mut num {
                    num.restart();
                }
            },
            if let Some(num) = num {
                match num.value().as_ref() {
                    Some(num) => match num.as_ref() {
                        Ok(num) => rsx! { p { "Next number is {num}" } },
                        Err(e) => rsx! { p { "Loading next number failed, {e}" } },
                    },
                    None =>  rsx! { p { "Loading..." } }
                }
            }
        }
    }
}

#[server]
async fn get_next_number() -> Result<u32, ServerFnError> {
    static mut NUM: u32 = 0;
    unsafe {
        NUM += 1;
        Ok(NUM)
    }
}
```

**Environment:**
 - Dioxus version: master
 - Rust version: nightly
 - OS info: NixOS
 - App platform: fullstack

**Questionnaire**

ealmloff:
#2440 is caused by restarting a future that has already finished. Something like this was happening:
1) resource runs (`TaskId(0)`)
2) resource finishes. The task id is recycled
3) A new task is spawned for the async event handler (`TaskId(0)`)
4) Inside that task, the resource cancels its old future that was already recycled `TaskId(0)`
5) The resource also spawns a new future in an empty id. `TaskId(0)` which we are currently in was already recycled so the new future is again spawned in `TaskId(0)`
6) The async event handler finishes and since the future is resolved it removes the task. This removes `TaskId(0)`
7) Now the second run of the resource which was spawned on `TaskId(0)` has already been dropped so it is never polled

This PR fixes that issue by switching the resource's task to None when the future is finished. It also applies the fix to use_future and use_coroutine.

Fixes #2440


jkelleyrtp:
Can we use the `current_scope` from the runtime to get this information rather than overloading the component functions with the threadlocal? I'm just worried about panics/unwinding or other effects leaving things in a weird state between two threadlocals - would prefer to keep it all on the one runtime threadlocal.

We should have a `current_scope` function on the runtime somewhere that we can check the name/typeid of the current component against the one we're calling.

jkelleyrtp:
Closed because generational indices fixed this in #2488

PR #2450, opened 2024-05-24
Set up example scraping for docs.rs

ealmloff:
This PR sets up [example scraping](https://doc.rust-lang.org/rustdoc/scraped-examples.html) for docs.rs. This should help provide examples for some less documented items in desktop and the router. It also switches to `launch` instead of `launch_desktop` where possible and adds required features for desktop to make it less confusing when trying to run those examples on wasm

Screenshot 2024-05-24 at 11 17 00 AM


jkelleyrtp:
This is so cool!

PR #2456, opened 2024-05-27
Make use of `#[doc]` field attributes in `Props` derive macro

MrGVSV:
## Objective

Currently, it doesn't seem like you can pass the docs from a prop field to its method generated via the `Props` derive. The only way to achieve this is to duplicate the doc into a dedicated `#[props(doc = "My doc comment")]` attribute.

This, however, is both inconvenient and error-prone since users will need to handle the formatting manually and all in a single attribute.

## Solution

Pass along all `#[doc]` attributes on a prop field to the respective function. This allows the documentation to be shared by that method and show up on hover when using the component.

Here's a before/after comparison in Rust Rover:

| Before | After |
| ------ | ----- |
| dioxus_prop_docs--before | dioxus_prop_docs--after |

---

## Changelog

- Removed `#[props(doc)]` attribute
- Documentation on prop fields should now be visible from within `rsx!` invocations

### DIFF
--- a/packages/core-macro/src/props/mod.rs
+++ b/packages/core-macro/src/props/mod.rs
@@ -271,7 +271,7 @@ mod field_info {
     #[derive(Debug, Default, Clone)]
     pub struct FieldBuilderAttr {
         pub default: Option<syn::Expr>,
-        pub doc: Option<syn::Expr>,
+        pub docs: Vec<String>,
         pub skip: bool,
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
Thanks! I'm not sure why this wasn't already implemented properly - I had figured we were already passing along the docs directly.

PR #2460, opened 2024-05-29
Improve inline docs

### LINKED ISSUE #2390: Docs have a couple outdated examples
**Problem**

I found a couple instances where the docs don't match the state of the codebase.

- [ ] [VirtualDom.rebuild](https://docs.rs/dioxus-core/0.5.1/dioxus_core/prelude/struct.VirtualDom.html#method.rebuild) takes one param, but the example has zero params.
- [ ] Main crate doc for dioxus-ssr has [an example](https://docs.rs/dioxus-core/0.5.1/dioxus_core/prelude/struct.VirtualDom.html#method.rebuild) that uses `dioxus_ssr::Renderer.render_element`, but it [doesn't exist](https://docs.rs/dioxus-ssr/latest/dioxus_ssr/renderer/struct.Renderer.html#implementations).

Presumably they're outdated due to refactoring. I made an attempt to find if the issue was already reported.

**Steps To Reproduce**

- View the docs linked previously.

**Expected behavior**

All symbols mentioned in examples actually exist. Some kind of auto-compiling of examples would be pretty cool, if possibly overengineered.

**Screenshots**

N/A

**Environment:**
 - Current live docs in docs.rs for version `0.5.1`.

**Questionnaire**
I don't actually know the answer (am still learning) or I would send a PR myself. I suspect the surrounding explanation might also need an update.

The [source of render_element](https://docs.rs/dioxus-ssr/latest/src/dioxus_ssr/lib.rs.html#25-34) might be most of the answer for both. We have `dom.rebuild(&mut NoOpMutations);` which is probably the right thing if the vdom edits aren't important? The [examples/ssr.rs](https://github.com/DioxusLabs/dioxus/blob/main/examples/ssr.rs) is related but not really the answer we're looking for.

ealmloff:
The dioxus docsite is pretty good at getting you started with dioxus, but our inline and `docs.rs` docs are currently pretty sparse in comparison. This PR improves the inline docs for the `core`, `html`, `hooks`, and `signals` crates

Goals:
- Provide more examples for common items
  - [x] eval examples
  - [x] onmounted examples
  - [x] attributes rsx syntax
  - [x] elements rsx syntax
  - [x] resource
  - [x] memo
  - [x] reactivecontext
  - [x] Routable
  - [x] FromQuery
  - [x] FromQueryArgument
  - [x] FromRouteSegment
  - [x] FromRouteSegments
  - [x] ToRouteSegments
  - [x] use_effect
  - [x] rsx
- Make common items more discoverable with `doc(alias)`
  - [x] aliases for html names -> rust names on attributes
  - [x] use_state -> use_signal
  - [x] ref -> onmounted
- Improve trait not implemented errors with diagnostics
  - [x] closure maybe async return
  - [x] props not implemented
  - [x] component not implemented
  - [x] Routable/FromQuery/FromQueryArgument/FromRouteSegment/FromRouteSegments/ToRouteSegments
  - [x] Dependency
- Provide better references for common errors on methods where they can occur
  - [x] event handle errors
    - [x] `function requires argument type to outlive 'static`
    - [x] `use of moved value: your_value value used here after move`
  - [x] spawn errors
    - [x] `async block may outlive the current function, but it borrows value, which is owned by the current function`
    - [x] `use of moved value: value. move occurs because value has type YourType, which does not implement the Copy trait`
- [x] Update dioxus crate readme
  - Mention renderer features, signals, and the router
  - Don't go into detail about custom hooks etc. Just link to more complete docs

Fixes #2390


jkelleyrtp (approved):
Wow. So much work went into this. 

At some point we should propagate more of this onto the guide itself too.

PR #2461, opened 2024-05-29
Detect components called as functions

ealmloff:
This PR adds a runtime error for when components are called as functions. It will warn the user if they do something like this:
```rust
#[component]
fn MyComponent() {
    rsx! { "hello world" }
}

fn app() {
    rsx! { {MyComponent()} }
}
```


jkelleyrtp (approved):
Thanks for changing this, looks cleaner/better now

PR #2465, opened 2024-05-31
Switch from an absolute ::dioxus to the relative dioxus_core

pyrrho:
This resolves an edge-case that may arise if you (for some _weird_ reason) rename the dioxus dependency with the Cargo.toml `package = ...` key.

Consider a Cargo.toml with:
```
dioxus_but_named_wrong = { version = "0.5", features = ["web", "router"], package = "dioxus" }
```
And a main.rs with:
```
use dioxus_but_named_wrong::prelude::*;
```

If you then use the Routable derive macro, the compiler will error with:
```
error[E0433]: failed to resolve: could not find `dioxus` in the list of imported crates
 --> src/main.rs:6:17
  |
6 | #[derive(Clone, Routable, Debug, PartialEq)]
  |                 ^^^^^^^^ could not find `dioxus` in the list of imported crates
  |
  = note: this error originates in the derive macro `Routable` (in Nightly builds, run with -Z macro-backtrace for more info)
```

By switching this `quote!`'d return type from the root-level `::dioxus` namespace to the relative `dioxus_core`, we allow the generated code to use whatever `dioxus_core` symbols are imported from the `dioxus_you_can_name_this_however_you_want::prelude::*`.

### DIFF
--- a/packages/router-macro/src/lib.rs
+++ b/packages/router-macro/src/lib.rs
@@ -689,7 +689,7 @@ impl RouteEnum {
                     #(#site_map,)*
                 ];
 
-                fn render(&self, level: usize) -> ::dioxus::prelude::Element {
+                fn render(&self, level: usize) -> dioxus_core::Element {
                     let myself = self.clone();
                     match (level, myself) {
                         #(#matches)*


ealmloff (approved):

ealmloff:
Thanks! We need to support other names for a `dioxus`-like package in case you are using `dioxus-lib`

PR #2474, opened 2024-06-04
Update openid example

samtay:
I updated this example, largely keeping the existing logic even though some of it is a bit weird (many components are purely for effects rather than markup).

I also:

- left it outside of the workspace
- left it outside of CI
- converted the storage usage to sdk (a bit circular, but not really a problem with this package existing outside of the workspace)

This does mean that the example will almost certainly go out of date again. It would probably be best to add at least a `cargo check` on this in CI.


ealmloff:
Storage signals let us get rid of the manual synchronization code. They will automatically update the storage when you write to them. If we use them instead of the global signals, it simplifies things a lot

jkelleyrtp:
Much better, thanks! I meant to fix this in the 0.5 upgrade but it was kinda messy so we just excluded it from tests.

jkelleyrtp (approved):

PR #2480, opened 2024-06-05
Remove value attribute instead of resetting it

### LINKED ISSUE #2478: Conditionally exclude value attribute with non-null types (e.g. f64) from rendered HTML
The [daisyUI progress bar](https://daisyui.com/components/progress/#indeterminate-without-value) indeterminate animation only displays when the `<progress>` tag does not set the `value` attr. However, Dioxus' `progress` RSX element always sets `value` to some `f64` if `value` is declared in the RSX, defaulting to `0` for `None`. E.g. the following compiles but does not show the indeterminate animation:
```rust
let val: Option<f64> = None;
rsx! {
    progress { class: "progress", value: val } // renders as <progress class="progress" value="0" />
}
```
The desired behavior in this case is for `None` values to not set the attribute at all, i.e. render as `<progress class="progress">`, but that may break other uses, so it could be opt-in with some syntax, e.g.:
```rust
rsx! {
    progress { class: "progress", value: None }        // renders as <progress class="progress" value="0" />
    progress { class: "progress", value?: None }       // renders as <progress class="progress" />
}
```
My current workaround is conditionally rendering the `progress` element to remove the attribute:
```rust
rsx! {
    if let Some(value) = value {
        progress { class: "progress", value: val }
    } else {
        progress { class: "progress" }
    }
}
```

ealmloff:
This changes the logic for `None` values for the `value` attribute. We were previously just resetting the value which causes issues if the attribute is actually an attribute instead of a property. This PR changes `None` to both remove the attribute and reset the property

Fixes #2478

### DIFF
--- a/packages/interpreter/src/unified_bindings.rs
+++ b/packages/interpreter/src/unified_bindings.rs
@@ -98,6 +98,7 @@ mod js {
                 switch (field) {
                     case "value":
                         node.value = "";
+                        node.removeAttribute("value");
                         break;
                     case "checked":
                         node.checked = false;


jkelleyrtp:
I believe we need to change the TS too?

ealmloff:
> I believe we need to change the TS too?

I don't think we have any code for removing attributes in typescript. The code for setting attributes is there because we bind to it with both wasm bindgen and sledgehammer bindgen for template creation, but removing attributes only gets used by sledgehammer bindgen

jkelleyrtp (approved):

PR #2488, opened 2024-06-07
switch from slab to slotmap for tasks to fix the ABA problem

ealmloff:
This is an alternate solution to #2440 that addresses the wider [ABA problem](https://en.wikipedia.org/wiki/ABA_problem) with tasks by introducing generationals to tasks.

This PR is non-breaking so it can be backported to 0.5. `ElementId` and `ScopeId` have similar problems, but they are intentionally not changed in this PR because the inner `usize` type is public, and too small for generations on wasm where `usize` is `u32`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2253,6 +2253,7 @@ dependencies = [
  "rustversion",
  "serde",
  "slab",
+ "slotmap",
  "tokio",
  "tracing",
  "tracing-fluent-assertions",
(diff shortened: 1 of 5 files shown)


jkelleyrtp (approved):
Looks good, I think we might have some hidden race conditions somewhere that this will uncover, so it'll be good to put this on the radar before releasing the next patch/minor version.

ealmloff:
> This PR is non-breaking so it can be backported to 0.5. `ElementId` and `ScopeId` have similar problems, but they are intentionally not changed in this PR because the inner `usize` type is public, and too small for generations on wasm where `usize` is `u32`

Switching `ElementId` and `ScopeId` to slotmap would require a key type for slotmap that is FFI friendly. It would also require rewriting most of the tests in core. Holding off on that change for now

SkiFire13:
> Switching `ElementId` and `ScopeId` to slotmap would require a key type for slotmap that is FFI friendly.

Note that `slotmap`'s `Key` trait offers a way to convert to/from `KeyData`, which in turn provides a pair of `as_ffi`/`from_ffi` functions to convert it to/from a `u64` that can be used for FFI purposes.

PR #2494, opened 2024-06-09
fix: Add missing name attribute for <slot>

### LINKED ISSUE #2493: Missing HTML <slot> name attribute
**Problem**

The HTML `<slot>` element should accept the `name` attribute.

See <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot>

**Steps To Reproduce**

```rust
use dioxus::prelude::*;

#[component]
pub fn Plop() -> Element {
  rsx! {
      slot { name: "plop" }
  }
}
```

Trigger `error[E0599]: no function or associated item named `name` found for struct `slot` in the current scope`

**Expected behavior**

It should compile

**Environment:**
 - Dioxus version: 0.5.1
 - Rust version: 1.78
 - OS info: MacOs
 - App platform: `web`

**Questionnaire**

- [X] I would like to fix and I have a solution

ilaborie:
Fix #2493

### DIFF
--- a/packages/html/src/elements.rs
+++ b/packages/html/src/elements.rs
@@ -1600,7 +1600,9 @@ builder_constructors! {
     /// Build a
     /// [`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot)
     /// element.
-    slot None {};
+    slot None {
+        name: String DEFAULT,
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thank you!

PR #2507, opened 2024-06-11
Add CI step to test packages with debug assertions off

ealmloff:
This adds an extra step that tests packages with debug assertions off (but without optimizations) to test the faster paths in core and other packages. This should prevent regressions like https://github.com/DioxusLabs/dioxus/commit/79e18c2d62166f388641cd6e20a82bef2506535a


jkelleyrtp:
So much better, thank you thank you!

PR #2512, opened 2024-06-12
Remove implicit optional dependency features

### LINKED ISSUE #2513: Remove unintentionally exposed features from the `dioxus` crate
By using [saecki/crates.nvim](https://github.com/saecki/crates.nvim) I can see a lot of features that Dioxus exposes. Most of them are not intended to be exposed.

image
image

The solution (for the most part) is to use `dep:` syntax wherever possible to remove all of the implicit features. I believe similar tools will show the same result. And for a better DX, Dioxus really should only expose a few features. Something like:

image

https://doc.rust-lang.org/cargo/reference/features.html

ealmloff:
Optional dependencies add implicit features to crates if they are never used with `dep:`. This PR switches all of our optional dependencies to use `dep:` to remove those useless features

Closes #2513

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,8 +195,8 @@ fullstack = ["dioxus/fullstack"]
 axum = ["dioxus/axum"]
 server = ["dioxus/axum"]
 web = ["dioxus/web"]
-collect-assets = ["manganis"]
-http = ["reqwest", "http-range"]
+collect-assets = ["dep:manganis"]
+http = ["dep:reqwest", "dep:http-range"]
 
 [[example]]
 name = "login_form"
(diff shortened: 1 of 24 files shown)


Andrew15-5:
Woah, that's a lot of `dep:` additions! I knew that it's not gonna be small, but that's more than I expected.

I also found out about `cargo-whatfeatures` according to which now the feature list looks very clean (and this is the tool that I will use for all my next projects too):

```
╰─ c whatfeatures --manifest-path .
workspace for dioxus
├─ dioxus = "0.5.2"
│ └─ features
│   ├─ default
│   │ ├─ hooks
│   │ ├─ hot-reload
│   │ ├─ html
│   │ ├─ launch
│   │ ├─ macro
│   │ └─ signals
│   ├─ axum
│   ├─ desktop
│   ├─ fullstack
│   ├─ hooks (default)
│   ├─ hot-reload (default)
│   ├─ html (default)
│   ├─ launch (default)
│   ├─ liveview
│   ├─ macro (default)
│   ├─ mobile
│   ├─ router
│   ├─ signals (default)
│   ├─ ssr
│   ├─ static-generation
│   ├─ third-party-renderer
│   └─ web
```

But I can only check via the editor plugin when there will be a new crate release. Can't wait to see the difference!

PR #2515, opened 2024-06-13
Make it easier to provide context to fullstack

### LINKED ISSUE #1688: add ability to pass external state to fullstack server
## Specific Demand

currently, [the fullstack template](https://github.com/DioxusLabs/dioxus-template/blob/main/src/main.rs#L187-L193) uses a basic `LaunchBuidler` setup and calls `launch` on it to start up the backend server with the ability to server SSR'ed Dioxus pages. this works great. however, this functionality is very limiting since the user is unable to supply extra layers, extensions, and data to the (in my case) Axum server that ends up running. [The current solution](https://github.com/DioxusLabs/dioxus/discussions/1683#discussioncomment-7741397) is to take the code that is run in the `launch` method's source code, bring that in to your project, and work from there. this works fine but is less than ideal in terms of future upgrades to the library. it'd be great to be able to supply data, extensions, layers, etc. to the `LaunchBuilder` as an abstraction from needing to fuss with the underlying Dioxus server setup.

## Implement Suggestion

it could be something relatively simple as adding methods to the `LaunchBuilder` struct for adding each of these to the internal `server_cfg` and then on `launch`, add them into the correct place if they have been set. however, this might not be super scalable since each server library might have their own specific implementation for Data vs Extention vs etc. however, each method would theoretically be easy to write. maintainability long term could be an issue.

### LINKED ISSUE #2131: Dioxus fullstack lacking a formal way to provide state
## Feature Request

In https://github.com/DioxusLabs/dioxus/pull/1944, `DioxusRouterExt::register_server_fns_with_handler` was removed without providing a suitable replacement.
I used this method to provide state to Axum so that it could be extracted in the `#[server(..)]` functions. I know it is otherwise possible using Axum layers, but that isn't particularly ergonomic at the moment.

Observed in Dioxus 0.5.0-alpha-2
## Implement Suggestion

It was suggested in [Discord](https://discord.com/channels/899851952891002890/1220693416212430898/1220693416212430898) that the following function could be made public or form the basis of another public function:
https://github.com/DioxusLabs/dioxus/blob/b19a546c0ac639a5cf9e942fff7729ad30e91a5e/packages/fullstack/src/axum_adapter.rs#L485

What strikes me as odd is that the argument's type signature is `impl Fn() + 'static + Clone + Send,` when I think it should be `impl Fn(&mut DioxusServerContext) + 'static + Clone + Send,`, so that there is a context to modify.

ealmloff:
This PR changes the context function in the launch builder to provide context throughout the fullstack application including in server functions. You can now consume context in server functions with an extractor:
```rust
LaunchBuilder::new().with_context(1234567890u32).launch(app);

#[server]
async fn get_server_data() -> Result<String, ServerFnError> {
    let FromContext(context): FromContext<u32> = extract().await?;
    let text = reqwest::get("https://httpbin.org/ip").await?.text().await?;
    Ok(format!("context {context} Your IP address is: {text} 🌎"))
}
```

Closes #2131
Closes #1688

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -204,12 +204,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
 
 [[package]]
-name = "anymap"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "33954243bd79057c2de7338850b85983a44588021f8a5fee574a8888c6de4344"
-
-[[package]]
 name = "anymap2"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2130,6 +2124,7 @@ dependencies = [
 name = "dioxus"
 version = "0.5.2"
 dependencies = [
+ "axum",
  "criterion 0.3.6",
  "dioxus",
  "dioxus-config-macro",
@@ -2416,7 +2411,6 @@ dependencies = [
 name = "dioxus-fullstack"
 version = "0.5.2"
 dependencies = [
- "anymap",
  "async-trait",
  "axum",
  "base64 0.21.7",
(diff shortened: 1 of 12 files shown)


jkelleyrtp (approved):

scullionw:
how would .with_context be used to pass resources that require async to be built?

like if we want to store a database connection? (let conn = connect_to_db().await)

something like:

LaunchBuilder::new().with_context_future(async { connect_to_db().await }).launch(app);

?

PR #2518, opened 2024-06-14
add custom decorations example

RobertasJ:
This pr adds an example of a project with a decorationless window, which adds back the functionality of resizing, and the window bar but with rsx.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -235,105 +235,105 @@ doc-scrape-examples = true
 
 [[example]]
 name = "backgrounded_futures"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "calculator_mutable"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "calculator"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "clock"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "crm"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "custom_html"
-required-features = ["desktop"]
+required-features = ["default"]
 doc-scrape-examples = true
 
 [[example]]
 name = "custom_menu"
-required-features = ["desktop"]
+required-features = ["default"]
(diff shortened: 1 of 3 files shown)


ealmloff:
On MacOs, this example opens a blank window (without decorations) that loads forever and refuses to exit

jkelleyrtp:
Also - I'd like to not include the entire project here - I think we can get away with just a single example in the examples folder.

RobertasJ:
I will get this down to one file. But i cant test anything on MacOS myself to understand the issue ealmoff found. 

For now i will get this down into one file.

RobertasJ:
It seems like it is quite a halsste to get it down to one file because it dioxus_free_icons uses the latest crates.io dioxus which isnt compatible with the current one on github. Should i reimplement the functionality of dioxus_free_components instead?

RobertasJ:
Ok, ive gotten this down to one file. I over the some of the dioxus-free-icons code for the buttons on the window bar. I cannot test the issue with MacOS myself sadly as i do not own one. 

Ive also noticed that there are some issues with the examples. They had desktop in the required-features field instead of default, which is what i guess they were supposed to be set to. This made it possible to run the examples.

RobertasJ:
Ive now fixed this so it will work on mac. It wont look native but it works.

jkelleyrtp:
Hey, sorry this sat so long! I think this example might be better served as a dedicated repository rather than in our tree, at least in its current form.

Including tailwind and the icon trait just makes this PR too messy / complex. Examples should be smaller and tighter.

I think the PR also points out some weak spots of Dioxus that we need to fix. Examples are basically us showing the "blessed way" of doing things with the framework. This PR does lots of eval and polling which is not what we want to teach.

Can you take some of the learnings from this PR and spin them out into separate issues? It'd be great to highlight exactly where dioxus-desktop is deficient so we know where to focus improvements.

RobertasJ:
From what ive found, all of these issues stem from depending on tao, which has half of its issues because of GTK. I think if i was to create an issue it would be called "Switch from tao".

jkelleyrtp:
> From what ive found, all of these issues stem from depending on tao/wry (im havent looked at which one dioxus depends on), which has half of its issues because of GTK. I think if i was to create an issue it would be called "Switch from tao/wry".

I'd like to switch from tao to winit #2706 but we do use quite a bit of wry's functionality and I'm unsure if we still need gtk.

RobertasJ:
I am not sure what you would directly use gtk for, from what i see its exposed by tao.

Issue #2312, opened 2024-04-15
`dx serve` prints "Failed to parse file: No such file or directory" and starts rebuild sometime on static string change (Neovim)


Andrew15-5:
**Problem**

`dx serve` (I guess starting from 0.5.x) now gives an error:

```
2024-04-15T14:10:04.619499Z ERROR dioxus_cli::server: Failed to parse file: No such file or directory (os error 2)
```

when changing a static string in the `rsx` block (or probably anything else). In rare cases, it triggers a rebuild.

After some debugging, I tried VSCodium and after each save absolutely nothing happened, and at that point I realized, it's https://github.com/typst/typst/issues/2633#issuecomment-1837183367 all over again. And, of course, the workaround is the same:

```lua
local rust_group = vim.api.nvim_create_augroup("rust_group", {})
vim.api.nvim_create_autocmd({ "FileType" }, {
  pattern = "rust",
  callback = function()
    vim.opt_local.backupcopy = "yes"
  end,
  group = rust_group,
})
```

I think it is also caused by notify crate (same as with Typst). So either some configuration must be changed for the watcher to work better, or the upstream crate should accommodate the "fix" of the default when saving files in Neovim. Some new option or something. Otherwise, I think this won't be resolved (there is probably no chance Neovim will change its default behavior).

**Steps To Reproduce**

- `dx new` for desktop + tailwind
- `cd`
- `nvim src/main.rs` (Neovim 0.9.4)
- `dx serve`
- try changing some static string in the `rsx` block
- see error or sudden rebuild

**Expected behavior**

No errors or rebuilds.

**Environment:**
 - Dioxus version: 0.5.4, main
 - Rust version: 1.79.0-nightly
 - OS info: Pop!_OS
 - App platform: desktop

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [x] I don't have time to fix this right now, but maybe later

Andrew15-5:
I don't.

FelipeFTN:
I'm having the same issue here.
**Environment:**
    Dioxus version: 0.5.4, main
    Rustc version: 1.80.0-nightly
    OS info: Manjaro-i3
    App platform: web
    Neovim version: v0.9.5

The workaround from https://github.com/typst/typst/issues/2633 seems to work fine on my neovim lua files:
```lua
-- Rust dioxus workaround
vim.opt_local.backupcopy = "yes"
```

[...]

PR #2528, opened 2024-06-16
Experimental fix for [issue 2312]

JustSimplyKyle:
This pr is an experimental fix for https://github.com/DioxusLabs/dioxus/issues/2312
it doesn't change the underlying logic, just the events that's tracked.
(this is just me adding random events to the track list until it fixes the hot-reload issue)
This works under `helix` and `vscode`, not sure if `neovim` has the same behaviour
@Andrew15-5 mind testing?

### DIFF
--- a/packages/cli/src/server/mod.rs
+++ b/packages/cli/src/server/mod.rs
@@ -90,11 +90,26 @@ fn watch_event<F>(
 ) where
     F: Fn() -> Result<BuildResult> + Send + 'static,
 {
+    use notify::event::EventKind;
+    use notify::event::ModifyKind;
+    use notify::event::RemoveKind;
+    use notify::event::RenameMode;
     // Ensure that we're tracking only modifications
-    if !matches!(
-        event.kind,
-        notify::EventKind::Create(_) | notify::EventKind::Remove(_) | notify::EventKind::Modify(_)
-    ) {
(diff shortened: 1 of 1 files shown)


Andrew15-5:
If we have an image or something called `img.png` and then we want to rename it. If the file name is changed to `image.png` **first** in the Rust code, then the `dx` will show the unknown file error. After renaming the file (`mv img.png image.png`) the watcher won't see it and the workflow will be broken. Maybe re-saving the Rust file will fix this.

This is just an example that I come up with on the spot. This means that the potential FS manipulations can degrade the DX for everyone just so that you can use the default saving behavior in Neovim. I don't think this is worth it, if what I describe above is true. I don't really even know why other watchers (although I haven't used much of them, or I just can't remember them) don't complain about the Neovim's backup saving shenanigans. I only faced this problem with Rust/[notify](https://crates.io/crates/notify)-based watchers (Dioxus/Typst CLI).

I can test this later (in a few days?), now I'm mostly busy.

jkelleyrtp:
In our recent CLI overhaul I incorporated your changes into our watcher - will test against the original issue.

Andrew15-5:
https://github.com/DioxusLabs/dioxus/issues/2312#issuecomment-2247186938

So, for that commit this isn't even necessary.

PR #2535, opened 2024-06-17
Update `dioxus_desktop::Config` to also allow for asynchronous custom protocol handlers

d3rpp:
This is a very simple PR, that is tested to work on my Fedora machine, although I'm unable to do any real testing on other OSs.

For some reason `wry` doesn't like using synchronous custom protocols, I'm sure you've experienced that since it appear's you're also using the asynchronous one, so I've added an extra config option that adds asynchronous ones.

this is gonna be a case of certified ISO 9000 - Works on my machine, as such I encourage others to test on their machines.

Also added a QoL fix that allows the passing of anything that implements `ToString` instead of a string, since it may be a safe assumption that most of the time, it'd be passed in as a string literal.


jkelleyrtp (approved):
Looks good, thank you!

PR #2540, opened 2024-06-17
add aria-current to link

uzytkownik:


jkelleyrtp (changes requested):
Tests are failing since we snapshot test the HTML and it's not expecting the aria label. Can we fix that test?

uzytkownik:
@jkelleyrtp how do I run locally? I did `cargo test` and `cargo clippy` before submission.

ealmloff:
There is a list of commands CI runs here: https://dioxuslabs.com/learn/0.5/contributing#before-you-contribute. The main command for testing is:
```
cargo test --all --tests
```

uzytkownik:
@ealmloff It might be better to add CONTRIBUTING.md. Plus the code seems to be outdated:

```
% cargo fmt -- src/**/**.rs
Error: file `src/**/**.rs` does not exist
```

ealmloff (approved):

Issue #1867, opened 2024-01-25
cli hot reload does not work for WSL


JocoNagy:
**Problem**

When I run 'dx serve' with hot reload in WSL, it does not detect any changes.

**Steps To Reproduce**

- Install Dioxus CLI on WSL
- run 'dx serve --hot-reload'
- change any element or attribute

**Expected behavior**

The page should be updated with the changes

**Environment:**
 - Dioxus version: 0.4.3
 - Rust version:  1.75.0
 - OS info: Microsoft Windows 11 Pro / WSL2 
 - App platform: web

**Additional notes**

Its probably related to [this](https://github.com/notify-rs/notify/issues/254) issue in notify-rs
I had similar problem with trunk, using explicit '--poll' flag helps there.

PR #2544, opened 2024-06-19
Fix hot-reloading on Windows

matthunz:
I found hot-reloading doesn't seem to be working for me on Windows because the initial output of `rsx` has template names like:
```
src/main.rs:9:5:0
```
while hot-reload updates have different names such as:
```
src\main.rs:9:5:0
```

This converts the template path to a UNIX-style path to be compatible with the `file!` macro. This seems to fix hot-reload on my machine but I'm not sure how robust this is just yet. I'd love some help testing!


matthunz:
Maybe related? https://github.com/DioxusLabs/dioxus/issues/1867 I'll try to check if this was the issue

jkelleyrtp:
Thanks!

PR #2547, opened 2024-06-20
Add missing packages in flake.nix

uzytkownik:
A few packages were missing from flake.nix needed to build tests.


ealmloff (approved):
Thanks, we updated these dependencies in the installation docs, but they weren't updated in the flake

PR #2552, opened 2024-06-22
Update the examples link on the README

hardBSDk:


ealmloff (approved):
Thanks! (We shouldn't be linking to the example projects repo anymore because it is out of date and most of the examples have been moved into the examples folder)

Issue #2486, opened 2024-06-07
hot reload does refresh rsx! if it contains \n


gusinacio:
I think I found the problem. The problem was not related to the file, but the content of the strings in the whole rsx!

I generated some code in v0, and there were some `\n` in the middle of the strings. The whole rsx stops reloading if one of  the strings contains `\n`.

https://github.com/DioxusLabs/dioxus/assets/4844547/b0327733-0373-4bfa-bf43-16a2b231c889

_Originally posted by @gusinacio in https://github.com/DioxusLabs/dioxus/issues/2485#issuecomment-2155248154_

ealmloff:
This seems similar to https://github.com/DioxusLabs/dioxus/issues/1586

ealmloff:
fixed by #2725

jkelleyrtp:
Did you test it locally? @ealmloff

ealmloff:
> Did you test it locally? @ealmloff

Yes, hot reloading this snippet works on desktop:
```rust
pre {
     "hello\tworld"
     "hello\nworld"
}
```

PR #2557, opened 2024-06-27
Partially Fix Hotreload

### LINKED ISSUE #2180: Tailwind classes are not hot reloaded
**Problem**

The blog post says that tailwind classes can now be hot reloaded, but they currently require a full recompile

**Steps To Reproduce**

- Clone the dioxus repo (rev 2461790baf302766f35861c3c85269ba90add247)
- Run:
```sh
cd examples/tailwind
dx serve
```
- Run in another terminal:
```sh
cd examples/tailwind
npx tailwindcss -i ./input.css -o ./public/tailwind.css --watch
```
- Change a tailwind class
- Observe a normal rust recompile

**Expected behavior**

In the blog post we mention that tailwind hot reloading is supported:
Screenshot 2024-03-29 at 8 32 45 AM

**Environment:**
 - Dioxus version: `master`
 - Rust version: `nightly`
 - OS info: MacOS
 - App platform: `web`

**Questionnaire**

DogeDark:
- Normalizes all template names to follow `foo/bar/baz.rs`
- Fixes a caching issue on Web with re-sending the templates that occurred before a full reload.

This gets hot reloading working on Web for the Tailwind example. Desktop requires more work as it's affected by #1586 and #2486. 

I noticed that there were two separate socket systems between Web & Desktop. I think these should be merged into the Web's socket system with the communication models being standardized between them. This would also allow hot reload to a mobile device from the host machine in the future.

Half-Fixes #2180 but should fix any tailwind/hotreload in cases where `\n` or `\t` aren't used in RSX.

**Testing:**
To test if the template path name standardization logic works correctly between platforms.
- [x] Windows
- [x] Linux (I'll test this)
- [ ] Mac


DogeDark:
Looks like these issues might already be fixed in #2258.

jkelleyrtp:
I'm not normalizing path names so that's still a useful change. If this is something we can backport then I'd still like to accept this PR so we can get it out to Dioxus 0.5 users!

jkelleyrtp:
Oh, I think I *did* #1586 in my PR. We were to_string the wrong thing and it wasn't escaping stuff correctly. Need to check,

DogeDark:
Maybe we could backport a fix for that too?

PR #2574, opened 2024-07-02
Remove deprecated macros: inline_props, format_args_f!, render!

jkelleyrtp:
Removes macros that we deprecated in previous versions of Dioxus.

Currently we now export three macros:
- `#[rsx]`
- `#[component]`
- `#[derive(Properties)]`


packages/core-macro/tests/ifmt.rs
```diff
@@ -1,34 +0,0 @@
-use dioxus_core_macro::*;
```
ealmloff, line 1: Could we keep testing the behavior of weird formatted expressions with rsx if we don't expose the format macro separately?

PR #2596, opened 2024-07-07
Use RecommendedWatcher.

LiamKarlMitchell:
Resolves building on Windows with https://github.com/DioxusLabs/dioxus/pull/2258/commits


LiamKarlMitchell:
I was seeing an error.
```
error[E0432]: unresolved import `notify::FsEventWatcher`                                                                                                                                                          
   --> packages\cli\src\server\watcher.rs:11:16
    |
11  |     EventKind, FsEventWatcher,
    |                ^^^^^^^^^^^^^^ no `FsEventWatcher` in the root
```

PR #2603, opened 2024-07-08
CLI: Add a signal handler to reset the cursor on interrupt

FragrantArmpit:
In CLI we use `cargo-generate` for `init` and `new`.

`cargo-generate` uses things like [Select](https://docs.rs/dialoguer/latest/dialoguer/struct.Select.html) from [dialoguer](https://crates.io/crates/dialoguer).

Those things hide your terminal's cursor for aesthetic purposes.

If you press `Ctrl + C` while those things are active, you will end up with a cursorless terminal.

See: https://github.com/console-rs/dialoguer/issues/294

This PR adds a `ctrlc` signal handler to reset the cursor and exit.


jkelleyrtp:
Thanks! Much better

Issue #2595, opened 2024-07-07
Dioxus v5 no longer hides window decorations when disabled (works in v4)


imgurbot12:
**Problem**

Disabling Window Decorations for Dioxus Desktop using the latest release of `v0.5.1` does not actually hide the default decorations anymore.

**Steps To Reproduce**

Here's a simple script to replicate the issue

```rust
use dioxus::prelude::*;

fn app() -> Element {
    rsx! { "Hello World!" }
}

fn main() {
    let window = dioxus_desktop::WindowBuilder::default().with_decorations(false);
    let config = dioxus_desktop::Config::default().with_window(window);
    LaunchBuilder::desktop().with_cfg(config).launch(app);
}
```

**Expected behavior**

In Version `v0.4.3` the Decorations are hidden as intended
```rust
use dioxus::prelude::*;

fn app<'a>(cx: Scope<'a, ()>) -> Element<'a> {
    cx.render(rsx! { "Hello World!" })
}

fn main() {
    let builder = dioxus_desktop::WindowBuilder::new().with_decorations(false);
    let config = dioxus_desktop::Config::new().with_window(builder);
    dioxus_desktop::launch_cfg(app, config);
}
```

**Environment:**
 - Dioxus version:  `dioxus = { version = "0.5.1", features = ["desktop"] }`
 - Rust version:   `rustc 1.79.0 (129f3b996 2024-06-10)`
 - OS info:  Ubuntu 24.04 LTS
 - App platform:   Desktop

**Questionnaire**

- [X] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

imgurbot12:
I thought this might be an upstream issue related to wry/tao since that happens to be the library behind the actual window rendering and is the source of `WindowBuilder` but running the wry [simple example](https://github.com/tauri-apps/wry/blob/dev/examples/simple.rs) doesn't include any decorations even when including `with_decorations(true)`

```rust
let window = WindowBuilder::new()
        .with_decorations(false)
        .build(&event_loop)
        .unwrap();
```

Changing it from true/false doesn't effect decorations whatsoever and none appear so I have no idea where the decorations from dioxus are even coming from.

Running tao's [decorators example](https://github.com/tauri-apps/tao/blob/dev/examples/decorations.rs) has similar issues. Decorations neither appear or reappear as they are always missing no matter what I do.

[...]

PR #2613, opened 2024-07-09
fix (desktop): hide menu when window decorations are disabled

imgurbot12:
fixes https://github.com/DioxusLabs/dioxus/issues/2595

thanks to @ealmloff for his direction on fixing the issue


jkelleyrtp (approved):
Looks good, thanks!

PR #2625, opened 2024-07-11
Re-Export Manganis & Update Examples

DogeDark:
Re-Exports Manganis in Dioxus prelude. This creates a new (default) feature called `asset`. The following items are re-exported:
```rs
use dioxus::prelude::{
    // The mg! (or re-exported asset!) macro needs `manganis` in scope.
    self,
    // manganis::classes! macro
    classes
    // manganis::mg! macro
    asset
    // manganis::ImageAsset
    ImageAsset
    // manganis::ImageType
    ImageType

};   
```

This PR also:
- Updates the styled examples to use Manganis' upcoming filesystem fallback to provide CSS assets.
- Special handling has been added to correct Manganis filesystem fallback paths.

**Potential Problems**
- Dioxus apps using the latest version will always show a build warning that the `dx` cli wasn't used with Manganis whether or not manganis is actually used. (regular cargo build)

**Depends On:** https://github.com/DioxusLabs/collect-assets/pull/37


DogeDark:
Had to remove a few definitions of the examples to be scraped as for some reason `cargo check/clippy` runs it in the directory of the package's `Cargo.toml`, causing the example asset paths to be invalid. 

Not sure of a great way to fix this but I'm open to suggestions. We could also revert to `include_str` or no styling for those examples.

jkelleyrtp (approved):

PR #2629, opened 2024-07-12
Better `expect` error messages

matthunz:
As per @ealmloff 's suggestion, this creates a `RuntimeError` struct that impls `std::Error` and `fmt::Display` (as `"Must be called from inside a Dioxus runtime."`). This replaces the original `.expect` error messages of `to be in a Dioxus runtime`.

Core methods, such as `current_scope_id`, that previously returned `Option<T>` now return `Result<T, RuntimeError>`.
This also adds `#[track_caller]` on functions using `Runtime::current()` to output better source locations.

Some methods do still return an `Option` but I'm not sure which should be converted yet, maybe we can refactor more over time if this is the right approach?


ealmloff:
I agree, `to be in a dioxus runtime` is a very vague error message. Looking back at this code, the methods that can fail when they are not in a runtime like `current_scope_id` should probably return a `Result<_, RuntimeError>` instead of `Option<_>` as well. Then we wouldn't need to duplicate `expect` everywhere. It would also be useful to mention the [RuntimeGuard](https://docs.rs/dioxus-core/latest/dioxus_core/prelude/struct.RuntimeGuard.html) which is generally how the error can be fixed

matthunz:
Sounds good  :+1: I really like that approach @ealmloff 

I added the new error type here: https://github.com/DioxusLabs/dioxus/blob/2e786d147488f0e682d084f2c4e4b3bf5dcbe019/packages/core/src/runtime.rs#L273-L296

ealmloff (approved):
Much better, thanks!

I added a little help message to help fix a common situation I have seen people run into with websys closures/dylibs

PR #2647, opened 2024-07-19
[feat:host-env] To deploy fullstack on AWS Fargate, added `HOST` environment to expose service externally

hackartists:
- Fargate is the only way to deploy of dioxus fullstack on AWS.
  - However, `127.0.0.1` blocks to expose port to ESC Service.

Currently, we can change host address by `addr` function. However, to use it, we need to handle complicated entry point at apps.

### DIFF
--- a/packages/fullstack/src/config.rs
+++ b/packages/fullstack/src/config.rs
@@ -32,7 +32,10 @@ impl Default for Config {
             #[cfg(feature = "server")]
             server_fn_route: "",
             #[cfg(feature = "server")]
-            addr: std::net::SocketAddr::from(([127, 0, 0, 1], 8080)),
+            addr:  match option_env!("HOST") {
+                Some(addr) => addr.parse().unwrap(),
+                None => std::net::SocketAddr::from(([127, 0, 0, 1], 8080)),
+            },
             #[cfg(feature = "server")]
             server_cfg: ServeConfigBuilder::new(),
             #[cfg(feature = "web")]


ealmloff:
You can also change the address fullstack uses with the `--addr` command line option on the `dioxus-cli` or the the server binary dioxus fullstack builds

hackartists:
> You can also change the address fullstack uses with the `--addr` command line option on the `dioxus-cli` or the the server binary dioxus fullstack builds

does `dx build` also support the `--addr`? I can't see `--addr` option in `dx build --help`.
When deploying its binary to fargate, we need `--addr` option in `dx build` or in built biyard. @ealmloff

ealmloff:
> > You can also change the address fullstack uses with the `--addr` command line option on the `dioxus-cli` or the the server binary dioxus fullstack builds
> 
> does `dx build` also support the `--addr`? I can't see `--addr` option in `dx build --help`. For deploying its binary to fargate, we need `--addr` option in `dx build` or in built biyard. @ealmloff

`--addr` is supported in `dx serve` and when running the final server binary standalone

hackartists:
> > > You can also change the address fullstack uses with the `--addr` command line option on the `dioxus-cli` or the the server binary dioxus fullstack builds
> > 
> > 
> > does `dx build` also support the `--addr`? I can't see `--addr` option in `dx build --help`. For deploying its binary to fargate, we need `--addr` option in `dx build` or in built biyard. @ealmloff
> 
> `--addr` is supported in `dx serve` and when running the final server binary standalone

I meant that the final server binary is `dist/{service}` binary after `dx build --release`.
I think we need to use the `dist` and built binary for docker, instead of `dx serve`

ealmloff:
> I meant that the final server binary is `dist/{service}` binary after `dx build --release`. I think we need to use the `dist` and built binary by `dx build --release` for production, instead of `dx serve`

Yes, you can pass the --addr to the binary built with dx serve. `dist/{service} --addr 0.0.0.0 --port 1111`

hackartists:
> > I meant that the final server binary is `dist/{service}` binary after `dx build --release`. I think we need to use the `dist` and built binary by `dx build --release` for production, instead of `dx serve`
> 
> Yes, you can pass the --addr to the binary built with `dx build`. `dist/{service} --addr 0.0.0.0 --port 1111`

@ealmloff I've tested on binary built with `dx build --release --platform fullstack` of `dx v0.5.6`.
However, `--addr 0.0.0.0 --port 5000` does not work for me.

ealmloff:
`--addr 0.0.0.0 --port 5000` for fullstack only works in the git version of dioxus if you are using the launch function or launch builder

hackartists:
> `--addr 0.0.0.0 --port 5000` for fullstack only works in the git version of dioxus if you are using the launch function or launch builder

Thank you for the information. However, it seem not to work with environment variables. I tried to build below.
```
VERSION=v0.1.0 COMMIT=b0c608b LOG_LEVEL=debug DATE="" dx build --release --platform fullstack
```

And the environment variables will used in server code below
```
#[server(endpoint = "/version", input=GetUrl)]
pub async fn version() -> Result<String, ServerFnError> {
    Ok(match option_env!("VERSION") {
        Some(version) => match option_env!("COMMIT") {
            Some(commit) => format!("{}-{}", version, commit),
            None => format!("{}", version),
        },
        None => match option_env!("DATE") {
            Some(date) => date.to_string(),
            None => "unknown".to_string(),
        },
    }
    .to_string())
}
```

And the web build cause panic as below
```
thread 'tokio-runtime-worker' panicked at /home/hackartist/.cargo/git/checkouts/dioxus-1e619ce595d3799d/40f936d/packages/cli/src/builder/web.rs:96:18:
called `Result::unwrap()` on an `Err` value: failed reading '/home/hackartist/data/devel/github.com/biyard/photodek/dist/photodek.wasm'

Caused by:
    No such file or directory (os error 2)
2024-07-23T02:58:02.484577Z ERROR dioxus_cli::builder::web: Bindgen build failed: JoinError::Panic(Id(31), ...)                                                     
thread 'tokio-runtime-worker' panicked at /home/hackartist/.cargo/git/checkouts/dioxus-1e619ce595d3799d/40f936d/packages/cli/src/builder/web.rs:96:18:
called `Result::unwrap()` on an `Err` value: failed reading '/home/hackartist/data/devel/github.com/biyard/photodek/dist/photodek.wasm'

Caused by:
    No such file or directory (os error 2)
Error: 🚫 Building project failed:

Caused by:
    Failed to build project
```

`no dx git version` does work properly. @ealmloff

hackartists:
> `--addr 0.0.0.0 --port 5000` for fullstack only works in the git version of dioxus if you are using the launch function or launch builder

@ealmloff Thank you for the information.
I've tested with `git version`. However, the `git version` seems to be unstable, right?
I've installed `dx` with git version, right now. But it doesn't generate any binary in `dist` directory. In fact, I previously used `dx git version`. However, from yesterday, it couldn't serve my project with `address already used in` error, but practically no program binding that port by `sudo netstat -nltp`.
And `dx cargo version` does work properly and generate binary file in `dist` on the same project.

ealmloff:
The versions of the dioxus-cli and dioxus itself must match because they communicate with each other at build time. The new version of the CLI proxies fullstack in dev mode. The old version of dioxus and the CLI could both be trying to serve on the same port. Updating dioxus should fix that issue

hackartists:
> cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli

Thank you. Finally, I've check it out and it does work with below.
- Cargo.toml: `dioxus-html` for `file-engine` feature.
```
[dependencies]
dioxus = { git = "https://github.com/DioxusLabs/dioxus", version = "0.5",  features = ["router", "fullstack"] }
dioxus-html = { git = "https://github.com/DioxusLabs/dioxus", version = "0.5" }
```

- Running script
```
dist/bootstrap --addr 0.0.0.0 --port 8080
```

I will close this PR.

PR #2656, opened 2024-07-20
Fix rsx autocomplete and diagnostics in the root; provide better completions after attributes are finished

ealmloff:
A few small fixes to rsx:
1) Provide diagnostic errors on the root rsx if an error is encountered
2) Always parse elements/components in the root rsx as elements or components, never as attributes. This lets us only complete to elements or components. It also fixes failing silently with attributes on spreads in the root rsx
3) After we encounter the first spread, element or component, stop parsing attributes. This lets autocomplete `t` in the following code to `table` instead of `title` (which would not be valid):
```rust
dialog {
     width: "100%",
     h {}
     t
}
```


jkelleyrtp (approved):
Looks good, merge whenever :)

packages/autofmt/src/writer.rs
```diff
@@ -82,7 +82,8 @@ impl<'a> Writer<'a> {
 
         write!(self.out, "{name} {{")?;
 
-        self.write_rsx_block(attributes, spreads, children, brace)?;
+        let brace = brace.unwrap_or_default();
+        self.write_rsx_block(attributes, spreads, children, &brace)?;
```
jkelleyrtp, line 86: just fyi we'll be inserting a brace on nodes parsed as elements

jkelleyrtp (changes requested):
tests faling presumably due to parsing of children being changed.

see rsx_usage

packages/rsx/src/node.rs
```diff
@@ -79,13 +79,12 @@ impl Parse for BodyNode {
         // this is an Element if path match of:
         //
         // - one ident
-        // - followed by `{`
         // - 1st char is lowercase
         // - no underscores (reserved for components)
         //
         // example:
         // div {}
-        if stream.peek(Ident) && stream.peek2(Brace) {
```
jkelleyrtp, line 88: this change breaks parsing module::Component {}

jkelleyrtp (approved):

Issue #2615, opened 2024-07-10
Auto-fill crashes desktop app


rhaskia:
**Problem**
When clicking on the edge autofill for an input, I get the following panic message and my app crashes:
```
thread 'main' panicked at C:\Users\user\.cargo\registry\src\index.crates.io-6f17d22bba15001f\dioxus-html-0.5.1\src\transit.rs:41:17:
Failed to deserialize event data for event keydown:  MissingField(
    "char_code",
)
'Map(
    {},
)'
```

**Steps To Reproduce**
- Have an input, with previously entered data or however edge does that.
- Click into the input field, and the autofill should show up
- Click on any of the "Saved Data" and it should crash

**Expected behavior**
For the app not to crash, and the value to be put into the input.

**Screenshots**
![image](https://github.com/DioxusLabs/dioxus/assets/94507929/259bce73-a5b2-49ac-84e9-70faca4a74b5)

**Environment:**
 - Dioxus version:  0.5.1
 - Rust version: 1.79.0-nightly
 - OS info: Windows 10
 - App platform: Desktop

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

ealmloff:
If the browser doesn't give us any data for the event, we should probably not fire the event at all. Instead of panicing when deserialization fails, we can just return the serialization error and log it

ealmloff:
Fixed in https://github.com/DioxusLabs/dioxus/pull/2664

Issue #2385, opened 2024-05-02
pointermove events on Windows crash with InvalidType


esimkowitz:
**Problem**

The `onpointermove` event appears broken on Windows. When a component that binds to this event is present, the moment the user moves the cursor, the main thread panics with an InvalidType error. 

**Steps To Reproduce**

- Create a component that binds to the `onpointermove` event.
- On Windows, run `dx serve --platform desktop`
- Access the component with the bound event
- Try moving the mouse
- Observe that the application crashes

**Expected behavior**

Binding to the `onpointermove` event should return an `Event<PointerData>` event to the event handler.

**Screenshots**

N/A

**Environment:**
 - Dioxus version: 0.5.1
 - Rust version: 1.78.0
 - OS info: Windows 11 (22631.3527)
 - App platform: desktop

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

DogeDark:
I am unable to reproduce it on Windows 10. Perhaps it's a Windows 11 issue? I used this code snippet (is this what causes the crash for you?)

```rust
rsx! {
    div {
        style: "width: 10vw; height: 10vh; background-color: red;",
        onpointermove: move |data| {
            println!("hi | {:?}", data);
        }
    }
}
```

esimkowitz:
> I am unable to reproduce it on Windows 10. Perhaps it's a Windows 11 issue? I used this code snippet (is this what causes the crash for you?)

Yeah this is what caused the issue for me. I can try to dig into it a bit more and see if there's some field that is getting added for Windows 11 that we're missing.

Doublonmousse:
Same issue here, It segfaults on any input (mouse or pen input) on windows 11.

Doublonmousse:
All fields are named correctly, the issue is types

Doublonmousse:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">




</head>

<body link="#467886" vlink="#96607D">


[...]

PR #2664, opened 2024-07-21
Handle deserialize errors without panic

Septimus:
This addresses the panics shared in #2385 and #2615. It does not attempt to fix the deserialization issues.

Rather than panicking we return the error and log it


ealmloff (approved):
Thank you! This is a much better way to handle event deserialization errors

PR #2675, opened 2024-07-22
Don't panic when not using manganis

luveti:


jkelleyrtp (approved):

jkelleyrtp:
This is good in theory but magnanis can't really be disabled when running with the CLI - it seems like there's a bug in the magnanis-cli-support crate that's causing it to crash and/or not make that output file. We shouldn't be unwrapping there, but we also shouldn't be panicking upstream. This really isn't the right fix for that issue - can you do some sleuthing to figure out what's actually causing the MG json file to not exist?

luveti:
I dug into manganis, and have fixed the issue: https://github.com/DioxusLabs/manganis/pull/46. 

Just a classic NixOS issue :)

This explains why the `optimizing assets` step was taking so long, I'm guessing the CLI is timing out after the script failed. Now it completes instantly!

PR #2678, opened 2024-07-22
Fix ReadOnlySignal double drop

### LINKED ISSUE #2331: Diffing cloned Components with readonlysignal panics on double drop
**Problem**

Any props with the T -> ReadOnlySignal<T> logic fail to diff if you try to render it multiple times:

This is the signal version of the EventHandler problem fixed in #2298 

**Steps To Reproduce**

Run this code:
```rust
use dioxus::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {
    if generation() < 5 {
        println!("Generating new props");
        needs_update();
    }

    let read_only_signal_rsx = rsx! {
        // # Creation
        // This creates a new read only signal which is owned by the props
        // There is currently no Clone requirement to use this syntax
        // 
        // # Diffing
        // When memorizing the props, if the value is different then value is taken out of the signal
        TakesReadOnlySignal { sig: generation() as i32 }
    };

    rsx! {
        // When this is diffed, the value is moved out of read_only_signal_rsx
        {read_only_signal_rsx.clone()}
        // When this is diffed, it panics because the value has already been moved out of the signal in the previous line
        {read_only_signal_rsx}
    }
}

#[component]
fn TakesReadOnlySignal(sig: ReadOnlySignal<i32>) -> Element {
    rsx! {}
}
```

**Expected behavior**

**Screenshots**

**Environment:**
 - Dioxus version: `master`
 - Rust version: `nightly`
 - OS info: MacOS
 - App platform: `all`

**Questionnaire**

ealmloff:
Fixes #2331

### DIFF
--- a/packages/core-macro/src/props/mod.rs
+++ b/packages/core-macro/src/props/mod.rs
@@ -586,6 +586,17 @@ mod struct_info {
         }
 
         fn memoize_impl(&self) -> Result<TokenStream, Error> {
+            // First check if there are any child owned fields, if there are not, we can just use the partialEq impl
+            if !self.has_child_owned_fields() {
+                return Ok(quote! {
+                    let equal = mounted == new;
+                    if !equal {
+                        *mounted = new.clone();
+                    }
+                    equal
+                });
+            }
+
             // First check if there are any ReadOnlySignal fields, if there are not, we can just use the partialEq impl
             let signal_fields: Vec<_> = self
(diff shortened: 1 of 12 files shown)


ealmloff:
This adds a bound to ReadOnlySignal<T> in props that we can avoid by adding an extra level of indirection to signals in props instead

Issue #2666, opened 2024-07-22
Remove Dioxus CLI NASM dependency


opensource-inemar-net:
**Problem**

The installation of the dioxus-cli using:
cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli

on a windows computer fails due to a compilation problem.

error: failed to run custom build command for `aws-lc-sys v0.20.0`

Caused by:
  process didn't exit successfully: `C:\Users\yyyyyyy\AppData\Local\Temp\cargo-install370lfC\release\build\aws-lc-sys-a9bd54e5d3d757e6\build-script-main` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX
  cargo:rerun-if-env-changed=AWS_LC_SYS_INTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_CMAKE_BUILDER
  cargo:rerun-if-env-changed=AWS_LC_SYS_STATIC

  --- stderr
  AWS_LC_SYS_NO_ASM=1
  AWS_LC_SYS_CMAKE_BUILDER=1
  Missing dependency: cmake
  thread 'main' panicked at C:\Users\yyyyyy\.cargo\registry\src\index.crates.io-6f17d22bba15001f\aws-lc-sys-0.20.0\builder/main.rs:289:38:
  called `Result::unwrap()` on an `Err` value: "Required build dependency is missing. Halting build."
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile `dioxus-cli v0.5.2 (https://github.com/DioxusLabs/dioxus#ba64ee04)`, intermediate artifacts can be found at `C:\Users\yyyyyy\AppData\Local\Temp\cargo-install370lfC`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

**Steps To Reproduce**

- windows machine without cmake and nasm installed
- cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli
- 

**Expected behavior**

[...]

PR #2682, opened 2024-07-23
Remove Dioxus CLI NASM dependency #2666

opensource-inemar-net:
This should fix: https://github.com/DioxusLabs/dioxus/issues/2666


jkelleyrtp:
Thanks! I think aws-lc-sys still shows up in the crate graph due to ureq but that needs to be fixed in manganis.

opensource-inemar-net:
Hi !

on my windows machine aws-lc-sys is gone when checking out current main.

ureq already disabled it in their cargo toml.
rustls = { version = "0.23.5", optional = true, default-features = false, features = ["ring", "logging", "std", "tls12"] }

attached cargo tree from windows system.

Luckily no more work needed to fix the issue.
[tree.txt](https://github.com/user-attachments/files/16402314/tree.txt)

Thank you for the fast merging.

PR #2685, opened 2024-07-23
TUI tweaks

matthunz:
- [x] Tab ordering should be 1-indexed
- [x] Flip to the build tab on build errors
- [x] Keybinds
  - [x] Clear the current screen: `c`
  - [x] Reload the current app: `r`
  - [x] Open the current app: `o`
- [x] Added placeholder to `/` screen for now


jkelleyrtp (approved):

packages/cli/src/serve/output.rs
```diff
@@ -159,7 +162,7 @@ impl Output {
     /// Why is the ctrl_c handler here?
     ///
     /// Also tick animations every few ms
-    pub async fn wait(&mut self) -> io::Result<()> {
+    pub async fn wait(&mut self) -> io::Result<bool> {
```
jkelleyrtp, line 165: In the future we should provide a comment as to what the boolean actually means - or communicate this through a different mechanism.

PR #2689, opened 2024-07-23
Fix: File Watcher Ignore List

DogeDark:
We were feeding the `ignore` crate the wrong data and it would build an empty ignore list. Fixing this greatly speeds up `serve` start times as we aren't checking every file in folders like `target`.


jkelleyrtp:
THANK YOU! I figured we were doing something wrong here.

PR #2690, opened 2024-07-23
CI: re-reorder free disk space action to clear free up space earlier

jkelleyrtp:
Attempt to fix CI for the release check action by clearing disk space earlier.


jkelleyrtp:
Playwright won't pass but the rest will.

PR #2702, opened 2024-07-25
CLI: Toasts & Tweaks

DogeDark:
**Changes**
- Adds toasts messages for full rebuilds on the web platform. These will appear in the top right corner for 5 seconds and are dismissable with a click. In the future, this would be a great configurable CLI setting for those who do not want it.
- On the desktop platform the splash screen will no longer open in the browser. It is still available on the devserver.
- A few naming and doc additions in `server.rs` file.

<details>
<summary>Toasts Preview</summary>

</details>


jkelleyrtp (approved):
Looks good! One thing I noticed is that when fullstack reloads, the server is not yet ready so the proxy fails and we end up in an error page state.

That happens on main too, so not required for this PR.

A follow up here would be maybe a spinner animation or something for the "rebuilding..." case and then keep the toast alive until the page reloads (or is cleared manually).

PR #2705, opened 2024-07-25
Switch to a pool of dynamic values for hot reloading

### LINKED ISSUE #2077: Removing components with hotreloading enabled can lead to page crash
**Problem**

If you have a series of components and remove one or multiple of them, the hotreloading resolver in core will crash, presumably because the scope we've removed no longer exists during reconciliation.

Not sure what the solution is.

```rust
use dioxus::prelude::*;

fn main() {
    dioxus::launch(|| {
        rsx! {
            div {
                "hi!!"
                Child {}
                Child {}
                Child {}
            }
        }
    })
}

#[component]
fn Child() -> Element {
    let mut counter = use_signal(|| 0);
    rsx! {
        div {
            "hi"
            button { onclick: move |_| { counter += 1 } }
        }
    }
}
```

And then delete one of the components

**Environment:**
 - Dioxus version: [e.g. v0.17, `master`]
 - Rust version: [e.g. 1.43.0, `nightly`]
 - OS info: [e.g. MacOS]
 - App platform: [e.g. `web`, `desktop`]

**Questionnaire**

ealmloff:
Instead of relying on the template structure for hot reloading, we can hot reload VNodes with signals.

We create a list of baked dynamic objects in each rsx block at compile time and then map that list into a VNode with a template. That template is driven by a signal that can be hot reloaded

Compared to the current version of hot reloading:
1) This is simpler to diff - we only need to match up rsx blocks
2) This is less tied to core - we no longer need special diffing logic to dfs node/attribute paths
3) This lets us create new dynamic text nodes and text attributes on the fly as long as the formatted segments are the same
4) This lets us move dynamic text between different attributes and nodes
5) It also implements hot reloading rsx that starts as empty

https://github.com/user-attachments/assets/3463ae4a-3ebd-4d83-a982-c1591af65de2

Fixes #2077


packages/rsx/src/template_body.rs
```diff
@@ -81,8 +79,14 @@ pub struct TemplateBody {
     pub template_idx: DynIdx,
     pub node_paths: Vec<NodePath>,
     pub attr_paths: Vec<(AttributePath, usize)>,
+    pub dynamic_text_segments: Vec<FormattedSegment>,
     pub diagnostics: Diagnostics,
-    current_path: Vec<u8>,
+}
+
+impl Default for TemplateBody {
+    fn default() -> Self {
```
jkelleyrtp, line 87: maybe no default here - and just use this in the to_tokens directly? This might be confusing if we try to ::default() somewhere

jkelleyrtp (approved):
A few places where some more comments might be helpful. Maybe worth commenting/reorganizing the diff.rs file slightly. Otherwise looks good.

PR #2710, opened 2024-07-26
chore(cli): slimmed down codebase by using vergen

Andrew15-5:
Auxiliary files had to be created to use `built` crate. Now they are removed.

Now any build-time info can be grabbed from env var via `env!()`. For some additional data only a new feature must be added with a corresponding line in the `build.rs` to enable it.

One concern is that the number of built crates increased by 91. But I think practically all of them are tiny, so it wouldn't affect compile time much, if at all.

I run 2 tests and the first time the difference was 5 m 34 s vs. 9 m 17 s (I think it was an "improvement"). The second time it was 5 m 32 s vs. 5 m 41 s. 2.7% slower, without proper testing. I think it doesn't matter since you have to wait 5.5 minutes either way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1181,9 +1181,6 @@ name = "built"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c6a6c0b39c38fd754ac338b00a88066436389c0f029da5d37d1e01091d9b7c17"
-dependencies = [
- "git2",
-]
 
 [[package]]
 name = "bumpalo"
@@ -1306,7 +1303,7 @@ dependencies = [
  "env_logger 0.11.3",
  "fs-err",
  "git2",
- "gix-config",
+ "gix-config 0.36.1",
  "heck 0.5.0",
  "home",
  "ignore",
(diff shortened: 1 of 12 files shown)


jkelleyrtp:
Thanks for the PR! However... vergen brings in a lot of extra dependencies when compared to built (just git2). I get that the vergen folks want to use gix instead of libgit2 but I'm not super convinced we should bring on their larger dependency tree just to get away from that.

There are other changes in this PR that are nice if you want to still include those, otherwise I don't think it's worth switching over.

jkelleyrtp (changes requested):
Thanks for the PR! However... vergen brings in a lot of extra dependencies when compared to built (just git2). I get that the vergen folks want to use gix instead of libgit2 but I'm not super convinced we should bring on their larger dependency tree just to get away from that.

There are other changes in this PR that are nice if you want to still include those, otherwise I don't think it's worth switching over.

Andrew15-5:
I've checked how much unique and common dependencies does `vergen-git2`, `vergen-gix`, `vergen-gitcl`, and `built` have with the codebase. `built` has less common ones but also much less unique ones.

`vergen-gix`:
- unique: 72
- common: 53

`vergen-git2`:
- unique: 21
- common: 27

`vergen-gitcl`:
- unique: 16
- common: 15

`built`:
- unique: 6
- common: 13

I don't understand where exactly `links = ` is used so that with `vergen-git2` it fails to build, but with `vergen-gitcl` it only increases dependency count by 22 (856). It requires `git` binary to be installed, but I haven't checked what happens if it isn't. Maybe it fails to run `git` command or something similar. But realistically no one would compile the CLI without having `git` installed. What do you think about this compromise?

Since [you've said](https://discord.com/channels/899851952891002890/1265730874066997258/1265803673678774272):
> I just pulled in built for the git hash, but there's definitely more data I want to export onto our new tui help screen

what exactly do you want to use? Because for now I can probably hand-roll a few env var generators myself since we only need to get commit hash and build profile. It's pretty easy and this approach is used in Typst CLI. This way the dependency count will go down by about 6 (828?). So it's a win-win, but hand-made though still extensible (depends on what other info you need).

jkelleyrtp:
Not worth the churn and the extra dependency count - built seems fine for now. We can revisit this if built becomes an issue.

PR #2715, opened 2024-07-26
Feat: `always_on_top` CLI Setting

### LINKED ISSUE #2354: Add ability to configure "always on/pin to top" option for the Dioxus Desktop
## Feature Request

The "always on top" feature is blocking the file choose window (`<input type="file">`), which is supposed to be a part of the app. So not only it blocks other dev windows (editor, `dx` and `tailwind` commands), but also itself. And when moving to the virtual desktop with that window and an editor, it will always refocus on that window, instead of the previously focused editor, which is just not a great DX, to put it mildly.

## Implement Suggestion

Probably a toggle either in the `Dioxus.toml` file or in the `dioxus::desktop::Config::new()`.

DogeDark:
Adds a new CLI setting for configuring the always-on-top CLI setting for debug-mode Dioxus desktop.

Closes #2354

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10382,7 +10382,7 @@ source = "git+https://github.com/DioxusLabs/warnings#9889b96cccb6ac91a8af924cfee
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.72",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Perfect - thank you :)

PR #2716, opened 2024-07-26
Fix #2378: CLI MSRV and add MSRV to CI

### LINKED ISSUE #2378: Dioxus CLI MSRV
**Problem**

Can't compile dioxus-cli v0.5.4. What is the MSRV for cli? I can't find any information about it. What is the MSRV for Dioxus itself?

`Updating crates.io index
  Downloaded dioxus-cli v0.5.4
  Downloaded 1 crate (144.1 KB) in 0.18s
  Installing dioxus-cli v0.5.4
error: failed to compile `dioxus-cli v0.5.4`, intermediate artifacts can be found at `/tmp/cargo-installLEyIVT`

Caused by:
  failed to select a version for `env_logger`.
      ... required by package `dioxus-cli v0.5.4`
  versions that meet the requirements `^0.11.3` are: 0.11.3

  the package `dioxus-cli` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
   It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.

  failed to select a version for `env_logger` which could resolve this conflict`

**Steps To Reproduce**

- cargo install dioxus-cli 

**Expected behavior**

Compile successful.

**Environment:**
 - cargo 1.67.1 - can't update it for embedded target

**Questionnaire**

jkelleyrtp:
- Fix: bump msrv up to be valid for cli
- add binstall check to ci
- adjust msrv of tomls

Unfortunately due to a number of changes in our upstream and our own syntax uses, we have to move our msrv up to 1.79

It'd be nice to support lower versions

Fixes #2378

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,16 +53,7 @@ jobs:
         with:
           cache-all-crates: "true"
           save-if: ${{ github.ref == 'refs/heads/main' }}
-      # https://github.com/foresterre/cargo-msrv/blob/4345edfe3f4fc91cc8ae6c7d6804c0748fae92ae/.github/workflows/msrv.yml
-      - name: install_cargo_msrv
-        run: cargo install cargo-msrv --all-features
-      - name: version_of_cargo_msrv
-        run: cargo msrv --version
-      - name: run_cargo_msrv
-        run: cargo msrv --output-format json verify -- cargo check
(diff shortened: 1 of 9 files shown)


ealmloff (approved):
Nice! This has been on the todo list for a while

PR #2719, opened 2024-07-27
Follow-up for error messages in core

matthunz:
This follows https://github.com/DioxusLabs/dioxus/pull/2629 to replace `.unwrap` with `panic!` for clearer error messages.

Previously when unwrapped the message outputted was:
```
thread 'main' panicked at src/main.rs:62:49:
Called `Result::unwrap()` on an `Err` value: RuntimeError
```

But with `#[track_caller]` and `panic!` we get:
```
thread 'main' panicked at src/main.rs:62:49:
Must be called from inside a Dioxus runtime.

Help: Some APIs in dioxus require a global runtime to be present.
If you are calling one of these APIs from outside of a dioxus runtime
(typically in a web-sys closure or dynamic library), you will need to
grab the runtime from a scope that has it and then move it into your
new scope with a runtime guard.

For example, if you are trying to use dioxus apis from a web-sys
closure, you can grab the runtime from the scope it is created in:

use dioxus::prelude::*;
static COUNT: GlobalSignal<i32> = Signal::global(|| 0);

#[component]
fn MyComponent() -> Element {
    use_effect(|| {
        // Grab the runtime from the MyComponent scope
        let runtime = Runtime::current().expect("Components run in the Dioxus runtime");
        // Move the runtime into the web-sys closure scope
        let web_sys_closure = Closure::new(|| {
            // Then create a guard to provide the runtime to the closure
            let _guard = RuntimeGuard::new(runtime);
            // and run whatever code needs the runtime
            tracing::info!("The count is: {COUNT}");
        });
    })
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10382,7 +10382,7 @@ source = "git+https://github.com/DioxusLabs/warnings#9889b96cccb6ac91a8af924cfee
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.71",
+ "syn 2.0.72",
 ]
 
 [[package]]
--- a/examples/rsx_usage.rs
+++ b/examples/rsx_usage.rs
@@ -95,6 +95,12 @@ fn app() -> Element {
                     }
                 }
             }
+
+            // dangerous_inner_html for both html and svg
+            div { dangerous_inner_html: "<p>hello dangerous inner html</p>" }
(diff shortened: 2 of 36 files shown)


jkelleyrtp (approved):
Much nicer output :)
It might be worth eventually adding some std-err tests on core to make sure we don't accidentally drift in the future since these are rather innocuous `unwrap_or_else`-es.

Issue #2723, opened 2024-07-28
dx fmt panics if faulty orphan file exists


opensource-inemar-net:
**Problem**

Project which contains a .rs file which is not referenced by lib.r or main.rs directly or indirect and contains faulty syntax in rsx!.
When running dx fmt panic occurs:

C:\aPerf\dioxus\test2>dx fmt
thread '<unnamed>' panicked at packages\autofmt\src\lib.rs:72:65:
called `Result::unwrap()` on an `Err` value: Error("expected identifier, found keyword `let`")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

**Steps To Reproduce**

[test2.zip](https://github.com/user-attachments/files/16402338/test2.zip)

- create a project
- create a file with .rs extension and faulty rsx!
- run dx fmt

**Expected behavior**

There are two possible enhancments:

Simple:
 - dx fmt should output the offending file and the error location instead of panic

Possible Advanced:
 - dx fmt should only format files which are actually part of the build

**Environment:**
 - Dioxus version:  main
 - Rust version:    1.79.0
 - OS info:         windows

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [X ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Next week or so I will provide a pull request with the simple solution.

PR #2724, opened 2024-07-28
This fixes issue https://github.com/DioxusLabs/dioxus/issues/2723

opensource-inemar-net:
https://github.com/DioxusLabs/dioxus/issues/2723

In case we have a syntax error during parsing of rsx! we do the following:
- write out error msg to stderr unfortunately we are not able to provide info about the fiel which contains the error.
- stop parsing and returned the onmodified file

As this is done also in other cases (silent continue) these seems to be acceptable


jkelleyrtp (approved):
Thank you! Good catch

PR #2725, opened 2024-07-28
fix issue 1586

i123iu:
fix issue [1586](https://github.com/DioxusLabs/dioxus/issues/1586)


jkelleyrtp (approved):
Thanks - figured that this would be it and glad it's such an easy fix.

i123iu:
should also fix [this](https://github.com/DioxusLabs/dioxus/issues/2486)

Issue #2729, opened 2024-07-29
Desktop window state (position, state) not saved


i123iu:
**Problem**

When closing the app window on Windows desktop, the window state (position, size) is not saved. 

**Steps To Reproduce**

- Create new desktop app
- `dx serve`
- the window is at position A
- move the window to B
- close it
- restart the app
- the position stays A

**Expected behavior**

It stays B as is written [here](https://github.com/DioxusLabs/dioxus/blob/main/packages/desktop/src/app.rs#L432)

**Environment:**
 - Dioxus version: 0.5 and master
 - Rust version: rustc 1.80.0
 - OS info: Windows 10
 - App platform: desktop

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [X] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

PR #2730, opened 2024-07-29
desktop app saves window position fix

i123iu:
fixes [this](https://github.com/DioxusLabs/dioxus/issues/2729)
`app.handle_loop_destroyed` is called after the window is destroyed so it cannot be saved
added another save right before the last window is destroyed


jkelleyrtp:
Good catch - there'd be no window for us to persist when the final window is closed!

Thanks!

jkelleyrtp (approved):

Issue #2265, opened 2024-04-06
dx serve does not stop after desktop app is closed


umnovI:
**Problem**

After closing a desktop app `dx serve` stops only if pressed ctrl+c

**Steps To Reproduce**

- Create a desktop app through cli with tailwindcss true and router false
- Start the app with `dx serve`
- Close the app's window
- See cli still being used by dx

**Expected behavior**

dx should stop

**Screenshots**

https://github.com/DioxusLabs/dioxus/assets/30232564/90745630-6a65-4f0a-bd74-2383d503a9be

**Environment:**
 - Dioxus version: dioxus = { version = "0.5", features = ["desktop"] }
 - Rust version: rustc 1.77.1 (7cf61ebde 2024-03-27)
 - OS info: win11
 - App platform: desktop

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [x] I don't have time to fix this right now, but maybe later

PR #2734, opened 2024-07-29
Fix 2265: close tui on success, custom tui subscriber

jkelleyrtp:
Fix: #2265
Fix: #2735 

Also adds a bit of coloring for different message sources.


DogeDark:
Not sure if this is exactly what you had in mind but it works.

jkelleyrtp:
I'm getting panics when trying to run this for fullstack.

```
     Running `/Users/jonkelley/Development/dioxus/target/cli-dev/dx serve`
2024-07-31T00:05:57.751582Z  INFO dx::builder::cargo: 🚅 Running build [Desktop] command...
2024-07-31T00:05:57.751575Z  INFO dx::builder::cargo: 🚅 Running build [Desktop] command...
thread 'main' panicked at packages/cli/src/serve/output.rs:232:72:
called `Option::unwrap()` on a `None` value
```
The tracing log for this platform is coming in before its registered by the build engine causing a race condition + panic. We should use entry().unwrap_or_default() or shell out to the push_log method and make it use the entry api instead.

ealmloff (approved):

Issue #2604, opened 2024-07-08
Unnecessary spaces after `dx fmt`


chungwong:
**Problem**

**Steps To Reproduce**

Before `dx fmt`
```rs
    rsx! {
        if let Some(Some(record)) = &*records.read_unchecked() {
            {
                let (label, value): (Vec<String>, Vec<f64>) = record.iter().rev().map(|d| (d.model.clone().unwrap(), d.row_total)).collect();

                rsx! {
                    BarChart {
                        id: "bar-plot".to_string(),
                        x: value,
                        y: label,
                    }
                }
            }
        }
    }
```
After `dx fmt`
```diff
         if let Some(Some(record)) = &*records.read_unchecked() {
             {
                 let (label, value): (Vec<String>, Vec<f64>) = record.iter().rev().map(|d| (d.model.clone().unwrap(), d.row_total)).collect();
-
+            
                 rsx! {
                     BarChart {
                         id: "bar-plot".to_string(),

```
**Expected behavior**
Should not add the extra spaces.

**Screenshots**

**Environment:**
 - Dioxus version: dioxus = { version = "0.5", features = ["fullstack", "router"] }
 - Rust version:   rustc 1.79.0 (129f3b996 2024-06-10)
 - OS info:  Debian 12
 - App platform:   fullstack

istudyatuni:
I'm facing the same issue with dx v0.6.1

Issue #2341, opened 2024-04-19
dx fmt deletes comments in non-expression postion


magikmw:
**Problem**

Using `dx fmt` on source files is deleting comments, both in rsx macros and standard rust when using `--all-code`.  
This is especially egregious because standard `rustfmt` formatter is really messing up any files with rsx in them, so `dx fmt` is the only viable alternative.

**Steps To Reproduce**

- Put any comments into a file - either inside rsx or elsewhere
- Run `dx fmt -f yourfile.rs`
- Comments deleted

**Expected behavior**

Comments should be left alone.  
I'm not entirely sure how `rustfmt` does it, but with my limited experience with it, it seems to keep comments intact and relatively in the relevelant place after formatting.

**Screenshots**

n/a

**Environment:**
 - Dioxus version: 0.5.1
 - Rust version: rustc 1.77.2 (25ef9e3d8 2024-04-09)
 - OS info: Fedora 40 beta
 - App platform: n/a (but I'm making a fullstack app)

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [x] I don't have time to fix this right now, but maybe later

DogeDark:
I am unable to reproduce this using the git version of the CLI. Can you test the git version of the CLI on your end?

You can install the git version with:
```
cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli
```

magikmw:
I did run the command to install git version and checked again:

```diff
-                // TODO: Implement event handling here -> listen to eve
nts in child
                 let accept_keys = [Key::Enter, Key::Character(" ".to_string())];
-                if accept_keys.contains(&evt.key()){
+                if accept_keys.contains(&evt.key()) {
                     log::info!("{} pressed in {}", evt.key(), "{props.labelText}");
                     log::info!("{} pressed in {}", evt.key(), "{props.labelText}");
                 }
             },
             class: "flex flex-1 h-full w-3/4 items-center justify-center btn btn-outline my-1 py-1",
-            div { class: "flex-none py-1", // TODO: Refactor into separate component
+            div { class: "flex-none py-1",
                 Icon { width: 24, height: 24, icon: props.icon, title: "{props.iconTitle}" }
-            },
+            }
             "{props.labelText}"
-            input { // TODO: Refactor into separate component
+            input {

```

dx -V says:

`dioxus 0.5.2`

DogeDark:
[...]

Issue #2240, opened 2024-04-04
Improve error handling in `dx fmt`


matthunz:
## Feature Request

Currently when `dx fmt` fails, I usually see a panic with little indication to what's causing the error. While formatting is being worked out I think it'd be helpful to see a line number and maybe a section of the code causing the error, so we can all help narrow down formatting bugs.

## Implement Suggestion

Replace `.unwrap()` by returning a custom error type that wraps `anyhow::Error`.
Maybe something like
```rust
pub struct Error {
  span: Option<Span>,
  cause: anyhow::Error
}
```

Issue #1492, opened 2023-09-25
Make dioxus formatting respect rustfmt settings


Natrocx:
## Specific Demand

Dioxus formatting (as in `dx fmt`) only applies default values for settings like tab size and maximum line length creating a break in style between "normal" rust code and the code inside `rsx!` macros.

## Implement Suggestion

Read the content of rustfmt.toml (and possibly overwrite configured values if other values are specified in dioxus.toml if that happens to make sense).

jkelleyrtp:
Closed by 
https://github.com/DioxusLabs/dioxus/pull/1597

PR #2736, opened 2024-07-29
Fix: #2604, Fix: #2240, Fix: #2341, Fix #1355 - Better error handling and and spaces handling in autofmt

### LINKED ISSUE #1355: `#[dxfmt::skip]` similar to `#[rustfmt::skip] `
## Specific Demand

I would like a macro to flag portions of code that you don't want `dx fmt` to touch.
This would be handy on codebases that you would want `rustfmt` to format your structs:

```rust
my_thing: Thing = Thing {...}
```

And `dx fmt` would just format this as a component.

Then, `rustfmt` and Dioxus would fight over the formatting.
I can always turn `rustfmt` off but I cannot do the same with Dioxus formatter.

## Implement Suggestion

Similar to `#[rustfmt::skip]`

jkelleyrtp:
This fixes a few bugs in autofmt such that fewer comments get eaten.

- [x] Fix: #2604 
- [x] Fix: #2240
- [x] Fix: #1355 
- [x] Fix: #1488 
- [x] Fix: #2341
- [x] Fix: #1492

However, it still doesn't fix comments getting eaten in handlers - this will require rustfmt integration

We now respect rustfmt::skip settings as well as rustfmt.toml


packages/autofmt/src/collect_macros.rs
```diff
@@ -3,37 +3,66 @@
...
-        file,
-    );
+    let mut collector = MacroCollector::new(&mut macros);
+    MacroCollector::visit_file(&mut collector, file);
     macros
 }
 
 struct MacroCollector<'a, 'b> {
     macros: &'a mut Vec<CollectedMacro<'b>>,
+    skip_stack: Vec<()>,
```
ealmloff, line 19: This could be usize

packages/autofmt/src/prettier_please.rs
```diff
@@ -1,153 +1,159 @@
...
+                i.path = syn::parse_str(MARKER).unwrap();
+                i.tokens = Default::default();
+
+                // Push out the indent level of the formatted block if it's multiline
+                // Normally we'd use a `\n` as the delimiter, but we actually want to make sure
+                // any block-level things are indented properly
+                //
+                // we should really just be parsing the block here and using the same heuristic
+                // rsx_call uses (short strings are inlined, all others are multiline
+                if formatted.contains('{') {
```
ealmloff, line 50: Will this format "{hello world}" as a multi-line string?

packages/autofmt/src/writer.rs
```diff
@@ -114,66 +119,441 @@ impl<'a> Writer<'a> {
...
+
+    pub fn write_body_nodes(&mut self, children: &[BodyNode]) -> Result {
+        let mut iter = children.iter().peekable();
+
+        while let Some(child) = iter.next() {
+            if self.current_span_is_primary(child.span().start()) {
+                self.write_comments(child.span().start())?;
+            };
+            self.out.tab()?;
+            self.write_ident(child)?;
```
ealmloff, line 246: This should be write_body?

packages/autofmt/src/writer.rs
```diff
@@ -114,66 +119,441 @@ impl<'a> Writer<'a> {
...
+                write!(self.out, " ")?;
+
+                self.write_attributes(attributes, spreads, true, brace, has_children)?;
+
+                if !children.is_empty() && !attributes.is_empty() {
+                    write!(self.out, " ")?;
+                }
+
+                for child in children.iter() {
+                    self.write_ident(child)?;
```
ealmloff, line 355: Do we need to insert spaces between children here?

packages/autofmt/src/writer.rs
```diff
@@ -114,66 +119,442 @@ impl<'a> Writer<'a> {
...
+        let cond = self.unparse_expr(&if_chain.condition);
+        write!(self.out, "if {cond} {{ ")?;
+        self.write_attribute_value(&if_chain.then_value)?;
+        write!(self.out, " }}")?;
+        match if_chain.else_value.as_deref() {
+            Some(AttributeValue::IfExpr(else_if_chain)) => {
+                write!(self.out, "else ")?;
+                self.write_attribute_if_chain(else_if_chain)?;
+            }
+            Some(other) => {
```
ealmloff, line 546: I think this should be `else {{ `

packages/autofmt/src/writer.rs
```diff
@@ -262,13 +650,16 @@ impl<'a> Writer<'a> {
                     .unwrap_or_default();
                 condition_len + value_len + if_len + brace_len + space_len + else_len
```
ealmloff, line 651: Not caused by this PR, but the else_len calculation is wrong

packages/autofmt/src/writer.rs
ealmloff: No because we only do single line when there is one child

packages/rsx/src/literal.rs
```diff
@@ -151,6 +151,14 @@ pub struct HotReloadFormattedSegment {
     pub dynamic_node_indexes: Vec<DynIdx>,
 }
 
+impl HotReloadFormattedSegment {
+    /// This mnethod is very important!
+    /// Deref + Spanned + .span() methods leads to name coliisions
```
ealmloff, line 156: typos

ealmloff (approved):

PR #2745, opened 2024-07-30
Workspace support for `dx fmt`

matthunz:
Adds `dx fmt -p my_package` and `dx fmt --package my_package` to specify formatting a package in a Cargo workspace.

Fixes https://github.com/DioxusLabs/dioxus/issues/1547

This uses the new workspace logic in `DioxusCrate` to get the path to a specific package in a workspace. Maybe we should split up `DioxusCrate` into smaller parts here eventually instead of re-using it entirely?


jkelleyrtp (approved):
Let's see if we can get `check` wired up with the `package` field on `target_args` and then I think everything has workspace support!

PR #2756, opened 2024-07-31
Fix the base path in the CLI

ealmloff:
This is an alternative to https://github.com/DioxusLabs/dioxus/pull/2395 with absolute links

Both `/basepath` and `/basepath/` work

Closes https://github.com/DioxusLabs/dioxus/issues/2285


jkelleyrtp (approved):
We should *really* set up a harness so these things stay more stable.

PR #2757, opened 2024-07-31
Feat: Auto Default Into

### LINKED ISSUE #2373: Call `.into()` automatically for the default value when deriving `Props`
When using `#[props(into, default = "...")]`, we should call `.into()` automatically for the default value. It will be more conveninent.
```rust
#[derive(Clone, PartialEq, Props)]
pub struct FormFieldContainerProps {
    /// The class attribute for the component.
    #[props(into, default = "field is-horizontal".into())]
    pub class: Class,
    /// A class to apply to the field label container.
    #[props(into, default = "field-label".into())]
    pub field_label_class: Class,
    /// A class to apply to the field body container.
    #[props(into, default = "field-body".into())]
    pub field_body_class: Class,
    /// A class to apply to the `label` element.
    #[props(into, default = "label".into())]
    pub label_class: Class,
    /// The label content.
    #[props(into)]
    pub label: SharedString,
    /// The children to render within the component.
    children: Element,
}
```

DogeDark:
Allows automatic `into()` for the default value in props.

```rs
// E.g. instead of 

#[derive(Clone, PartialEq, Props)]
pub struct SomeCoolProps {
    #[props(into, default = "abc".to_string())]
    pub data: String,

    #[props(into, default = true.into())]
    pub other_data: u64,

    #[props(into, default = true.into())]
    pub other_other_data: u16,
}

// you can now do this

#[derive(Clone, PartialEq, Props)]
pub struct SomeCoolProps {
    #[props(into, default = "abc")]
    pub data: String,

    #[props(into, default = true)]
    pub other_data: u64,

    #[props(into, default = true)]
    pub other_other_data: u16,
}
```

Closes #2373

### DIFF
--- a/packages/core-macro/src/props/mod.rs
+++ b/packages/core-macro/src/props/mod.rs
@@ -1373,10 +1373,10 @@ Finally, call `.build()` to create the instance of `{name}`.
                     quote!(let #name = self.#name;)
                 } else if let Some(ref default) = field.builder_attr.default {
                     // If field has `into`, apply it to the default value.
-                    let into = if field.builder_attr.auto_into { 
+                    let into = if field.builder_attr.auto_into {
                         quote!{ .into() }
                     } else if field.builder_attr.auto_to_string {
-                        quote!{ .to_string() }  
+                        quote!{ .to_string() }
                     } else {
                         quote!{}
                     };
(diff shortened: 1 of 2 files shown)


jkelleyrtp (changes requested):
Looks great! Do you mind adding a a testcase somewhere in the codebase - either in rsx_usage or in core_macro/tests (or anywhere really...) that stresses this feature and maybe links to the issue, just so we don't accidentally break this in the future?

jkelleyrtp (approved):

PR #2772, opened 2024-08-02
Bubble errors from the server to the client on suspense boundaries

### LINKED ISSUE #2570: Errors don't bubble through a SuspenseBoundary
**Problem**

The following code produces an unwrap error:

```
use serde::{Deserialize, Serialize};
use strum::{AsRefStr, Display, EnumString};

use crate::prelude::*;

#[derive(
    Clone, Copy, Eq, PartialEq, Debug, Serialize, Deserialize, AsRefStr, Display, EnumString,
)]
#[strum(serialize_all = "snake_case")]
pub enum ServerError {
    Unauthenticated,
}

impl std::error::Error for ServerError {}

#[component]
pub fn App() -> Element {
    rsx! {
        ErrorBoundary {
            handle_error: |_| rsx! {
                "Hmm, something went wrong."
            },
            Home {}
        }
    }
}

#[component]
pub fn Home() -> Element {
    rsx! {
        SuspenseBoundary {
            fallback: |context: SuspenseContext| rsx! {
                div {
                    "Loading..."
                }
            },
            Stuff {
                id: "floob_123",
            }
        }
    }
}

#[component]
pub fn Stuff(id: String) -> Element {
    let Some(floob) = use_floob(id)?.suspend()?()? else {
        return rsx! {
            "No floob"
        };
    };
    rsx! {
        {floob}
    }
}

#[server]
pub async fn floob(id: String) -> Result<Option<String>, ServerFnError<ServerError>> {
    tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
    Err(ServerFnError::WrappedServerError(ServerError::Unauthenticated))
}

pub fn use_floob(id: String) -> Result<Resource<Result<Option<String>, ServerFnError<ServerError>>>, RenderError> {
    use_server_future(move || {
        to_owned![id];
        async move {
            floob(id).await
        }
    })
}
```

Running it gives the following error:
image

Changing Home to this code, however, works:

```
#[component]
pub fn Home() -> Element {
    rsx! {
        SuspenseBoundary {
            fallback: |context: SuspenseContext| rsx! {
                div {
                    "Loading..."
                }
            },
            ErrorBoundary {
                handle_error: |_| rsx! {
                    "Hmm, something went wrong inside of the suspense boundary."
                },
                Stuff {
                    id: "floob_123",
                }
            }
        }
    }
}
```

The error can't bubble through the SuspenseBoundary correctly.

**Environment:**
 - Dioxus version: `master`
 - Rust version: `v1.81.0-nightly`
 - OS info: `Arch Linux`
 - App platform: `fullstack`

**Questionnaire**

ealmloff:
This PR handles suspense boundaries throwing an error into their suspended parent. Instead of trying to rerun the error boundary that is frozen, we send the error to the client and continue bubbling the error there

Fixes #2570


jkelleyrtp (approved):
We chatted about this at length re: the behavior of inserting error boundaries on-the-fly.

While it's not ideal, it's fine for now. Ultimately we don't want to pay that cost while streaming and only pay that cost when the error actually propagates, but there's some tradeoffs there about how interconnected core needs to be with streaming.

We should definitely mark this as "todo" but not stop the fix for 0.6.

PR #2774, opened 2024-08-02
Add argument to `use_callback`

### LINKED ISSUE #2371: use_callback should take a closure that accepts an argument
## Feature Request
Currently there's no way to pass in an argument to the closure that use_callback wraps. This should be easy enough to add. I think the only reason we didn't add it is to make the type signature of UseCallback a bit more compact, but that's not so important.

matthunz:
Adds a generic argument to the `use_callback` hook

Fixes #2371

```rs
let f = use_callback(|n| println!("{}", n));
f(0)
```


jkelleyrtp (approved):
Seems about right - I expected there to be more of a cascade of type errors but I guess all of our built-in callbacks don't take a type.

PR #2775, opened 2024-08-02
Add new crate to workspace members

### LINKED ISSUE #2350: dx new in existing multicrate workspace
## Feature Request

When creating a new project with `dx new`, after the directory is created `cargo fmt` is ran. If this is in a workspace where the user has not pre-added the crate to workspace members list, the user is presented with a cargo fmt error message.

Everything else runs fine, and the command has an exit code of 0, as everything is fine, it just looks worrying.

## Implement Suggestion

Not sure how possible it is, currently investigating, but if the user is in a workspace either
- the user is asked in the template whether to add the newly created crate to members or not
- or; add it anyway without asking

matthunz:
When inside a Cargo workspace, this now adds newly created crates (with `dx new` or `dx init`) to the root workspace `members` array.

Fixes #2350

#### Unsolved issue:
Running `dx new my_crate` in an empty virtual cargo workspace (e.g. `members = []`) will still produce an error after the crate is created.

`cargo metadata` seems to return an error in this case, without the crate information. In this case I suggest we either leave the current behavior, and maybe trim the error messages, or cancel creating the new crate entirely until the manifest is properly setup.


jkelleyrtp (approved):
Had to make changes so `dx new packages/blah` actually put the crate into the packages folder.

matthunz:
> Had to make changes so `dx new packages/blah` actually put the crate into the packages folder.

My fault I didn't know you could do that with cargo 🤔 thanks for the help here

PR #2786, opened 2024-08-05
Re-export the axum feature flag under "server"

ealmloff:
This PR re-exports the axum feature flag on dioxus under the server feature. Now that we no longer support multiple backends for fullstack, it doesn't make sense to pick a specific backend to enable on the server. The old feature flag is kept for compatibility


packages/dioxus/Cargo.toml
```diff
@@ -58,6 +58,7 @@ ssr = ["dep:dioxus-ssr", "dioxus-router?/ssr", "dioxus-config-macro/ssr"]
 liveview = ["dep:dioxus-liveview", "dioxus-config-macro/liveview", "dioxus-router?/liveview"]
 static-generation = ["dep:dioxus-static-site-generation", "dioxus-config-macro/static-generation"]
 axum = ["dioxus-fullstack?/axum", "dioxus-fullstack?/server", "dioxus-static-site-generation?/server", "ssr", "dioxus-liveview?/axum", "dep:axum"]
+server = ["axum"]
```
jkelleyrtp, line 61: does it make sense to still have `axum` as a feature name here instead of just renaming it? Or we're trying to retain backwards compat?

jkelleyrtp (approved):
yay!

packages/dioxus/Cargo.toml
ealmloff: It is just there for backwards compatibility. I'm not sure if we can mark a feature as deprecated? Maybe through a build script?

Issue #2783, opened 2024-08-03
Suspense diffing fails when switching components on the client


DonAlonzo:
**Problem**

Hydration errors when navigating _from_ a route to a route that has a use_server_future. Does **not** appear when refreshing the page, or going directly to the URL.

**Steps To Reproduce**

1. Run `dx serve`
1. Go to http://localhost:8080
1. Click the links until hydration error appears

`main.rs`

```
use dioxus::prelude::*;

#[cfg(feature = "web")]
fn main() {
    dioxus::web::launch::launch_cfg(App, dioxus::web::Config::new().hydrate(true));
}

#[cfg(feature = "server")]
#[tokio::main]
async fn main() {
    let app = axum::Router::new()
        .serve_dioxus_application(ServeConfig::builder().build(), App)
        .into_make_service();

    let listener = tokio::net::TcpListener::bind("0.0.0.0:8080")
        .await
        .unwrap();

    axum::serve(listener, app).await.unwrap();
}

#[derive(Routable, Clone)]
pub enum Route {
    #[route("/")]
    Home,
    #[route("/page")]
    Page,
}

#[component]
pub fn App() -> Element {
    rsx! {
        ErrorBoundary {
            handle_error: |_| rsx! {
                "Hmm, something went wrong."
            },
            SuspenseBoundary {
                fallback: move |_| rsx! {
                    "Loading..."
                },
                Router::<Route> {}
            }
        }
    }
}

#[component]
pub fn Home() -> Element {
    let _calculation = use_server_future(calculate)?.suspend();
    rsx! {
        Link {
            to: Route::Page,
            "Go to page"
        }
    }
}

#[component]
pub fn Page() -> Element {
    let _calculation = use_server_future(calculate)?.suspend();
    rsx! {
        Link {
            to: Route::Home,
            "Go to home"
        }
    }
}

#[server]
pub async fn calculate() -> Result<u32, ServerFnError> {
    Ok(1 + 1)
}
```

`Cargo.toml`

```
[package]
name = "hydration-navigation-error"
edition = "2021"
version = "0.1.0"

[dependencies]
axum = { version = "0.7.5", optional = true }
dioxus = { version = "0.6.0-alpha.1", features = ["fullstack", "router"] }
tokio = { version = "1.38.0", features = ["net", "sync"], optional = true }

[features]
default = []
server = [
  "axum",
  "dioxus/axum",
  "tokio",
]
web = [
  "dioxus/web",
]
```

[...]

PR #2789, opened 2024-08-05
Fix suspense diffing

ealmloff:
Fixes https://github.com/DioxusLabs/dioxus/issues/2783


jkelleyrtp (approved):
Guess it was a race condition with pulling nodes out and diffing them? Good find.

PR #2792, opened 2024-08-06
Synchronous prevent default

### LINKED ISSUE #85: Need for synchronous PreventDefault across Tao/Wry boundary

### LINKED ISSUE #1662: `prevent_default` is not respected after bubbling events
**Problem**

In Liveview, `prevent_default` on the Link component does not work:

https://github.com/DioxusLabs/dioxus/blob/d9220d4e42b86114096a639ba0b308062b6d763a/packages/router/src/components/link.rs#L248

As a result, the link treats every route, internal or external, as an external link and the page reloads upon redirect.

**Steps To Reproduce**

Set up a router and Links in Liveview.

Example:
```
render! {
        div {
            for movie in movies {
                Link {
                    key: "{movie.id}",
                    to: match movie.screening {
                        Some(screening) => Route::Screening { id: screening },
                        None => Route::Movie { id: movie.id },
                    },
                    "{movie.name}"
                }
            }
        }
    }
```

**Expected behavior**

A Link to an internal route should be treated as an internal route. The expected behavior can be simulated by commenting out this line:

https://github.com/DioxusLabs/dioxus/blob/d9220d4e42b86114096a639ba0b308062b6d763a/packages/router/src/components/link.rs#L247)

**Screenshots**

N/A

**Environment:**
 - Dioxus version: `master`
 - Rust version: 1.76.0-nightly
 - OS info: Arch Linux
 - App platform: Liveview

**Questionnaire**

### LINKED ISSUE #2383: Ctrl/cmd click does not open in a new tab/window
**Problem**

`Link` elements are `a` tags in rendered html, but have a `dioxus-prevent-default="onclick"` which I assume is not looking at the `ctrl`/`cmd` state when clicked to replicate the normal browser behavior. https://support.mozilla.org/en-US/questions/1336268 is a discussion of this problem in general (dioxus is not the first and only framework with this issue, but it is fixable.)

If you are not familiar with this workflow: ctrl clicking a "normal" link on the vast majority of websites will open the link in a new browser tab/window, but not so with `Link` elements in a current dioxus site. Consistency with typical browser behavior/usage seems desirable. (Thus this issue is being tagged as a bug rather than feature request.)

It's easy enough to find where normal link behavior is blocked: https://github.com/DioxusLabs/dioxus/blob/472031d1f51557601797d1cb469f09e4f3e0ca71/packages/web/src/dom.rs#L84

And I believe this is where link clicking is replaced: https://github.com/DioxusLabs/dioxus/blob/472031d1f51557601797d1cb469f09e4f3e0ca71/packages/router/src/components/link.rs#L265

But I'm not sure how quite to work with the router and/or browser to handle this behavior properly on the web -- especially given dioxus's cross platform nature of desktop vs web, etc where ctrl click opening a new tab perhaps does not make sense except only in a browser. It seems like some kind of override is needed for web to handle this consistently with that the typical behavior unique for that platform (browsers). Given pointers on where the appropriate (i.e. acceptable to the team) place to make such an override would be, I'll happily attempt to make the changes myself.

**Steps To Reproduce**

- `dx new` with router
- `dx serve` that new app
- open up http://localhost:8080/
- ctrl click the "Go to blog" link at the top

**Expected behavior**

Ctrl clicking the "Go to blog" link should open in a new tab/window (depending on your browser config).

For comparison, go to https://www.google.com/ and ctrl click the "About" link in the top left, it will open in a new tab/window.

**Environment:**
 - App platform: `web`

**Questionnaire**

ealmloff:
Fixes #85 
Fixes #2383 
Fixes #1662
Fixes https://github.com/DioxusLabs/docsite/issues/11


packages/desktop/src/webview.rs
```diff
@@ -1,17 +1,130 @@
...
+
+    pub fn try_handle_event(
+        &self,
+        request: wry::http::Request<Vec<u8>>,
+    ) -> Result<Vec<u8>, serde_json::Error> {
+        let response = match serde_json::from_slice(request.body()) {
+            Ok(event) => self.handle_html_event(event),
+            Err(err) => {
+                tracing::error!("Error parsing user_event: {:?}", err);
+                println!("as string: {:?}", request);
```
jkelleyrtp, line 58: println

jkelleyrtp (changes requested):
Thoughts re: keep handling of dioxus-prevent-default especially with regards to Link {} component since it no longer handles that.

PR #2793, opened 2024-08-06
chore(cli): improved package version & commit hash usage

Andrew15-5:
The valuable part from the dismissed #2710. `CARGO_PKG_VERSION` is a [standard environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates).

### DIFF
--- a/packages/cli-config/src/lib.rs
+++ b/packages/cli-config/src/lib.rs
@@ -74,11 +74,10 @@ pub static CURRENT_CONFIG: once_cell::sync::Lazy<
     match serde_json::from_str(config) {
         Ok(config) => Ok(config),
         Err(err) => {
-            let mut cli_version = crate::build_info::PKG_VERSION.to_string();
+            let mut cli_version = env!("CARGO_PKG_VERSION").to_string();
 
             if let Some(hash) = crate::build_info::GIT_COMMIT_HASH_SHORT {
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Hey, thanks for the PR!

This PR isn't addressing any open issues, fix any bugs, or meaningfully clean up any parts of the codebase, so I'm going to close it.


We already use the GIT_COMMIT_HASH_SHORT field from dx_build_info so I think it makes sense to keep the PKG_VERSION that built-rs gives us. 
```
        crate::dx_build_info::PKG_VERSION,
        crate::dx_build_info::GIT_COMMIT_HASH_SHORT.unwrap_or("was built without git repository")
```

Andrew15-5:
[You said yourself](https://github.com/DioxusLabs/dioxus/pull/2710#pullrequestreview-2211021545) that there are still some useful changes:
> There are other changes in this PR that are nice if you want to still include those, otherwise I don't think it's worth switching over.

These are the only other changes besides the vergen-gix swap. So I don't understand how you switch between "there are some nice changes" and "it doesn't address anything or meaningfully cleans up the codebase". I specifically created it because of your comment. We agreed that `hash.trim_start_matches('g')` is something weird that shouldn't be used, so I fixed it.

I can exclude the `env!("CARGO_PKG_VERSION")` change.

PR #2794, opened 2024-08-06
Fix autocomplete in rsx

ealmloff:
I'm not exactly sure why, but putting the release version of rsx first breaks autocomplete. Maybe the span being used in an inactive part of rsx first links autocomplete to that part the rsx?


jkelleyrtp (approved):
sigh.....

PR #2799, opened 2024-08-07
Fix nested rsx expansion by not using template titles

jkelleyrtp:
Templates are now hashed by their content's pointers rather than an arbitrary name (which is wrong due to nested expansion), fixing issues with template merging in core that breaks lookbook integration.


packages/core/src/diff/node.rs
```diff
@@ -402,9 +402,7 @@ impl VNode {
     }
 
     fn templates_are_different(&self, other: &VNode) -> bool {
-        let self_node_name = self.template.id();
-        let other_node_name = other.template.id();
-        self_node_name != other_node_name
+        self.template == other.template
```
ealmloff, line 405: This function got inverted. It now checks if they are the same. I'm not sure where it is used

packages/core/src/nodes.rs
```diff
@@ -1103,6 +1101,13 @@ impl IntoAttributeValue for manganis::ImageAsset {
     }
 }
 
+// #[cfg(feature = "manganis")]
```
ealmloff, line 1104: Is this intentional?

packages/core/src/virtual_dom.rs
```diff
@@ -612,24 +602,29 @@ impl VirtualDom {
     /// ```
     #[instrument(skip(self, to), level = "trace", name = "VirtualDom::rebuild")]
     pub fn rebuild(&mut self, to: &mut impl WriteMutations) {
-        self.flush_templates(to);
+        self.maybe_rebuild(Some(to));
+    }
+
+    /// Rebuild the virtual dom without handling any of the mutations
+    pub fn maybe_rebuild<M: WriteMutations>(&mut self, to: Option<&mut M>) {
```
ealmloff, line 609: It might make sense to make the virtual dom itself generic over mutations at some point? Then you could remove the generic on rebuild_in_place

ealmloff (changes requested):
Much nicer overall! Hydration is a bit cleaner and nested rsx just works by default. I noticed a few changes that don't look intentional, but other than that it looks good

ealmloff:
This shaves off `~3kb` for a minimal web example

ealmloff (approved):

Issue #2817, opened 2024-08-11
Assets served to Webview2 by dioxus-desktop can be blocked for violating CORS policy


daixiwen:
**Problem**

When dioxus-desktop serves assets to Webview2, it doesn't set the Access-Control-Allow-Origin header, and some assets can be blocked for violating the CORS policy.
I'm not sure if this is the same thing that was mentioned in the megathread #2788 "We don't properly set cors/CSP/XSRF headers on assets to clients in dev" ?

**Steps To Reproduce**

Create an application with dioxus-desktop on windows.
Make the app use a stylesheet served as an asset.
Make that stylesheet import another asset (either another css file, or a font for example)

The first stylesheet will be loaded, but the other asset will be blocked for violating CORS policy.

This is only a problem with Webview2, the webkit2gtk on Linux works perfectly in that scenario.

**Expected behavior**

Both assets should be loaded by Webview2

**Environment:**
 - Dioxus version:  v0.5 & master
 - Rust version:    1.79.0
 - OS info:         Windows
 - App platform:    desktop

**Questionnaire**
- [ ] I'm interested in fixing this myself but don't know where to start
- [x] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

I fixed it by adding the Access-Control-Allow-Origin header to all assets returned by dioxus, the same way it is done for the index request

daixiwen:
The fix has been merged in main. Thanks!

PR #2818, opened 2024-08-11
adding the Access-Control-Allow-Origin header to assets

daixiwen:
This fixes the issue #2817

I've just added the Access-Control-Allow-Origin header to all assets returned by dioxus, the same way it is done for the index request. 

This removes the problem with Webview2 blocking some local assets due to CORS policy violation.


jkelleyrtp (approved):

jkelleyrtp:
Thanks!

Issue #2788, opened 2024-08-05
Megathread: cli 0.6.0-alpha quirks/issues/ux/perf


jkelleyrtp:
**Problem**

We have a number of quirks / bugs in the CLI post-rewrite that need to be resolved before releasing to the general public.

This issue is where we put them.

- [x] Fullstack server takes a moment to start leading to the default-router's "connection refused" screen
- [x] The tui sidebar makes it impossible to copy logs - needs to be moved to a row #2846
- [x] The tui scrolling is just janky/slow/hacky *Improved* in #2846
- [x] The tui "modal" is not complete at all and needs more helpful info (summary of config, urls for dev server, process names, etc)
- [x] Missing logs for downloading/compiling dependency crates
- [x] tracing annotations make it harder for vscode to implement jump-to-file
- [x] Fullstack is just broken due to a port forwarding issue
- [x] Asset collection causes unnecessary recompiles due to thrashing of env vars in build scripts
- [x] CWD for bundle (used by icon search) is just wrong leading to #2344 
- [x] Collecting and optimizing assets should be optional/skipped in dev mode in favor of just copying assets #2844
- [x] wasm-bindgen crashing nukes our TUI
- [x] We don't invalidate literals in hotreloading that are the same as the original, meaning you can hotreload true -> false but not back to true
- [x] We don't send down old hotreloaded templates to new clients anymore
- [ ] We don't properly set cors/CSP/XSRF headers on assets to clients in dev
- [ ] we should print the resolved config on a panic in the cli for better debugging

Andrew15-5:
> - [ ] Fullstack is just broken due to a port forwarding issue

Is this the same as:

- [ ] "Failed to find index.html" Fullstack panic

```
[app] thread 'main' panicked at /cargo/git/checkouts/dioxus-1e619ce595d3799d/bd58a92/packages/fullstack/src/serve_config.rs:91:37:
[app] 2024-08-04T12:04:46.395679Z  WARN dioxus_cli_config: A library is trying to access the crate's configuration, but the dioxus CLI was not used to build the application.
[app] Failed to find index.html. Make sure the index_path is set correctly and the WASM application has been built.: Os { code: 2, kind: NotFound, message: "No such file or directory" }
[app] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

?

- [ ] Some messages sometimes are written on top of the UI

[...]

PR #2822, opened 2024-08-12
Fix (and rewrite) `dx init`/`dx new`

Andrew15-5:
Fixes issue in https://github.com/DioxusLabs/dioxus/issues/2788#issuecomment-2270538452. I have found few semantic breakages and bugs, but upon trying to fix stuff and dig deeper to get the whole image of what the current state of `init`/`new` subcommands is trying to do, I stumbled upon even more bugs and performance issues (calling `MetadataCommand` from incorrect CWD causes a 2 s slowdown!). Here is my [monologue](https://discord.com/channels/899851952891002890/936277970644000778/1272266828894437376) with a little bit more context.

Here I am cooking a new version of those subcommands. The workspace change will be mostly untouched as I'm not familiar with it, and it doesn't overlap with the problem at hand.

The first commit contains a ton of commented code, which most likely breaks the `new` command (at least in some way). But I fully rewrote (refactored + added a few things) the `init` command. Had to get an absolute path without resolving symbolic links, so I used a `path-absolutize` crate (for free). Second commit adds one dev dependency for running and testing commands (including `dx`), it's pretty small. I rarely wrote tests before, but these once look very nice. I think I balanced abstraction and clarity pretty well.

One thought I now have (since my testing strategy has changed from the beginning) is to put `init` subcommmand tests to its own file instead of the main one.

Notes (to not forget):
- Currently no way to use non-default branch (tag/revision) when using a cargo-generate template. Need to add/copy `branch`, `tag`, `revision` options from cargo-generate. Should this be a separate PR or it can be included here? It by itself probably will be very small.


Andrew15-5:
I also added the `branch` option to be able to use (unmerged) branches from a remote template (git) repository (https://github.com/DioxusLabs/dioxus/issues/2788#issuecomment-2270538452). As a complete set, also added `revision` and `tag` options. This will open a possibility for a wider range of reproducible projects/bugs.

Andrew15-5:
Current implementation of `dx init` and `dx new` follows the behavior of `cargo init` and `cargo new` respectively. Specifically, how `path` and `name` work. This is technically a breaking change, but for most use cases almost nothing has changed. IMO, the most noticeable change is that `path` is required for `dx new`. I laid out the problem [here](https://discord.com/channels/899851952891002890/936277970644000778/1274110035928027207).

Andrew15-5:
Another notes:
- `vcs` not only practically was set to None in both branches of the condition, but it also was loosely dependent on `metadata` which feels like not a good thing to do. Moreover, cargo-generate won't create a VCS if already inside VCS. And lastly, if anything, we can have a `--vcs` option to use it directly as intended.
- the mutual exclusivity of some options is enforced by clap, so I'm not sure if tests for this should be written, but my gut feeling says that they should be, just in case.
- [here](https://github.com/DioxusLabs/dioxus/pull/2822/files#diff-03993b5e442e077ccb1304ee14eb355c979828aace31c370b2dddc7cbe6ec209R113-R127) I'm pretty confident that I set a correct current directory for metadata command, but not on 100%.
- I could've compressed the tests more by substituting repetitive code with a macro, so that it will show panic in the correct place (2 places), but it seems a bit less readable, including more verbose logs.
- maybe use a local (in-repo) micro-template for the tests? This will remove the necessity for Internet connection and that you must pass the `--subtemplate` option until https://github.com/cargo-generate/cargo-generate/issues/1181 is fixed. Reducing things that can go wrong is always better.

Andrew15-5:
I found that `Error::CargoMetadata` variant is gone, so I recreated it in the place where it used before it was removed.

I was surprised that the tests failed after I initially thought that I'm done. But apparently I was not the first one: https://github.com/assert-rs/assert_cmd/issues/131. So because of this [limitation](https://docs.rs/assert_cmd/2.0.16/assert_cmd/cargo/#limitations) I used `escargot` instead of `assert_cmd`. I did use the new (fancy) `LazyLock` for static one-time cargo build, which is expensive to call.

The tests still pass without any changes, so that's cool. :)

Do we need some sort of "workspace" `dx new`/`dx init` tests?

Also, I updated the `Cargo.lock` file, which is now has some crates with newer versions. Is it fine, or should I revert them?

Andrew15-5:
Oh no... It's still not in the stable version (for some platforms?).
https://github.com/DioxusLabs/dioxus/actions/runs/11802634307/job/32878728014?pr=2822

Andrew15-5:
I also think that renaming `Create` to `New` would be a bit more intuitive, since it's responsible for the `dx new` subcommand.

jkelleyrtp:
Thanks! Will take a look tomorrow - might be able to get it into the release.

jkelleyrtp (approved):
Actually, all looks good, thank you!

Andrew15-5:
Oh, wow, OK. Great, but what about questions from https://github.com/DioxusLabs/dioxus/pull/2822#issuecomment-2471110440 and https://github.com/DioxusLabs/dioxus/pull/2822#issuecomment-2471282670?

jkelleyrtp:
> Do we need some sort of "workspace" dx new/dx init tests?

Eventually, sure, plus a whole slew of other tests for the CLI. But for now it's not a big deal - the code here isn't too crazy to make it worth not shipping it in 0.6.

> I also think that renaming Create to New 

We could but it doesn't seem terribly important to do that either.

Andrew15-5:
Alright, and the last question from the https://github.com/DioxusLabs/dioxus/pull/2822#issuecomment-2471110440. I guess slightly more up-to-date random crates isn't an issue either? I thought that this could add some potential problems, but maybe it's fine since all the current tests are passing.

PR #2825, opened 2024-08-13
Update README.md

chikoYEAT:
extra '


jkelleyrtp (commented):
The apostrophe is because that usage of Dioxus is possessive.

PR #2832, opened 2024-08-13
Always Add Children Builder On Props

### LINKED ISSUE #2281: Tweak: always generate a `children` method on component prop builders
## Feature Request

Right now we can't hotreload component bodies going from something to nothing or nothing to something since that's functionally adding/removing a .children() method.

In debug or dev mode we should always pass in children as either some or none into the builder, even if the component doesn't do anything with those children.

This would be breaking (might be able to find a way around it).

DogeDark:
Always adds a children builder method on a component's props.

Closes #2281


jkelleyrtp:
closing this but will re-open it at a later date!

Issue #2815, opened 2024-08-10
bug: Panic on diff/node.rs


marc2332:
**Problem**

```sh
thread 'main' panicked at C:\Users\mespi\.cargo\git\checkouts\dioxus-1e619ce595d3799d\0bb81ab\packages\core\src\diff\node.rs:651:9:
Node mount should be valid
```

**Steps To Reproduce**

```rs

fn app() -> Element {
    rsx!(
        Comp {
            for i in 0..3 { // <---- This makes it crash.
                p { 
                    key: "{i}",
                    "aaa {i}"
                }
            }
        }
		// Comp {
        //     p { "Hello, World!" } // <-- But this works
        // }
    )
}

#[component]
fn Comp(children: Element) -> Element {
    let mut show = use_signal(|| false);

    rsx!(
        div {
            onclick: move |_| show.toggle(),
            height: "50%",
            width: "100%",
            if show() {
                {children}
            }
        }
    )
}
```

**Expected behavior**

Not crash

**Environment:**
 - Dioxus version:  `main`
 - Rust version:    1.80
 - OS info:         Windows
 - App platform:    `freya`

PR #2834, opened 2024-08-13
Fix mount use after removal/Simplify mounts a bit

ealmloff:
Fixes https://github.com/DioxusLabs/dioxus/issues/2815. We were trying to use a mount that had already been removed. This PR fixes that issue by removing the mount from the vnode as we remove the mount from the virtual dom.

It also removes the distinction between `VNode::clone` vs `VNode::clone_mounted` which was unused


packages/core/src/diff/iterator.rs
```diff
@@ -145,15 +145,15 @@ impl VirtualDom {
             // we need to find the right "foothold" though - we shouldn't use the "append" at all
             if left_offset == 0 {
                 // insert at the beginning of the old list
-                let foothold = &old[old.len() - right_offset];
+                let foothold = &new[old.len() - right_offset];
```
ealmloff, line 148: I'm not entirely sure about this logic. Should we be removing old here? Has it already been removed?

jkelleyrtp (approved):

PR #2835, opened 2024-08-13
Add version number to package section of Cargo.toml to fix issue #2833

### LINKED ISSUE #2833: Fullstack examples fail because of missing "version" in "package" section of main Cargo.toml
**Problem**

Running `dx serve --platform fullstack` in, for instance, `packages/fullstack/examples/hello-world` gives the following:

```
% dx serve --platform fullstack
2024-08-13T21:52:10.618439Z  INFO dx: You appear to be creating a Dioxus project from scratch; we will use the default config
- ⚙ Compiling registry+https://github.com/rust-lang/crates.io-index#futures-util@0.3.30
  👑 Build done.
thread '<unnamed>' panicked at /home/jack/.cargo/registry/src/index.crates.io-6f17d22bba15001f/manganis-cli-support-0.2.4/src/manifest.rs:53:70:
called `Result::unwrap()` on an `Err` value: Workspace(Parse(Error { inner: Error { inner: TomlError { message: "missing field `version`", raw: [raw omitted for brevity], keys: ["package"], span: Some(6520..6908) } } }))
```

**Steps To Reproduce**

- `cd packages/fullstack/examples/hello-world`
- `dx serve --platform fullstack

**Expected behavior**

No errors, at least

**Screenshots**

**Environment:**
 - Dioxus version: 0.6.0-alpha.2
 - Rust version: 1.80.0
 - OS info:  Ubuntu 24.04
 - App platform:  fullstack

**Questionnaire**

jacklund:
Fix for Issue #2833. 

This adds a version number to the `[package]` section of the `Cargo.toml`, which resolves the issue.

Admittedly, not the best solution, given that the version number there and in the `[workspace]` section will have to be kept in sync. I looked around for a better one and couldn't find it, but I'm very open to suggestions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2598,7 +2598,7 @@ dependencies = [
 
 [[package]]
 name = "dioxus-examples"
-version = "0.0.0"
+version = "0.6.0-alpha.2"
 dependencies = [
  "base64 0.21.7",
  "ciborium",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,6 +183,7 @@ documentation = "https://dioxuslabs.com"
 keywords = ["dom", "ui", "gui", "react", "wasm"]
 rust-version = "1.79.0"
 publish = false
+version = "0.6.0-alpha.2"
 
 [dependencies]
 manganis = { workspace = true, optional = true }


jkelleyrtp:
Thanks! This works but we should really just be fixing manganis to not panic on missing version numbers :-)

jkelleyrtp (approved):

PR #2836, opened 2024-08-13
feat(cli): added Ctrl+Z (SIGTSTP) support

Andrew15-5:
Fixes one entry from https://github.com/DioxusLabs/dioxus/issues/2788#issuecomment-2270538452 by adding support for Ctrl+Z. [I did face a few problems](https://discord.com/channels/899851952891002890/936277970644000778/1272989881433460836). But overall it works.

It adds `(signal)` to the zsh's suspend job log message, which isn't there when SIGTSTP is used in other programs. Another thing is that when it is stopped in the middle of some building process (at least during Wasm shenanigans) it can ruin the build (with a panic message) and the `dx serve` must be restarted. This should be explored more because suspending shouldn't break anything _ideally_.

The quirk that is interesting is where the execution is continued. This is nice to know as I used a condition check in the main loop before and didn't understand why no logs from it showed up. The less code, the cleaner it gets!

It looks like there is only 1 new package according to `Cargo.lock`. Nice.

I added a few notes in different places. Like using private field from `screen` variable isn't great.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2458,6 +2458,8 @@ dependencies = [
  "rustls 0.23.12",
  "serde",
  "serde_json",
+ "signal-hook",
+ "signal-hook-tokio",
  "subprocess",
  "syn 2.0.74",
  "tar",
@@ -5822,7 +5824,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Andrew15-5:
I finally remembered to test it from just recipe. Yeah, it doesn't work like that. Or rather, I have to press Ctrl+Z twice. I think, once to suspend `dx` and once to suspend `just`. So I'm not sure if it's OK to use this implementation of the feature. It's definitely better than nothing, but it isn't perfect (it's very close though).

Andrew15-5:
@jkelleyrtp, ready for initial review, thoughts, verdict. Still has some debugging stuff.

jkelleyrtp:
Hey - I'm cleaning up the PR list and this got clobbered by a bunch of changes we made regarding the CLI.

Feel free to re-open this if you want. Definitely give the new CLI a test drive first though, since this might be fixed with the new render code.

Andrew15-5:
> Feel free to re-open this if you want. Definitely give the new CLI a test drive first though, since this might be fixed with the new render code.

Sure! But did you know that a regular user can't re-open if someone with write permissions/member/owner has closed the PR (or issue)? So this means that _you_ need to re-open it for me _if_ I want it.

Andrew15-5:
I tried the new CLI, and it still doesn't support terminal stop signal. :(

PR #2838, opened 2024-08-14
Router: Non-Closure Redirects

### LINKED ISSUE #2591: Allow non-closure expression in redirect
## Feature Request

Currently `#[redirect]` requires closure. This caused problem when running clippy - rust-lang/rust-clippy#13057 as it tries to change it to simple expression.

DogeDark:
Allows support for a few non-closure redirect types:
```rs
enum Route {
    ...

    // Directly using route variants with static fields.
    #[redirect("/hi", Route::Hi {})] 
    #[redirect("/thecoolestperson", Route::User { name: "Doge" })]
    // Paths to functions
    #[redirect("/hi2", Route::default)]
    #[redirect("/coolroute", Route::the_coolest_route)]
    
    // This is unchanged
    #[redirect("/hi3", || Route::Hi {})]
    #[redirect("/somealias/:id", |id: u32|, Route::Blog { id })] 
    ...

}
```

Closes #2591

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,7 +270,7 @@ checksum = "0ae92a5119aa49cdbcf6b9f893fe4e1d98b04ccbf82ee0584ad948a44a734dea"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -292,7 +292,7 @@ dependencies = [
  "argh_shared",
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -351,7 +351,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "swc_macros_common",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -520,7 +520,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -586,7 +586,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -603,7 +603,7 @@ checksum = "6e0c28dcc82d7c8ead5cb13beb15405b57b8546e93215673ff8ca0349a028107"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
@@ -674,7 +674,7 @@ checksum = "3c87f3f15e7794432337fc718554eaa4dc8f04c9677a950ffe366f20a162ae42"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.74",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ealmloff:
I'm worried that allowing only redirects with no dynamic paths to be extracted into a separate function will be confusing. If this works:
```rust
#[redirect("/:id", |id: usize| Route::Home)]
```
And this works:
```rust
#[redirect("/static", static_redirect)]
```
Then it seems like this should work:
```rust
#[redirect("/:id", id_redirect)]
fn id_redirect(id: usize) -> Route {
    Route::Home
}
```

It currently doesn't because we rely on the names matching between the dynamic segments and redirect function. Functions don't have named arguments so we can't allow extracting the function directly. However, if we just call the function with arguments parsed in order from the route, we can extract any number of dynamic segments. This would be a breaking change, but I feel like most redirect closures will end up accepting arguments in the order the dynamic segments are defined anyway, so it seems fine.


We could use a trait to extra the argument types from the unnameable closure types:

```rust
use std::str::FromStr;

fn main() {
    // #[redirect("/:route", |_: i32| Route::Home)] could expand to something like this: 
    {
        let route = "/1234";
        let mut segments = route.split('/');
        let redirect = |_: i32| Route::Home;
        if let Ok(parsed) = parse_nth_argument::<1, _, _>(redirect, segments.next().unwrap()) {
            move_to_route(redirect(parsed));
        }
    }

    // #[redirect("/:id/:name", my_redirect_function)] could expand to something like this:
    {
        fn my_redirect_function(_: i32, _: String) -> Route {
            Route::Home
        }
        let route = "/1234/Bob";
        let mut segments = route.split('/');
        let redirect = my_redirect_function;
        if let Ok(__arg_1) = parse_nth_argument::<1, _, _>(redirect, segments.next().unwrap()) {
            if let Ok(__arg_2) = parse_nth_argument::<2, _, _>(redirect, segments.next().unwrap()) {
                move_to_route(redirect(__arg_1, __arg_2));
            }
        }
    }
}

trait FunctionArguments<const N: usize, Marker = ()> {
    type Argument;
}

macro_rules! impl_function_arguments {
    ($n:literal, $marker:ident, [$($number:literal),+], [$($variable:ident),+]) => {
        #[doc = concat!("A marker type for functions with ", $n, " arguments")]
        pub struct $marker;

        impl_function_arguments!(inner, $n, $marker, [$($number),+], [$($variable),+], [$($variable),+]);
    };
    (inner, $n:literal, $marker:ident, [], [], [$($variable_again:ident),*]) => {};
    (inner, $n:literal, $marker:ident, [$first_number:literal $(, $number:literal)*], [$first_variable:ident $(, $variable:ident)*], [$($variable_again:ident),*]) => {
        impl<Function: Fn($($variable_again,)*) -> R, R, $($variable_again,)*> FunctionArguments<$first_number, ($marker, $($variable_again,)*)> for Function {
            type Argument = $first_variable;
        }

        impl_function_arguments!(inner, $n, $marker, [$($number),*], [$($variable),*], [$($variable_again),*]);
    };

}

impl_function_arguments!(1, OneArgument, [1], [A]);
impl_function_arguments!(2, TwoArguments, [1, 2], [A, B]);
impl_function_arguments!(3, ThreeArguments, [1, 2, 3], [A, B, C]);
impl_function_arguments!(4, FourArguments, [1, 2, 3, 4], [A, B, C, D]);
impl_function_arguments!(5, FiveArguments, [1, 2, 3, 4, 5], [A, B, C, D, E]);
impl_function_arguments!(6, SixArguments, [1, 2, 3, 4, 5, 6], [A, B, C, D, E, F]);
impl_function_arguments!(
    7,
    SevenArguments,
    [1, 2, 3, 4, 5, 6, 7],
    [A, B, C, D, E, F, G]
);

// Parse an arbitrary argument of a function. Checked at compile time.
fn parse_nth_argument<const N: usize, M, F>(_: F, text: &str) -> Result<F::Argument, <F::Argument as FromStr>::Err>
where
    F: FunctionArguments<N, M>,
    F::Argument: FromStr,
{
    text.parse()
}

enum Route {
    Home,
    About
}

fn move_to_route(route: Route) {}
```

jkelleyrtp:
I think the actual bug in the original issue is that we're not passing attributes to items in our macros, so the clippy lint exclusion isn't passing. Since this isn't as simple of a change as we were hoping, I'm just going to close this.

PR #2839, opened 2024-08-14
Fix cloned rsx double drop

### LINKED ISSUE #2331: Diffing cloned Components with readonlysignal panics on double drop

ealmloff:
Fixes #2331


packages/core-macro/tests/values_memoize_in_place.rs
```diff
@@ -161,3 +161,52 @@ fn spreads_memorize_in_place() {
     assert!(props.memoize(&CompProps::builder().build()));
     assert_eq!(props.attributes, vec![]);
 }
+
+// Regression test for https://github.com/DioxusLabs/dioxus/issues/2331
+#[test]
+fn cloning_read_only_signal_components_work() {
```
matthunz, line 167: Awesome! This test looks like it covers an issue I was having awhile back 👍

matthunz (approved):

jkelleyrtp (approved):

PR #2846, opened 2024-08-15
CLI Fixes & Move CLI Info Bars

DogeDark:
- Increases scroll speed to two lines per scroll.
- Adds error handling for manganis failure that still lets the build run.

Moves TUI info bars to the bottom of the screen:

image

### DIFF
--- a/packages/cli/src/builder/cargo.rs
+++ b/packages/cli/src/builder/cargo.rs
@@ -19,6 +19,7 @@ use manganis_cli_support::AssetManifest;
 use manganis_cli_support::ManganisSupportGuard;
 use std::fs::create_dir_all;
 use std::path::PathBuf;
+use tracing::error;
 
 impl BuildRequest {
     /// Create a list of arguments for cargo builds
@@ -222,7 +223,10 @@ impl BuildRequest {
                 cargo_args,
                 Some(linker_args),
             )?;
-            let assets = asset_manifest(&build);
+            let Some(assets) = asset_manifest(&build) else {
+                error!("the asset manifest was not provided by manganis and we were not able to collect assets");
+                return Err(anyhow::anyhow!("asset manifest was not provided by manganis"));
+            };
             // Collect assets from the asset manifest the linker intercept created
             process_assets(&build, &assets, &mut progress)?;
             // Create the __assets_head.html file for bundling
@@ -230,8 +234,7 @@ impl BuildRequest {
 
             Ok(Some(assets))
         })
-        .await
-        .unwrap()
+        .await?
     }
 
     pub fn copy_assets_dir(&self) -> anyhow::Result<()> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2095,7 +2095,7 @@ dependencies = [
  "cssparser-macros",
  "dtoa-short",
  "itoa 1.0.11",
- "phf 0.11.2",
+ "phf 0.8.0",
  "smallvec",
 ]
 
@@ -5049,7 +5049,7 @@ dependencies = [
  "httpdate",
  "itoa 1.0.11",
  "pin-project-lite",
- "socket2 0.5.7",
+ "socket2 0.4.10",
  "tokio",
  "tower-service",
  "tracing",
--- a/packages/cli/src/assets.rs
+++ b/packages/cli/src/assets.rs
@@ -19,13 +19,13 @@ use walkdir::WalkDir;
 /// The temp file name for passing manganis json from linker to current exec.
 pub const MG_JSON_OUT: &str = "mg-out";
 
-pub fn asset_manifest(build: &BuildRequest) -> AssetManifest {
+pub fn asset_manifest(build: &BuildRequest) -> Option<AssetManifest> {
     let file_path = build.target_out_dir().join(MG_JSON_OUT);
-    let read = fs::read_to_string(&file_path).unwrap();
+    let read = fs::read_to_string(&file_path).ok()?;
     _ = fs::remove_file(file_path);
     let json: Vec<String> = serde_json::from_str(&read).unwrap();
 
-    AssetManifest::load(json)
+    Some(AssetManifest::load(json))
 }
 
 /// Create a head file that contains all of the imports for assets that the user project uses
--- a/packages/cli/src/serve/output.rs
+++ b/packages/cli/src/serve/output.rs
@@ -1,3 +1,4 @@
+use super::{Builder, Server, Watcher};
 use crate::{
     builder::{
         BuildMessage, MessageSource, MessageType, Stage, TargetPlatform, UpdateBuildProgress,
@@ -36,7 +37,8 @@ use tokio::{
 };
 use tracing::Level;
 
-use super::{Builder, Server, Watcher};
+// How many lines should be scroll on each mouse scroll or arrow key input.
+const SCROLL_SPEED: u16 = 2;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum LogSource {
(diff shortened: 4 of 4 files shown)


FragrantArmpit:
I think tabs are a bad idea because the user would have to switch back and forth between their code terminal and their TUI terminal whereas if there were no tabs the user could just move the TUI terminal to a second monitor or whatever and never touch it (although as of right now I have to reload it manually to see any non-hot-reloadable changes, hopefully this will change in the future).

I suggest we follow what `docker-compose` does: there is just one tab and each "stream" of logs has a different color (in case of `docker-compose` each "stream" is the output of a particular container and in our case it could be the logs from the CLI itself, the logs from building target X (for fullstack projects we build for `web` and `server` targets simultaneously), etc. This can probably be implemented by having a `tracing` span with a field to differentiate "streams" and set appropriate color.

FragrantArmpit:
i.e. 

image

packages/cli/src/builder/cargo.rs
```diff
@@ -99,9 +101,12 @@ impl BuildRequest {
 
     pub(crate) async fn build(
         &self,
-        mut progress: UnboundedSender<UpdateBuildProgress>,
+        mut progress: UnboundedSender<BuildProgressUpdate>,
     ) -> Result<BuildResult> {
-        tracing::info!("🚅 Running build [Desktop] command...");
+        tracing::info!(
+            dx_src = ?MessageSource::Build,
+            "🚅 Running build [Desktop] command..."
```
FragrantArmpit, line 108: `[Desktop]` in `🚅 Running build [Desktop] command...` is currently hard-coded.
```suggestion
            "🚅 Running build [{}] command...", self.target_platform)
```

DogeDark:
@FragrantArmpit 
> I think tabs are a bad idea...

I've heard similar from a couple of other people. I don't plan to change this behavior in this PR but I do think it's something that should be discussed. If you could create a new issue about it, that would be great!

packages/cli/src/serve/output/render.rs
```diff
@@ -0,0 +1,277 @@
...
+                            Level::TRACE => level_span.black(),
+                            Level::DEBUG => level_span.light_magenta(),
+                            Level::INFO => level_span.light_blue(),
+                            Level::WARN => level_span.light_yellow(),
+                            Level::ERROR => level_span.light_red(),
+                        };
+
+                        let mut out_line = vec![source_span, level_span];
+                        for span in line.spans {
+                            out_line.push(span);
```
jkelleyrtp, line 134: Per UX-review, we should only mark the first line with the source tag - all other lines should be at the right separation distance (as per measured by the first line) but not have the `[build]` tag.

```
[build] INFO: Optimized static asset /Users/jonkelley/Developm... [minified]                                                                              ▐
[build] INFO: Copying public assets to the output directory...                                                                                            ▐
[build] INFO: Running wasm-bindgen                                                                                                                        ▐
[build] WARN: The `web.resource` config has been deprecated in favor of head components and will be removed in a future release. Instead of including     ▐
assets in the config, you can include assets with the `asset!` macro and add them to the head with `head::Link` and `Script` components.                  ▐
[build] WARN: To migrate to head components, remove `web.resource.style` and include the following rsx in your root component:                            ▐
[build] WARN: ```rust                                                                                                                                     ▐
[build] WARN: rsx! {                                                                                                                                      ▐
[build] WARN:     head::Link { rel: "stylesheet", href: asset!(css("./assets/main.css")) }                                                                ▐
[build] WARN: }                                                                                                                                           ▐
```
[build] INFO: Optimized static asset /Users/jonkelley/Developm... [minified]                                                                              ▐
[build] INFO: Copying public assets to the output directory...                                                                                            ▐
[build] INFO: Running wasm-bindgen                                                                                                                        ▐
[build] WARN: The `web.resource` config has been deprecated in favor of head components and will be removed in a future release. Instead of including     ▐
assets in the config, you can include assets with the `asset!` macro and add them to the head with `head::Link` and `Script` components.                  ▐
[build] WARN: To migrate to head components, remove `web.resource.style` and include the following rsx in your root component:                            ▐
[build] WARN: ```rust                                                                                                                                     ▐
[build] WARN: rsx! {                                                                                                                                      ▐
[build] WARN:     head::Link { rel: "stylesheet", href: asset!(css("./assets/main.css")) }                                                                ▐
[build] WARN: }                                                                                                                                           ▐
```

To something like:

```
[build] WARN: The `web.resource` config has been deprecated in favor of head components and will be removed in a future release.
              To migrate to head components, remove `web.resource.style` and include the following rsx in your root component:                            
              ```rust                                                                                                                                     
              rsx! {                                                                                                                                      
                  head::Link { rel: "stylesheet", href: asset!(css("./assets/main.css")) }                                                                
              }                                                                                                                                           
```

**Review by jkelleyrtp — CHANGES_REQUESTED** [review]
Overall the code quality here is great :)

A few things discovered:
- copy-paste lines is borked with the scrollbar - we probably need to indicate number of lines on the top/bottom of the frame so they don't get caught
- mousecapture captures the line selection. if we keep mouse capture then we need to implement our own text selection (which maybe is fine? - how hard could this logic be? 😆)
- mousecapture also captures copy/paste. We probably could add this back, but then it's an uphill battle for unicode support
- if we do override mouse selection then it might be useful to select entire lines at a time when doing multitline - IE we don't have to write overly complex line selection logic: just select the containing lines or just the sub-line.
- Removing tabs is the right answer, but we do need some way of filtering
- we should be printing the launch URL or put it in the powerline. we could do a really short url (dx.local or [::]:8081) so it doesn't take up much space
- Some warnings (like the web.resource deprecation warning) are really annoying. I don't need/want to see them on every rebuild. Perhaps we could just do what cargo does and overwrite an ongoing line with status updates before committing the message. And then we don't show warnings on hot rebuilds: only errors.

**Comment by Andrew15-5** [comment]
> * [x]  Removal of the "listening at x...". Pressing `[o] open` does the trick and the possibility to remove it was mentioned so I did so anyways (too lazy to make it responsive with the terminal size). **Let me know if we do want to keep it.**

> * [x]  Print launch URL in console.

So, is the URL still present? If so, then maybe it's fine. But I want to see it. Does it compile now?

**Comment by Andrew15-5** [comment]
Compiled, started serving and pressing arrow keys:

```
thread 'main' panicked at packages/cli/src/serve/output.rs:435:53:
attempt to subtract with overflow
```

https://github.com/dioxuslabs/dioxus/blob/0103d51d6d46de96e18ac7ade46ae7737e4ef1b1/packages/cli/src/serve/output.rs#L435

**Comment by Andrew15-5** [comment]
When logs are already on "assets optimization", the status message still shows the compiling that has been finished a long time ago:

image

It should skip showing stages that already have been done. Logs and status message should be in sync.

**Comment by Andrew15-5** [comment]
> So, is the URL still present? If so, then maybe it's fine. But I want to see it. Does it compile now?

```
[dev]  INFO: Development server listening at http://127.0.0.1:8080
```

image

I guess this is how it looks now? Since we have a wrapper server starting at the very start, it's OK to send this message at the start.

DogeDark:
@Andrew15-5 Thanks for the report about the overflow issue with arrow keys. I am looking into that now. The other issue with the build messages is that it is building for fullstack and since two builds are running, they can conflict with the message output. What happened here is one build finished quicker and started optimizing but the status bar (compiling x%..) was showing the status for the slower build. I do not plan to fix this just yet as we wanna get this PR wrapped up.

Andrew15-5:
I also would like this path (and others if they will appear) to be relative to the project dir. This will make it smaller and more readable. Same for issue logs. So instead it will be `[target/dx-dist-name]`.

image

DogeDark:
@Andrew15-5 I think that would be best in a separate PR once this and the other big PR are merged. If you could post your request in the CLI megathread issue, that would be great.

jkelleyrtp (approved):

Andrew15-5:
There is a typo, btw: https://github.com/DioxusLabs/dioxus/actions/runs/10845472485/job/30096405681?pr=2846.

jkelleyrtp:
I ended up finding a fix to the vscode scrollback issue so we can actually select lines again.

The custom solution just wasn't working for me on macos and seemed too complex to keep in the CLI. 

I also ended up moving the sections around for a slightly different appearance to be more "TUI-esque":

Screenshot 2024-09-13 at 1 12 13 AM

I think longer term we want to hide all warnings from cargo so the build log isn't so noisy, and then maybe add a keybind to toggle "verbose" mode.

Andrew15-5:
> I think longer term we want to hide all warnings from cargo so the build log isn't so noisy, and then maybe add a keybind to toggle "verbose" mode.

Yeah, that's interesting. But it should mention somewhere that "All warnings are suppressed" if some warnings appear. It's a great alternative to the manual workaround: https://users.rust-lang.org/t/suppress-warnings-from-the-cargo-command/10536.

Issue #2807, opened 2024-08-08
`error: [warning] Elements and text nodes do not need to be separated by commas.` should be a warning


Niedzwiedzw:
- Dioxus version:  0.6.0-alpha.2 (master, commit b97e607aca0bd8a12971fcf403908addf9a45dd7)

PR #2869, opened 2024-08-20
Fix: clippy, dont throw error on commas

jkelleyrtp:
Fix:
#2807

### DIFF
--- a/packages/core/src/suspense/component.rs
+++ b/packages/core/src/suspense/component.rs
@@ -316,7 +316,7 @@ impl SuspenseBoundaryProps {
             // Store the (now mounted) children back into the scope state
             let scope_state = &mut dom.scopes[scope_id.0];
             let props = Self::downcast_from_props(&mut *scope_state.props).unwrap();
-            props.children = children.clone();
+            props.children.clone_from(&children);
 
             let scope_state = &mut dom.scopes[scope_id.0];
             let suspense_context = scope_state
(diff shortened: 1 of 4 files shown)


packages/core/src/suspense/component.rs
```diff
@@ -435,7 +435,7 @@ impl SuspenseBoundaryProps {
             // Store the (now mounted) children back into the scope state
             let scope_state = &mut dom.scopes[scope_id.0];
             let props = Self::downcast_from_props(&mut *scope_state.props).unwrap();
-            props.children = children.clone();
+            props.children.clone_from(&children);
```
jkelleyrtp, line 438: new clippy lint not firing on our current ci setup

packages/rsx/src/ifmt.rs
```diff
@@ -399,10 +391,11 @@ mod tests {
         let input = syn::parse2::<IfmtInput>(quote! { "hello {world} {world} {world()}" }).unwrap();
         println!("{}", input.to_string_with_quotes());
 
+        // raw expressions are always static!
         let input =
             syn::parse2::<IfmtInput>(quote! { r#"hello {world} {world} {world()}"# }).unwrap();
         println!("{}", input.to_string_with_quotes());
-        assert!(!input.is_static());
+        assert!(input.is_static());
```
ealmloff, line 398: This seems wrong... The input has a bunch of dynamic segments. It shouldn't be static

ealmloff (approved):

PR #2872, opened 2024-08-21
Add placeholder doc comments to macro-generated enums

pyrrho:
This removes some yellow squiggles present in workspaces/crates that build with
```
[workspace.lints.rust]
missing_docs = "warn"`
```
image

I don't feel that I'm familiar enough with the structure of these enums to suggest real doc comments, so I just tossed in placeholders. Would love to see some more descriptive text replace the TODOs.

Another option for resolving the warnings would be to replace all the placeholders with an `#[allow(missing_docs)]` attribute added above the two `quote!`d `pub enum`s on lib.rs:689 and segment.rs:300

### DIFF
--- a/packages/router-macro/src/lib.rs
+++ b/packages/router-macro/src/lib.rs
@@ -643,7 +643,11 @@ impl RouteEnum {
                     let error_name = route.error_ident();
                     let route_str = &route.route;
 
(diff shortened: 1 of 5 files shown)


pyrrho:
Thank you _so much_ for finishing and landing this PR!

PR #2886, opened 2024-08-26
Fix hot reloading components with keys

### LINKED ISSUE #2881: Component with `key` attribute does not pass a `ReadOnlySIgnal<bool>` attribute properly
**Problem**

With latest main in web, this app will display `falsefalse` when I would expect `truefalse`. Specifically, the `is_on: true` is not getting passed or getting ignored when there is a `key` on the component.

```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;

#[derive(Clone, Routable, Debug, PartialEq)]
enum Route {
    #[route("/")]
    Home {},
}

fn main() {
    launch(App);
}

fn App() -> Element {
    rsx! {
        Router::<Route> {}
    }
}

#[component]
fn Comp(
    is_on: ReadOnlySignal<bool>,
) -> Element {
    rsx! {
        "{is_on()}"
    }
}

#[component]
fn Home() -> Element {
    rsx! {
        Comp {
            key: "A",
            is_on: true,
        }
        Comp {
            key: "B",
            is_on: false,
        }
    }
}
```

If you remove the `key` from the first `Comp` it will display what I would expect: `truefalse`. But I need to have a key on these elements.

**Environment:**
 - Dioxus version: main
 - App platform: web

**Questionnaire**

ealmloff:
We were adding `key`s from component to the component literal pool which caused indexing issues because keys don't show up as props

Fixes #2881

### DIFF
--- a/packages/rsx/src/assign_dyn_ids.rs
+++ b/packages/rsx/src/assign_dyn_ids.rs
@@ -1,6 +1,7 @@
 use crate::attribute::Attribute;
 use crate::{
-    AttributeValue, BodyNode, HotLiteral, HotReloadFormattedSegment, Segment, TemplateBody,
+    AttributeName, AttributeValue, BodyNode, HotLiteral, HotReloadFormattedSegment, Segment,
+    TemplateBody,
 };
 
 /// A visitor that assigns dynamic ids to nodes and attributes and accumulates paths to dynamic nodes and attributes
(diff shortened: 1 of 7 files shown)


ealmloff:
> Hmm, I cherry picked these onto my local and it did not fix #2881

I reverted a bit too much while fixing tests. Should be fixed now

rogusdev:
Ha, now it's inverted: `falsetrue`

ealmloff:
Should be fixed now:
Screenshot 2024-08-26 at 11 49 46 AM

rogusdev:
Yeah, works for me, thanks! (Including in my actual project)

matthunz (approved):

PR #2890, opened 2024-08-26
Restore Nix devShell support

srid:
- #2547 introduced a change that broke the devShell on macOS
- Also update nixpkgs & rust-overlay so we get newer Rust compiler that is needed to compile `dioxus-cli`.

Tested using `nix develop -c cargo build -p dioxus-cli` which produces a functioning `target/debug/dx`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,44 +5,26 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1709336216,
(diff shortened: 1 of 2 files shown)


ealmloff (approved):

jkelleyrtp:
Can you follow this PR up with a CI additional that tests nix support so we don't break it in the future?

srid:
> Can you follow this PR up with a CI additional that tests nix support so we don't break it in the future?

@jkelleyrtp Yup - https://github.com/DioxusLabs/dioxus/pull/2910

PR #2892, opened 2024-08-27
Fix static generation serve

ealmloff:
Static generation was looking for an environment variable that the CLI no longer sets when serving. This PR fixes that issue and makes dx serve just work

### DIFF
--- a/packages/cli/src/builder/fullstack.rs
+++ b/packages/cli/src/builder/fullstack.rs
@@ -98,7 +98,7 @@ impl BuildRequest {
 
     fn new_server(serve: bool, config: &DioxusCrate, build: &Build) -> Self {
         let mut build = build.clone();
-        if build.profile.is_none() {
+        if build.profile.is_none() && !build.release {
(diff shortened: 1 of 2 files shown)


srid:
`dx serve` seems to work **at first**

image

But it eventually crashes with:


image

srid:
With [0c88b39](https://github.com/DioxusLabs/dioxus/pull/2892/commits/0c88b39c765fbf3fb421a4636890c949926f0557), `dx build --release` succeeds, but the static files cannot be opened in the browser when served with a plain http-server (it just serves the file statically from `./dist`, like GitHub pages would). I get a page with title, but the body is empty; with these console errors:

image

`dx serve` on the other hand, continues to fail:

image

Even without `--release`:

image


@ealmloff Is there something I'm missing here?

ealmloff:
`dx serve` and `dx build` are kind of confusing for static site generation. Currently build is set up to build the wasm and static site generation binary, but not run it. It should be running it as well. Marking this as a draft until I get more intuitive behavior implemented

PR #2897, opened 2024-08-27
Fix partialeq implementation for coroutine

ealmloff:
Fixes #2895


marc2332:
🤦🏻‍♂️ My bad, looks like I did not test it properly and went too fast, and the CI trolled us 🙏🏻

marc2332 (approved):

PR #2902, opened 2024-08-29
Allow hooks in the expression of a match statement and if statement

ealmloff:
Fixes `dx check` for this code snippet:
```rust
if use_signal(|| true) {
     println!("123")
}

match use_signal(|| true) {
    true => println!("false"),
    false => println!("true")
}
```

And report errors for this code snippet:
```rust
use_hook(|| use_signal(|| 0));
spawn(async move { use_signal(|| 0) });
```

### DIFF
--- a/packages/check/src/check.rs
+++ b/packages/check/src/check.rs
@@ -208,7 +208,11 @@ impl<'ast> syn::visit::Visit<'ast> for VisitHooks {
                 .unwrap_or_else(|| i.span())
                 .into(),
         )));
-        syn::visit::visit_expr_if(self, i);
+        // only visit the body and else branch, calling hooks inside the expression is not conditional
+        self.visit_block(&i.then_branch);
+        if let Some(it) = &i.else_branch {
+            self.visit_expr(&(it).1);
+        }
         self.context.pop();
     }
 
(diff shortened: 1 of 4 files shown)


DogeDark:
Does Rust evaluate every condition in an if statement? E.g.
```rs
let some_value = false;

// Would the use_signal still get checked(ran) despite the conditional already failing?
if some_value && use_signal(|| true) {
    println!("123");
}
```

ealmloff:
> Does Rust evaluate every condition in an if statement? E.g.
> 
> ```rust
> let some_value = false;
> 
> // Would the use_signal still get checked(ran) despite the conditional already failing?
> if some_value && use_signal(|| true) {
>     println!("123");
> }
> ```

No, that case is still unchecked. Compared to if statements and closures, I think it is more rare, but we should lint for that eventually as well. Note it isn't specific to if statements, this code doesn't print anything either:
```rust
fn main() {
    let val = false && returns_true();
    assert!(!val)
}

fn returns_true() -> bool {
    panic!("this will not be run")
}
```

matthunz (approved):
Looks great! I never thought of using hooks like that but I'm sure I'll give it a go now. I posted https://github.com/DioxusLabs/dioxus/issues/2940 to eventually add specific lints for `&&` and `||` in `if` and `match`

PR #2910, opened 2024-08-31
ci: Add workflow for building `flake.nix`

srid:
This will build all flake outputs (using [omnix](https://omnix.page/om/ci.html)). It will also run `cargo build` inside the Nix devShell to make sure that it works, since that seems to be our primary intended use case of Nix in this repo.


matthunz:
Looks great! I'll merge this since Jon already requested it https://github.com/DioxusLabs/dioxus/pull/2890#issuecomment-2319845388

matthunz (approved):

Issue #2858, opened 2024-08-18
Add `relatedTarget` property to `FocusEvent`


Aaron3154518:
## Feature Request

I have a use case when I need to know what element is being focused when another element is being blurred. The stack overflow solution below uses the `relatedTarget` property to do this, however this property isn't exposed by dioxus. Can we get this field?

Stack overflow solution: https://stackoverflow.com/questions/39439115/how-to-execute-click-function-before-the-blur-function/57983847#57983847
`relatedTarget`: https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget

ealmloff:
This could be added when the `target` field is added to events as a `ElementRef` (currently called `MountedData`.

miguelpduarte:
Is this being worked on? This is quite useful when figuring out e.g. whether to close a custom dropdown by checking which element is focused.

PR #2911, opened 2024-09-01
 Add the `onvisible` event handler to `Element`

ASR-ASU:
This PR add the capability to handle the visible events for a given `Element`:

```rust
#[component]
fn MyComponent() {
    rsx! {
        div {
            onvisible: move |evt| {
	        let data = evt.data();

                println!("bounding_client_rect={:?}", data.get_bounding_client_rect());
                println!("intersection_ratio={:?}", data.get_intersection_ratio());
                println!("intersection_rect={:?}", data.get_intersection_rect());
                println!("is_intersecting={:?}", data.is_intersecting());
                println!("root_bounds={:?}", data.get_root_bounds());
                println!("target={:?}", data.get_target());
                println!("time={:?}", data.get_time());
            }
        }
    }
}
```

For now, the [IntersectionObserverEntry.target](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/target) is not used/serialized. If it make sense to use it, how can we serialize it? Could we use the `data-dioxus-id`?


ASR-ASU:
`ElementRef` could do the trick.

packages/desktop/src/webview.rs
```diff
@@ -110,6 +110,7 @@ impl WebviewEdits {
                     data.into_any()
                 }
             }
+            dioxus_html::EventData::Visible(_) => data.into_any(),
```
ealmloff, line 113: This looks the same as the branch following it. Should it be merged or do we need to handle converting visible data to any in a different way?

packages/html/src/events/visible.rs
```diff
@@ -0,0 +1,325 @@
...
+        self.inner.is_intersecting()
+    }
+
+    /// Get the rect for the intersection observer's root
+    pub fn get_root_bounds(&self) -> VisibleResult<PixelsRect> {
+        self.inner.get_root_bounds()
+    }
+
+    /// Get the element whose intersection with the root changed
+    pub fn get_target(&self) -> VisibleResult<ElementId> {
```
ealmloff, line 47: `ElementId` isn't very useful in user-facing code. It is mainly used in core/the renderers. We could use `MountedData` here which wraps the `ElementId` or `web_sys::Element` with accessors.

The target here will always be the element that the event listener is attatched to right? If that is true, then it might be better to wait until we add `target() -> MountedData` to all events together.

packages/html/src/events/visible.rs
```diff
@@ -0,0 +1,325 @@
...
+        self.inner.get_target()
+    }
+
+    /// Get a timestamp indicating the time at which the intersection was recorded
+    pub fn get_time(&self) -> VisibleResult<f64> {
```
ealmloff, line 52: Could we return an instant here instead of a float? I'm guessing seconds, but it isn't clear what the unit is looking at the function signature.

ealmloff (commented):
Thanks for working on this! I had a few questions about some APIs, otherwise the implementation looks good

packages/desktop/src/webview.rs
ASR-ASU: Indeed. I'm going to remove this branch.

packages/html/src/events/visible.rs
ASR-ASU: This is also my understanding (cf. https://css-tricks.com/a-few-functional-uses-for-intersection-observer-to-know-when-an-element-is-in-view/). Are you referring to the https://github.com/DioxusLabs/dioxus/issues/2858?

ealmloff: > Are you referring to the https://github.com/DioxusLabs/dioxus/issues/2858?
Yes, both `relatedTarget` and `target` need the mounteddata for the element

Issue #2908, opened 2024-08-31
Why do you use tokio::time::sleep in docs if it is not supported by wasm?


gekh:
Just some examples, everywhere is `tokio::time::sleep`
https://dioxuslabs.com/learn/0.5/reference/use_coroutine
https://dioxuslabs.com/blog/release-050
https://github.com/DioxusLabs/dioxus/blob/9d099a7f3fa78ee9d87aae06ad1aae03fded6c19/examples/eval.rs#L16

I ended up by using `wasmtimer::tokio::*`, but it took me a lot of time and nerves to find it. 
Why do you have tokio in docs and examples? Why you do not have your own sleep implementation like Leptos?

mchant:
Thanks!! Just ran into this issue myself.

Heeten:
I ran into this issue as well, it's used in the `use_future` doc at https://docs.rs/dioxus/latest/dioxus/prelude/fn.use_future.html

PR #2912, opened 2024-09-01
Use of `async_std::task::sleep` instead of `tokio::time::sleep` in examples

ASR-ASU:
Proposal for https://github.com/DioxusLabs/dioxus/issues/2908 issue. Replacing tokio allows the following examples to run on wasm platforms:
- backgrounded_futures
- clock
- control_focus
- eval
- future
- signals
- streams

What about the documentation?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2603,6 +2603,7 @@ dependencies = [
 name = "dioxus-examples"
 version = "0.6.0-alpha.2"
 dependencies = [
+ "async-std",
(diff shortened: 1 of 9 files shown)


ealmloff (approved):
Thank you! I switched to just always using web_time in the clock example. It will fall back to std on any non-wasm platform

ealmloff:
> What about the documentation?

We should modify the documentation in the same way. If the code snippets we use can be cross platform, that would be better

PR #2934, opened 2024-09-06
Added `dx clean` stats

Andrew15-5:
https://github.com/DioxusLabs/dioxus/issues/2788#issuecomment-2270538452:

> `dx clean` doesn't show any progress or how many GiB it has removed.

Currently, I just naively parsed the output of `cargo clean` and then summed it with Dioxus-specific cleanings.

1. `DioxusCrate::new()` uses Internet connection when using `dx clean`
2. Does `cargo clean` **only** remove the `${CARGO_TARGET_DIR:-target}`? If so, then I can directly grab its size which will eliminate string parsing and the call to `cargo clean` (optionally).
3. Currently, can only print the total removed size, but not the total number of files removed.
4. Need to create 2 static hashmaps by using `once_cell` (can be removed if MSRV is 1.80, IIRC).

### DIFF
--- a/packages/cli/src/cli/clean.rs
+++ b/packages/cli/src/cli/clean.rs
@@ -1,6 +1,8 @@
 use crate::DioxusCrate;
 use anyhow::Context;
 use build::TargetArgs;
+use once_cell::sync::Lazy;
+use std::collections::HashMap;
 
 use super::*;
 
@@ -9,8 +11,36 @@ use super::*;
 #[clap(name = "clean")]
 pub struct Clean {}
 
+static N_TO_IEC_UNIT_MAP: Lazy<HashMap<i32, &str>> = Lazy::new(|| {
+    HashMap::<i32, &str>::from([
+        (1, "K"),
+        (2, "M"),
+        (3, "G"),
+        (4, "T"),
+        (5, "P"),
+        (6, "E"),
+        (7, "Z"),
+        (8, "Y"),
+    ])
+});
+
+static IEC_UNIT_TO_N_MAP: Lazy<HashMap<&str, i32>> = Lazy::new(|| {
+    HashMap::<&str, i32>::from([
+        ("B", 0),
+        ("KiB", 1),
+        ("MiB", 2),
+        ("GiB", 3),
+        ("TiB", 4),
+        ("PiB", 5),
+        ("EiB", 6),
+        ("ZiB", 7),
+        ("YiB", 8),
+    ])
+});
+
 impl Clean {
     pub fn clean(self) -> anyhow::Result<()> {
+        // Tries to access Internet when offline.
         let dioxus_crate =
             DioxusCrate::new(&TargetArgs::default()).context("Failed to load Dioxus workspace")?;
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/clean.rs
```diff
@@ -9,8 +11,36 @@ use super::*;
...
+        ("MiB", 2),
+        ("GiB", 3),
+        ("TiB", 4),
+        ("PiB", 5),
+        ("EiB", 6),
+        ("ZiB", 7),
+        ("YiB", 8),
+    ])
+});
+
```
matthunz, line 40: Could we replace these runtime-allocated hashmaps with match statements?

E.g.
```rs
match iec {
   "B" => 0,
   "KiB" => 1,
   ...
}
```

packages/cli/src/dioxus_crate.rs
```diff
@@ -139,10 +140,18 @@ pub struct DioxusCrate {
...
+        cmd.lock_opts(LockOptions {
+            offline: true, // Don't access Internet when `dx clean` is run.
+            frozen: false,
+            locked: false,
+        });
         cmd.features(target.features.clone());
         let builder = krates::Builder::new();
+        // Tries to access Internet when offline.
         let krates = builder.build(cmd, |_| {})?;
         let package = find_main_package(target.package.clone(), &krates)?;
```
matthunz, line 156: Is this necessary here? I'm not sure `cargo clean` will attempt to access the internet, but I'm definitely not sure

matthunz (commented):
Looks great! :+1: 

Something I'm curious about here is maybe directly checking the folder sizes before and after a clean instead of parsing? :thinking: Just in case cargo changes its format or etc

Just a couple ideas I had poking around:

packages/cli/src/cli/clean.rs
Andrew15-5: Huh... I guess we could. Instead of `.get()` function.

packages/cli/src/dioxus_crate.rs
Andrew15-5: It wouldn't have been there if I didn't try calling `dx clean` on a place with no Internet connection. It was failing for some reason. That was the reason. I was thinking that this should be a separate issue. But I'm not sure. I think a global `--offline` (and others) flag should be added, and it should be enabled by default for `clean` subcommand.

So it's necessary if you are offline and want to just clean up the project dir. But an additional if statement is missing.

matthunz: Ohh ok that does sound like a great idea 👍 I’ll see if I can reproduce that error, that definitely sounds like the right fix

Andrew15-5:
> Something I'm curious about here is maybe directly checking the folder sizes before and after a clean instead of parsing?

Yes, but if you consider that the project can have a giant directory or is in a workspace or something, this can lead to a longer clean up time. But in the happy case, this is definitely one of the easiest things to do.

Directly checking this makes sure that we don't spend any extra time evaluating the size that will be removed. Since Rust apps are supposed to be fast, I think we can trade a bit of more verbose clean up code for maximum performance.

But again, if only `${CARGO_TARGET_DIR:-target}` is removed, then I can remove parsing `cargo clean`'s output. I definitely want this instead of what I'm doing now. I might have to ask this in rustaceans' chat or look at it's implementation (and maybe even copy something from it).

Andrew15-5:
It looks like it is indeed only ever removes the target dir and nothing else. But you can configure in more ways than I thought: https://doc.rust-lang.org/cargo/guide/build-cache.html. I'm not sure if looking for `config.toml` is a good idea, especially since there could be a few of those. I will mention it so that maybe in the future this can be implemented if someone is indeed using it.

Andrew15-5:
Now the command shows this:

```
[packages/cli/src/cli/clean.rs:28:9] target_dir = "/tmp/name/target"
[packages/cli/src/cli/clean.rs:38:9] target_dir_file_count = 1
[packages/cli/src/cli/clean.rs:38:9] target_dir_size = 1733
[packages/cli/src/cli/clean.rs:41:9] &out_dir = "/tmp/name/dist"
[packages/cli/src/cli/clean.rs:51:9] out_dir_file_count = 0
[packages/cli/src/cli/clean.rs:51:9] out_dir_size = 0
Removed 1 files, 1.69 KiB total
```

Because of that same file that is being added at the top of the function call. You can't get 0 files and bytes.

But at least now everything is counted manually and precisely. But `cargo clean` is now substituted with `walkdir`.

jkelleyrtp:
Honestly probably not, but if we end up bringing back the out_dir for `dx bundle` then it might be useful again.

Andrew15-5:
So... Do you wanna keep it for now even though it doesn't serve any useful purpose? And I don't know when or if the out dir will be brought back. I say we remove it. If it will be needed by then, then we can make it better right away (with stats and all that).

Issue #1177, opened 2023-07-10
Iterable Children


itsezc:
## Specific Demand

Iterating through the Children of a component is very complex currently, for instance:

```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;

#[derive(Props)]
pub struct BreadcrumbsProps<'a> {
  divider: Option<&'a str>,
  children: Vec<Element<'a>>
}

pub fn Breadcrumbs<'a>(cx: Scope<'a, BreadcrumbsProps<'a>>) -> Element {
  cx.render(rsx!(
    nav {
      role: "navigation",
      for child in &cx.props.children {
        span {
          p { cx.props.divider.unwrap_or("->") }
          child
        }
      }
    }
  ))
}

#[derive(PartialEq, Props)]
pub struct BreadcrumbProps<'a> {
  name: &'a str,
  link: &'a str
}

pub fn Breadcrumb<'a>(cx: Scope<'a, BreadcrumbProps<'a>>) -> Element {
  cx.render(rsx!(
    a {
      href: cx.props.link,
      cx.props.name
    }
  ))
}
```

So this should be consumable like so:

```
Breadcrumbs {
  divider: "/",
  Breadcrumb {
    name: "Level 1",
    link: "/"
  }
  Breadcrumb {
    name: "Level 2",
    link: "/level2"
  }
}
```

Additional context: https://discord.com/channels/899851952891002890/1127735881076330548

## Implement Suggestion

The syntax for such implementation is not very straight forward, making the DX very poor implementing components like this. Either this should be made easier, or a better approach should be documented on how to best tackle this issue with code examples.

Demonthos:
In the short term, using a Vec of Elements as children can work, it just makes using the component less clean:

```rust
#![allow(non_snake_case)]
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch(app)
}

fn app(cx: Scope) -> Element {
    render! {
        Breadcrumbs {
            divider: "/",
            child_routes: vec![
                render!{
                    Breadcrumb {
                        name: "Level 1",
                        link: "/"
                    }
                },
                render!{
                    Breadcrumb {
                      name: "Level 2",
                      link: "/level2"
                    }
                }
            ]
        }
    }
}

#[derive(Props)]
pub struct BreadcrumbsProps<'a> {
    divider: Option<&'a str>,
    child_routes: Option<Vec<Element<'a>>>,
}

[...]

PR #2946, opened 2024-09-09
Create `Visit` trait

matthunz:
Implements a zero-cost visitor for VNodes via a trait pattern instead of an iterator.

Alternative to https://github.com/DioxusLabs/dioxus/pull/2722 to solve https://github.com/DioxusLabs/dioxus/issues/1177.

Exports the `Visit` trait:
```rs
pub trait Visit {
    /// Visit a [`VNode`].
    fn visit_vnode(&mut self, vdom: &VirtualDom, vnode: &VNode) { ... }

    /// Visit a template node.
    fn visit_node(&mut self, vdom: &VirtualDom, vnode: &VNode, node: TemplateNode) { ... }

    /// Visit an element.
    fn visit_element(
        &mut self,
        vdom: &VirtualDom,
        vnode: &VNode,
        tag: &'static str,
        namespace: Option<&'static str>,
        attrs: &'static [TemplateAttribute],
        children: &'static [TemplateNode],
    ) { ... }

    /// Visit an element attribute.
    fn visit_attr(&mut self, vdom: &VirtualDom, vnode: &VNode, attr: &TemplateAttribute) { ... }

    /// Visit a dynamic node.
    fn visit_dynamic(
        &mut self,
        vdom: &VirtualDom,
        vnode: &VNode,
        index: usize,
        node: &DynamicNode,
    ) { ... }

    /// Visit a placeholder.
    fn visit_placeholder(&mut self, vdom: &VirtualDom, vnode: &VNode, index: usize) { ... }

    /// Visit a fragment.
    fn visit_fragment(&mut self, vdom: &VirtualDom, vnode: &VNode, roots: &[VNode]) { ... }

    /// Visit a text node.
    fn visit_text(&mut self, vdom: &VirtualDom, vnode: &VNode, text: &str) { ... }
}
```

And default methods under the `visit` module:
```rs
/// Default method to visit a [`VNode`].
pub fn visit_vnode<V: Visit + ?Sized>(visitor: &mut V, vdom: &VirtualDom, vnode: &VNode) { ... }

/// Default method to visit a template node.
pub fn visit_node<V: Visit + ?Sized>(
    visitor: &mut V,
    vdom: &VirtualDom,
    vnode: &VNode,
    node: TemplateNode,
) { ... }

/// Default method to visit an element.
pub fn visit_element<V: Visit + ?Sized>(
    visitor: &mut V,
    vdom: &VirtualDom,
    vnode: &VNode,
    tag: &'static str,
    namespace: Option<&'static str>,
    attrs: &'static [TemplateAttribute],
    children: &'static [TemplateNode],
) { ... }

/// Default method to visit a dynamic node.
pub fn visit_dynamic<V: Visit + ?Sized>(
    visitor: &mut V,
    vdom: &VirtualDom,
    vnode: &VNode,
    index: usize,
    node: &DynamicNode,
) { ... }

/// Default method to visit a placeholder.
pub fn visit_placeholder<V: Visit + ?Sized>(
    visitor: &mut V,
    vdom: &VirtualDom,
    vnode: &VNode,
    index: usize,
) { ... }

/// Default method to visit a fragment.
pub fn visit_fragment<V: Visit + ?Sized>(
    visitor: &mut V,
    vdom: &VirtualDom,
    vnode: &VNode,
    roots: &[VNode],
) { ... }
```

### DIFF
--- a/packages/core/src/lib.rs
+++ b/packages/core/src/lib.rs
@@ -26,6 +26,8 @@ mod scopes;
 mod suspense;
 mod tasks;
 mod virtual_dom;
+
+/// Visitor for VNodes.
 pub mod visit;
 
 mod hotreload_utils;
@@ -63,6 +65,7 @@ pub(crate) mod innerlude {
     pub use crate::suspense::*;
     pub use crate::tasks::*;
     pub use crate::virtual_dom::*;
+    pub use crate::visit::{self, Visit};
 
     /// An [`Element`] is a possibly-none [`VNode`] created by calling `render` on [`ScopeId`] or [`ScopeState`].
     ///
@@ -92,13 +95,13 @@ pub mod prelude {
         provide_context, provide_error_boundary, provide_root_context, queue_effect, remove_future,
         schedule_update, schedule_update_any, spawn, spawn_forever, spawn_isomorphic, suspend,
         throw_error, try_consume_context, use_after_render, use_before_render, use_drop, use_hook,
-        use_hook_with_cleanup, with_owner, AnyValue, Attribute, Callback, Component,
+        use_hook_with_cleanup, visit, with_owner, AnyValue, Attribute, Callback, Component,
         ComponentFunction, Context, Element, ErrorBoundary, ErrorContext, Event, EventHandler,
         Fragment, HasAttributes, IntoAttributeValue, IntoDynNode, OptionStringFromMarker,
         Properties, ReactiveContext, RenderError, Runtime, RuntimeGuard, ScopeId, ScopeState,
         SuperFrom, SuperInto, SuspendedFuture, SuspenseBoundary, SuspenseBoundaryProps,
         SuspenseContext, SuspenseExtension, Task, Template, TemplateAttribute, TemplateNode, VNode,
-        VNodeInner, VirtualDom,
+        VNodeInner, VirtualDom, Visit,
     };
 }
 
--- a/packages/core/src/lib.rs
+++ b/packages/core/src/lib.rs
@@ -26,6 +26,7 @@ mod scopes;
 mod suspense;
 mod tasks;
 mod virtual_dom;
+pub mod visit;
 
 mod hotreload_utils;
 
--- a/packages/core/src/visit.rs
+++ b/packages/core/src/visit.rs
@@ -126,9 +126,11 @@ pub fn visit_dynamic<V: Visit + ?Sized>(
 ) {
     match node {
         DynamicNode::Component(component) => {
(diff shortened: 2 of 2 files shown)


ealmloff:
I see two main use cases for #1177:
1) Renderers like `dioxus-tui` need to traverse the virtual dom and don't care about caching too much
2) Components like `head::Title` need to look through their children and find a specific node. This could also allow more lints like warning if an outlet is not found in the children of a router

Right now any general version of vnode traversal only really useful for `1` because it requires direct access to the `VirtualDom`. I don't think we can create a unified solution to both until we move component state to the runtime

The visitor approach lets you avoid matching over the node type which is nice, but the API is a bit odd if you don't know about templates which seems common for `2`. You still need to deal with static and dynamic attribute values separately and there are indexes floating around the methods to link Template and VNode state together

I think the visitor API works better for `1`. It fits an immediate mode renderer very nicely. Still a bit rough around the edges because of templates, but a lot better than how you would implement this today. Renders like dioxus-ssr also seem like a natural fit, but I don't think we would use this in practice because we need to cache the template and avoid traversing it every time

What do you think about combining the two APIs? The visitor could visit the more abstract version of nodes that the cursor API provides. Then for use case `2` you can just use the cursor directly

matthunz:
> Right now any general version of vnode traversal only really useful for `1` because it requires direct access to the `VirtualDom`. I don't think we can create a unified solution to both until we move component state to the runtime

Makes sense 👍 If we can move component state I feel like this could easily be used without the `VirtualDom`

> The visitor approach lets you avoid matching over the node type which is nice, but the API is a bit odd if you don't know about templates which seems common for `2`.

Totally agree, I am wondering if we could ease that with docs like the `syn` crate https://docs.rs/syn/latest/syn/visit/index.html

> You still need to deal with static and dynamic attribute values separately and there are indexes floating around the methods to link Template and VNode state together

You make an interesting point here, I feel like we could change the `Visit` trait to visit both the same. Maybe `visit_dynamic` can delegate to `visit_attribute` or `visit_element`?

> What do you think about combining the two APIs? The visitor could visit the more abstract version of nodes that the cursor API provides. Then for use case `2` you can just use the cursor directly

That makes a lot of sense to me 👍 I personally think it would be nice to have an iterator for more dynamic situations (and easier `children.map(...)` etc) and a visitor for static situations (like renderers, debugging, and maybe lints now that you mention it)

jkelleyrtp:
Closing this since I'm not very committed to the structure of the rsx!{} call output (vnodes/nodes/etc) and am hesitant to support any iterators/traits until I feel like the rest is stable.

Currently it's possible to traverse the tree manually - as we do in a variety of places - but not very ergonomic.

PR #2947, opened 2024-09-10
add `disabled`, `form` and `name` attributes to fieldset

chungwong:
[fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attributes) is supposed to have `disabled`, `form` and `name` attributes.


ealmloff (approved):
Thanks!

Issue #2933, opened 2024-09-06
External component breaks fullstack with custom Tokio/Axum


chungwong:
**Problem**

**Steps To Reproduce**
Given two crates and they are using the latest commit https://github.com/DioxusLabs/dioxus/commit/9816b402fc6bfd6bd3aad38d2612709f16dff296 (latest or alpha2, doesn't make a difference)
1. A UI lib crate https://github.com/chungwong/rayquaza
2. A consumer crate https://github.com/chungwong/myrust which uses the ui crate

On its own, the ui crate is working good.
```bash
❯ git clone https://github.com/chungwong/rayquaza.git
❯ cd rayquaza/crates/ui

❯ dx-9816b40 -V
dioxus 0.6.0-alpha.2 (9816b40)

❯ npm install
❯ npx tailwindcss -i ./input.css -o ./assets/tailwind.css --watch

❯ dx-9816b40 serve --example tabs
```
You will see the the Tab is responsive and its content is changing on http://localhost:8080

---

Let's try to consume the ui crate, and run it as an example.
```bash
❯ git clone https://github.com/chungwong/myrust.git

❯ cd myrust
❯ npm install
❯ npx tailwindcss -i ./input.css -o ./assets/tailwind.css --watch

❯ dx-9816b40 serve --example tabs
```
On http://localhost:8080, the tab still works

## Issue
```bash
❯ cd myrust
❯ dx-9816b40 serve
```
On http://localhost:3000 ([port has to change](https://github.com/chungwong/myrust/blob/main/src/main.rs#L53), it cannot be 8080 anymore, may deserve another issue, it is conflicting with the new asset server which is also on 8080),
 you will notice the tab is no longer responive and just broken.

Naturally, one may think the UI crate got problem, is it? Let's test it.
Turn this [main](https://github.com/chungwong/myrust/blob/main/src/main.rs#L34-L36) on and let's run it again

```bash
❯ dx-9816b40 serve
```
This time, it is http://localhost:8080 again and you will see the tab works again.

The problem is so strange, there is no error, neither in the terminal or the browser side. It looks like it is related to the tokio/axum side but I cannot be sure.

**Environment:**
 - Dioxus version:  master@9816b4
 - Rust version:    1.80
 - OS info:   debian 12
 - App platform:    fullstack

ealmloff:
[...]

PR #2952, opened 2024-09-11
Improve documentation for the fullstack crate

ealmloff:
The inline fullstack docs are currently very sparse. Items are documented, but they lack examples which can make it difficult to understand how they should be used. This PR adds a bunch of examples to the fullstack crate inline docs


packages/fullstack/README.md
```diff
@@ -63,6 +63,81 @@ async fn get_meaning(of: String) -> Result<Option<u32>, ServerFnError> {
...
+    // Get the address the server should run on. If the CLI is running, the CLI proxies fullstack into the main address
+    // and we use the generated address the CLI gives us
+    let cli_args = dioxus_cli_config::RuntimeCLIArguments::from_cli();
+    let address = cli_args
+        .as_ref()
+        .map(|args| args.fullstack_address().address())
+        .unwrap_or_else(SocketAddr::from(([127, 0, 0, 1], 8080)));
+
+    // Set up the axum router
+    let serve_config = ServeConfig::new().unwrap();
```
chungwong, line 100: Maybe we shouldn't promo/demo the usage of unwrap here because of
1. Highly possible it is not what the users wants.
2. using `?` makes the code a little bit succinct

packages/fullstack/README.md
```diff
@@ -63,6 +63,81 @@ async fn get_meaning(of: String) -> Result<Option<u32>, ServerFnError> {
...
+// The entry point for the server
+#[cfg(feature = "server")]
+fn main() {
+    // Get the address the server should run on. If the CLI is running, the CLI proxies fullstack into the main address
+    // and we use the generated address the CLI gives us
+    let cli_args = dioxus_cli_config::RuntimeCLIArguments::from_cli();
+    let address = cli_args
+        .as_ref()
+        .map(|args| args.fullstack_address().address())
+        .unwrap_or_else(SocketAddr::from(([127, 0, 0, 1], 8080)));
```
chungwong, line 97: I am suggesting to add [this cli logic](https://github.com/DioxusLabs/dioxus/blob/main/packages/fullstack/src/launch.rs#L131-L138).

With [this changes](https://github.com/DioxusLabs/dioxus/issues/2933#issuecomment-2338105792), it will just cause friction as port 8080 will conflict with the asset server.

ealmloff: port 8080 is a fallback if the CLI doesn't pass the port for the proxy

packages/fullstack/docs/request_origin.md
```diff
@@ -0,0 +1,43 @@
...
+}
+```
+
+2. A request to a [`server`](crate::prelude::server) function called directly from the client (either on desktop/mobile or on the web frontend after the initial render)
+
+```rust
+#[server]
+async fn read_headers() -> Result<(), ServerFnError> {
+    // The server context only exists on the server, so we need to put it behind a server_only! config
+    server_only! {
```
Shadorain, line 24: Is `server_only!` redundant already being in a `#[server]` defined server function?

Shadorain (commented):
Great work here!

PR #2779, opened 2024-08-03
Fix bundling/path errors, sidecar support, iOS/Android simulator support, asset hotreload fixes, serverfn hotrelad, native tailwind support

jkelleyrtp:
Attempt fixes for:
- [x] https://github.com/DioxusLabs/dioxus/issues/2526
- [x] https://github.com/DioxusLabs/dioxus/issues/2386
- [x] https://github.com/DioxusLabs/dioxus/issues/2379
- [x] https://github.com/DioxusLabs/dioxus/issues/2429
- [x] https://github.com/DioxusLabs/dioxus/issues/2176
- [x] https://github.com/DioxusLabs/dioxus/issues/232
- [x] https://github.com/DioxusLabs/dioxus/issues/2692
- [x] https://github.com/DioxusLabs/dioxus/issues/2509
- [x] https://github.com/DioxusLabs/dioxus/issues/2312
- [x] https://github.com/DioxusLabs/dioxus/issues/2093
- [x] https://github.com/DioxusLabs/dioxus/issues/2762
- [x] https://github.com/DioxusLabs/dioxus/issues/2344 (because we're joining from cwd instead of the asset dir)
- [x] https://github.com/DioxusLabs/dioxus/issues/2487
- [x] https://github.com/DioxusLabs/dioxus/issues/2155
- [x] https://github.com/DioxusLabs/dioxus/issues/2304

Misc cleanup:
- [x] fix all cache thrashing
- [x] get mobile + fullstack working
- [x] change fullstack to be a binary yes/no thing
- [x] helpful warnings when running fullstack w/o server functions
- [x] refactor CLI staging to include bundles (and thus, bundled hotreloading)
- [x] lay groundwork for #2754
- [x] adds run command (like serve without the TUI)

Currently cutting out:
- [x] parse env!() as std::env::var or meta tags in html
- [x] drop dep count on cli-config
- [x] dioxus-hotreload should be dioxus-devtools

Pending to be cut out:
- [x] change "hotreload" crate to "devtools"
- [x] adds http-server command
- [x] adds doctor command for tooling
- [x] rename "hotreload" feature to "devtools"

Cut out of this PR (and merged):
- [x] move hotreload context into core-types
- [x] move bubbles fn into core-types
- [x] new core-types crate
- [x] new isrg crate https://github.com/DioxusLabs/dioxus/pull/2969
- [x] remove oidc example (wasn't compiling) https://github.com/DioxusLabs/dioxus/pull/2966
[...]

PR #2959, opened 2024-09-13
Chore: hoist example projects for discoverability

jkelleyrtp:
Hoist all of our example projects into a top-level "example-projects" folder, similar to NextJS and Leptos.

This should make discoverability easier.

Also removes the OIDC example since it wasn't working and wasn't properly updated for 0.5.

Cut out of #2779 to make that PR smaller

Closes https://github.com/DioxusLabs/example-projects/issues/45


packages/static-generation/benches/incremental.rs.old
```diff
@@ -1,207 +0,0 @@
-#![allow(unused)]
```
ealmloff, line 1: Is removing this benchmark intentional?

ealmloff: Oh, I missed the .old

ealmloff (approved):
I like these examples being at the top level, but the `/examples` vs `/example-projects` distinction doesn't seem that useful. Many of these examples are simple unstyled examples that happen to use a full project to make the cargo.toml easier to copy. It might be better to just keep the `example-projects` we moved over from the original repo in the folder and `fullstack-hackernews` and move the rest of the full crate examples into the examples folder

PR #2960, opened 2024-09-13
Move component state from the `VirtualDom` to the `Runtime`

matthunz:
First draft at moving component state from the `VirtualDom` to the `Runtime`.
This allows access to component state while a Dioxus app is running, allowing for `VNode` iterators/visitors, new lints, and debugging.

Replaces `ScopeState` with a wrapper around an inner struct that is hidden from the external API:
```rs
pub(crate) struct Inner { ... }

pub struct ScopeState {
    inner: Rc<RefCell<Inner>>
}
```

Helps with https://github.com/DioxusLabs/dioxus/issues/1177:
 - https://github.com/DioxusLabs/dioxus/pull/2946
 - https://github.com/DioxusLabs/dioxus/pull/2722

Tasks:
- [x] Move component data to the `Runtime`
- [x] Update crates that depend on `dioxus-core`
- [ ] Fix missing root `ScopeState` in `provide_context`

### DIFF
--- a/packages/core/src/arena.rs
+++ b/packages/core/src/arena.rs
@@ -81,15 +81,22 @@ impl VirtualDom {
     // Note: This will not remove any ids from the arena
     pub(crate) fn drop_scope(&mut self, id: ScopeId) {
         let height = {
-            let scope = self.scopes.remove(id.0);
-            let context = scope.state();
+            let scope = self.runtime.scopes.borrow_mut().remove(id.0);
+            let scope_ref = scope.borrow();
+            let context = scope_ref.state();
             context.height
         };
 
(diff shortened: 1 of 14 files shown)


jkelleyrtp:
Why are we doing this?

matthunz:
> Why are we doing this, now especially?

Hey I'm sorry if I misunderstood the task queue, I noticed #1177 in the high-priority backlog and figured now would be a good time to try and get it in before `v0.6`. This would help combine the two current proposed solutions https://github.com/DioxusLabs/dioxus/pull/2946 and https://github.com/DioxusLabs/dioxus/pull/2722

I should be able to finish this draft today, and start combining those other two PRs, if that does make sense now. Otherwise please let me know if I should take on any specific issues for `v0.6`!

PR #2964, opened 2024-09-15
Cut build times in half by re-orienting build graph and simplifying a number of features and configs

jkelleyrtp:
This PR pulls out the most important changes related to cutting down build times from #2779 

Namely it:
- Adds a core-types crate so `dioxus-core` doesn't get compiled for proc macros
- Splits apart rsx into rsx and rsx-hotreload 
- Cuts apart and simplifies dioxus-cli-config to not include any dependencies (just env vars!)
- Cuts apart dioxus-hotreload into dioxus-devtools and dioxus-devtools-types
- Drops build.rs and fixes rerun-if-changed tracking for lazy-js-bundle to solve some spurious cache thrashing

Things that need to be done:
- Manganis needs to be cut apart (currently pulls in 60 dependencies across proc macros and runtime)
- Manganis probably needs to be inlined into this repo - we should not be re-exporting something in our prelude that we don't own and have version locked
- The router needs to be split apart so no web-sys stuff shows up for desktop
- The html crate needs to be split apart so no websys stuff shows up desktop (and native stuff for web...)
- more? I figured out a lot of stuff in #2779 

Other outstanding stuff from that PR:
- reworking of how we project configs into apps
- reworking of how the link intercept approach works

## numbers

With the asset feature disabled (a big outstanding culprit for thrashing and compile times), we drop from 25-30s compile times to 11/12s on my machine.

The number of crates falls from:
- dioxus (default, no renderer): ~155 to 96
- web: ~350 to 220
- desktop: ~400 to 210

All of our platforms compile in relatively similar times.


jkelleyrtp:
I'm just going to trim down #2779 into a better state, there's too many things that are connnected to each other to split this one out, unfortunately :(

PR #2967, opened 2024-09-16
Simplify the launch builder

### LINKED ISSUE #2965: `launch` should not be in prelude and `LaunchBuilder` and associated configs need to be reworked
## Feature Request

The `launch` function for dioxus apps should not be in the prelude. Since this function is used only once in dioxus apps, it does not belong in the prelude which is intended as a utility support glob import.

To launch a dioxus app, you should go through the `dioxus::launch` function:

```rust
fn main() {
    dioxus::launch(app);
}

```

This is generally just "good practice" to only include the minimal set of functions/traits in the prelude, as well as making it possible to discover other, new functions by using intellisense. We can then make things like the LaunchBuilder accessible via the `launch` module:

```rust
dioxus::launch::builder()
    .with_cfg(|| MyConfig)
    .launch(app);
```

## Clean up the `launch` crate:

The launch system we have now is supported by a bunch of macro_rules that make understanding `launch` quite confusing.

By default, we even mark `launch` as deprecated when no features are enabled, making `launch` even more confusing when using it with a default rust-analyzer setup.

This is also wired up into a build.rs and env!() calls that break proper caching, leading to very long incremental compile times.

The new design:

- Should not lean into so many macro_rules
- Should not be so typesafe (with_cfg can take any `impl Config` but not be connected to the renderer)
- Should not export `launch_desktop`, `launch_mobile`, etc.
- Should not specify Send/Sync of Context and instead take a Send/Sync callback that produces a !Send/!Sync Context
- Should simply panic with a *very pretty message* if no platform is specified.
- Should not be so generic (renderers should just attempt to downcast a named config)

## Notes

I originally took a stab at this in #2779 but dropped those changes since they added 100+ changed files and should be in their own PR.

ealmloff:
Stacked on https://github.com/DioxusLabs/dioxus/pull/2952
Fixes https://github.com/DioxusLabs/dioxus/pull/2892
Fixes #2965


examples/custom_menu.rs
```diff
@@ -27,8 +27,8 @@ fn main() {
     // Create a desktop config that overrides the default menu with the custom menu
     let config = dioxus::desktop::Config::new().with_menu(menu);
 
-    // Launch the app with the custom menu
-    LaunchBuilder::new().with_cfg(config).launch(app)
+    //dioxus::launch the app with the custom menu
```
liigo, line 30: shouldn't replace it

examples/hello_world.rs
```diff
@@ -2,17 +2,17 @@
 //!
 //! In this example we:
 //! - import a number of important items from the prelude (launch, Element, rsx, div, etc.)
-//! - define a main function that calls the launch function with our app function
+//! - define a main function that calls thedioxus::launch function with our app function
```
liigo, line 5: the space shouldn't be replaced

examples/hello_world.rs
```diff
@@ -2,17 +2,17 @@
...
 //! - define an app function that returns a div element with the text "Hello, world!"
 //!
 //! The `launch` function is the entry point for all Dioxus apps. It takes a function that returns an Element. This function
-//! calls "launch" on the currently-configured renderer you have. So if the `web` feature is enabled, it will launch a web
-//! app, and if the `desktop` feature is enabled, it will launch a desktop app.
+//! calls "launch" on the currently-configured renderer you have. So if the `web` feature is enabled, it willdioxus::launch a web
```
liigo, line 9: spaces (and more)

examples/mobile_demo/src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub fn main() -> Result<()> {
 
     // Right now we're going through dioxus-desktop but we'd like to go through dioxus-mobile
     // That will seed the index.html with some fixes that prevent the page from scrolling/zooming etc
-    LaunchBuilder::mobile()
+   dioxus::LaunchBuilder::mobile()
```
liigo, line 53: space

packages/autofmt/tests/samples/many_exprs.rsx
```diff
@@ -8,7 +8,7 @@ use std::{
 use tokio::time::sleep;
 
 fn main() {
-    LaunchBuilder::desktop().launch(app);
+   dioxus::LaunchBuilder::desktop().launch(app);
```
liigo, line 11: bad indents

PR #2983, opened 2024-09-20
Fix links with set base_path & make external links usable without modifier

LeWimbes:
This pull request addresses the following issues:

- **Internal Links:** When a `base_path` is set in `Dioxus.toml`, internal links only worked with a simple left click; CTRL + left click failed because the `href` only contained the absolute path to the next page without the `base_url`.
- **External Links:** External links required CTRL + left click due to the set `action` overwriting the normal left click behavior. This issue is independent of the `base_path`.

Additionally, I noticed that [this commit](https://github.com/DioxusLabs/dioxus/commit/3fb1f739d13967da9ee8db9f17a91fea321c9df2) broke base paths in general, but this pull request does not address that issue.

This is one of my first pull requests, so please let me know if there are any changes or improvements needed. Thank you!

### DIFF
--- a/packages/router/src/components/link.rs
+++ b/packages/router/src/components/link.rs
@@ -264,11 +264,15 @@ pub fn Link(props: LinkProps) -> Element {
     let do_default = onclick.is_none() || !onclick_only;
 
     let action = move |event: MouseEvent| {
+        // Only handle internal links
+        if is_external {
+            return;
+        }
         // Only handle events without modifiers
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
Hey! Thanks for the PR. We broke some of the basepath logic recently and I'm going to go fix it, so I scrubbed that from this PR but will incorporate it in a follow-on later today.

Issue #2990, opened 2024-09-22
Memo referenced in event handler is evaluating too soon?


rogusdev:
**Problem**

From discord https://discord.com/channels/899851952891002890/1287445518590677085/1287445518590677085

With this minimal repro (below), if you click one checkbox and then click the other one, the browser console shows a panic:

```
panicked at src/main.rs:39:14:
called `Result::unwrap()` on an `Err` value: AlreadyBorrowedMut(AlreadyBorrowedMutError { borrowed_mut_at: Location { file: "src/main.rs", line: 51, col: 26 } })
```

It is complaining about this (the data.write() bit is line 51, col 26):
```rust
                onclick: move |_| {
                    data.write().insert(k, Data {
                        is_on: !is_on(),
                    });
                },
```

And the panic is from this (the data.read() bit is line 39, col 14):
```rust
    let is_on = use_memo(move ||
        data
            .read()
            .get(&k)
            .map(|data| data.is_on)
            .unwrap_or_default()
    );
 ```

The full app:
```rust
#![allow(non_snake_case)]

use std::collections::BTreeMap;

use dioxus::prelude::*;
use dioxus_logger::tracing::{info, Level};

#[derive(Clone, Routable, Debug, PartialEq)]
enum Route {
    #[route("/")]
    Home {},
}

fn main() {
    // Init logger
    dioxus_logger::init(Level::INFO).expect("failed to init logger");
    info!("starting app");
    launch(App);
}

fn App() -> Element {
    rsx! {
        Router::<Route> {}
    }
}

#[derive(Default, Debug)]
pub struct Data {
    pub is_on: bool,
}

#[component]
fn Comp(
    k: u32,
    data: Signal<BTreeMap<u32, Data>>,
) -> Element {
    let is_on = use_memo(move ||
        data
            .read()
            .get(&k)
            .map(|data| data.is_on)
            .unwrap_or_default()
    );

    rsx! {
        div {
            "{k}"

            label {
                onclick: move |_| {
                    data.write().insert(k, Data {
                        is_on: !is_on(),
                    });
                },

                input {
                    type: "checkbox",
                    checked: is_on,
                }
            }
        }
    }
}

#[component]
fn Home() -> Element {
    let data: Signal<BTreeMap<u32, Data>> = use_signal(|| [
        (1, Data {
            is_on: false,
        }),
        (2, Data {
            is_on: true,
        }),
    ].into_iter().collect());

    rsx! {
[...]

PR #2992, opened 2024-09-23
Fix memo dirty flag after unrelated writes

ealmloff:
We use a dirty flag to keep track of if memos are dirty and need to be rerun before the next async tick. However, we were only resetting that flag when the value of the memo changed. This PR fixes that issue

Fixes https://github.com/DioxusLabs/dioxus/issues/2990

### DIFF
--- a/packages/signals/tests/memo.rs
+++ b/packages/signals/tests/memo.rs
@@ -1,148 +1,172 @@
-// TODO: fix #1935
+#![allow(unused, non_upper_case_globals, non_snake_case)]
+use dioxus_core::NoOpMutations;
+use std::collections::HashMap;
+use std::rc::Rc;
 
-// #![allow(unused, non_upper_case_globals, non_snake_case)]
-// use dioxus_core::NoOpMutations;
-// use std::collections::HashMap;
-// use std::rc::Rc;
+use dioxus::html::p;
+use dioxus::prelude::*;
+use dioxus_core::ElementId;
+use dioxus_signals::*;
+use std::cell::RefCell;
 
-// use dioxus::html::p;
-// use dioxus::prelude::*;
-// use dioxus_core::ElementId;
-// use dioxus_signals::*;
-// use std::cell::RefCell;
+#[test]
+fn memos_rerun() {
+    let _ = simple_logger::SimpleLogger::new().init();
 
-// #[test]
-// fn memos_rerun() {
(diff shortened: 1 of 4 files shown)


rogusdev:
Can confirm that this fixes the issue in my actual project, thank you!

packages/signals/tests/memo.rs
```diff
@@ -1,148 +1,178 @@
...
-//         let current_counter = counter.borrow();
-//         assert_eq!(current_counter.component, 2);
-//         assert_eq!(current_counter.memo, 3);
-//     }
-// }
+#![allow(unused, non_upper_case_globals, non_snake_case)]
+use dioxus_core::NoOpMutations;
+use std::collections::HashMap;
+use std::rc::Rc;
+
```
matthunz, line 5:
```suggestion
use std::sync::atomic::{AtomicBool, Ordering};
```

packages/signals/tests/memo.rs
```diff
@@ -1,148 +1,178 @@
...
+        let current_counter = counter.borrow();
+        assert_eq!(current_counter.component, 2);
+        assert_eq!(current_counter.memo, 3);
+    }
+}
+
+// Regression test for https://github.com/DioxusLabs/dioxus/issues/2990
+#[test]
+fn memos_sync_rerun_after_unrelated_write() {
+    static PASSED: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
```
matthunz, line 151:
```suggestion
    static PASSED: AtomicBool = AtomicBool::new(false);
```

packages/signals/tests/memo.rs
```diff
@@ -1,148 +1,178 @@
...
+
+        rsx! {
+            div {}
+        }
+    });
+
+    dom.rebuild_in_place();
+    dom.mark_dirty(ScopeId::APP);
+    dom.render_immediate(&mut NoOpMutations);
+    assert!(PASSED.load(std::sync::atomic::Ordering::SeqCst));
```
matthunz, line 177:
```suggestion
    assert!(PASSED.load(Ordering::SeqCst));
```

matthunz (approved):

packages/signals/tests/memo.rs
matthunz: Just my personal preference feel free to merge without it 😄 

Looks like solid code!

PR #2995, opened 2024-09-23
remove `value` attr from HTML select tag

chungwong:
`value` was added in [c961bf0779](https://github.com/DioxusLabs/dioxus/commit/c961bf0779#diff-e62d9776cc79c4b0ce740fbc99604b53f028061d209aa30ef8990c9a0ed7f441R1073). It is not a valid (attribute)[https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attributes] to the select tag.

The attribute doesn't do anything but only cause confusion to users as the only correct way to set a value natively is via the selected attr in option
```html
<select>
    <option value="1">Something</option>
    <option value="2" selected="selected">Something else</option>
</select>
```

### DIFF
--- a/packages/html/src/elements.rs
+++ b/packages/html/src/elements.rs
@@ -1553,8 +1553,6 @@ builder_constructors! {
     /// [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)
     /// element.
     select None {
-        // defined below
-        // value: String,
         autocomplete: String DEFAULT,
         autofocus: Bool DEFAULT,
         disabled: Bool DEFAULT,
@@ -1563,7 +1561,6 @@ builder_constructors! {
         name: Id DEFAULT,
         required: Bool DEFAULT,
         size: usize DEFAULT,
-        value: String volatile,
     };
 
     /// Build a


ealmloff:
`value` is not an attribute on the select element, but it is a property. In dioxus, the `value` attribute sets the `value` property for controlled inputs. Here is an example of the `value` attribute on select element:
```rust
use dioxus::prelude::*;

fn main() {
    launch(App);
}

#[component]
fn App() -> Element {
    let value = 2;
    rsx! {
        select {
            value,
            option { value: "1", "something" }
            option { value: "2", "something else" }
        }
    }
}
```

Screenshot 2024-09-23 at 8 15 33 PM

chungwong:
How does `value` work for multi select? Last time I tried it didn't work.

ealmloff:
> How does `value` work for multi select? Last time I tried it didn't work.

I don't think the javascript `value` property works with `multiple="true"`. You need to set each value on the options manually. Dioxus could accept an array and apply each value in the array to the select element, but today it just forwards the value to the javascript property

chungwong:
> but today it just forwards the value to the javascript property
😮 This is new to me, I thought it was purely about HTML only.

Thanks for the review and I am closing this now.

PR #3014, opened 2024-09-29
feat: support for sync global signals

JustSimplyKyle:


JustSimplyKyle:
Good point, after re-evaluating my code, I indeed can refactor my code to not use external runtime threads.

PR #3040, opened 2024-10-07
Use &Event for Vdom::handle_event

nicoburns:
This removes the need for an `Rc` and allows code calling `handle_event` to access the `prevent_default` and `stop_propagation` status without a clone


nicoburns:
C

PR #3046, opened 2024-10-09
Fix: Windows File Drag-N-Drop

### LINKED ISSUE #2954: Dropping files (`ondrop`) doesn't work on Windows
**Problem**

Dropping files (`ondrop`) doesn't work on Windows, the even doesn't seem to get triggered.

**Steps To Reproduce**

- `cargo r --example file_upload`
- Click on "Choose Files" (the file input) and select a .txt file -> That works fine but:
- Now click "Clear Files" and drag the same .txt file from Windows Explorer and drop it onto the "Drop files here" area
- Nothing happens!

**Expected behavior**

It should load the file just like when opening it by clicking on the file input and selecting it through the file dialog.

**Screenshots**

image

**Environment:**
 - Dioxus version:  latest `main` branch, commit `b20db13e84227b76741e007b2763adc8237ada48`
 - Rust version:    `rustc 1.82.0-nightly (8b3870784 2024-08-07)`
 - OS info:         Windows 11
 - App platform:    desktop

DogeDark:
This fixes file drop handling on Windows. I do not believe this fixes every drag event, but it is enough to get file drag-n-drop working. 

#### The Problem
Windows webview fails to provide HTML-native drag events when a wry drag & drop handler is provided. We also can't remove the wry drag & drop handler because the webview does not provide file data to the native HTML events.

#### The Solution
Use the events given by the wry drag & drop handler to mimic HTML-native events using eval.

Closes #2954


packages/desktop/src/webview.rs
```diff
@@ -232,10 +233,93 @@ impl WebviewInstance {
...
                 file_hover.set(evt);
+
+                // Windows webview blocks HTML-native events when the drop handler is provided.
+                // The problem is that the HTML-native events don't provide the file, so we need this.
+                // Solution: this glue code to mimic drag drop events.
+                #[cfg(windows)]
+                {
+                    use dioxus_html::prelude::eval;
+                    file_hover.set(evt.clone());
+
```
jkelleyrtp, line 253: Instead of wrapping the vdom in a RefCell, can we just make a new message type on WindowUserEvent that executes this with an &mut access to dom? AFAIK `eval` is not sync anyways

jkelleyrtp (approved):

PR #3047, opened 2024-10-10
Remove dependency on atty

tpoliaw:
It is no longer maintained and has been replaced by functionality in the
stdlib (std::io::IsTerminal).

Removes two warnings from cargo-audit for being unmaintained and being
[potentially unsound](https://rustsec.org/advisories/RUSTSEC-2021-0145.html).


ealmloff (approved):
Thanks!

Issue #1657, opened 2023-11-22
Dioxus Router’s `on_update` callback is not called when using browser’s back/forward buttons


iggyzuk:
**Problem**

Router's `on_update` callback is not called when going back or forward using the browser's buttons, it only works when you click on links.

**Steps To Reproduce**

- Create a project using the code below
- In terminal `dx serve`
- In browser go to `localhost:8080`
- Open developer tools > console
- Click on a link – notice the message works: "--- route update ---"
- Go back using the browser key – notice no message now
- Go forward using the browser key – notice no message either

**Expected behavior**

`on_update` should be called when clicking on links and using the browser's back and forward buttons.

**Environment:**
 - Dioxus version: [0.4.0]
 - Dx serve version: [0.4.1]
 - Rust version: [1.73.0]
 - OS info: [MacOS]
 - App platform: [`web`]

**Code**

Use the [example code](https://dioxuslabs.com/learn/0.4/router/example/full-code) and add a log in the `RouterConfig::on_update`.

```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;
use dioxus_router::prelude::*;

// ANCHOR: router
#[derive(Routable, Clone)]
#[rustfmt::skip]
enum Route {
    #[layout(NavBar)]
        #[route("/")]
        Home {},
        #[nest("/blog")]
            #[layout(Blog)]
                #[route("/")]
                BlogList {},
                #[route("/post/:name")]
                BlogPost { name: String },
            #[end_layout]
        #[end_nest]
    #[end_layout]
    #[nest("/myblog")]
        #[redirect("/", || Route::BlogList {})]
        #[redirect("/:name", |name: String| Route::BlogPost { name })]
    #[end_nest]
    #[route("/:..route")]
    PageNotFound {
        route: Vec<String>,
    },
}
// ANCHOR_END: router

pub fn App(cx: Scope) -> Element {
    render! {
        Router::<Route> {
            config: move || RouterConfig::default()
            .on_update(move |_| {
                log::info!("--- route update ---");
                None
            })
        }
    }
}

#[inline_props]
fn NavBar(cx: Scope) -> Element {
    render! {
        nav {
            ul {
                li { Link { to: Route::Home {}, "Home" } }
                li { Link { to: Route::BlogList {}, "Blog" } }
            }
        }
        Outlet::<Route> {}
    }
}

#[inline_props]
fn Home(cx: Scope) -> Element {
    render! {
[...]

PR #3057, opened 2024-10-14
Specialize `HistoryProvider::on_update` for web

matthunz:
Removes the need for `Send + Sync` bounds by specializing the `HistoryProvider::on_update` for `!Send` `!Sync`.

Fixes https://github.com/DioxusLabs/dioxus/issues/1657 by triggering `switch_route` on the `popstate` event.

```rs
/// Callback for history updates.
pub enum HistoryCallback {
    /// Local callback.
    ///
    /// This callback is only used for `dioxus-web`.
    Local(Rc<dyn Fn()>),
    /// Shared callback.
    ///
    /// This callback is used for all other Dioxus platforms to enable multi-threaded callbacks.
    Shared(Arc<dyn Fn() + Send + Sync>),
}
```

### DIFF
--- a/packages/history/src/lib.rs
+++ b/packages/history/src/lib.rs
@@ -20,6 +20,28 @@ pub fn provide_history_context(history: Rc<dyn History>) {
     provide_context(history);
 }
 
+/// Callback for history updates.
+pub enum HistoryCallback {
+    /// Local callback.
+    ///
+    /// This callback is only used for `dioxus-web`.
+    Local(Rc<dyn Fn()>),
+    /// Shared callback.
+    ///
+    /// This callback is used for all other Dioxus platforms to enable multi-threaded callbacks.
+    Shared(Arc<dyn Fn() + Send + Sync>),
+}
+
+impl HistoryCallback {
+    /// Run this callback.
+    pub fn run(&self) {
+        match self {
+            Self::Local(callback) => callback(),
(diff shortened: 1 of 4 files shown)


ealmloff:
> #[cfg(feature = "web")]
> fn updater(&mut self, callback: Rc<dyn Fn()>) { ... }
> 
> #[cfg(not(feature = "web"))]
> fn updater(&mut self, callback: Arc<dyn Fn() + Send + Sync>) { ... }

Specializing based on features is generally problematic because it makes the features nonn-additive. If a library activates the "web" feature on the router and I add that library as a dependency, then my code could break if I do something like this:
```rust
let updater = updater();
std::thread::spawn(|| updater);
```

We already have this issue in a couple different places in the dioxus crates, but we have been slowly working to remove them. For example, the default history for the router is non-additive, but will be removed in #400.

Would it be possible to implement the fix without changing the signature of `updater` based on feature flags?

jkelleyrtp:
I think now that the router has moved to be provided by each platform, we can simply implement this in the web crate with 0 configs.

matthunz:
Thanks! After chatting on Discord I totally get why we don't want mututally-exlcusive features.

I have a new version here using an enum to keep the public API the same. There is a hidden `#[cfg(target_arch = "wasm32"]` though that I hope makes sense in this context 🤔

PR #3059, opened 2024-10-14
bump manganis to alpha.3

jkelleyrtp:


jkelleyrtp:
Adds in FolderAsset and the fix that drops the build.rs checking config flags

PR #3093, opened 2024-10-24
WGPU using child windows

vyeevani:
This is a minor change from the code perspective to allow child windows to be created and some fixes for using the transparency feature. However, the implications of this are pretty staggering as they allow for the use of other rendering toolboxes.

Tested on macOS.


examples/wgpu.rs
```diff
@@ -0,0 +1,233 @@
...
+    desktop: DesktopContext,
+    #[borrows(desktop)]
+    #[not_covariant]
+    resources: GraphicsResources<'this>,
+}
+
+fn app() -> Element {
+    let mut graphics_resources = use_resource(move || async {
+        let context = GraphicsContextAsyncBuilder {
+            desktop: use_window(),
```
ealmloff, line 175: You can't use hooks here because this will not run in the same order every time the component runs. This should be `window()`

vyeevani: I'm confused why would this not be run in the same order every time? I checked and shifting the use_window outside the use_resource wouldn't break anything but I'm just trying to wrap my head around dioxus and it's reactive model as well.

vyeevani: nvm I just realized you can't use reactive stuff inside a use_resource given its execution is outside the normal flow

vyeevani:
Slightly more interesting example of how this could play out: https://github.com/vyeevani/dioxus_native_graphics

vyeevani:
https://github.com/user-attachments/assets/082ef776-dad2-4872-8e1c-b230574ec234

examples/wgpu.rs
```diff
@@ -0,0 +1,227 @@
...
+
+async fn setup_triangle<'a>(
+    context: &'a DesktopContext,
+) -> (
+    wgpu::Surface<'a>,
+    wgpu::Device,
+    wgpu::RenderPipeline,
+    wgpu::Queue,
+) {
+    let window = &context.window;
```
jkelleyrtp, line 26: Just return the `GraphicsResources` struct directly here.

examples/wgpu.rs
```diff
@@ -0,0 +1,227 @@
...
+        rpass.set_pipeline(&render_pipeline);
+        rpass.draw(0..3, 0..1);
+    }
+
+    queue.submit(Some(encoder.finish()));
+    frame.present();
+}
+
+struct GraphicsResources<'a> {
+    surface: wgpu::Surface<'a>,
```
jkelleyrtp, line 158: I'm not too familiar with wgpu but DesktopContext is `Clone`, so do we even need the lifetime?

examples/wgpu.rs
```diff
@@ -0,0 +1,227 @@
...
+                })
+            },
+        }
+        .build()
+        .await
+    });
+
+    let desktop_context = use_window();
+    graphics_resources.read().as_ref().map(|_| {
+        desktop_context.window.request_redraw();
```
jkelleyrtp, line 194: Should we be doing this on every render? Maybe a use_effect or whenever the contents change (in say, the wry_event_handler)

packages/desktop/src/app.rs
```diff
@@ -192,6 +194,25 @@ impl App {
         }
     }
 
+    pub fn resize_window(&self, size: PhysicalSize<u32>) {
```
jkelleyrtp, line 197: Isn't this implicit with resizing? Or does that not work with child windows?

Cargo.toml
```diff
@@ -240,6 +240,8 @@ rand = { version = "0.8.4", features = ["small_rng"] }
 form_urlencoded = "1.2.0"
 async-std = "1.12.0"
 web-time = "1.1.0"
+wgpu = "0.19"
+ouroboros = "*"
```
jkelleyrtp, line 244: can we move this to [dependencies] with an `optional=true` flag like `http` and then add `wgpu` to the required features list?

jkelleyrtp (changes requested):
Thanks for the PR!

A few things:
- this wgpu version is out of date (0.9 vs 0.12)
- Are we sure we need the resizing code?
- I think we don't need the lifetimes in the wgpu code (or add a 'static lifetime)
- Can we gate `wgpu` behind a `wgpu` feature on dioxus-examples so we don't end up compiling it?
- cargo.lock needs to be resolved for merge conflicts.

jkelleyrtp:
I struggled to get merge conflicts fixed for this PR since the github CLI was checking out the wrong branch. I ended up fixing things in #3257 and adding you as a co-author so you get commit credits :)

Thanks for the PR!

vyeevani:
Awesome! Sorry I hadn't been very active on this, got pulled into some other stuff. thank you for seeing it through :)

PR #3098, opened 2024-10-25
Dioxus Playground Hot Reload Support

DogeDark:
This adds hot reload support to the web renderer through the `window.postMessage` system for `dioxus-playground`. The `devtools-playground` feature enables this system. 

Required by https://github.com/DioxusLabs/playground/pull/13


jkelleyrtp:
I think `dioxus-devtools` should simply expose the hotreload message pipe by default without the additional feature. I do not want to add any more features or `[cfg(feature = "xyz")]`. Switching over to `devtools` from `hotreload` should open the "semantic door" to more changes that turn on bells and whistles that get disabled in release.

packages/web/src/devtools.rs
```diff
@@ -237,3 +242,26 @@ pub(crate) fn invalidate_browser_asset_cache() {
...
     }
 }
+
+/// Initialize required devtools for dioxus-playground.
+///
+/// This listens for window message events from other Windows (such as window.top when this is running in an iframe).
+#[cfg(feature = "devtools-playground")]
+fn playground(tx: UnboundedSender<HotReloadMsg>) {
+    let window = web_sys::window().unwrap();
+    window.set_onmessage(Some(
```
jkelleyrtp, line 252: we should also call the original onmessage if there was one

DogeDark: Do you mean to check if there is already a `onmessage` callback registered and to call it from inside our callback?

jkelleyrtp (approved):
Nice and small, thanks! :)

PR #3106, opened 2024-10-26
Fix launch path for the desktop platform

DasBlackfur:
Just a simple fix with the version number bumped to v0.5.8
Would be cool if this could get onto crates.io so theres a stable version that actually works with the desktop preset.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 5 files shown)


DasBlackfur:
Will push another fix for windows in the bundle command.

jkelleyrtp:
We probably don't want the Cargo.lock changes nor the version bump in this PR - I'll play with this and then cut those out.

PR #3078, opened 2024-10-20
fix fullstack build bug when --release

hackartists:
* rust 1.82.0 doesn't allow custom profile when release build
* Below command always cause panic from `cargo` due to using profile with release
```
dx build -p platform --release
```

PR #3115, opened 2024-10-27
Fix: CLI Release & Profile Conflict

DogeDark:
Fixes the conflict of CLI providing `--release` and `--profile something` to Cargo, which isn't allowed. This fix makes `--release` have priority.

Re-implements functionality of #3078 since it was accidentally overwritten.


matthunz (approved):
Looks solid 👍

Issue #3119, opened 2024-10-27
Panic when accessing current scope from `use_muda_event_handler`


matthunz:
**Problem**

`use_muda_event_handler` panics with a `RuntimeError` when accessing the current scope while handling events. The culprit seems to be `Runtime::with_current_scope`.

**Steps To Reproduce**

```rs
use dioxus::prelude::*;
use dioxus_desktop::use_muda_event_handler;

fn main() {
    dioxus::launch(app);
}

fn app() -> Element {
    use_muda_event_handler(|_event| {
        dbg!("A");

        spawn(async move {
            dbg!("B");
        });
    });

    rsx! {}
}
```

**Expected behavior**

In the latest alpha, a `RuntimeGuard` is created before running the event handler, so I would expect `dioxus::spawn` to run as usual.

**Screenshots**

```
thread 'main' panicked at /Users/matt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-core-0.6.0-alpha.3/src/global_context.rs:146:53:
called `Result::unwrap()` on an `Err` value: RuntimeError
```

**Environment:**
 - Dioxus version:  `main`
 - Rust version:   `nightly`
 - App platform:   Desktop

**Questionnaire**

I'm interested in fixing this myself but don't know where to start

PR #3120, opened 2024-10-28
Capture current scope ID in `use_muda_event_handler`

matthunz:
Captures the current scope during  `use_muda_event_handler` and later enters it with `Runtime::on_scope` to allow for functions like `dioxus::spawn` in the event handler.

Fixes https://github.com/DioxusLabs/dioxus/issues/3119

### DIFF
--- a/packages/desktop/src/hooks.rs
+++ b/packages/desktop/src/hooks.rs
@@ -45,14 +45,18 @@ pub fn use_wry_event_handler(
 pub fn use_muda_event_handler(
     mut handler: impl FnMut(&muda::MenuEvent) + 'static,
 ) -> WryEventHandler {
-    // move the runtime into the event handler closure
+    use dioxus_core::prelude::current_scope_id;
+
+    // Capture the current runtime and scope ID.
     let runtime = Runtime::current().unwrap();
+    let scope_id = current_scope_id().unwrap();
 
     use_wry_event_handler(move |event, _| {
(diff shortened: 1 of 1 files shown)


packages/desktop/src/hooks.rs
```diff
@@ -45,14 +45,18 @@ pub fn use_wry_event_handler(
...
     mut handler: impl FnMut(&muda::MenuEvent) + 'static,
 ) -> WryEventHandler {
-    // move the runtime into the event handler closure
+    use dioxus_core::prelude::current_scope_id;
+
+    // Capture the current runtime and scope ID.
     let runtime = Runtime::current().unwrap();
+    let scope_id = current_scope_id().unwrap();
 
     use_wry_event_handler(move |event, _| {
```
ealmloff, line 54: Instead of patching use_muda_event_handler, can you apply this patch to use_wry_event_handler so that both carry the scope with them. If the use_muda_event_handler closure has the runtime and scope, then use_wry_event_handler closure should inherit those fixes

matthunz: Sounds good 👍 that does feel better to keep the same behavior in `use_wry_event_handler`

Looks to be working in https://github.com/DioxusLabs/dioxus/pull/3120/commits/e44041a20b6f769059212ae5d7fcc923f027f840

ealmloff (approved):

Issue #2138, opened 2024-03-25
Re-add SystemTray from Tauri


jkelleyrtp:
## Feature Request

Tauri made a refactor that pulled out muda, globalhotkey, and system tray into their own crates, which meant we lost that builtin functionality. We've added back muda and globalhotkey, but not tray yet.

Refactor:
https://github.com/tauri-apps/tao/pull/778/

The crate:
https://github.com/tauri-apps/tray-icon/

We should add tray back into the desktop renderer so users can build system trays with dioxus.

cosmincatalin:
Would being able to close the main window but keeping the app running be part of this as well?

emadbaqeri:
> Would being able to close the main window but keeping the app running be part of this as well?

Yeah, I think we should register a handler for `app_request_close` or something like this and run the app in the background. I think it works the way that `tauri` handles. I think having a tray icon without app running in the background does not have a point.

GetPsyched:
Has there been any update on this since? I would love to have this feature!

Issue #3113, opened 2024-10-27
WindowCloseBehaviour::LastWindowHides


Klemen2:
**Problem**

When setting WindowCloseBehavior to LastWindowHides the window stays visible, only the content of the window disappears.

**Steps To Reproduce**

- Add the following config to the application
```
Config::new().with_close_behaviour(WindowCloseBehaviour::LastWindowHides)
```
- Close the app

**Expected behavior**

Window actually hides

**Screenshots**

- Before
![image](https://github.com/user-attachments/assets/1e3e1c24-3c19-42d7-80cf-055ab95ca0a7)
- After
![image](https://github.com/user-attachments/assets/6278a918-070b-4ad7-a894-9f5a69c67110)

**Environment:**
 - Dioxus version:  0.6.0-alpha.3
 - Rust version:  1.82.0
 - OS info:   Windows 11
 - App platform:  desktop

PR #3123, opened 2024-10-28
SystemTray

Klemen2:
Closes: #2138 

- Initialize a tray icon with
`init_tray_icon(default_tray_icon(), None);`

Default behaviour:
- left click to show the app
- right click to show context menu

Unfortunatly due to #3113 you have to use `use_window().set_visible(false);` to actually hide the app at least on windows 11


packages/desktop/src/app.rs
```diff
@@ -134,6 +138,29 @@ impl App {
             _ => (),
         }
     }
+    #[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
+    pub fn handle_tray_menu_event(&mut self, event: tray_icon::menu::MenuEvent) {
```
jkelleyrtp, line 142: What's happening here? Should we be firing any handlers?

Klemen2: The thing is it could be the same as https://github.com/DioxusLabs/dioxus/blob/733d16236986d778c2951defcb939a9f97ac1147/packages/desktop/src/app.rs#L115-L140 
but I don't know how useful that would be to open dev tools and make the window float on top from the tray instead of the menu so if there is some event that would make sense to be in the tray let me know, otherwise I can remove that.

jkelleyrtp (approved):

jkelleyrtp:
Thanks!

PR #3137, opened 2024-10-28
use DragEvent not MouseEvent for drag events

### LINKED ISSUE #3133: Drag + drop events downcast to MouseEvent, when they should go to web_sys::DragEvent
**Problem**

Drag events in dioxus currently downcast to web_sys::MouseEvent, when they should go to web_sys::DragEvent, per https://docs.rs/dioxus-web/0.6.0-alpha.3/dioxus_web/trait.WebEventExt.html#associatedtype.WebEvent-4 according to @ealmloff 

See all the drag* + drop events in https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/drag_event -- they all take DragEvent

**Steps To Reproduce**

See my repro code and explanation in discord thread: https://discord.com/channels/899851952891002890/1300590718766813344/1300591028499255306

**Environment:**
 - Dioxus version: main
 - App platform: web

**Questionnaire**

I'm interested in fixing this myself but don't know where to start

rogusdev:
Resolves #3133 maybe?

### DIFF
--- a/packages/web/src/events/drag.rs
+++ b/packages/web/src/events/drag.rs
@@ -63,13 +63,13 @@ impl PointerInteraction for Synthetic<DragEvent> {
 
 impl HasMouseData for Synthetic<DragEvent> {
     fn as_any(&self) -> &dyn std::any::Any {
-        self
+        &self.event
     }
 }
 
 impl HasDragData for Synthetic<DragEvent> {
     fn as_any(&self) -> &dyn std::any::Any {
-        self
+        &self.event
     }
 }
 
--- a/packages/web/src/events/drag.rs
+++ b/packages/web/src/events/drag.rs
@@ -4,71 +4,78 @@ use dioxus_html::{
     },
     HasDragData, HasFileData, HasMouseData,
 };
-use web_sys::MouseEvent;
+use wasm_bindgen::JsCast;
+use web_sys::{DragEvent, MouseEvent};
 
 use super::{Synthetic, WebEventExt};
 
 pub(crate) struct WebDragData {
-    raw: Synthetic<MouseEvent>,
+    drag: Synthetic<DragEvent>,
+    mouse: Synthetic<MouseEvent>,
 }
 
 impl WebDragData {
-    pub fn new(raw: MouseEvent) -> Self {
+    pub fn new(raw: DragEvent) -> Self {
+        let mouse = raw
+            .clone()
+            .dyn_into::<MouseEvent>()
+            .expect("Inconceivable! DragEvent is not MouseEvent?");
         Self {
-            raw: Synthetic::new(raw),
+            drag: Synthetic::new(raw),
+            mouse: Synthetic::new(mouse),
         }
     }
 }
 
 impl HasDragData for WebDragData {
     fn as_any(&self) -> &dyn std::any::Any {
-        &self.raw as &dyn std::any::Any
+        &self.drag as &dyn std::any::Any
     }
 }
 
 impl HasMouseData for WebDragData {
     fn as_any(&self) -> &dyn std::any::Any {
-        &self.raw as &dyn std::any::Any
+        &self.mouse as &dyn std::any::Any
     }
 }
 
 impl PointerInteraction for WebDragData {
     fn trigger_button(&self) -> Option<dioxus_html::input_data::MouseButton> {
-        self.raw.trigger_button()
+        self.mouse.trigger_button()
     }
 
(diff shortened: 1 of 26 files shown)


rogusdev:
Needs to convert the DragEvent into MouseEvent for some of the functions now

matthunz:
Looks great! 👍 I just edited this to match the other web events that aren't using a wrapper struct (like `WebDragData`) and simplify because `web_sys::DragEvent` derefs to a `web_sys::MouseEvent`

matthunz (approved):

rogusdev:
@matthunz sorry I did not see your changes before I force pushed my updates, because I found an issue. Please re-apply your updates you had done before, or just recommend them and I can add them.

@ealmloff your changes in https://github.com/DioxusLabs/dioxus/commit/cda8bb24f103c5de91f2859b40c884d8db73023c#diff-517df2174967af5678c41e224b2d9a1f1e63e7fcdfe46bc4108c14bdf4defff2R25 broke `downcast`, I suspect for all event types. This is because `self.raw` is no longer a `web_sys::*Event`, instead they are all `Synthetic<web_sys::*Event>` and Synthetic is crate visibility only. So the `as_any` called by `downcast` needs to change to return the actual web_sys event, and not the Synthetic wrapper, otherwise outside code cannot access the wrapped web_sys events.

Given the last bit, I am not sure if this needs a larger set of changes to all the events, or if this is fine for now, or what. Please advise.

packages/web/src/events/drag.rs
```diff
@@ -4,71 +4,78 @@ use dioxus_html::{
...
     fn as_any(&self) -> &dyn std::any::Any {
-        &self.raw as &dyn std::any::Any
+        &self.drag.event as &dyn std::any::Any
     }
 }
 
 impl HasMouseData for WebDragData {
     fn as_any(&self) -> &dyn std::any::Any {
-        &self.raw as &dyn std::any::Any
+        &self.mouse.event as &dyn std::any::Any
```
rogusdev, line 38: Same idea here for the wrapper, and accessing the wrapped web_sys event

ealmloff:
> @ealmloff your changes in https://github.com/DioxusLabs/dioxus/commit/cda8bb24f103c5de91f2859b40c884d8db73023c#diff-517df2174967af5678c41e224b2d9a1f1e63e7fcdfe46bc4108c14bdf4defff2R25 broke downcast, I suspect for all event types. This is because self.raw is no longer a web_sys::*Event, instead they are all Synthetic<web_sys::*Event> and Synthetic is crate visibility only. So the as_any called by downcast needs to change to return the actual web_sys event, and not the Synthetic wrapper, otherwise outside code cannot access the wrapped web_sys events.
> 
> Given the last bit, I am not sure if this needs a larger set of changes to all the events, or if this is fine for now, or what. Please advise.

Yes, applying that change to all events would be great. Many people will be downcasting to the web-sys counterpart from previous releases, so if we can not change the type it downcasts to that is better

matthunz:
Oh my fault I didn't mean to get in the way @rogusdev I just re-applied those changes I hope that's all good.

I also added your fix for downcasting to all other events

rogusdev:
@matthunz thanks! Your cleanup looks promising :) that said, I see a failing ci job now. Any idea what that is about?

rogusdev:
@matthunz I re-added the web_sys event reference for the drag.rs as_any-s because those got lost with the re-applying of the old commit

rogusdev:
I also notice that web/events/mouse has these tidbits:

```rust
impl HasFileData for Synthetic<MouseEvent> {}

impl HasDragData for Synthetic<MouseEvent> {
    fn as_any(&self) -> &dyn std::any::Any {
        &self.event
    }
}
```

Perhaps these were just because of the old dragevent implementation and we can get rid of these now?...

matthunz:
> @matthunz I re-added the web_sys event reference for the drag.rs as_any-s because those got lost with the re-applying of the old commit

Ah man I totally missed that thanks!

matthunz:
> @matthunz thanks! Your cleanup looks ~promising~ like a significant improvement! :) that said, I see a failing ci job now. Any idea what that is about?

Awesome! (: it looks like we might have a CI issue - looks like some links in various READMEs aren’t loading correctly for the runner

rogusdev:
Do you want to approve for me to merge as-is and you fix the readmes separately?

ealmloff (approved):

ealmloff:
Thanks, this looks much better. The readme links should be fixed automatically when the docsite is back to normal

PR #3139, opened 2024-10-29
feat: move ssg to CLI to fix docsite

jkelleyrtp:
Remove ssg target in favor of a flag on the CLI.

Keeps our target list small and ensures we always use the proper fullstack setup for SSG.


jkelleyrtp:
This PR contains a much larger cleanup/refactor than I want to do now for the 0.6 release.

However, I think it's good to compactify/cleanup the fullstack crate long term, so we should revisit this for 0.7

Issue #3141, opened 2024-10-29
Crashing upon file changes with the mega commit


chungwong:
**Problem**

With the latest main which is up to 5f34822a66aacf13f3d7d221b93283d273f61ed6, I believe it was introduced by https://github.com/DioxusLabs/dioxus/commit/7ec3453ca36347a22c15634ac60eb6c4127fd57a
```
dx serve --example hello_world
```
Now make a change to examples/hello_world.rs
```diff
 fn app() -> Element {
+    "";
     rsx! {
         div { "Hello, world!" }
     }
```
It crashes at the second you save the change. 
Note: I made a few changes for logging and the line number is affected. Originally it panics at [line 191](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/serve/runner.rs#L191)
```
[packages/cli/src/serve/runner.rs:194:17] &rust_file = "/home/chung/repos/dioxus/examples/hello_world.rs"
thread 'main' panicked at packages/cli/src/serve/runner.rs:196:16:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Originally I thought there must be some error in the path but it is not the case, the path is valid.
```
❯ ls -l /home/chung/repos/dioxus/examples/hello_world.rs
-rw-r--r-- 1 chung chung 809 Oct 30 10:07 /home/chung/repos/dioxus/examples/hello_world.rs
```

jkelleyrtp:
Keeping this open until a fix is verified.

chungwong:
The above error *No such file or directory* was editing with Vim

I tested it again with gedit and the error is different
```
09:34:55 [dev] Full rebuild: /examples/hello_world.rs
09:35:03 [dev] Build failed: I/O Error: Text file busy (os error 26)

╭────────────────────────────────────────────────────────────────────────────────────────── /:more ╮
│  App:     ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           Platform: Desktop Linux                       │
│  Bundle:  ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           App features: ["desktop"]                     │
│  Status:  Failed                                   ServerFns at: no server address               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

jkelleyrtp:
Fixed by the read-blue work #3608

PR #3142, opened 2024-10-30
fix: don't panic while reading file in hotreload

jkelleyrtp:
Fixes: https://github.com/DioxusLabs/dioxus/issues/3141

I think this is because some editors might use `.bk` files but I'm not exactly sure why a file wouldn't show up.

I can't replicate this locally on my mac.


chungwong:
Just to add more context to the issue, I am using Vim. I will test it with gedit tomorrow.

PR #3150, opened 2024-10-31
Fix: Link Checker

DogeDark:
GitHub Pages `404.html` returns HTTP 404 on the new docsite site but didn't previously(?).

This drastically reduces the effectiveness of having a link checker.

### DIFF
--- a/examples/dog_app.rs
+++ b/examples/dog_app.rs
@@ -11,10 +11,11 @@ use dioxus::prelude::*;
 use std::collections::HashMap;
 
 fn main() {
-    dioxus::launch(app);
+    dioxus::launch(App);
 }
 
-fn app() -> Element {
+#[component]
+fn App() -> Element {
     // Breed is a signal that will be updated when the user clicks a breed in the list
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
Allowing a 404 response is basically turning the link checker off - we should just go fix the routes on the docsite (which are blocked due to other reasons, unfortunately)

DogeDark:
@jkelleyrtp It's not the routes that are wrong. For some reason, the new changes on the docsite have made it no longer work with the `404.html` page. The only difference between now and then that I could find is we used a custom `index.html` instead of the pre-generated one from `DX` - which would've been copied to `404.html`

jkelleyrtp:
We set our 404 page on github pages to return the same html as index.html, so technically those routes *are* returning 404s.

This is because SSR is currently disabled on the docsite while I fix ISRG in #3139

DogeDark:
Oh, I finally understand. SSR is generating all the `index.html`s for each route. I thought originally that we used the `404.html` for everything.

Issue #2526, opened 2024-06-16
Git version of dioxus fails to load collected assets using the Desktop template


JustSimplyKyle:
**Problem**
When doing `dx serve`
Assets are collected to the folder `assets`
```
.rw-r--r-- 23k kyle 16 Jun 19:45 -N header.svg
.rw-r--r-- 652 kyle 16 Jun 19:45 -N main.css
```
But spits out this error.
```
2024-06-16T11:50:11.217950Z  WARN dioxus_desktop::protocol: Assets built with manganis cannot be preloaded (failed to read "/home/kyle/coding/test-case/__assets_head.html"). This warning may occur when you build a desktop application without the dioxus CLI. If you do not use manganis, you can ignore this warning: No such file or directory (os error 2).
```
and the example comes out like this:
![image](https://github.com/DioxusLabs/dioxus/assets/68589851/1cb98bbb-87a9-4ec7-bff5-a72d566f9e9f)

**Steps To Reproduce**

-  `dx new test-case -o styling=Vanilla -o router=false --subtemplate Desktop`
-  `cd test-case`
-  `cargo add dioxus --features desktop --git https://github.com/DioxusLabs/dioxus`
-  `cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli`
-  `dx serve`

**Expected behavior**

To launch correctly with assets loading.

**Temporary fix**
- `dx build`
- `cd dist`
- `./test-case`

**Environment:**
 - Dioxus version: master, including the cli
 - Rust version: 1.78.0
 - OS info: Arch Linux 6.9.4
 - App platform: Desktop

JustSimplyKyle:
Additionally, if we compile to web instead, it works flawlessly.

JustSimplyKyle:
I can still reproduce it on my end, with the CLI and desktop being the latest git version; I have ensured that through this following command:
- `cargo update && cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli`

i123iu:
Hi, I've also had this issue and found something: 
- I believe the issue is in the function [`running_in_dev_mode`](https://github.com/DioxusLabs/dioxus/blob/main/packages/desktop/src/protocol.rs#L253) which decides by checking the `CARGO` environmental variable
```rust
fn running_in_dev_mode() -> bool {
    // If running under cargo, there's no bundle!
    // There might be a smarter/more resilient way of doing this
    std::env::var_os("CARGO").is_some()
}
```
[...]

PR #3161, opened 2024-11-02
Make wasm-bindgen parallel/disable log, speed up 2-10x

jkelleyrtp:
- wip: pull out incr linker intercept and debug=true for wasm for faster builds
- make walrus parallel


JustSimplyKyle:
It seems like this pr brought back https://github.com/DioxusLabs/dioxus/issues/2526

jkelleyrtp:
> It seems like this pr brought back https://github.com/DioxusLabs/dioxus/issues/2526

You need to make sure that your assets are actually referenced by a Link or Stylsheet component - we don't implicitly create an asset head for you anymore unless you reference the assets explicitly in the final app.

PR #3175, opened 2024-11-04
Perform a full rebuild when a hot reload is not possible

ealmloff:
The CLI currently tries to hot reload changes to your `Cargo.toml` even though it is not used as an asset in the application. It should trigger a full rebuild because it is not linked as an asset. Other build time assets like an external lalrpop grammar that drives a build script should also trigger a full rebuild

### DIFF
--- a/packages/cli/src/serve/runner.rs
+++ b/packages/cli/src/serve/runner.rs
@@ -169,6 +169,9 @@ impl AppRunner {
                     // todo(jon): don't hardcode this here
                     let asset_relative = PathBuf::from("/assets/").join(bundled_name);
                     assets.push(asset_relative);
+                } else {
+                    // This was not an asset that we could hot reload, perform a full rebuild
+                    return None;
                 }
             }
         }


jkelleyrtp:
I feel like we should special case certain files for this... I've much preferred the flow where `r` performs the manual rebuild rather than the CLI always attempting full rebuilds when it can't autodetect the reloads.

We *could* parse the `.d` files which contain information about files the rust compiler is tracking itself, or frequently poll `cargo check` or something and listen to the rebuild information.

Also, FWIW, whenever the `cargo.toml` and `dioxus.toml` files change, we're not rebuilding our "understanding of the world" so our cargo metadata is out of date.

jkelleyrtp:
I added support for full rebuilds when the cargo.toml changes in a recent pr.

I don't want us to always do full rebuilds when a file changes that we don't understand since that seems like a potentially worse DX than requiring the occasional `r.`

I'm confident we can walk the `.dep` files eventually as our source of filesystem watching as a better replacement for our too-fuzzy system that we have today.

PR #3176, opened 2024-11-04
Remove the stylesheet component

ealmloff:
This PR removes the stylesheet component. The `Style` component already renders a stylesheet `link` in the head if you pass in a link instead of inline CSS. `Stylesheet` is just a less general version of the `Style` component. The two components have a very similar name and the exact same purpose which could cause confusion

Reverts https://github.com/DioxusLabs/dioxus/pull/3138


nicoburns (approved):
Makes sense to me!

spookyvision:
unlike `Stylesheet`, `Style` does not seem to set the `rel="stylesheet"` attribute, which on my system (Firefox 133.0.3 on Mac) causes the CSS to not apply

ealmloff:
> unlike `Stylesheet`, `Style` does not seem to set the `rel="stylesheet"` attribute, which on my system (Firefox 133.0.3 on Mac) causes the CSS to not apply

This PR fixes that issue

spookyvision:
> This PR fixes that issue ([here](https://github.com/DioxusLabs/dioxus/pull/3176/files#diff-f59b79f6b42573ae44ec6f6e5cf266ee809a53e7eee177eb6ce6279a8c27d2fbR108))

Missed that, apologies!

jkelleyrtp:
I don't want to remove the Stylesheet component. Just coming from a world of html I'm not sure I would know to reach for `Style` over `Stylesheet` so if `Stylesheet` is *just* an alias to `<link rel="stylesheet">` then that's fine. 

But, we did need to fix the bug with Style not applying the `rel` attribute, so I went ahead and just brought those changes in.

PR #3179, opened 2024-11-05
Remove debug printlns

Rahul721999:
Remove debug printlns

Changes made:
- Removed leftover debug `println` statements from `protocol.rs` module.
- These are unsuitable for production environments.


ealmloff (approved):
Oops, good catch

PR #3180, opened 2024-11-05
Add support for user-provided event loops & event handlers

Aandreba:
I was trying to write an application that has a combination of Dioxus & OpenGL windows but that was made impossible by the fact that the only way to create multiple windows is by creating windows with WebViews.

Having the ability to manually provide an event loop should allow user to create as many windows as they want before the dioxus window start, and allow more flexibility.

Custom event handlers were implemented to allow the handling of events generated by the manually created windows.

### DIFF
--- a/packages/desktop/src/app.rs
+++ b/packages/desktop/src/app.rs
@@ -53,8 +53,11 @@ pub(crate) struct SharedContext {
 }
 
 impl App {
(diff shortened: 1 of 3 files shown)


jkelleyrtp (approved):
Looks good, thank you!

Issue #3080, opened 2024-10-21
with “dioxus::desktop::window().new_window()” to create a new window，The program will crash and crash.


uefva:
After clicking on the input box, the program froze and crashed. I tried using examples/popup.rs and encountered the same problem. My computer is running on Windows 11 system

```
pub fn create_topology_handle(mut address: Signal<String>) -> Coroutine<String> {
    use_coroutine(move |mut rx: UnboundedReceiver<String>| async move {
        while let Some(message) = rx.next().await {
            address.set(message);
        }
    })
}

pub fn topology_window(address: Signal<String>) {
    let tx = create_topology_handle(address).tx();
    dioxus::desktop::window().new_window(
        VirtualDom::new_with_props(
            topology_popup,
            Rc::new(move |s| match tx.unbounded_send(s) {
                Ok(_) => {}
                Err(error) => {
                    println!("{}", error);
                }
            }),
        ),
        Default::default(),
    );
}

pub fn topology_popup(send: Rc<dyn Fn(String)>) -> Element {

```

```
thread 'main' panicked at C:\Users\songy\.cargo\registry\src\rsproxy.cn-0dccff568467c15b\dioxus-core-0.6.0-alpha.3\src\diff\node.rs:70:49:
already borrowed: BorrowMutError
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library\std\src\panicking.rs:652
   1: core::panicking::panic_fmt
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library\core\src\panicking.rs:72
   2: core::cell::panic_already_borrowed
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library\core\src\cell.rs:785
   3: core::cell::RefCell<slab::Slab<dioxus_core::nodes::VNodeMount> >::borrow_mut<slab::Slab<dioxus_core::nodes::VNodeMount> >
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23\library\core\src\cell.rs:1076
   4: dioxus_core::nodes::VNode::move_mount_to
             at C:\Users\songy\.cargo\registry\src\rsproxy.cn-0dccff568467c15b\dioxus-core-0.6.0-alpha.3\src\diff\node.rs:70
[...]

PR #3182, opened 2024-11-06
Workaround:  to Prevent Crashes (#3080) during new_window creation

Rahul721999:
- create new window in a seperate thead, without interfaring the main thread or getting program crashed
- Add changes in example code snipets (multiwindow.rs, popup.rs) with new multithreaded way to create additional window
- Updated documentation to include best practices and guidance on multi-threading when using new_window() method.

### DIFF
--- a/packages/desktop/src/protocol.rs
+++ b/packages/desktop/src/protocol.rs
@@ -94,8 +94,6 @@ fn serve_asset(request: Request<Vec<u8>>) -> Result<Response<Vec<u8>>> {
         let bundle_root = get_asset_root();
         let relative_path = uri_path.strip_prefix("/").unwrap();
         uri_path = bundle_root.join(relative_path);
-        println!("bundle root: {bundle_root:?}");
-        println!("uri path: {uri_path:?}");
     }
 
     // If the asset exists, then we can serve it!


ealmloff:
`spawn` runs the future on the same thread. I'm guessing it happens to fix this issue because it waits to execute the spawned task until there is a break in rendering. We need to fix the underlying error instead of relying on that behavior in `spawn`. We could probably fix the issue without requiring the function to be used in spawn by queueing the window creation for the next `Poll` event on the window instead of running it immediately

Rahul721999:
Thanks @ealmloff !

Thank you for the suggestion! I understand now that, instead of relying on `spawn`, we should utilize an event poll system to `queue `the window creation, which would align better with the current architecture.

I'll give this approach a try. Before I dive in, could you guide me a bit on how to create or use the `poll `event in this context? If there's already an implementation of a similar `poll event` elsewhere in the codebase, could you point me to that section? It would be really helpful as a reference to implement it correctly.

Thanks again for the direction!

Rahul721999:
Hi @ealmloff,

I've been working on the suggestions you provided, and I wanted to share my findings so far:

> (if I understood correctly)

1. For events, we're currently using `tao_waker`, which is initialized during the creation of `WebViewInstance`.
2. `WebViewInstance::new()` is called initially during app launch.
3. Later, when attempting to create a `new window`, `WebViewInstance` is being invoked again, but it gets stuck within the `window.build()` method.

You suggested queuing the `create new window` event, but I'm unclear on the exact location for this in the code. I tried using `send_event(NewWindow_event)`, but it seems that this only queues the event—it doesn’t actually create a new window. To open a window, we still need to create a `WebViewInstance`, which is where the process gets stuck.

Any further guidance on where  & how to queue this event would be greatly appreciated. Thanks for your help!

Issue #3147, opened 2024-10-30
Bug with nested Link?...


rogusdev:
**Problem**

I updated to latest main, previously using start of oct. ~I have not identified which PR changed this.~ Commit that introduced this behavior is https://github.com/DioxusLabs/dioxus/commit/baf5a95cae3fc6d4277d3f0a8713b1c35cf48218 . I have a repro of behavior that used to work, and now doesn't, and I assume it's a bug -- but maybe I just need to do things differently. If it's not a bug, I would say some documentation or something is needed.

Specifically, when I have a nav bar that links to some pages, that themselves link back to each other, the URL bar gets updated, but it does not render the new view. ~No errors are logged.~ A warning is logged to the console, that seems directly related to internals of the commit linked above (read only signal, trying to set a signal value during render).

**Steps To Reproduce**

Screenshots below of the repro steps with this code (using dioxus main):

```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;
use dioxus_logger::tracing::{info, Level};

#[derive(Clone, Routable, Debug, PartialEq)]
enum Route {
    #[route("/")]
    Home {},
    #[route("/blog/:id")]
    Blog { id: i32 },
    #[route("/view/:id")]
    View { id: u32 },
}

fn main() {
    // Init logger
    dioxus_logger::init(Level::INFO).expect("failed to init logger");
    info!("starting app");
    launch(App);
}

fn App() -> Element {
    rsx! {
        Router::<Route> {}
    }
}

#[component]
fn Blog(id: i32) -> Element {
    rsx! {
        Link { to: Route::Home {}, "Go to counter" }
        "Blog post {id}"
    }
}


#[derive(Debug, PartialEq, Clone)]
pub struct Contents {
    pub id: u32,
    pub children: Vec<Contents>,
}

fn contents_for(id: u32) -> Contents {
    match id {
        1 => Contents {
            id: 1,
            children: Vec::from([
                Contents {
                    id: 2,
                    children: Vec::new()
                },
                Contents {
                    id: 3,
                    children: Vec::new()
                }
            ])
        },
        2 => Contents {
            id: 2,
            children: Vec::new()
        },
        3 => Contents {
            id: 3,
            children: Vec::new()
        },
        _ => Contents {
            id: 0,
            children: Vec::new()
        }
    }
}

#[component]
fn Home() -> Element {
    rsx! {
        Nav {
            id: None,
        }

        div {
            "Start viewing:"
        }

        div {
[...]

PR #3194, opened 2024-11-11
Fix write on render warning with read only signal props

ealmloff:
Fixes the warning in https://github.com/DioxusLabs/dioxus/issues/3147


rogusdev:
Works for me, thanks

PR #3195, opened 2024-11-11
Restore manganis optimizations

### LINKED ISSUE #3229: Manganis 0.6.0-alpha.5 using image() in asset! results in `macro expansion ignores token , and any following`
**Problem**

When attempting to use the `asset!` macro as described in documentation [here](https://github.com/DioxusLabs/dioxus/blob/af53354b28981ff84d4e533798f1f8920371ef92/packages/manganis/manganis-macro/src/asset.rs#L26-L33) an error is thrown not allowing the usage of image() to specify image format & size of the asset.

```
20:35:26 [cargo] error: macro expansion ignores token `,` and any following
  --> src/main.rs:38:34
   |
37 |       const AUTUMN_LOGO: Asset = manganis::asset!(
   |  ________________________________-
38 | |         "/assets/autumn-logo.png",
   | |                                  ^
39 | |         asset::image().format(ImageType::Avif).size(64, 64)
40 | |     );
   | |_____- caused by the macro expansion here
   |
   = note: the usage of `manganis::asset!` is likely invalid in expression context
```

Setting the type of `AUTUMN_LOGO` to `ImageAsset` results in mismatched types error, unknown if this is even the intended solution though.
```
error[E0308]: mismatched types
  --> src/main.rs:37:37
   |
37 |       const AUTUMN_LOGO: ImageAsset = manganis::asset!(
   |  _____________________________________^
38 | |         "/assets/autumn-logo.png",
39 | |         asset::image().format(ImageType::Avif).size(64, 64)
40 | |     );
   | |_____^ expected `ImageAsset`, found `Asset`
   |
   = note: this error originates in the macro `manganis::asset` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**Steps To Reproduce**

- Use asset! macro with image() to specify image format & size
```
    const AUTUMN_LOGO: Asset = manganis::asset!(
        "/assets/autumn-logo.png",
        asset::image().format(ImageType::Avif).size(64, 64)
    );
```

This [GitHub repo](https://github.com/hyziri/dioxus_manganis_bug) is a minimal example of where this bug can be replicated.

**Environment:**
 - Dioxus version:  0.6.0-alpha.5 & also occurs in 0.6.0-alpha.4
 - Rust version:    occurs in both stable & nightly
 - OS info:         Linux
 - App platform:    occurs in both web & fullstack

**Questionnaire**

I'm interested in fixing this myself but don't know where to start

ealmloff:
This PR will restore the bulk of manganis with a slightly different serialization approach. It also restores all of the documentation that was removed in #2779. Instead of parsing the builder inside of the macro and serializing that into json, this PR serializes the builder directly. This should make it easier to add new asset types to manganis and make the macro significantly more flexible.

Previously the macro would only accept literals, now it will be able to accept any const type:
```rust
const ASSET: Asset = asset!("/assets/image.png");
const OPTIONS: ImageAssetOptions = ImageAssetOptions::new().with_format(ImageFormat::Avif);
const AVIF_ASSET: Asset = asset!("/assets/image.png", OPTIONS);

```

The path itself still needs to be a literal string so we can validate and hash the contents at compile time, but the rest of the settings are just const rust

closes #3229


packages/const-serialize/README.md
```diff
@@ -0,0 +1,57 @@
...
+    };
+    if !serialize_eq(&data, &deserialized) {
+        panic!("data mismatch");
+    }
+}
+```
+
+## How it works
+
+`const-serialize` relies heavily on well defined layouts for the types you want to serialize. The serialization format is the linear sequence of unaligned bytes stored in the order of the fields, items or variants of the type. Numbers are stored in little endian order.
```
ealmloff, line 45: Highly recommend reading this before reviewing const-serialize. Const rust is very limited which makes some of the code look odd. The serialization framework ends up looking very similar to zero copy serialization frameworks like:
- postcard-forth (a reimplementation of postcard serialization by the original author): https://onevariable.com/blog/pods-from-scratch
- zerocopy: https://github.com/google/zerocopy
Unlike zero copy serialization, const serialize normalizes endianness across platforms

DogeDark:
I tried testing this on Windows 10 and I got this behavior:
- assets are optimized and the optimized assets are placed in my top-level assets folder (nothing about assets in the target folder)
- the link returned by manganis for use in `img` src/`document::Link` is weird/doesn't work: 
```
http://127.0.0.1:8080/assets///?\C:\Users\Darka\Documents\Projects\DioxusLabs\temp\web-manganis-test-3195\assets\main-e6ddccc1e64cc12d.css=
```

ealmloff:
> I tried testing this on Windows 10 and I got this behavior:
> 
> * trying to use a folder asset and it errors with access denied
> * assets are optimized and the optimized assets are placed in my top-level assets folder (nothing about assets in the target folder)
> * the link returned by manganis for use in `img` src/`document::Link` is weird/doesn't work:
> 
> ```
> http://127.0.0.1:8080/assets///?\C:\Users\Darka\Documents\Projects\DioxusLabs\temp\web-manganis-test-3195\assets\main-e6ddccc1e64cc12d.css=
> ```

All of those issues should be fixed now (tested on windows 11). There was a path separator issue in manganis on windows

packages/cli/Cargo.toml
```diff
@@ -88,6 +88,76 @@ prettyplease = { workspace = true }
...
+swc_ecma_parser = { version = "=0.149.1", default-features = false }
+swc_ecma_preset_env = { version = "=0.217.0", default-features = false, features = [
+    "serde",
+] }
+swc_ecma_transforms = { version = "=0.239.0", default-features = false }
+swc_ecma_transforms_base = { version = "=0.145.0", default-features = false }
+swc_ecma_transforms_classes = { version = "=0.134.0", default-features = false }
+swc_ecma_transforms_compat = { version = "=0.171.0", default-features = false }
+swc_ecma_transforms_macros = { version = "=0.5.5", default-features = false }
+swc_ecma_transforms_module = { version = "=0.190.0", default-features = false }
```
jkelleyrtp, line 143: it really would not be terrible to cut out the asset system (Manifest, optimizer, etc) into its own crate again... and maybe with some feature flags so we can gate `dev` mode `dx` to not have to include all this.

packages/cli/src/build/bundle.rs
```diff
@@ -377,57 +380,89 @@ impl AppBundle {
 
         let asset_dir = self.build.asset_dir();
 
-        // First, clear the asset dir
-        // todo(jon): cache the asset dir, removing old files and only copying new ones that changed since the last build
-        _ = std::fs::remove_dir_all(&asset_dir);
+        // First, clear the asset dir of any files that don't exist in the new manifest
         _ = create_dir_all(&asset_dir);
+        // Create a set of all the paths that new files will be bundled to
```
jkelleyrtp, line 385: it might be a good idea to do all this under `tokio::spawn_blocking()` now that we might be doing some serious optimizations.

jkelleyrtp (changes requested):
Small tweaks:
- might be good to pull out the assets module from the CLI into its own crate 
- some of the code we generate in the macro could be moved into a const function or two
- some docs in some places refernce the old `mg!()` or are incomplete, presumably since they're already incomplete.
- it would be nice if the serialized bytes were valid json but I get how the const_serialize can't do that in a general case.
- double check any issues with us not properly cleaning the build dirs - I want to make sure there's no weirdness with leftover html/assets/gradle files still showing up between builds leaving sticky state behind.

Overall much better!

packages/cli/src/build/bundle.rs
ealmloff: Good catch, I didn't notice that got removed when assets were removed.

packages/manganis-core/src/images.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    /// Make the asset preloaded
+    ///
+    /// Preloading an image will make the image start to load as soon as possible. This is useful for images that will be displayed soon after the page loads or images that may not be visible immediately, but should start loading sooner
+    ///
+    /// ```rust
+    /// # use manganis::{asset, Asset, ImageAssetOptions};
+    /// const _: Asset = asset!("/assets/image.png", ImageAssetOptions::new().with_preload(true));
```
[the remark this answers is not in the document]
ealmloff: ImageAsset doesn't exist. We can't support metadata about assets like lqip without injecting it in the linker, so I didn't add it in this PR because of the time constraint

packages/manganis-core/Cargo.toml
```diff
@@ -13,5 +13,12 @@ keywords = ["dom", "ui", "gui", "assets"]
 # be careful with dependencies you add here - these need to get compiled for the proc macro and therefore
```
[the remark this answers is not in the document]
ealmloff: It doesn't change the diff much but I pulled that change out into https://github.com/DioxusLabs/dioxus/pull/3245. Most of manganis wasn't in the repo at all.

packages/manganis-macro/src/asset.rs
```diff
@@ -0,0 +1,213 @@
...
+                // The source is used by the CLI to copy the asset
+                const __ASSET_SOURCE_PATH: &'static str = #asset_str;
+                // The options give the CLI info about how to process the asset
+                const __ASSET_OPTIONS: manganis::AssetOptions = #options.into_asset_options();
+                // We calculate the bundled path from the hash and any transformations done by the options
+                // This is the final path that the asset will be written to
+                const __ASSET_BUNDLED_PATH: manganis::macro_helpers::const_serialize::ConstStr = manganis::macro_helpers::generate_unique_path(__ASSET_SOURCE_PATH, __ASSET_HASH, &__ASSET_OPTIONS);
+                const __ASSET_BUNDLED_PATH_STR: &'static str = __ASSET_BUNDLED_PATH.as_str();
+                // Create the asset that the crate will use
+                const __ASSET: manganis::BundledAsset = manganis::BundledAsset::#constructor(__ASSET_SOURCE_PATH, __ASSET_BUNDLED_PATH_STR, __ASSET_OPTIONS);
```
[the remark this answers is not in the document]
ealmloff: I could inline some constants, but I think that would make things less readable the inputs/steps would be less documented.

I added some more comments about why we need each step. We are running into some limitations of const rust here. We can't return &'static str we generate in const function and into_asset_options can't easily be turned into a trait

packages/cli/Cargo.toml
ealmloff: Agreed. I would like to pull out asset optimization and resolution into a separate crate, but I don't want to block the release and it would take longer to make an more stable API. How about a follow up PR after more pressing issues like the base path are fixed? I have an issue ready once this is merged

ealmloff:
> it would be nice if the serialized bytes were valid json but I get how the const_serialize can't do that in a general case.

I think that is possible, but it would require a lot more logic to parse and format strings and numbers at compile time. I'm not sure how that would effect compile times, but it could be worth exploring in the future

I don't love having a bespoke serialization format. https://github.com/jamesmunns/postcard is a much simpler well defined serialization format that might be easier to target than json. I think it is pretty similar to what we are currently generating

DogeDark:
> All of those issues should be fixed now (tested on windows 11). There was a path separator issue in manganis on windows

I still have all the same issues on the latest commit except the folder asset now panics as todo during the CLI's asset stage

ealmloff:
> * assets are optimized and the optimized assets are placed in my top-level assets folder (nothing about assets in the target folder)
> * the link returned by manganis for use in `img` src/`document::Link` is weird/doesn't work:

Can you share the code you used that causes those issues with the latest version? This code works as I would expect on windows:
```rust
use dioxus::prelude::*;

fn main() {
    launch(|| {
        let asset = asset!("/assets/img.png");
        rsx! {
            img {
                src: "{asset}"
            }
        }
    })
}
```

DogeDark:
@ealmloff https://github.com/DogeDark/manganis-3195

ealmloff:
> @ealmloff https://github.com/DogeDark/manganis-3195

Thanks, that example is working now as well. I was testing windows on desktop where path separator in the CLI is the same as the target platform path separator. WASM has unix path separators so it was stripping the wrong path when cross building to web

DogeDark:
> Thanks, that example is working now as well.

I can confirm, it works!

jkelleyrtp (approved):

PR #3198, opened 2024-11-12
Fix: CLI Fullstack Executable File Extension

DogeDark:
CLI did not add `.exe` extension to the `server` file copied to the target-dist folder.


jkelleyrtp:
thank you!

PR #3215, opened 2024-11-14
Feat: Android Dynamic Arch Support

DogeDark:
Adds support for specifying an Android target arch with the `--arch` flag. If no `--arch` flag is included, the CLI will attempt to automatically detect the arch using `adb` and if that fails, the arch defaults to Arm64.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -114,10 +114,16 @@ impl BuildRequest {
         // We don't want to overwrite the user's .cargo/config.toml since that gets committed to git
         // and we want everyone's install to be the same.
         if self.build.platform() == Platform::Android {
-            let linker = self
+            let ndk = self
                 .krate
-                .android_linker()
+                .android_ndk()
                 .context("Could not autodetect android linker")?;
+            let linker = self
+                .build
+                .target_args
+                .arch
+                .unwrap_or_default()
+                .android_linker(&ndk);
 
             tracing::trace!("Using android linker: {linker:?}");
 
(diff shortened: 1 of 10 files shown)


packages/cli/src/build/request.rs
```diff
@@ -536,7 +548,13 @@ impl BuildRequest {
                 .join("src")
                 .join("main")
                 .join("jniLibs")
-                .join("arm64-v8a"),
+                .join(
+                    self.build
+                        .target_args
+                        .arch
+                        .unwrap_or_default()
```
jkelleyrtp, line 555: I would replace `.arch` with `.arch()` similar to how we do `.platform()`

packages/cli/src/cli/target.rs
```diff
@@ -50,3 +50,127 @@ pub(crate) struct TargetArgs {
...
+        // Try auto detecting arch through adb.
+        static AUTO_ARCH: OnceCell<Option<Arch>> = OnceCell::new();
+
+        *AUTO_ARCH.get_or_init(|| {
+            // TODO: Wire this up with --device flag. (add `-s serial`` flag before `shell` arg)
+            let output = std::process::Command::new("adb")
+                .arg("shell")
+                .arg("uname")
+                .arg("-m")
+                .output();
```
jkelleyrtp, line 85: We can't use std::process since this blocks the thread - can we somehow reorient this to be async + lazy?

packages/cli/src/build/verify.rs
```diff
@@ -134,9 +134,19 @@ impl BuildRequest {
...
+            let Some(ndk) = self.krate.android_ndk() else {
+                break 'a true;
+            };
+
+            let linker = self.build.target_args.arch().android_linker(&ndk);
+
+            !linker.exists()
+        };
+
+        if failed {
```
jkelleyrtp, line 147: this is weird code, can't we just use an unwrap_or or a .map or something that doesn't involve named breaks?

jkelleyrtp (approved):
looks great, thank you!

Issue #3148, opened 2024-10-30
Fullstack 0.6.0-alpha.3: (When custom rust flags) Build failed: Failed to parse asset manifest from "/var/folders...


tgrushka:
**Problem**

**EDIT:** Only when custom cargo config/flags/linker.

`[dev] Build failed: Failed to parse asset manifest from "/var/folders/...`

**Steps To Reproduce**

**EDIT:** 0. Have a custom rust config in ~/.cargo/config.toml, i.e.:
```
[build]
target = "aarch64-apple-darwin"

[target.aarch64-apple-darwin]
linker = "/usr/bin/clang"
rustflags = [
    "-C", "relocation-model=static",
    "-C", "link-arg=--ld-path=/usr/local/bin/ld64.sold"
]
```

1. `cargo install dioxus-cli --force --git 'https://github.com/DioxusLabs/dioxus#main'`
2. `dx new dx-6-test` → fullstack → true → Vanilla
3. `cd dx-6-test; dx serve`
4. Build almost completes, then above error.

**Expected behavior**

Expect build to finish and serve.
Behaves / serves properly when choosing **web** instead of **fullstack**.
**EDIT:** I commented out my ~/.cargo/config.toml and now it builds fine! Would be nice to use faster linker but I see there are other issues about this.

**Environment:**
 - Dioxus version: `dioxus 0.6.0-alpha.3 (5d99889)`
 - Rust version:   `rustc 1.82.0 (f6e511eec 2024-10-15)`
 - OS info:    macOS 15.0.1 (24A348)
 - App platform:    **fullstack** (not an issue on **web**)

**Trace output**

```
...
14:41:45 [dev] Build update: Progress { stage: Compiling { platform: Server, current: 349, total: 351, krate: "dioxus_router" } }
14:41:45 [dev] Build update: Progress { stage: Compiling { platform: Server, current: 350, total: 351, krate: "dioxus" } }
14:41:46 [dev] Build update: BuildFailed { err: Other(Failed to parse asset manifest from "/var/folders/5f/lt49hfxs2f708tdrn5vd31200000gn/T/.tmpU2MG3l"

Caused by:
    EOF while parsing a value at line 1 column 0

Stack backtrace:
   0: std::backtrace::Backtrace::create
   1: anyhow::context::<impl anyhow::Context<T,E> for core::result::Result<T,E>>::with_context
   2: dx::assets::AssetManifest::load_from_file
[...]

PR #3228, opened 2024-11-17
feat: traverse the target directory instead of intercepting the linker to collect assets

jkelleyrtp:
it came to me in a dream...

instead of intercepting the linker, simply traverse the `deps` and `incremental` directories for their artifacts.

This drastically speeds up asset bundling times and is much nicer when used with LTO since we are basically forced to compile twice with LTO.

This is also generally better for folks who run `cargo check` with our profiles wrt cache thrashing.

Fixes: #3148

### DIFF
--- a/packages/cli/src/assets.rs
+++ b/packages/cli/src/assets.rs
@@ -24,8 +24,8 @@ impl AssetManifest {
     }
 
     /// Fill this manifest with a file object/rlib files, typically extracted from the linker intercepted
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
Tested against android/ios/web 👍

Builds are *so* much faster now. Would like to add caching of the rlib parsing but it's already sooo fast. In the future maybe we can do it in parallel.

PR #3239, opened 2024-11-18
Fix: CLI Windows Write `exe` Errors

DogeDark:
The CLI fails to write the new executable on Windows because the old one is still running during rebuilds.

This PR introduces a `kill_all()` method on the `AppRunner`. Ideally, the app would be killed right before the bundler starts, but I wasn't sure how to best get a handle down to the bundler. Instead, I opted for this simpler, but perhaps slightly less DX solution (since the app is closed during rebuilds and the dev can't admire their desktop app or interact with serverfns).

### DIFF
--- a/packages/cli/src/serve/mod.rs
+++ b/packages/cli/src/serve/mod.rs
@@ -107,6 +107,9 @@ pub(crate) async fn serve_all(mut args: ServeArgs) -> Result<()> {
                 } else if runner.should_full_rebuild {
                     tracing::info!(dx_src = ?TraceSrc::Dev, "Full rebuild: {}", file);
 
+                    // Kill runnning executables
+                    runner.kill_all();
+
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
This is good for now, but a better solution to this problem might be:
- to write the executables with a hash attached to the name that we save in the BuildArtifacts that gets removed
- split apart build and bundle into two separate phases that only kills when we start bundling

When builds take a while it is nice to still have the app interactive. Hopefully one day our binary patching is good enough for most cases though.

DogeDark:
@jkelleyrtp If it sounds good to you, I will gate this behind Windows and I'll add the preferred fix into Notion for a future patch release so that it's ready for 0.6, but can be properly fixed when there is more time.

jkelleyrtp (approved):

PR #3250, opened 2024-11-24
fix: Support external link targets without router

marc2332:


marc2332:
Duplicate of https://github.com/DioxusLabs/dioxus/pull/3633 (lol)

PR #3274, opened 2024-12-03
Generate the `dioxus_html/src/elements.rs` with codegen

rambip:
For now, it is just initial work.

Motivation:
- easier to read and to contribute, in particular for people who would like to use dioxus with other elements.
-  if it works, `dioxus-cli` would not need to be compiled using the exact same version of `dioxus_elements`. The idea is that the cli could look at the dependencies, find what element library is used, get the `elements.toml`, and use it to do hot-reloading.

Missing:
- syntax to specify the same attributes for a lot of elements
- what name to give to the `a` element in the `svg` namespace.


jkelleyrtp:
Hi there, just closing this since it's been up for a while. Please feel free to re-open it or make a new PR if anything updates here!

PR #3282, opened 2024-12-04
Fix: Flush Cargo Logs on Build Failure

DogeDark:
Using `dx build`, when Cargo fails, it early returns and doesn't process any remaining `BuildUpdate`s in the channel. This change flushes any `BuildUpdate::CompilerMessage` in the `BuildFailed` handler so that logs are outputted correctly.

### DIFF
--- a/packages/cli/src/build/builder.rs
+++ b/packages/cli/src/build/builder.rs
@@ -249,6 +249,13 @@ impl Builder {
                     return Ok(bundle);
                 }
                 BuildUpdate::BuildFailed { err } => {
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
we might also need this for regular `dx serve`?

DogeDark:
> we might also need this for regular `dx serve`?

I believe the `.finished()` method is only used by the build/bundle commands. Since `dx serve` continues running after a build fails, it should continue processing the build logs.

Issue #3276, opened 2024-12-03
SSR produces HTML that defeats the purpose of SSR


ochrons:
**Problem**

With streaming rendering, the HTML that SSR produces is not useful. The whole point of SSR is to provide an HTML page that user can view (and search engines index) before any JS/WASM code gets executed on the client.

But in its current form, the content is streamed into a separate `div` that is marked `hidden` with an attribute, making it invisible to the user (and probably to some extent, to search engines). Nothing is shown to the user until WASM code is loaded and executed.

**Steps To Reproduce**

- create a fullstack web app with a server function call using `use_server_future` (might not even be needed)
- load the page into a browser with network throttling on

**Expected behavior**

SSR should produce HTML that is ready to be rendered without any additional script execution.

**Screenshots**

![image](https://github.com/user-attachments/assets/86b2877d-3101-48cd-b8fa-bedacb7cc2d2)

**Environment:**
 - Dioxus version: 0.6.0-rc.0
 - Rust version:    1.83.0
 - OS info:         WSL
 - App platform:    Fullstack app

**Questionnaire**

ealmloff:
Dioxus currently supports only out of order streaming which loads chunks of HTML from the server as the data finishes loading, regardless of order. Out of order streaming currently requires javascript to stream in the content. (there are demos of streaming without JS with the declarative shadow dom, but this creates too many limitations about the structure of the streaming for use in dioxus)

Search engines should still be able to read the content of the streaming chunks of the page even though it is hidden. We use the same approach for streaming as NextJS. They have more documentation about tests they have done with out of order streaming and seo [here](https://vercel.com/guides/does-streaming-affect-seo#does-streaming-affect-seo-and-can-streamed-content-be-indexed).

The other approach to streaming is which it sounds like you were expecting is in order streaming. In order streaming doesn't require javascript, but it needs to load elements in order from the start of the html which can mean most of the page is blocked from loading by an unimportant suspense boundary near the top of the HTML. Fullstack doesn't support this approach today, but it may be added in the future.

Both approaches to streaming hydrate the page as the content is streamed in from the server which significantly improves the time to interactive (for sites that require WASM). Much more of the page can be interactive quickly with out of order streaming because it doesn't impose order restrictions

ochrons:
[...]

PR #3288, opened 2024-12-04
Expose streaming options to choose between full prerendering and out of order streaming

ealmloff:
This PR makes the out of order streaming option optional to allow fullstack sites to support content loading without javascript. All of the playwright tests are still running with out of order streaming mode enabled because that is the more difficult case to get right

Closes https://github.com/DioxusLabs/dioxus/issues/3276


jkelleyrtp (approved):
Thank you for doing this!

ochrons:
Thanks for quickly addressing this issue!

ochrons:
Verified this with my Dioxus app (with 0.6.0 release) and now the HTML is served as expected. Nice work!

PR #3290, opened 2024-12-05
update Leptos example in README.md

gbj:
As of 0.7, Leptos supports the same kind of unkeyed, VDOM-like diff that Dioxus does in the comparison example given in the README.

I only updated the example, but the changes to the example mean that the rest of the surrounding text no longer makes sense.

Personally, I'd suggest just removing it; Leptos does provide a (keyed) For component helper, but the unkeyed behavior is now identical with the Dioxus unkeyed behavior, the only real difference between the two examples being that the Leptos one uses an iterator and the Dioxus one uses its special rsx `for` syntax, so the comments about losing reactivity and special control-flow components being required are not particularly accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,23 +295,25 @@ fn Counters() -> Element {
 ```rust
 fn Counters() -> impl IntoView {
     let counters = RwSignal::new(vec![0; 10]);
-
+    
     view! {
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Thanks! I just went ahead and nixed the whole comparison.

https://github.com/DioxusLabs/dioxus/pull/3292

PR #3294, opened 2024-12-05
Feat: Manganis SCSS Support

DogeDark:
Adds basic SCSS/SASS support to Manganis.

### DIFF
--- a/packages/cli-opt/src/css.rs
+++ b/packages/cli-opt/src/css.rs
@@ -75,7 +75,7 @@ pub(crate) fn process_scss(
 
 /// Logger for Grass that re-uses their StdLogger formatting but with tracing.
 #[derive(Debug)]
-pub struct ScssLogger {}
+struct ScssLogger {}
 
 impl grass::Logger for ScssLogger {
     fn debug(&self, location: SpanLoc, message: &str) {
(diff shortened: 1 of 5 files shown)


packages/manganis/manganis/src/macro_helpers.rs
```diff
@@ -74,14 +74,38 @@ pub const fn generate_unique_path(
...
+    while i < left.len() {
+        if left[i] != right[i] {
+            return false;
+        }
+        i += 1;
+    }
+
+    true
+}
+
```
jkelleyrtp, line 108: can we just do the b"scss" slice compare thing?

packages/cli-opt/src/css.rs
```diff
@@ -40,3 +43,57 @@ pub(crate) fn minify_css(css: &str) -> String {
...
+            output_path.display()
+        )
+    })?;
+
+    Ok(())
+}
+
+/// Logger for Grass that re-uses their StdLogger formatting but with tracing.
+#[derive(Debug)]
+pub struct ScssLogger {}
```
jkelleyrtp, line 78: doesn't need to be public

jkelleyrtp (approved):
Looks good! Might be worth adding a scss asset to one of the doc-tests/doc-comments in manganis itself to assert that the extension gets changed.

When we migrated magnanis to this repo we unfortunately dropped our test suite -  we need to add that back at some point.

PR #3318, opened 2024-12-09
Rewrite all the `dioxus_elements` logic for more modular `hotreload` support

rambip:
It is very much a draft.
For now, this channel is mainly a way for me to write down my reasoning about what has to be done.

The final goal of this PR is to get to a state where:
- the knowledge about the elements is not baked into the cli
- the user can easily add or change (and maybe delete) the elements we wants to use in the rsx
- the rsx hotreload and the compilation does not take more time
- maybe use a trait-based element definition to allow the user to add it's own element without having `dioxus_elements` in scope
- the code for hotreload is simpler (no `HotReloadCtx` trait)

I also have a different PR #3274 that generates the elements spec with codegen, and this work is almost usable right now. I will have to merge the 2 at one point.

### DIFF
--- a/packages/cli/src/elements_metadata.rs
+++ b/packages/cli/src/elements_metadata.rs
@@ -1,0 +1,37 @@
+use manganis_core::linker::LinkSection;
+use manganis_core::BundledMetadata;
+use object::{read::archive::ArchiveFile, File as ObjectFile, Object, ObjectSection};
+use serde::{Deserialize, Serialize};
+
+/// Fill this manifest with whatever tables might come from the object file
+fn collect_elements_metadata(&mut self, obj: &ObjectFile) -> anyhow::Result<HasMap<ConstStr, ConstStr>> {
+    for section in obj.sections() {
+        let Ok(section_name) = section.name() else {
+            continue;
+        };
+
+        // Check if the link section matches the asset section for one of the platforms we support. This may not be the current platform if the user is cross compiling
+        let matches = LinkSection::ALL
+            .iter()
+            .any(|x| x.link_section == section_name);
+
+        if !matches {
+            continue;
+        }
+
+        let bytes = section
+            .uncompressed_data()
+            .context("Could not read uncompressed data from object file")?;
+
+        let mut buffer = const_serialize::ConstReadBuffer::new(&bytes);
+        while let Some((remaining_buffer, asset)) =
+            const_serialize::deserialize_const!(BundledMetadata, buffer)
+        {
+            self.assets
+                .insert(asset.absolute_source_path().into(), asset);
+            buffer = remaining_buffer;
+        }
+    }
+
+    Ok(())
+}
--- a/packages/manganis/manganis-core/src/metadata.rs
+++ b/packages/manganis/manganis-core/src/metadata.rs
@@ -1,0 +1,63 @@
+use const_serialize::{ConstStr, SerializeConst};
+use std::path::PathBuf;
+
+#[derive(
+    Debug,
+    PartialEq,
+    PartialOrd,
+    Clone,
+    Copy,
+    Hash,
+    SerializeConst,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct BundledMetadata {
+    pub key: ConstStr,
+    pub value: ConstStr,
+}
+
+impl BundledMetadata {
+    #[doc(hidden)]
+    /// This should only be called from the macro
+    /// Create a new asset
+    pub const fn new(
+        absolute_source_path: &'static str,
+        bundled_path: &'static str,
+        options: AssetOptions,
+    ) -> Self {
+        Self {
+            absolute_source_path: ConstStr::new(absolute_source_path),
+            bundled_path: ConstStr::new(bundled_path),
+            options,
+        }
+    }
+}
+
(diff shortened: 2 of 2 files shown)


rambip:
ROADMAP:

rambip:
First difficulty: where must the element namespace be static, and when does it have to do dynamic ?
Clearly, since the cli reads the element specification at runtime, it's element specifications are dynamic.

Right now, the fileds of `TemplateNode` for `tag` and `namespace` are static, and I think it will not work for our approach.

rambip:
Right now, the hotreloading is enabled by the following trait (https://github.com/DioxusLabs/dioxus/blob/main/packages/core-types/src/hr_context.rs):

```rust
pub trait HotReloadingContext {
    fn map_attribute(
        element_name_rust: &str,
        attribute_name_rust: &str,
    ) -> Option<(&'static str, Option<&'static str>)>;
    fn map_element(element_name_rust: &str) -> Option<(&'static str, Option<&'static str>)>;
}
```

Meaning that the namespace of the attribute depends on the parent element. If we have to call `manganis::register_attribute` for each attribute inside each element, the amount of static data generated will be huge.

I have to find a way to compress it, like pointing to an attribute inside the definition of an element

ealmloff:
> First difficulty: where must the element namespace be static, and when does it have to do dynamic ? Clearly, since the cli reads the element specification at runtime, it's element specifications are dynamic.
> 
> Right now, the fields of `TemplateNode` for `tag` and `namespace` are static, and I think it will not work for our approach. We could `Box::leak` it but it's not very elegant.

We currently leak the string in debug mode during hot reloading. Templates are fairly small and the leak only happens in debug mode, so it hasn't been an issue so far, but we have also discussed switching to `Cow<'static, str>` in the future. That would be a much more broad breaking change that will require a lot of changes to the core crate and tests. I think it would be better to just continue leaking for now to keep the scope of this PR more manageable.

Currently, the CLI sends a [HotReloadedTemplate](https://github.com/DioxusLabs/dioxus/blob/main/packages/core/src/hotreload_utils.rs#L332-L347) message to the front end for hot reloading. That is then combined with the [dynamic pool](https://github.com/DioxusLabs/dioxus/blob/main/packages/core/src/hotreload_utils.rs#L248-L266) to create a full VNode

> Right now, the hotreloading is enabled by the following trait (https://github.com/DioxusLabs/dioxus/blob/main/packages/core-types/src/hr_context.rs):
> 
> ```rust
> pub trait HotReloadingContext {
>     fn map_attribute(
>         element_name_rust: &str,
>         attribute_name_rust: &str,
>     ) -> Option<(&'static str, Option<&'static str>)>;
>     fn map_element(element_name_rust: &str) -> Option<(&'static str, Option<&'static str>)>;
> }
> ```
> 
> Meaning that the namespace of the attribute depends on the parent element. If we have to call `manganis::register_attribute` for each attribute inside each element, the amount of static data generated will be huge.
> 
> I have to find a way to compress it, like pointing to an attribute inside the definition of an element

The svg and global attributes traits could be pulled out into a separate "attribute group" section with a name and then each element could include one or more attribute group. For example, a `td` element would include `colspan`, and `rowspan` as special attributes and the string, or hash of `GlobalAttribute` for the global attribute group

rambip:
Is there a way to derive the parsing code such that `parse_macro_input!(ElementSpecification { stuff }) == ElementSpecification {stuff }` ?

https://github.com/DioxusLabs/dioxus/blob/793f833bb0714034d93f938a9c918fd894a3364a/packages/core-types/src/elements.rs#L17

rambip:
I have seen that the manganis support crate for the cli has moved to https://github.com/DioxusLabs/dioxus/blob/main/packages/cli-opt/src/lib.rs

Is renaming `AssetManifest` to `ManganisData` and adding a `metadata` field in this file ok ?
I don't know if the metadata (where the element specification will be stored) have to be read each time the assets have to be read.

rambip:
> > I have seen that the manganis support crate for the cli has moved to https://github.com/DioxusLabs/dioxus/blob/main/packages/cli-opt/src/lib.rs
> > Is renaming `AssetManifest` to `ManganisData` and adding a `metadata` field in this file ok ?
> 
> From the linker side, it might be easier to model the metadata as a separate variant of the asset enum so we can re-use all of the extraction logic. After we get the data out of the link sections splitting it into a separate field would be good. This PR will already be breaking, so changing public fields isn't an issue. The name `ManganisData` is a bit more vague than `AssetManifest` and you could say the rsx macro definition is an asset provided to the CLI. Renaming things should be easy enough with rust analyzer once the rest of the PR is working
> 
> > I don't know if the metadata (where the element specification will be stored) have to be read each time the assets have to be read.
> 
> I think both metadata and assets will need to be extracted every time the build runs before serve finishes, so they can share the same link section

But an important property of an asset is it's path, what path do I set for metadata ? The current rust file ?

ealmloff:
> But an important property of an asset is it's path, what path do I set for metadata ? The current rust file ?

Sorry, I was thinking of the old structure of manganis. Path used to be optional. It could be a different link section or a different top level enum that is either an asset or metadata.

jkelleyrtp:
Hi there, just closing this since it's been up for a while. Please feel free to re-open it or make a new PR if anything updates here!

PR #3321, opened 2024-12-10
feat: Reexport the `Warning` trait in dioxus-signals

marc2332:
feat: Reexport the `Warning` trait in dioxus-signals to allow the use of `::allow()` without pulling the warnings crate

fixes: #3513

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3404,6 +3404,7 @@ dependencies = [
  "thiserror 1.0.69",
  "tokio",
  "tracing",
+ "warnings",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


jkelleyrtp:
Maybe we should be re-exporting warnings from the `dioxus` namespace?

marc2332:
> Maybe we should be re-exporting warnings from the `dioxus` namespace?
> 
> ```rust
> dioxus::warnings::allow()
> ```

In dioxus-lib as well if that was the case

jkelleyrtp:
Changed this to re-export warnings as `dioxus::warnings` under both dioxus and dioxus-lib

jkelleyrtp (approved):

Issue #3322, opened 2024-12-10
Desktop app can't be automatically re-opened by the `dx serve`


Andrew15-5:
When re-opening or re-building the app on Linux, it shows:

```
14:43:17 [dev] Build failed: Other(Failed to write main executable

Caused by:
    0: I/O Error: Text file busy (os error 26)
    1: Text file busy (os error 26))
```

**Steps To Reproduce**

- create a Dioxus app for desktop platform
- run `dx serve` on Linux/Unix-like system
- re-open/re-build the app
- see the error

**Expected behavior**

No errors.

**Screenshots**

![image](https://github.com/user-attachments/assets/435fa5ac-f09e-4a0f-972d-d51e6ffe89eb)

**Environment:**
 - Dioxus version: 0.6
 - Rust version: rustc 1.83.0-nightly (fb4aebddd 2024-09-30)
 - OS info: Pop!_OS 22.04 LTS
 - App platform: desktop

kingzcheung:
I'm encountering the same issue.

Andrew15-5:
I've rebased https://github.com/DioxusLabs/dioxus/pull/3323 onto v0.6.1, so you can use it with all the latest updates.

jeancf:
I am seeing the same issue on linux with desktop feature and dx 0.6.1. I am using rustc 1.83.0.

Andrew15-5:
@jeancf, have you tried [the fix](https://github.com/DioxusLabs/dioxus/pull/3323)?

jeancf:
I am happy to try the fix but I don't know how to do it.

Andrew15-5:
```sh
git clone --depth 1 -b fix-desktop-open-linux https://github.com/Andrew15-5/dioxus.git
cd dioxus
cargo install --path ./packages/cli
```

jeancf:
Thanks. I would not have figured it on my own.

The fix indeed closes and restarts the app after a rebuild.

After manually closing the app window, pressing `o` does not reopen it though, but I am not sure this fix was meant to address that as well.

crabdancing:
Can confirm this bug also on NixOS 24.11. Can also confirm that @Andrew15-5's patch fixes the issue. I can _also_ confirm that the app does steal focus when reloaded on my system too, unfortunately. It seems intrinsic problem to re-opening the window. I'm not sure how the tooling behaves on other systems.

[...]

PR #3323, opened 2024-12-10
fix(cli): desktop app now can be re-opened on Linux

Andrew15-5:
Fixes https://github.com/DioxusLabs/dioxus/issues/3322.

I added some tracing logs just in case someone else also needs them, like I did when debugging this issue.

IDK why this is added and so soon in the build cycle, but perhaps this can be replaced by the approach from this PR, but someone needs to check if replacing works.

https://github.com/DioxusLabs/dioxus/blob/74252f23d85359c96ed22014091b5cafd99127c5/packages/cli/src/serve/mod.rs?plain=1#L110-L113

### DIFF
--- a/packages/cli/src/build/bundle.rs
+++ b/packages/cli/src/build/bundle.rs
@@ -7,6 +7,7 @@ use dioxus_cli_opt::{process_file_to, AssetManifest};
 use manganis_core::AssetOptions;
 use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use std::collections::HashSet;
+use std::ffi::OsString;
 use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


istudyatuni:
Another solution is just remove target file and then copy. It also seems simpler to implement

Andrew15-5:
Interesting.

pros:

- only 2 cumulative operations in the code base, i.e., just need to add 1 operation to an existing one

cons:
- not an atomic operation, unlike `mv` (in Unix)

It doesn't require additional space for the 3rd binary. Well actually, since some process is still using the file, this would mean that after rm & cp there will be 2 visual copies and 1 more copy which is used by a process.

So this one is debatable. It will still require 2 operations for the Linux platform. And the code wouldn't become tremendously smaller, so I think we should keep the atomic approach.

jkelleyrtp:
I implemented a similar but different system in #3608 to solve this issue not just for linux but for all platforms that have this restriction.

Will add your name to the contribution list so you get the badge :) thank you for the PR!

Andrew15-5:
Thanks. I've tested that and it does work. Though I saw that in your PR [there is a place](https://github.com/DioxusLabs/dioxus/pull/3586/files#diff-4f814301cd0835cc4f3998188d78fd7a3f38a78216fa0eb54b5c03adc2db59f6L61-R61) where you replace `o` key action with `p` instead of adding it, so I thought that maybe manual opening is removed altogether, but apparently not. Maybe you should look at it. Looks like a typo or a bug, or maybe I don't understand something.

jkelleyrtp:
> Thanks. I've tested that and it does work. Though I saw that in your PR [there is a place](https://github.com/DioxusLabs/dioxus/pull/3586/files#diff-4f814301cd0835cc4f3998188d78fd7a3f38a78216fa0eb54b5c03adc2db59f6L61-R61) where you replace `o` key action with `p` instead of adding it, so I thought that maybe manual opening is removed altogether, but apparently not. Maybe you should look at it. Looks like a typo or a bug, or maybe I don't understand something.

I moved the 'o' key to the `/` extra panel since we occasionally tell the user via a log they can use 'o' to re-open.

Andrew15-5:
Ohh, yeah, that makes sense.

Andrew15-5:
> Will add your name to the contribution list so you get the badge :) thank you for the PR!

I ain't seeing no badge. ;-;

https://github.com/DioxusLabs/dioxus/releases/tag/v0.6.2

PR #3330, opened 2024-12-10
Fix: Windows Manganis Asset Hot Reload

DogeDark:
Manganis' asset manifest uses canonicalized assets and Windows may use the long-form path instead of regular `C:\`

### DIFF
--- a/packages/cli/src/serve/handle.rs
+++ b/packages/cli/src/serve/handle.rs
@@ -1,6 +1,7 @@
 use crate::{AppBundle, Platform, Result};
 use anyhow::Context;
 use std::{
+    fs,
     net::SocketAddr,
     path::{Path, PathBuf},
     process::Stdio,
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Works for me on macos

packages/cli/src/serve/handle.rs
```diff
@@ -187,8 +188,17 @@ impl AppHandle {
...
 
+        // Canonicalize the path as Windows may use long-form paths "\\\\?\\C:\\".
+        let changed_file = match fs::canonicalize(changed_file) {
+            Ok(c) => c,
+            Err(e) => {
+                tracing::debug!("Failed to canonicalize hotreloaded asset: {e}");
+                return None;
+            }
+        };
+
```
jkelleyrtp, line 199: I believe you can inspect_err + ok() this

jkelleyrtp (approved):

PR #3335, opened 2024-12-11
Feat: CLI `wasm-bindgen-cli` Installer

DogeDark:
This PR adds prebuilt binary (and fallback compile-from-source) installation of the wasm-bindgen CLI. This allows our CLI to build a project of *any* `wasm-bindgen` version as long as the installed `wasm-bindgen-cli` supports the required flags.

- A temporary folder is used to store installation artifacts.
- Installed binaries are stored in `{data_local_dir}/dioxus/wasm-bindgen`
- Installed binaries are named `wasm-bindgen-{version}{.exe}`

The installer attempts to install in this order:
1. Prebuilt release binaries on wasm-bindgen's GitHub
2. cargo-binstall
3. cargo-install

This could use testing on other platforms:
- [x] Windows
- [x] MacOS
- [x] Linux

image


jkelleyrtp (approved):
Cleaned it a bit + merged main but looks good otherwise!

PR #3347, opened 2024-12-13
Fix: `always-on-top` setting

DogeDark:
This PR fixes the `always-on-top` setting (the environment variable wasn't being set) and also makes a few revisions to make `CliSettings` the single source of truth for the settings it contains.


packages/cli/src/dioxus_crate.rs
```diff
@@ -25,7 +25,7 @@ pub(crate) static PROFILE_ANDROID: &str = "android-dev";
 pub(crate) static PROFILE_SERVER: &str = "server-dev";
 
 impl DioxusCrate {
```
jkelleyrtp, line 27: I'd like to merge this PR but I don't want to change the DioxusCrate::new() signature - I don't think the override stuff is worthwhile to include in this PR.

Can you strip this PR down to *just* fixing the always-on-top setting?

jkelleyrtp (changes requested):

jkelleyrtp (approved):

PR #3351, opened 2024-12-13
fix(cli): addr parameter of dx serve

rummatee:
When I run `dx serve --addr 0.0.0.0` in order to receive requests from any ip address, then currently the dev server tries to forward this request to 0.0.0.0, however since the proxied server is always started on the same machine, I believe that the request should always be forwarded to localhost.

Is that correct, or am I misunderstanding how the addr parameter should work?

I was also considering changing the address to localhost in the proxied_server_address function as well, so that also the proxied server would always only listen to localhost, but then I thought it might be useful to be able to reach it directly as well, when specifying the address as 0.0.0.0 and calling from another host.

### DIFF
--- a/packages/cli/src/serve/server.rs
+++ b/packages/cli/src/serve/server.rs
@@ -82,7 +82,7 @@ impl WebServer {
             .then(|| get_available_port(devserver_ip))
             .flatten();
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Sorry for letting this sit - this was fixed in a different PR a while back, but thanks for the PR!

PR #3362, opened 2024-12-15
feature(cli): dioxus-cli as lib

notdanilo:
### Problem

While it is possible to wrap dx using std::process::Command, a more flexible and programmable approach would be to reuse dioxus-cli as a library. Publishing dioxus-cli as a library would greatly enhance its usability.

### Solution

This PR exposes `dioxus-cli` as a library, making it possible to reuse `dx` instead of redirecting cli commands to it.

A good example is how `bolt-cli` uses `anchor-cli` as a library.
https://github.com/magicblock-labs/bolt/blob/main/crates/bolt-cli/src/lib.rs

### DIFF
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -141,6 +141,9 @@ wasm-opt = ["dep:wasm-opt"]
 path = "src/main.rs"
 name = "dx"
 
+[lib]
+path = "src/lib.rs"
+
 [dev-dependencies]
 escargot = "0.5"
 
--- a/packages/cli/src/cli/autoformat.rs
+++ b/packages/cli/src/cli/autoformat.rs
@@ -10,35 +10,35 @@ use std::{borrow::Cow, fs, path::Path};
 
 /// Format some rsx
 #[derive(Clone, Debug, Parser)]
-pub(crate) struct Autoformat {
+pub struct Autoformat {
     /// Format rust code before the formatting the rsx macros
     #[clap(long)]
-    pub(crate) all_code: bool,
+    pub all_code: bool,
 
     /// Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits
     /// with 1 and prints a diff if formatting is required.
     #[clap(short, long)]
-    pub(crate) check: bool,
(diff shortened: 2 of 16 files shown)


notdanilo:
Thanks for running the checks. I will fix them asap.

jkelleyrtp:
Thanks! Also - do you mind adding a `no_exhaustive` on any structs in the public API of the cli?

I'm okay with opening up the CLI as a library but that means the CLI needs to follow semver. We should try to keep our public API minimal in that case. This means marking things as no_exhaustive where possible and not making too many things public.

jkelleyrtp:
Going to close this since it's been open for a while... 

Also, having a little while to think about it, I'm not 100% sure we want to open up the CLI as a semver API. I'd be happy to accept a PR that exposes more of our types for Process/Command consumption, but I generally don't think people would even want to develop against the CLI. It's very heavy to develop against and does many weird things like overriding loggers, ctrl-c, sigkill, etc.

If you want to use the CLI from a script then I'd recommend using the types crates we provide to parse the output with JSON mode enabled.

notdanilo:
I had to put the web-component effort on hold. But I might come back to this as this is something that would make it easier for me to release a cli that depends on dioxus-cli

jkelleyrtp:
For sure - we could expose our CLI's arguments as an independent library instead of the entire CLI. We do that with the return types in the `dioxus-dx-wire-format` crate. I could see a `dx-driver` crate that wraps the `dioxus-dx-wire-format` and some Process work to allow scripting of DX

notdanilo:
But doesn't that mean that I need to ask for my cli user to download and install dioxus-cli?

jkelleyrtp:
You could download it for them with `cargo-binstall`. The CLI sets up lots of hooks that is going to cause trouble for the host program - the logging system, setting sigkill hooks, setting a panic handler, etc 

Tools like https://github.com/dustinblackman/cargo-run-bin can automate downloading the right dx version for you, allowing you to get a handle to the program and execute it.

notdanilo:
@jkelleyrtp thanks a lot for the suggestions. I didn't know these crates :)

PR #3369, opened 2024-12-16
fix(cli): get ip addr & port number from config

wiseaidev:
[I have been experimenting with Dioxus SSG recently](https://github.com/opensass/opensass/pull/22) and noticed that the port number and IP address are hardcoded. The port is set to `9999` instead of retrieving the values from the configuration.

With these changes, the port is now set to the env var value at compile time (defaults to 8080):

```sh
PORT=3000 dx serve --ssg 
```

### DIFF
--- a/packages/cli/src/build/bundle.rs
+++ b/packages/cli/src/build/bundle.rs
@@ -688,7 +688,9 @@ impl AppBundle {
         use futures_util::StreamExt;
         use tokio::process::Command;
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Good catch - I didn't think it would be important to pass the port/address through but in hindsight some features *do* care about these values (like side apis and cli_config)

jkelleyrtp (approved):

PR #3375, opened 2024-12-17
fix peek() example

spookyvision:
this PR aligns what the code snippet does with what the comment says it should do (peek and log the *event source*, not the *count value*). It also ensures `count += 1` is done for both buttons, because otherwise only the double button causes a log message.


jkelleyrtp:
Good catch, thank you!

jkelleyrtp (approved):

PR #3379, opened 2024-12-17
fix: throw error when asset is empty

jkelleyrtp:
While working on the docsite I found that if you passed `asset!("")` on accident, then we'd end up copying the entire current directory over. This is not good. It also causes new `dx serve` to hang infinitely since we get stuck on clearing the assets dir.

This fixes that by throwing an error if the path is empty and ensuring the final path is a child of the crate.

### DIFF
--- a/packages/manganis/manganis-macro/src/asset.rs
+++ b/packages/manganis/manganis-macro/src/asset.rs
@@ -11,12 +11,6 @@ use syn::{
 };
 
 fn resolve_path(raw: &str) -> Result<PathBuf, AssetParseError> {
-    if raw.is_empty() {
-        return Err(AssetParseError::InvalidPath {
-            path: PathBuf::new(),
-        });
-    }
-
     // Get the location of the root of the crate which is where all assets are relative to
     //
     // IE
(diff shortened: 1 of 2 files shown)


packages/manganis/manganis-macro/src/asset.rs
```diff
@@ -19,19 +25,28 @@ fn resolve_path(raw: &str) -> Result<PathBuf, AssetParseError> {
...
     // 2. absolute path to the asset
-    manifest_dir
-        .join(raw.trim_start_matches('/'))
-        .canonicalize()
-        .map_err(|err| AssetParseError::AssetDoesntExist {
-            err,
-            path: input.clone(),
-        })
+    let path =
+        std::path::absolute(manifest_dir.join(raw.trim_start_matches('/'))).map_err(|err| {
```
ealmloff, line 37: `std::path::absolute` does not check if the path exists (unlike canonicalize) so this error does not trigger when the path doesn't exist

packages/manganis/manganis-macro/src/asset.rs
```diff
@@ -19,19 +25,28 @@ fn resolve_path(raw: &str) -> Result<PathBuf, AssetParseError> {
...
+            AssetParseError::AssetDoesntExist {
+                err,
+                path: input.clone(),
+            }
+        })?;
+
+    // 3. Ensure the path is not the current dir or exist outside the current dir
+    if path == manifest_dir || !path.starts_with(manifest_dir) || !path.exists() {
+        return Err(AssetParseError::InvalidPath { path });
```
ealmloff, line 46: Instead, this case triggers which gives the wrong error message

ealmloff:
Looks good overall once the error message is fixed. I don't think the first if statement to check if the path should change behavior because the other changes have more robust checks that thow errors for `asset!("")`, `asset!("/../")`, `asset!("/")`, etc

ealmloff (approved):

Issue #3125, opened 2024-10-28
`oninput` fails to fire with chinese character on the latest git in desktop


JustSimplyKyle:
**Problem**

`oninput` event get losts when the input is a non-ascii character in desktop, on the latest git commit.
Interestingly, said problem can not be observed on web.

**Steps To Reproduce**

This behaviour can be observed on the latest git commit, but not on 0.5,
```rust
#![allow(non_snake_case)]

use dioxus::prelude::*;
use dioxus_logger::tracing::{info, Level};

fn main() {
    // Init logger
    dioxus_logger::init(Level::INFO).expect("failed to init logger");
    info!("starting app");
    launch(App);
}

#[component]
fn App() -> Element {
    // Build cool things ✌️
    let mut value = use_signal(String::new);

    rsx! {
        div {
            {value}
        }
        input {
            oninput: move |x| {
                value.set(x.value());
            },
            value
        }
    }
}

```
When you input anything that is non-ascii(e.g. Chinese 我), the `value` signal is not updated(as displayed by the block aboved).
In the development console, it complains about a type mismatch.
![image](https://github.com/user-attachments/assets/5b23554a-46de-4669-88f6-7db03d091e8d)

**Expected behavior**

The `oninput` signal be fired successfully.

**Environment:**
 - Dioxus version:  `master`
 - Rust version:   1.81.0
 - OS info:   6.11.5-arch1-1
 - App platform:   desktop

tdomhan:
I'm running into the same issue with any emoji in input/textarea elements.

glfeng318:
got the same issue

tdomhan:
I added a PR to fix this, which works for me locally (using any unicode characters like 我).

PR #3386, opened 2024-12-18
fix: Handle unicode in webview events.

tdomhan:
The request header does not support unicode characters leading to `Type: error` when calling `setRequestHeader` with a non-ascii string. This PR first encodes the data using base64 before setting the header. The downside is that this increases the content size by 33%. 

This addresses https://github.com/DioxusLabs/dioxus/issues/3125

Running the build updates all Javascript files. I assume this is due to the Typescript version. I'm using `5.7.2`. Let me know if I should be using a specific version.


jkelleyrtp:
Wow, thanks for the quick fix! Do we need to do base64 encoding? Is there a chance that our mime-type is just not set properly on the XHR request that we can fix by using `text/plain;charset=UTF-8` as the mime-type?

tdomhan:
I'll double check. In my initial attempts I wasn't able to get this work without the base64 workaround, but indeed if this could be avoided that'd be great and I hadn't changed the mime-type yet.

tdomhan:
I tried setting the mime type to `text/plain;charset=UTF-8` but run into the same issue. From what I'm gathering this would only affect the body and the header only allows `ISO-8859-1`.

jkelleyrtp:
Ah good point, we had to switch to the header from the body since wry was stripping the body on android. I'll check and see if that's still the case...

tdomhan:
Were you able to test whether this can be moved back to the body or would it be ok for the time being to move to base64? 
Potentially base64 could also be restricted to the header and Android and in the receiving part we could check whether the header exists and decode as base64 and otherwise use the body. What do you think?

jkelleyrtp (approved):

PR #3392, opened 2024-12-18
Make the file engine feature in the html crate a no-op

### LINKED ISSUE #3383: File engine does not exist on the server
**Problem**

This code should compile:
```rust
// dioxus = { version = "0.6", features = ["server", "fullstack"] }
use dioxus::prelude::*;
use dioxus_elements::FileEngine;

fn main() {}
```

**Environment:**
 - Dioxus version:  `master`
 - Rust version:    `nightly`
 - OS info:         MacOS
 - App platform:    html based platforms

**Questionnaire**

ealmloff:
This PR removes all usage of the the file engine feature flag in the html crate. This feature flag is useful to reduce the binary size of web builds which used to live in the html crate. Now that the bindings are moved to the web crate, we can remove the feature flag from dioxus-html.

This PR does not actually remove the feature flag because that would be a breaking change. This PR does enable a method on a trait when the `file_engine` trait is disabled, but that method has a default implementation so adding the method should be non-breaking.

I manually tested building the dioxus-web crate with the default features disabled and only the `mounted`, `file_engine`, or `hydrate` features enabled and all three builds succed.

Fixes #3383


jkelleyrtp (approved):
one day we can remove all the weird flags

PR #3393, opened 2024-12-18
fix(cli): Updated logic to resolve out_dir paths relative to workspace

Nathy-bajo:
This fix ensures the out_dir configuration in Dioxus.toml is respected for web builds, allowing users to specify custom output directories. Key updates include:

- Added out_dir handling to resolve both relative and absolute paths.
- Ensured the directory is created if missing, with error logging.
- Introduced an out_dir field in ApplicationConfig (default: "dist") and integrated it into DioxusConfig for backward compatibility.
Users can now customize the build output directory seamlessly.

### DIFF
--- a/packages/cli/src/config/app.rs
+++ b/packages/cli/src/config/app.rs
@@ -8,8 +8,15 @@ pub(crate) struct ApplicationConfig {
 
     #[serde(default)]
     pub(crate) sub_package: Option<String>,
+
+    #[serde(default = "out_dir_default")]
+    pub(crate) out_dir: PathBuf,
 }
 
 pub(crate) fn asset_dir_default() -> PathBuf {
     PathBuf::from("assets")
+}
+
+pub(crate) fn out_dir_default() -> PathBuf {
+    PathBuf::from("dist")  
 }
--- a/packages/cli/src/config/dioxus_config.rs
+++ b/packages/cli/src/config/dioxus_config.rs
@@ -24,6 +24,7 @@ impl Default for DioxusConfig {
             application: ApplicationConfig {
                 asset_dir: asset_dir_default(),
                 sub_package: None,
+                out_dir: out_dir_default(),
             },
             web: WebConfig {
                 app: WebAppConfig {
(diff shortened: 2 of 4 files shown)


richard-viney:
The code looks reasonable to me. Is it based on what was present in 0.5?

Nathy-bajo:
> The code looks reasonable to me. Is it based on what was present in 0.5?

No, I noticed the issue, thought of a solution, and then tried implementing it to fix the problem.

richard-viney:
All good, I was just curious how it ended up being by lost in the new version.

Is it possible to add a unit test for this functionality?

Nathy-bajo:
> All good, I was just curious how it ended up being lost in the new version.
> 
> Is it possible to add a unit test for this functionality?

Yes!

jkelleyrtp:
We dropped `outdir` as the primary build dir in 0.6 since that was causing lots of bugs on its own with multiple profiles/platforms/projects/etc.

To properly isolate builds from each other, dx needs to be somewhat opaque with its output structure.

You can use `dx bundle --json-output` and some jq to parse the output which might be better for automation cases.

Given how many folks rely on `outdir`, I think we should keep it though. However, my vote is to only have an outdir in release mode/bundle and/or make it a cli argument only.

`dx` now uses the build_dir for lots of stuff - not only emitting a `/public` dir and a server.

vnermolaev:
I encountered this issue while following the instructions in the [Dioxus Cookbook](https://dioxuslabs.com/learn/0.6/cookbook/publishing) for publishing a website on GitHub Pages. The corresponding entry should also be updated based on the decision made regarding `out_dir`.

jkelleyrtp:
I'm going to combine this with the #3572 PR to copy `dx bundle` artifacts to the outdir if it's specified. `dx build --release` doesn't generate the same structure as `dx bundle` and generally can't be relied upon to be stable in form between versions.

I don't want us to expose the internals of the dioxus build directories since it's very hard to keep those "clean." We have a similar problem to Cargo's `target` directory where we need to be able to change how we do builds between dioxus versions and don't want to expose the structure.

jkelleyrtp:
Implemented this in #3572  - you should've gotten co-author creds. Thank you!

PR #3395, opened 2024-12-18
chore(CLI): clarified `dx new` post-create message

### LINKED ISSUE #3394: Improve message after the project is created
## Feature Request

The default message

```
`cd` to your project and run `dx serve` to start developing.
If using Tailwind, make sure to run the Tailwind CLI.
More information is available in the generated `README.md`.
```

is misleading since it will cause an error if you just run `dx serve` with `is_fullstack=true`.

```sh
dx new test --subtemplate Bare-Bones -o is_fullstack=true -o is_router=false -o is_tailwind=false -o default_platform=web
```

## Implement Suggestion

#3395

Andrew15-5:
Fixes #3394

### DIFF
--- a/packages/cli/src/cli/create.rs
+++ b/packages/cli/src/cli/create.rs
@@ -210,7 +210,18 @@ pub(crate) fn post_create(path: &Path) -> Result<()> {
     let mut file = std::fs::File::create(readme_path)?;
     file.write_all(new_readme.as_bytes())?;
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
We fixed the bug in `dx` that is related to why this message needs to exist - you can create new projects that have fullstack and run `dx serve` now!

PR #3403, opened 2024-12-19
Fix/server macro defaults

Nathy-bajo:
Enhanced the #[server] macro documentation for improved clarity and usability. Key updates:

- Expanded explanations of arguments with examples (e.g., prefix, input, output).
- Showcased advanced features like middleware layers and context sharing.
- Refactored the macro implementation to replace hardcoded default values for better readability and maintainability.
- Enhanced error handling with detailed messages to help developers debug issues effectively.

### DIFF
--- a/packages/cli/src/config/app.rs
+++ b/packages/cli/src/config/app.rs
@@ -8,8 +8,15 @@ pub(crate) struct ApplicationConfig {
 
     #[serde(default)]
     pub(crate) sub_package: Option<String>,
+
+    #[serde(default = "out_dir_default")]
+    pub(crate) out_dir: PathBuf,
 }
 
 pub(crate) fn asset_dir_default() -> PathBuf {
     PathBuf::from("assets")
+}
+
+pub(crate) fn out_dir_default() -> PathBuf {
+    PathBuf::from("dist")  
 }
--- a/packages/cli/src/config/dioxus_config.rs
+++ b/packages/cli/src/config/dioxus_config.rs
@@ -24,6 +24,7 @@ impl Default for DioxusConfig {
             application: ApplicationConfig {
                 asset_dir: asset_dir_default(),
                 sub_package: None,
+                out_dir: out_dir_default(),
             },
             web: WebConfig {
                 app: WebAppConfig {
--- a/packages/cli/src/dioxus_crate.rs
+++ b/packages/cli/src/dioxus_crate.rs
@@ -97,9 +97,21 @@ impl DioxusCrate {
     /// is "distributed" after building an application (configurable in the
     /// `Dioxus.toml`).
     fn out_dir(&self) -> PathBuf {
-        let dir = self.workspace_dir().join("target").join("dx");
-        std::fs::create_dir_all(&dir).unwrap();
-        dir
+        // Resolve the out_dir based on the configuration or default
+        let out_dir_config = self.config.application.out_dir.clone();
+        // Resolve relative paths against the workspace root
+        let resolved_out_dir = if out_dir_config.is_relative() {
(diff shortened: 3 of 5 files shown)


ealmloff:
Thanks for working on this! It looks like some unrelated changes to base path leaked into this PR.

The two main fields that need documentation are the input and output fields. Since they effect both the encoding and the trait bounds on the inputs and outputs, it would be useful to include examples with the full server function, instead of just the encoding type. Here are a few examples of the bounds they imply:
```rust
/// Setting the type to `StreamingJson` means you need to return something that implements From<JsonStream<T>>
/// where T implements serde::serialize and serde::de::DeserializeOwned
#[server(output = StreamingJson)]
pub async fn json_stream_fn() -> Result<JsonStream<String>, ServerFnError> {
    todo!()
}

/// Setting the type to `StreamingText` means you need to return something that implements From<TextStream>
#[server(output = StreamingText)]
pub async fn text_stream_fn() -> Result<TextStream, ServerFnError> {
    todo!()
}

/// Setting the type to `PostUrl` means you need to return something that implements Serialize and Deserialize.
/// This uses `serde_qs` which imposes the following requirements:
/// - Be less than 5 levels deep
/// - Contain no `serde(flatten)` attributes
#[server(output = PostUrl)]
pub async fn form_fn() -> Result<TextStream, ServerFnError> {
    todo!()
}
```

Nathy-bajo:
> Thanks for working on this! It looks like some unrelated changes to base path leaked into this PR.
> 
> The two main fields that need documentation are the input and output fields. Since they effect both the encoding and the trait bounds on the inputs and outputs, it would be useful to include examples with the full server function, instead of just the encoding type. Here are a few examples of the bounds they imply:
> 
> ```rust
> /// Setting the type to `StreamingJson` means you need to return something that implements From<JsonStream<T>>
> /// where T implements serde::serialize and serde::de::DeserializeOwned
> #[server(output = StreamingJson)]
> pub async fn json_stream_fn() -> Result<JsonStream<String>, ServerFnError> {
>     todo!()
> }
> 
> /// Setting the type to `StreamingText` means you need to return something that implements From<TextStream>
> #[server(output = StreamingText)]
> pub async fn text_stream_fn() -> Result<TextStream, ServerFnError> {
>     todo!()
> }
> 
> /// Setting the type to `PostUrl` means you need to return something that implements Serialize and Deserialize.
> /// This uses `serde_qs` which imposes the following requirements:
> /// - Be less than 5 levels deep
> /// - Contain no `serde(flatten)` attributes
> #[server(output = PostUrl)]
> pub async fn form_fn() -> Result<TextStream, ServerFnError> {
>     todo!()
> }
> ```

I have added the examples as requested @ealmloff 😄

packages/cli/src/config/app.rs
```diff
@@ -8,8 +8,15 @@ pub(crate) struct ApplicationConfig {
...
+    #[serde(default = "out_dir_default")]
+    pub(crate) out_dir: PathBuf,
 }
 
 pub(crate) fn asset_dir_default() -> PathBuf {
     PathBuf::from("assets")
 }
+
+pub(crate) fn out_dir_default() -> PathBuf {
+    PathBuf::from("dist")
```
hackartists, line 21: It may break the previouse deployment script due to changing default build artifact directory. How about returning PathBuf for `target/dx`, which is current out directory in `dioxus_crate.rs:100`

Nathy-bajo: @hackartists Hello!
I think this is the PR you're referring to https://github.com/DioxusLabs/dioxus/pull/3393
And I already made that change there

PR #3404, opened 2024-12-19
Move web.resource.style log to a warning

ealmloff:
The deprecation warning was at the `debug` level which only gets logged if customize the logging level manually when serving (turning on trace mode once the CLI is running does not show any debug logs). Alternatively, we could move the warning to `dx check`


jkelleyrtp:
ended up rolling this into #3393

PR #3419, opened 2024-12-20
fix(fullstack): Allow configuration of the public_path

brianmay:
This was annoying me. I don't think the default is always sensible.

I ran the tests and they passed, but haven't yet tried running the code.

### DIFF
--- a/packages/fullstack/src/serve_config.rs
+++ b/packages/fullstack/src/serve_config.rs
@@ -215,6 +215,10 @@ impl TryInto<ServeConfig> for ServeConfigBuilder {
 
 /// Get the path to the public assets directory to serve static files from
 pub(crate) fn public_path() -> PathBuf {
+    if let Ok(path) = std::env::var("DIOXUS_PUBLIC_PATH") {
+        return PathBuf::from(path);
+    }
+
     // The CLI always bundles static assets into the exe/public directory
     std::env::current_exe()
         .expect("Failed to get current executable path")


brianmay:
I haven't tried to update the CLI tool, possible that should be updated also. Not sure. Haven't thought that far ahead yet.

brianmay:
Tested this change now using real code and it is working.

brianmay:
Could also replace `public_path()` to use some environment variable. Might make it easier to work with the CLI that way.

But being able to customize value from application I think is good also.

packages/fullstack/src/server/mod.rs
```diff
@@ -133,15 +134,15 @@ pub trait DioxusRouterExt<S> {
...
+    ///         .serve_static_assets(std::path::Path::new("/assets"))
     ///         // Server render the application
     ///         // ...
     ///         .into_make_service();
     ///     let listener = tokio::net::TcpListener::bind(addr).await.unwrap();
     ///     axum::serve(listener, router).await.unwrap();
     /// }
     /// ```
-    fn serve_static_assets(self) -> Self
+    fn serve_static_assets(self, pubic_path: &Path) -> Self
```
jkelleyrtp, line 145: Modifying this signature is not semver compatible meaning we wouldn't be able to release this in version 0.6.2.

I think we can add this with a default impl but since  DioxusRouterExt is not sealed, any changes to its signature will need to follow semver rules.

https://doc.rust-lang.org/cargo/reference/semver.html#trait-item-signature

jkelleyrtp (changes requested):
Overall looks good but adding the public_path into the signatures is semver breaking.

I think it'd be better to read an env var instead since I'd rather change the public_dir depending on my deploy setup rather than at compile time.

brianmay:
OK. I must admin I never considered being semver compatible.

Unless any objections, I will revert these changes for one that replaces `public_path()` with something that reads an environment variable instead. Which is totally fine for my purposes.

PR #3437, opened 2024-12-22
Make `Routable` derive macro hygienic

MrGVSV:
Makes the `Routable` derive macro hygienic by wrapping the output in an anonymous scope.

This allows multiple routers to be defined in the same scope if necessary (e.g. for testing purposes, macro generation, etc.).


jkelleyrtp:
Thank you!

PR #3441, opened 2024-12-23
fix(cli): show more shortcuts

### LINKED ISSUE #3430: Add an ability to pause watch
## Feature Request

Add a shortcut to dioxus cli to pause watching for files changes

Use case: some components always triggers full rebuilds, that's can be annoying when you in process of writing something

istudyatuni:
Close #3430

### DIFF
--- a/packages/cli/src/serve/mod.rs
+++ b/packages/cli/src/serve/mod.rs
@@ -59,7 +59,10 @@ pub(crate) async fn serve_all(mut args: ServeArgs) -> Result<()> {
                 • Press `ctrl+c` to exit the server
                 • Press `r` to rebuild the app
                 • Press `o` to open the app
+                • Press `p` to toggle automatic rebuilds
                 • Press `v` to toggle verbose logging
+                • Press `t` to toggle tracing
+                • Press `c` to clean logs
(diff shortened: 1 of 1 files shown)


packages/cli/src/serve/mod.rs
```diff
@@ -59,7 +59,10 @@ pub(crate) async fn serve_all(mut args: ServeArgs) -> Result<()> {
                 • Press `ctrl+c` to exit the server
                 • Press `r` to rebuild the app
                 • Press `o` to open the app
+                • Press `p` to toggle automatic rebuilds
                 • Press `v` to toggle verbose logging
+                • Press `t` to toggle tracing
+                • Press `c` to clean logs
                 • Press `/` for more commands and shortcuts
```
istudyatuni, line 66: Probably this should be
```suggestion
                • Press `/` for more commands
```
because there are no shortcuts

jkelleyrtp:
Sorry for letting this sit - I didn't think making the intro screen longer was the right thing to do here, so I took the liberty of implementing your change in #3586 as extra data on the help menu.
Screenshot 2025-01-17 at 6 05 27 PM

Issue #3083, opened 2024-10-22
Cargo.lock for dioxus-cli is outdates


uzytkownik:
**Problem**

Some packaging tools like nix use Cargo.lock to pick up versions of the dependencies. Currently dioxus-cli Cargo.lock uses 0.2.93 webasm-bindgen but rest of Dioxus uses 0.2.95 version of bindgen.

**Steps To Reproduce**

- Use packager using Cargo.lock
- Create example
- Run `dx serve`

**Expected behavior**

Sample application runs

**Actual behaviour**

`dx` fails due to mismatch of versions

**Environment:**
 - Dioxus version:  0.5.7
 - Rust version: 1.80.0
 - OS info: Nix
 - App platform: all

elanora96:
I can reproduce this as well on my machine as well

<details>
<summary>
<h1>flake.nix</h1>
</summary>

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
    flake-utils.url = "github:numtide/flake-utils";
  };

  outputs =
    {
      self,
      nixpkgs,
      flake-utils,
    }:
    flake-utils.lib.eachDefaultSystem (
      system:
      let
        pkgs = import nixpkgs { inherit system; };
      in
      {
        devShell =
          with pkgs;
          mkShell {
            buildInputs = [
              cargo
              rustc
              rustfmt
              rustPackages.clippy
              dioxus-cli
              lld
            ];
          };
      }
    );
}
```

</details>

<details>
<summary>
<h1>Enviroment</h1>
</summary>

```sh
$ dx --version && cargo --version && rustc --version && uname -r
dioxus 0.5.7
cargo 1.81.0 (2dbb1af80 2024-08-20)
rustc 1.81.0 (eeb90cda1 2024-09-04) (built from a source tarball)
6.11.4-arch1-1
```

</details>

<details>
<summary>
<h1>Output of `dx serve`</h1>
</summary>

```sh
$ dx serve
  👑 Build done.                                                                                        
thread 'main' panicked at src/builder.rs:188:14:
called `Result::unwrap()` on an `Err` value: 

it looks like the Rust project used to create this wasm file was linked against
version of wasm-bindgen that uses a different bindgen format than this binary:

  rust wasm file schema version: 0.2.95
     this binary schema version: 0.2.93

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating this binary or
[...]

PR #3442, opened 2024-12-23
Add a nix devShell for webapp development

j4ger:
In addition to the existing nix devShell config, which is for developing `dioxus` itself, this pull request adds a `web` devShell for developing web applications with dioxus.

Although `dioxus-cli` is available in nixpkgs, it does not update with main branch and it doesn't integrate well with versioned rust environment (due to the missing or non-functioning `rustup`). The new devShell config builds `dx` based on the repo and inserts a mock-up `rustup` to bypass `dx`'s toolchain verification process, allowing it to automatically install the corresponding versioned `wasm-bindgen` (I couldn't come up with a better way to make it use a `wasm-bindgen` provided by the flake, other than wrapping `dx` with modified `$XDG_STATE_HOME`, which would also interfere with other parts of the program using related envvars).

Fixes https://github.com/DioxusLabs/dioxus/issues/3083 (sort of?)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1722555600,
-        "narHash": "sha256-XOQkdLafnb/p9ij77byFQjDf5m5QYl9b2REiVClC+x4=",
+        "lastModified": 1733312601,
+        "narHash": "sha256-4pDvzqnegAfRkPwO3wmwBhVi/Sye1mzps0zHWYnP88c=",
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "rev": "8471fe90ad337a8074e957b69ca4d0089218391d",
+        "rev": "205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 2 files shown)


j4ger:
> Is there no option for `dioxus-cli` to be changed so you can override tool behavior directly with a CLI flag? Relying on compilation solution where `dx` fetches its own tools isn't a very good way of handling reproducibility, IMO, and would also break Nix pure build of Dioxus apps.
> 
> It would work as a temporary stopgap to allow impure dev env to work, but why not just PR to add a dedicated environment variable or CLI flag for controlling this? Has DioxusLabs already rejected such a solution?

I agree that the current behavior of `dx` does not comply with the strict "nix philosophy", as it fetches its own build artifacts which is not isolated, but at least its version is fixed (mostly, since it uses the version number as the identifier, not hash).

The ideal solution would be some sort of envvars to specify wasm-bindgen path, but the exact behavior of the verification process needs to be changed.

Although, it just came to me that the proper behavior would be something similar to the rust toolchain itself: `rustup` has its own mechanism of determining whether it's part of a "managed" installation, if so, it would not attempt to modify the environment in any way, but simply provide diagnosis.

crabdancing:
One problem I ran into is that your metadata is wrong.

```
nix run github:j4ger/dioxus#packages.x86_64-linux.dx
error: unable to execute '/nix/store/1drad9mli1zp3cqrqrxlqx3ydc2x5v3x-dioxus-cli-0.6.1/bin/dioxus-cli': No such file or directory
```

Because the package is named `dioxus-cli`, by default, it assumes the bin name is the same. Off the top of my head, I think what you want is `meta.mainProgram = "dx";` in your build expression. I could be misremembering / not accounting for naersk-specific details.

I have a fair bit of experience in Nix, so I'd be willing to help develop the flake further if you want. Exposing the local `dx` utility in the package flake is definitely a step in the right direction.

j4ger:
> One problem I ran into is that your metadata is wrong.
> 
> ```
> nix run github:j4ger/dioxus#packages.x86_64-linux.dx
> error: unable to execute '/nix/store/1drad9mli1zp3cqrqrxlqx3ydc2x5v3x-dioxus-cli-0.6.1/bin/dioxus-cli': No such file or directory
> ```
> 
> Because the package is named `dioxus-cli`, by default, it assumes the bin name is the same. Off the top of my head, I think what you want is `meta.mainProgram = "dx";` in your build expression. I could be misremembering / not accounting for naersk-specific details.
> 
> I have a fair bit of experience in Nix, so I'd be willing to help develop the flake further if you want. Exposing the local `dx` utility in the package flake is definitely a step in the right direction.

Thanks for pointing it out. Originally I was fiddling around with the workspace build mechanism since there are multiple crates in the repo. According to `naersk`'s document it should build the crate with the exact supplied name, but it didn't, so I had to use additional buildOptions.

Should be fixed now.

j4ger:
Closing as there are proposed alternative solutions in 0.6.2

PR #3455, opened 2024-12-26
Keep rebuilding toast alive longer

### LINKED ISSUE #2778: Keep `rebuilding` toast alive until dismissed manually or there's a build error
## Feature Request

The "your app needs to be rebuilt" toast is great, but it does dismiss a little early. I'd expect a loading spinner on the toast and then it stays alive until dismissed manually since the page is going to reboot anyways.

Also if the page rebuilding failed it'd be nice to get an error in the toast and/or bounce to the "build failed" screen when the toast is clicked.

liigo:
Fix #2778


jkelleyrtp (approved):
Thank you!

Issue #3417, opened 2024-12-20
Allow disabling CLI dependency downloads 


CathalMullan:
## Feature Request

For environments like [NixOS](https://github.com/NixOS/nixpkgs/pull/366178), auto-downloading binaries like `wasm-bindgen` isn't ideal.

Similar issue for the `rustup` usage for auto-downloading of toolchain targets. `rustup` may not even be present in certain environments.

Add a way to disable these downloads.

## Implement Suggestion

Add a CLI feature to disable downloads, similar to the "no_downloads" feature of [cargo-leptos](https://github.com/leptos-rs/cargo-leptos). Just look for the tools in $PATH instead.

Instead of using `rustup` to check what targets are currently installed, you could use `rustc --print sysroot` instead.

I have a (messy) example [here](https://github.com/CathalMullan/dioxus/commit/c043c3387669b486d4372c74c134b2f7f49d2637).

jkelleyrtp:
Closed by #3465

PR #3465, opened 2024-12-28
feat(cli): add feature to disable downloads

### LINKED ISSUE #3457: Option to stop Dioxus CLI from ignoring tooling in PATH
## Feature Request

As it stands, the `dx` utility is bundled with / automatically fetches its own `wasm-bindgen-cli` that it uses over the environment one. This is certainly useful on imperative systems, but on declarative ones like Nix, it seems actively harmful to get something to work reliably.

E.g., I see this message:

```
Caused by:
    0: Failed to generate wasm-bindgen bindings
    1:

       it looks like the Rust project used to create this Wasm file was linked against
       version of wasm-bindgen that uses a different bindgen format than this binary:

         rust Wasm file schema version: 0.2.99
            this binary schema version: 0.2.97
```

I add this to my dev environment:

```nix
            (pkgs.wasm-bindgen-cli.override {
              version = "0.2.99";
              hash = "sha256-1AN2E9t/lZhbXdVznhTcniy+7ZzlaEp/gwLEAucs6EA=";
              cargoHash = "sha256-DbwAh8RJtW38LJp+J9Ht8fAROK9OabaJ85D9C/Vkve4=";
            })
```

My Nix development environment is now correctly pinning the wasm-bindgen-cli version:

```sh
wasm-bindgen -V
wasm-bindgen 0.2.99
```

But `dx` ignores this and stubbornly keeps using the wrong version. 

## Implement Suggestion

It should be possible to pass a simple flag, or environment variable, to override `Dioxus CLI`'s behavior, so that rather than trying to handle which dependencies are used, it just trusts the environment (PATH, etc) to be correct. Perhaps a flag like `--use-local-tooling`? This ideally would stop any attempt to automatically fetch any external tooling, and instead simply look for it in PATH, and error immediately if needed but not present.

That way, reproducible build processes like Nix, Docker, and whatnot, can simply pin their own environment at a specific version, and have it just work.

This would (maybe?) obviate PR #3442, fix issue #3083, and pave the way toward better reproducible CI for all Dioxus apps.

CathalMullan:
Resolves #3457, #3417 

- Added 'no-downloads' feature to CLI.
- Switched to using rustc instead of rustup to determine what targets are installed.
- Lookup wasm-bindgen-cli from path when 'no-downloads' is enabled. 
- Don't try and install targets using rustup when 'no-downloads' is enabled.
- Enable 'no-downloads' feature for Nix CI job.

Would have liked to have added tests, but mocking PATH is a little tricky to do safely across multiple tests.

Mainly need this for [NixOS](https://github.com/NixOS/nixpkgs/pull/366178), but I suppose other environments may not want auto-downloading of tools either.

Also worth mentioning that the NixOS wiki now has a Dioxus page: https://wiki.nixos.org/wiki/Dioxus


jkelleyrtp:
Thanks for the PR! I made a few changes
- Remove the need for a trait - just compressed the two implementations together
- Wired up no-downloads to be available regardless of the feature. You can now add this dynamically with NO_DOWNLOADS=1 or set the no_downloads cli setting at runtime.
- left the original structure in-tact

Whens checks pass I'll merge and release in 0.6.2 shortly

jkelleyrtp (approved):

Issue #3342, opened 2024-12-11
Second window not closeable


tdomhan:
**Problem**

Trying the multiwindow example with 0.6 I was not able to close the new window that gets successfully opened.

- Create a new window similar to https://github.com/DioxusLabs/dioxus/blob/main/examples/multiwindow.rs
- Click on the x to close the window.

**Expected behavior**

The window to close

**Screenshots**

See: https://discord.com/channels/899851952891002890/1303247605849128960/1303247605849128960

**Environment:**
 - Dioxus version:  0.6
 - Rust version:   rustc 1.82.0-nightl
 - OS info:    Mac OS 
 - App platform:  desktop

bartoszkp:
Exactly the same issue seems to be present again in 0.6.3

arpan-vs:
I also got the same issue in 0.6.3 on windows 11.

PR #3466, opened 2024-12-29
Fix: WindowCloseBehaviour

Klemen2:
Closes: #3113
Fixes window not hiding for WindowCloseBehaviour:LastWindowHides, 

WindowCloseBehaviour:LastWindowHides had a comment "Closing the last window will not actually close it, just hide it" but it was implemented in a way that hides all the windows so I added WindowCloseBehaviour:WindowHides that does that, and changed LastWindowHides so that it only hides on the last one, and removes the rest.

Changed multiwindow example to include tray icon & WindowCloseBehaviour:WindowHides

Temporary fix for new window not closing #3342 or at least I think that there is something else going on here and hiding the closed window is just a temporary solution to this problem since it worked in 0.5 without that.

I also changed multiwindow example to use async move, if you don't use it it gets stuck on 
https://github.com/DioxusLabs/dioxus/blob/30f760ca1503a10694fee1cd345de8d0174d14c5/packages/desktop/src/webview.rs#L383

### DIFF
--- a/examples/multiwindow.rs
+++ b/examples/multiwindow.rs
@@ -9,7 +9,7 @@ use dioxus::desktop::*;
 use dioxus::prelude::*;
 fn main() {
     dioxus::LaunchBuilder::desktop()
-        .with_cfg(Config::new().with_close_behaviour(WindowCloseBehaviour::LastWindowHides))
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
Hey, thanks for the PR! I extracted your fixes into #3499 which should fix both issues. Will make sure to add your name to author list!

Issue #3467, opened 2024-12-29
Regression: dioxus-cli fails to create multiple processed copies of a same image with different options


Destroyerrrocket:
If you use two asset! macros, like:
```
use manganis::*;
pub struct BlogEntry {
    pub image_file_thumbnail: Asset,
    pub image_file_blog: Asset,
}

const BLOG: BlogEntry = BlogEntry {
    image_file_thumbnail: manganis::asset!(
        "./top.jpg",
        ImageAssetOptions::new()
            .with_size(ImageSize::Manual {
                width: 384,
                height: 384
            })
            .with_avif()
    ),
    image_file_blog: manganis::asset!(
        "./top.jpg",
        ImageAssetOptions::new().with_avif()
    ),
};
```
dioxus cli will generate, inconsistently, either one of the two images. 

**Steps To Reproduce**
Create a trivial project with the code above with a jpg
compile by `dx build --trace --verbose`
Observe that a single image appears in target/dx/<project>/debug/web/public/<assets_folder>

**Expected behavior**
One file for each configuration, instead of a single one.

My guess as to why this is happening is that the AssetManifest at [packages/cli-opt/src/lib.rs](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli-opt/src/lib.rs#L24) is mistakenly a hashmap over the absolute path of the original source of asset. This is wrong as this will mistakenly merge these resources. I think this would be easily fixed by changing this to a vector, or, alternatively, hash over the output directories as those are already made unique by mangaris.

Please do also note that in the past this used to work

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version:    1.82.0
 - OS info:         Arch Linux
 - App platform:   web

**Questionnaire**
I would like to fix and I think I have a solution

Destroyerrrocket:
While I can take a look at fixing this, I'd like some input on what/why this behavior was changed and overall whether my initial guess of the issue is in the right direction.

Destroyerrrocket:
Proposed fix in #3469

PR #3469, opened 2024-12-29
Make the asset collection keep track of all of the destinations (#3467)

Destroyerrrocket:
This fixes the issue I mentioned in #3467, where the use of two asset! macro usages was not being treated correctly by the handler.

BEFORE you merge: Please notice that the code at: packages/cli/src/serve/handle.rs is ""wrong"" (and it was wrong before as well). It does a simple file copy without checking if it needs to be processed in any way, which is likely to lead to issues when dealing with images. I have not encountered this problem though and I don't know what purpose it serves. If you consider this to be an issue, I'd encourage you to make an issue about it.

### DIFF
--- a/packages/cli-opt/src/lib.rs
+++ b/packages/cli-opt/src/lib.rs
@@ -20,8 +20,8 @@ pub use file::process_file_to;
 /// This will be filled in primarily by incremental compilation artifacts.
 #[derive(Debug, PartialEq, Default, Clone, Serialize, Deserialize)]
 pub struct AssetManifest {
-    /// Map of bundled asset name to the asset itself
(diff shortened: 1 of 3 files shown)


ealmloff:
> `HashMap<PathBuf, HashMap<PathBuf, BundledAsset>>`
It might make sense to just store `HashMap<PathBuf, HashSet<BundledAsset>>` instead of a nested hashmap since the `BundledAsset` struct already contains the bundled path and it looks like the hashmap value is never used.

I opened an issue here about hot reloaded assets not getting processed: https://github.com/DioxusLabs/dioxus/issues/3471

ealmloff:
This was fixed in https://github.com/DioxusLabs/dioxus/pull/3988

Issue #3233, opened 2024-11-18
dx bundle fail on windows


pojoin:
**ERROR Failed to bundle project: BundlerError**

```
371.923s  INFO Copying app to output directory...
 375.582s ERROR Failed to bundle project: BundlerError(
    Error {
        context: "error running candle.exe",
        source: GenericError(
            "failed to run C:\\Users\\hcq\\AppData\\Local\\tauri\\WixTools314\\candle.exe",
        ),
    },
)
 375.594s ERROR err=BundleFailed(BundlerError(error running candle.exe
```

photino:
The same error.

dunkyl:
I got it sort of working, after following workarounds:

0. (seems like many have already done this) added `icons/icon.ico`

1. I added to Dioxus.toml:
```toml
 [bundle]
publisher = "my example publisher"
identifier = "com.example.app"
```
2. I renamed `target/debug/<APPNAME>/bundle/windows/<APPNAME>` to `<APPNAME>.exe`

3. I manually ran WiX:
```sh
~\AppData\Local\tauri\WixTools314\light.exe -ext WixUIExtension -loc .\target/dx/<APPNAME>/bundle/windows\wix\x64\locale.wxl .\target\dx\<APPNAME>\bundle\windows\wix\x64\main.wixobj
```
This successfully generated a working installer, however, app lacked permission to create its WebView2 cache in the default installed location (`C:\Program Files\<APPNAME>\`). Installing, to a different location, under my user folder, worked fine.

dunkyl:
dx --version: `dioxus 0.6.0 (ccf500c)`

liigo:
The same error:

```shell
  17.829s  INFO Bundling app...
  17.852s  INFO Copying asset (0/3): C:\Users\liigo\tmp\hidx06\myhotdog\assets\header.svg
  17.852s  INFO Copying app to output directory...
  18.820s ERROR Failed to bundle project: BundlerError(
    Error {
        context: "error running candle.exe",
        source: GenericError(
            "failed to run C:\\Users\\liigo\\AppData\\Local\\tauri\\WixTools314\\candle.exe",
        ),
    },
)
  18.823s ERROR err=BundleFailed(BundlerError(error running candle.exe

Caused by:
    `failed to run C:\Users\liigo\AppData\Local\tauri\WixTools314\candle.exe`))
```

But tauri build works well (both tauri and dioxus use *the same wix* as i known):

```shell
[...]

PR #3482, opened 2025-01-02
the main exe to be bundled should have '.exe' extension on windows

liigo:
`hotdog.exe` instead of `hotdog` in directory `target\dx\hotdog\bundle\windows\`

this is one of the steps to fix https://github.com/DioxusLabs/dioxus/issues/3233 ( anothe one is to update dioxus-template, adding bundle.identifier to Dioxus.toml, https://github.com/DioxusLabs/dioxus-template/pull/58 )


jkelleyrtp:
Good find, thank you! We need to add bundling to CI to ensure we don't have future regressions.

PR #3483, opened 2025-01-02
Add helper methods to provide a single clone server only context

ealmloff:
The launch builder has several different variants of context provider methods for thread local context, and clone context. This PR adds similar methods to the server config for server-only context. It also adds examples of server only context created inside of the tokio runtime as doc examples in these methods


damccull:
Looking forward to this. Just found the manual boxing I need to make it work without this!

jkelleyrtp (approved):

PR #3484, opened 2025-01-02
Fix Routable::parent with hash segments and query params

ealmloff:
The `parent` and `is_child_of` methods are currently broken if the route has query or hash segments. This was causing an infinite loop in the docsite with hash fragments in https://github.com/DioxusLabs/docsite/pull/365.

This PR ignores hash segments and query segments when determining the parent/child relationship of routes. It also adds tests for the parent and is_child_of methods


jkelleyrtp (approved):
I wonder if it might be better to use a crate like URL/URI for parsing the various segments since it's likely we still have little bugs here and there.

Thanks for the test though!

PR #3485, opened 2025-01-02
remove version pin on tao

### LINKED ISSUE #3314: Can't move or resize window on Gnome + Wayland
**Problem**

When using dioxus desktop 0.6.0 on Gnome + Wayland the window can't be resized or moved. 

I think it's related to this tauri issue https://github.com/tauri-apps/tauri/issues/10686

The only workaround I found for now is to set GDK_BACKEND=x11 to run the window with XWayland. The problem only concerns Gnome + Wayland and not KDE + Wayland.

**Steps To Reproduce**
- be on Gnome + Wayland
- here is my very simple code :
```rust
use dioxus::desktop::Config;
use dioxus::prelude::*;

fn main() {
    LaunchBuilder::desktop()
        .with_cfg(
            Config::new().with_menu(None), // remove the default menu
        )
        .launch(App)
}

#[component]
fn App() -> Element {
    rsx! {HelloWorld {}}
}

#[component]
fn HelloWorld() -> Element {
    rsx! {span {"Hello, World!"}}
}
```

**Environment:**
 - Dioxus version:  0.6.0
 - Rust version:    1.83.0
 - OS info:         Arch Linux
 - App platform:    desktop

sehnryr:
Version 0.30.3 of tao includes the fix for the window resize issue on Wayland.

Fixes #3314

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13469,9 +13469,9 @@ dependencies = [
 
 [[package]]
 name = "tao"
-version = "0.31.0"
+version = "0.30.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc6b53216f32e60efc27dfa111268481e4dfba53e553e4cdebcaed9db36c11bb"
+checksum = "6682a07cf5bab0b8a2bd20d0a542917ab928b5edb75ebd4eda6b05cbaab872da"
 dependencies = [
  "bitflags 2.6.0",
  "cocoa 0.26.0",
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
This is a bit unfortunate - tao is part of our public API and bumping it would be a breaking change (ie 0.6 -> 0.7). A long-term way of fixing this would for dioxus to get onto a 1.0.0 so we could use the minor version for dependency upgrades or removing tao from our public API altogether (which we need to do for winit, eventually).

jkelleyrtp (approved):
Thank you!

PR #3493, opened 2025-01-04
fix(cli): Log errors when `adb` command fails

### LINKED ISSUE #3436: Android build doesn't fail when `adb` missing from path
**Problem**

The `dx serve --platform android` doesn't check for `adb` in path before trying to launch.  If it's missing, it will `Build complete successfully` and not launch.

**Steps To Reproduce**

Remove `$HOME/Library/Android/sdk/platform-tools` from path i.e. `adb`.  Then run `dx serve --platform android` with an Android emulator started.  The app will build but not launch and not inform the user.

**Expected behavior**

An error should occur if `adb` is not found.

**Environment:**
 - Dioxus version: .6.0.1
 - Rust version:    latest
 - OS info:         Mac OSX
 - App platform:    android

Plebshot:
As described in #3436, not having `adb` in the path (or having other problems with `adb`) fails silently.

The problem here is that the two `adb` commands for installing and starting the app are done on a new tokio background task without logging the result. Errors are propagated to the tasks handle, but the handle is detached and therefore never checked.

I wasn't sure what the best way to handle detached errors in the CLI is, so I went with simply logging errors instead of propagating them. Some open questions remain:

- Should we explicitly add `adb` as requirement in the docs for the [mobile docs](https://dioxuslabs.com/learn/0.6/guides/mobile/)?
- Should we check for system dependencies beforehand for clearer error messages?

Fixes #3436

### DIFF
--- a/packages/cli/src/serve/handle.rs
+++ b/packages/cli/src/serve/handle.rs
@@ -128,7 +128,7 @@ impl AppHandle {
 
             // https://developer.android.com/studio/run/emulator-commandline
             Platform::Android => {
-                self.open_android_sim(envs).await?;
+                self.open_android_sim(envs).await;
                 None
             }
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
Thanks for the PR! 

Eventually we should attach this task to a task pool or something that propagates a failure out, but the logging error is fine for now.

PR #3494, opened 2025-01-04
Get server address from cli in fullstack-auth example. 

### LINKED ISSUE #3363: When using the Axum authentication example, the devserver proxy does not work
**Problem**

The server with the authentication layer is listening on port 3000, and works when using `dx serve`. However, the hot reloader server on port 8080 does not proxy to the other server and throws the error "Backend connection failed. The backend is likely still starting up. Please try again in a few seconds."

I think the issue is [here](https://github.com/DioxusLabs/dioxus/blob/86082cec5cef0721b67f0f4f98e16d66285132a3/packages/cli/src/serve/server.rs#L80) which becomes the address to forward to [here](https://github.com/DioxusLabs/dioxus/blob/86082cec5cef0721b67f0f4f98e16d66285132a3/packages/cli/src/serve/server.rs#L384)

**Steps To Reproduce**

Running the fullstack-auth example has the same issue 

**Expected behavior**

The dev sever should use the correct port to forward requests to. 

**Environment:**
 - Dioxus version:   `master` 
 - Rust version:   1.81.0
 - OS info:         Windows 10
 - App platform:    `fullstack`

**Questionnaire**
I would like to fix and I have a solution 

Add an argument to the serve command to set the inner server address

charles-r-earp:
Fixes #3363.

The example is served to a hardcoded port 3000, while the cli expects to connect to 8080 by default. Using `dioxus_cli_config::fullstack_address_or_localhost`,  the proper address can be used by the server. This is consistent with the documentation for [dioxus-fullstack](https://github.com/DioxusLabs/dioxus/tree/main/packages/fullstack#axum-integration).

Note that because of the issue fixed in #3359, what works is `dx serve --client-features dioxus-web/mounted --open`, where the `mounted` feature enables the web-sys feature `DomRect`, which enables `DomRectReadOnly`.

### DIFF
--- a/examples/fullstack-auth/Cargo.toml
+++ b/examples/fullstack-auth/Cargo.toml
@@ -11,6 +11,7 @@ publish = false
 dioxus-web = { workspace = true, features = ["hydrate"], optional = true }
 dioxus = { features = ["fullstack"], workspace = true }
 dioxus-fullstack = { workspace = true }
+dioxus-cli-config = { workspace = true, optional = true }
 axum = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["full"], optional = true }
 tower-http = { workspace = true, features = ["auth"], optional = true }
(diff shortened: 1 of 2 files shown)


jkelleyrtp (approved):
Thank you!

PR #3508, opened 2025-01-07
Fix the event handler memory leak

### LINKED ISSUE #3421: bug(Memory Leak): Event handlers are not dropped properly
Same issue as in https://github.com/DioxusLabs/dioxus/issues/2272

I have simplified the code a bit, but both leak memory:

```rust
#![allow(non_snake_case)]
use dioxus::prelude::*;

    fn app() -> Element {

            loop {
                tokio::time::sleep(std::time::Duration::from_nanos(1)).await;
                if val == 70 {
                    count.set(0);
                } else {
                    count.set(val + 1);
                }
            }
        }));

        rsx! {
            for el in 0..*count.read() {
                div {
                    key: "{el}",
                    div {
                        onclick: move |_| { println!("click"); },
                    }
                }
            }
        }
    }

        .enable_all()
        .build()?
        .block_on(async {
            loop {
                vdom.wait_for_work().await;

            }
        })
}
```

`counter5.exe` is same code running with Dioxus 0.5. and `counter.exe` is with 0.6.

image

ealmloff:
The event handler into implementation uses the current owner which defaults to the current scope. That causes a memory leak when repeatedly creating new event handlers on the same scope. This PR fixes the issue by overriding the owner to the closure

Fixes #3421

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3641,6 +3641,7 @@ dependencies = [
  "serde",
  "slab",
  "slotmap",
+ "sysinfo",
  "tokio",
  "tracing",
  "tracing-fluent-assertions",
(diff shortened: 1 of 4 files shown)


jkelleyrtp (approved):
It would be worthwhile to add child components to the memory leak test since we tend to get issues on that boundary. Otherwise looks good - but we should definitely find some time to clean up the owner stuff a bit more for the next breaking release.

Issue #3504, opened 2025-01-06
DX bundle fails when no assets are copied


ealmloff:
**Problem**

If you try to bundle a dioxus desktop project without any assets the CLI throws a file system error.

**Steps To Reproduce**

- Clone and bundle this repo: https://github.com/ealmloff/dx-bundle-empty-panic

**Expected behavior**

Bundling a project without assets should work as normal

**Environment:**
 - Dioxus version:  `0.6`
 - Rust version:    `nightly`
 - OS info:         MacOS
 - App platform:    `desktop`

BleemIs42:
PR created: [#3519](https://github.com/DioxusLabs/dioxus/pull/3519)

PR #3519, opened 2025-01-08
Fix dx bundle fails when no assets are copied

BleemIs42:
Fix the bug that `dx bundle` fails when no assets are copied report by [#3504](https://github.com/DioxusLabs/dioxus/issues/3504)


jkelleyrtp (approved):

jkelleyrtp:
Thank you!

PR #3521, opened 2025-01-08
Fix ssg race condition

### LINKED ISSUE #3518: Static site generation race condition
**Problem**

Running static generation is a race condition between a 10 second timer and the site actually rendering. This caused the docsite build to silently fail and deploy the broken site

**Steps To Reproduce**

- Create a new project with this code:
```rust
// [dependencies]
// dioxus = { version = "0.6.0", features = ["fullstack", "router"] }
// tokio = { version = "1.42.0", features = ["full"], optional = true }
// tracing = "0.1.41"

// [features]
// default = []
// web = ["dioxus/web"]
// server = ["dioxus/server", "dep:tokio"]

use dioxus::prelude::*;

fn main() {
    LaunchBuilder::new()
        .with_cfg(server_only! {
            ServeConfig::builder().incremental(
                    IncrementalRendererConfig::new()
                        .static_dir(std::path::PathBuf::from("./docs"))
                        .clear_cache(false)
                )
            .build()
            .expect("Unable to build ServeConfig")
        })
        .launch(|| {
            rsx! {
                Router<Route> {}
            }
        })
}

#[derive(Routable, Clone, PartialEq, Eq, Hash)]
enum Route {
    #[route("/")]
    Home,
    #[route("/other_route")]
    OtherRoute,
}

fn Home() -> Element {
    use_server_future(pause)?;

    rsx! {
        div {
            h1 {"Hello, world!"}
        }
    }
}

fn OtherRoute() -> Element {
    rsx! {
        div {
            h1 {"Hello, world!"}
        }
    }
}

#[server]
async fn pause() -> Result<(), ServerFnError> {
    tokio::time::sleep(std::time::Duration::from_secs(60)).await;
    Ok(())
}

#[server(endpoint = "static_routes")]
async fn static_routes() -> Result<Vec<String>, ServerFnError> {
    Ok(Route::static_routes()
        .into_iter()
        .map(|route| route.to_string())
        .collect::<Vec<_>>())
}
```
- Run `dx build --ssg`
- Observe the generated site has no content

**Expected behavior**

SSG should not silently fail to build when it takes more than 10 seconds

**Environment:**
 - Dioxus version:  `0.6`
 - Rust version:    `nightly`
 - OS info:         MacOS
 - App platform:    `ssg`

ealmloff:
Instead of a constant 1 second initial boot delay and 10 ssg rendering second delay, this PR tries to connect to the server every second for up to 5 seconds and waits until all routes are fully rendered before finishing ssg.

- [x] Fix implemented and tested on both reproduction and docsite
- [x] Playwright regression test

Fixes #3518 
Stacked on #3488


jkelleyrtp (approved):
Looks good! I didn't realize we just needed to await the .text() stream.

PR #3525, opened 2025-01-08
Fix: Asset Cache On Windows & Asset Status

DogeDark:
**Problem**
Due to Windows' long-form path variants, cached assets were being deleted causing long rebuilds for apps with lots of assets e.g. docsite.

The asset status would jump around a lot in terms of progress.

**Fixes**
We use the `dunce` crate to canonicalize assets in a few places, replacing `fs::canonicalize`. The changes work well on Windows, but it would be good to test other platforms. There may be other places we should canonicalize.

The asset status now tracks the completed assets instead of the current asset.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3479,6 +3479,7 @@ dependencies = [
  "dioxus-rsx-hotreload",
  "dioxus-rsx-rosetta",
  "dirs",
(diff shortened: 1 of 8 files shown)


jkelleyrtp (approved):
Looks good, thank you!

PR #3531, opened 2025-01-09
Optimize wasm bindgen asset with manganis

### LINKED ISSUE #2124: Optimize initial loading of web resources
## Feature Request

Currently loading a web app built with Dioxus has some performance issues especially under high latency networks. Looking at the waterfall analysis of a Dioxus load (under "Fast 3G" option in Edge browser) you'll see following:

image

Disregarding the long download time for the WASM file (debug build), we can see several places that can be improved. The root cause of the unnecessary delay is the sequential loading of static resources, which slows things down even when the resources are small, due to network latency.

## Implement Suggestion

There are two improvements that will have significant effect and are fairly trivial to implement.

### Bundle the JS files into a single asset, avoiding the dynamic loading of six separate files.

This change alone will cut about a second of loading time in a high latency network. Under the current strategy the browser can only download additional script files once it has started executing the previous ones and encounters the `import` statement.

Browsers typically restrict the number of concurrent connections, so this creates additional latency when there are many files to download and it needs to wait for a connection to become available. The bundle file should have a content hash in its name.

### Instruct the browser to preload assets in the HEAD section.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload for details.

By introducing `<link rel=preload href="main.js" as="script" />` preloading hint for every resource needed later, the browser will start loading these immediately, instead of waiting for the rest of the index.html to be processed. Note that for the WASM file you'll need to use `crossorigin` as well since it's a `fetch`.

For example I tested by manually editing the `index.html` to prove that the browser indeed starts downloading files earlier:
```html
  <link rel="preload" href="/./assets/dioxus/Zhef_bg.wasm" as="fetch" crossorigin>
  <link rel="preload" href="/./assets/dioxus/Zhef.js" as="script">
  <link rel="preload" href="./snippets/dioxus-interpreter-js-e8d4f2300616f4bc/inline0.js" as="script">
  <link rel="preload" href="./snippets/dioxus-interpreter-js-e8d4f2300616f4bc/src/js/common.js" as="script">
  <link rel="preload" href="./snippets/dioxus-web-aa233b06656766d9/inline0.js" as="script">
  <link rel="preload" href="./snippets/dioxus-web-aa233b06656766d9/inline1.js" as="script">
  <link rel="preload" href="./snippets/dioxus-web-aa233b06656766d9/src/eval.js" as="script">
  ```
image

This is still not optimal due to the limit of concurrent downloads imposed by the browser.

The preload aspect should also be made available for the developers who might be adding resources via `Dioxus.toml`

### Add a hash to the WASM file name

Adding a content hash to the name of the WASM file will enable efficient caching of the resource by browsers and CDNs alike. This will not help with the initial cold load, but will allow the browser (and CDN) to cache the file indefinitely. Of course the caching headers will also need to be set by the server.

### LINKED ISSUE #3277: Suboptimal initial load of web app
**Problem**

When a Dioxus web app is initially loaded by the browser it's loaded slower than necessary.

**Steps To Reproduce**

- 
- 
- 

**Expected behavior**

Resources needed by the Dioxus app should be loaded with minimal amount of requests and utilize browser preloading capabilities.

**How to improve**

Here is a screenshot of a typical page load in Chrome using a slow network to make the problem more visible

image

(don't mind about CSS being loaded twice with different names)

First issue is that `web.js` is downloaded twice, because the `preload` attribute is not respected by the browser due to missing `crossorigin` attribute, which is also visible as a warning in console
```
101:17 A preload for 'http://127.0.0.1:8080/wasm/web.js' is found, but is not used because 
the request credentials mode does not match. Consider taking a look at crossorigin attribute.
```

Second problem is the number of JS files loaded in a cascading manner. These should all be bundled into a single JS file and served at once. There is also the issue of max 6 TCP connections with HTTP/1.1 that causes stalling, but that's not a real issue in production environment with HTTP/2 or HTTP/3. Even if the connection has high bandwidth but there is a long latency (server on the other side of the globe), these JS file loads will be the bottleneck.

Finally, application assets like CSS should be compressed with Brotli, which is already done to the Dioxus WASM and JS files.

**Environment:**
 - Dioxus version:  0.6.0-rc.0
 - Rust version:    1.83
 - OS info:         WSL
 - App platform:    fullstack web

**Questionnaire**

### LINKED ISSUE #3339: Preload option has no effect on CSS assets
**Problem**

Using `with_preload` on CSS assets has no effect since the `rel` attribute is set to `stylesheet` instead of `preload`. For example

```rust
const TAILWIND_CSS: Asset = asset!("/public/tailwind.css", CssAssetOptions::new().with_preload(true))
```
generates
```html
<link rel="stylesheet" href="/assets/tailwind-627b9919aa94f0b5.css" type="text/css"/>
```

**Steps To Reproduce**

- Use `CssAssetOptions::new().with_preload(true)` on a CSS asset
- Check the generated HTML

**Expected behavior**

This describes how to get this working https://stackoverflow.com/questions/61889044/how-to-use-both-rel-preload-and-rel-stylesheet-for-the-same-tag

### LINKED ISSUE #3350: wasm files are not hashed, leading to caching issues on deployments
**Problem**

After redeploying my site the page will not open, reporting the error Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #272 “wbg” “__wbindgen_closure_wrapper17305”. function import requires a callable

So far I've been able to fix this by clearing my browser cache, but I'm not sure why this is happening

**Steps To Reproduce**

I haven't been able to reproduce this bug consistently yet, it triggers by chance when redeploying the site

ealmloff:
This PR uses the manganis bundling system to hash and optimize the assets wasm bindgen generates. It also expands cli-opt to support bundling multiple js files together which is useful for the javascript files wasm bindgen outputs. 

Requests for hello world app with old version of DX:
Screenshot 2025-01-09 at 1 13 12 PM

Requests for hello world app with new version of DX:
Screenshot 2025-01-09 at 1 13 29 PM

Closes #3350
Closes #3339
Closes #3277
Closes #2124


packages/cli/Cargo.toml
```diff
@@ -7,7 +7,7 @@ description = "CLI for building fullstack web, desktop, and mobile apps with a s
 repository = "https://github.com/DioxusLabs/dioxus/"
 license = "MIT OR Apache-2.0"
 keywords = ["mobile", "gui", "cli", "dioxus", "wasm"]
-rust-version = "1.79.0"
+rust-version = "1.81.0"
```
ealmloff, line 10: swc requires a newer version of rust to build. I'm not sure how much we care about the rust version the CLI requires since we are leaning into binstall everywhere, but if this an issue I could look for alternatives

ealmloff:
Marking this as a draft until we find the latest version of swc settings that are compatible with a broader set of javascript like https://github.com/DioxusLabs/dioxus/pull/3539

packages/fullstack/src/server/mod.rs
```diff
@@ -276,6 +287,25 @@ where
     }
 }
 
+type MappedAxumService<S> = MapResponse<
+    S,
+    fn(Response<ServeFileSystemResponseBody>) -> Response<ServeFileSystemResponseBody>,
+>;
+
+fn cache_response_forever<
```
ealmloff, line 295: Make sure the html files are not cached

ealmloff: This only serves hashed assets. HTML files are generated by ssr

ealmloff:
This PR also fixes a few caching issues with manganis/the CLI:
- If you interrupt the CLI while it is optimizing a large folder and then restart it. The CLI would think the optimization was finished and use the partially finished folder
- If you upgrade the CLI and run it on an old project, the assets would not be re-optimized. That means any new CLI releases that fix optimization issues would not be applied until the target directory is cleared

jkelleyrtp (approved):

ealmloff:
Caching the fullstack asset folder with a long reset is problematic for legacy assets which are placed in the folder without a hashed name. That change is pulled out of this PR in favor of a followup that somehow communicates which assets are cached and not

PR #3533, opened 2025-01-09
Fix initialising tray icons on the macos desktop platform

PhilTaken:
Depending on multiple versions of the `muda` crate - one directly and one via a transitive dependency in the tray-icon crate - leads to the following error when attempting to initialize a tray icon on the desktop platform, specifically on MacOS:

```
   6.858s  INFO Build completed successfully in 2882ms, launching app! 💫   7.871s  INFO [macos]: 2025-01-10 00:16:01.736 dimsum[19111:45581742] +[IMKClient subclass]: chose IMKClient_Modern
   7.876s  INFO [macos]:   7.876s  INFO [macos]: thread 'main' panicked at /Users/philippherzog/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/muda-0.15.3/src/platform_impl/macos/mod.rs:934:1:
   7.876s  INFO [macos]: could not create new class MudaMenuItem. Perhaps a class with that name already exists?
```

Without any prior knowledge on how the `muda` crate or MacOS tray icons work, it appears that `muda` generates some sort of class definition for MacOS whose identifier needs to be unique. Multiple versions of the crate will then lead to duplicate definitions and thereby the above crash at runtime. 

Since this only breaks at runtime and only on the MacOS target it's far from trivial to add a test to ensure this doesn't break again on the next `cargo update`. I'll open a related ticket in the muda repository, maybe they can think of a more robust solution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3702,7 +3702,7 @@ dependencies = [
  "infer 0.11.0",
  "jni",
  "lazy-js-bundle",
- "muda 0.11.5",
+ "muda",
  "ndk",
  "ndk-context",
  "ndk-sys 0.6.0+11769913",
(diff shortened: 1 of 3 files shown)


jkelleyrtp:
Unfortunately I don't think we can bump muda in this minor version (0.6.x). Muda is exported from our public API via `dioxus::desktop::muda` so if you want to use muda in your crate I recommend using the one we re-export. Adding muda to your cargo.toml will cause the issue which is unfortunate but probably not something we can address.

Is there an issue with the 0.14.x series of muda that we *need* to upgrade to?

PhilTaken:
> Adding muda to your cargo.toml will cause the issue which is unfortunate but probably not something we can address.

that is not the case, dioxus and dioxus-desktop requiring different versions of the muda crate is the issue here

PhilTaken:
Another option here might be downgrading tray-icon to a version that depends on the same version of muda as dioxus such that the exported version of muda remains the same but neither solution is great and neither can be kept in sync easily aside from maybe parsing cargo tree in CI

jkelleyrtp (approved):

PR #3555, opened 2025-01-13
docs: demonstrate extending both global and element attributes

Plebshot:
This PR improves the docs on extending props a bit.

Before it was unclear to me, how one can extend both the global attributes and an element's extension attributes.

The statement `you can use any attribute that would normally appear on the button element` was also a bit misleading, given that the example only extends the global attributes. So I generalized original example to use a div instead of a button.


ealmloff (approved):

jkelleyrtp (approved):
Thank you!

Issue #3546, opened 2025-01-12
Dx serve failed to run project migrated from 0.5 to 0.6.1 after updating CLI


Anakael:
**Problem**

I have a project which was created on dioxus 0.5. After updating dependency to 0.6 it continued to working file. But after I've update CLI to version 0.6.1 it started to fail on dx serve (alson on dx check) command with error:

```
0.373s ERROR Loading crate
   0.373s ERROR Found package NodeIndex(391)
   0.373s ERROR err=Other(Failed to find target train-solfege-dx)
```

![image](https://github.com/user-attachments/assets/6c778b4d-504c-43b0-a2b5-9cdfa8b0350d)

But if I create new project with CLI 0.6.1 - it works fine.

**Steps To Reproduce**

- Update CLI from 0.5 to 0.6.1
- Try to run project migrated from 0.5 to 0.6.1

**Expected behavior**

No error

**Screenshots**

![image](https://github.com/user-attachments/assets/baa63f76-0f22-4cb9-a5b1-dc7247fa33bf)

**Environment:**
 - Dioxus version: 0.6.1
 - Rust version: 1.81.0
 - OS info: Arch linux
 - App platform: desktop

**Questionnaire**

Anakael:
I've figured out what cause the problem. I have workspace with multiple project and one of them is dioxus project. And it's package name differs from bin name, so condition here can not find it. https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/dioxus_crate.rs#L59
After setting bin name to package name or set --bin param it started to work.
Is it intended behavior or it should find bin in current package?

ealmloff:
The error message should be more clear in the git version of the CLI with #3505. The CLI should tell you it can't find the binary in the error with some suggested alternatives now.

It looks like `cargo run` will default to running the bin target if there is only one. If there is more than one, it uses the `default-run` key. Dioxus should probably mimic that behavior

PR #3561, opened 2025-01-13
fix: cli pakage name search

Anakael:
Resolves https://github.com/DioxusLabs/dioxus/issues/3546

Now, if `bin` param not specified for dx it firstly fallback `default-run` target, secondly to single bin target thirdly to package name.

`default-run` case:
image
image

single bin case:
image
image


jkelleyrtp (approved):
Thanks for the PR! I don't use default_run too often so we never really hit it. This looks good to me.

PR #3569, opened 2025-01-14
Fix repeated subscription to dropped scope leak

### LINKED ISSUE #3526: bug: Potential memory leak when calling components
**Problem**

It seems that simply calling components leaves a small leak that `v0.5` didn't have, its true that the following code is very awkward and definitely not real world, but the root issue might be affecting other scenarios that are actually more common and with worse outcomes.

**Steps To Reproduce**

```rs
#![allow(non_snake_case)]
use dioxus::prelude::*;

    fn app() -> Element {
        

        rsx! {
            Spammer {}
            Spammer {}
            Spammer {}
            Spammer {}
        }
    }

    fn Spammer() -> Element {

            loop {
                tokio::time::sleep(std::time::Duration::from_nanos(1)).await;
                if val == 70 {
                    count.set(0);
                } else {
                    count.set(val + 1);
                }
            }
        }));

        rsx! {
            for el in 0..*count.read() {
				// To avoid the leak simply replace this by `div {`
                Comp {
                    key: "{el}",
                }
            }
        }
    }
	

    #[component]
    fn Comp() -> Element {
        rsx!(
            div { }
        )
    }

        .enable_all()
        .build()?
        .block_on(async {
            loop {
                vdom.wait_for_work().await;

            }
        })
}
```

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version:    1.82
 - OS info:         Windows
 - App platform:   just core crates, no renderer

ealmloff:
When reactive scopes drop, they didn't remove themselves from signal's subscriber list until the signal was written to. That causes issues with the built in hot reloading signal which is never written to. This PR removes the subscriptions when the reactive scope drops

Fixes #3526


packages/core/src/reactive_context.rs
```diff
@@ -311,3 +311,15 @@ struct Inner {
...
+
+impl Drop for Inner {
+    fn drop(&mut self) {
+        let self_ = self
+            .self_
+            .expect("self_ should be set during initialization");
+        let old_subscribers = std::mem::take(&mut self.subscribers);
+        for subscriber in old_subscribers {
+            subscriber.0.lock().unwrap().remove(&self_);
+        }
```
jkelleyrtp, line 323: can we make these not unwrap/expect and instead use options/if/else?

I'm sure it won't be an issue but just want to guard against panics in drops since that can cause an abort.

jkelleyrtp (changes requested):

jkelleyrtp (approved):

marc2332:
Nice work!

Issue #3564, opened 2025-01-14
Output Directory Issue with dx bundle --platform web --outdir mydir


Dicklessgreat:
#### Problem
When running the command `dx bundle --platform web --outdir mydir`, the output files are unexpectedly placed in `/path/to/my/cargoproj/target/dx/form/release/web/public` instead of the specified `/path/to/my/cargoproj/mydir`.

#### Expected Behavior
The output files should be placed in `/path/to/my/cargoproj/mydir` as specified in the command.

#### Environment
- **Platform**: Web
- **OS**: 
  ```
  Distributor ID: Ubuntu
  Description:    Ubuntu 24.04.1 LTS
  Release:        24.04
  Codename:       noble
  Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 GNU/Linux
  ```
- **dx version**: `dioxus 0.6.1 (c2952a7)`

#### Steps to Reproduce
1. Run the command `dx bundle --platform web --outdir mydir` in the project directory.
2. Check the output directory.

### Additional Information
I made project with `fullstack` option.

PR #3571, opened 2025-01-15
Fix the issue for outdir not work when run dx bundle --platform web --outdir mydir

BleemIs42:
Fix [#3564](https://github.com/DioxusLabs/dioxus/issues/3564)


BleemIs42:
Wrong branch

PR #3578, opened 2025-01-16
Manganis CSS Modules Support

DogeDark:
Adds a `css_module!()` macro that generates a struct for type-safe, globally unique, classes and ids. This reuses the `asset` macro with a new `AssetOption` variant for CSS modules.

```rs
css_module!(Styles = "/mycss.module.css");
// css_module!(pub Styles = "/mycss.module.css");
// css_module!(pub Styles = "/mycss.module.css", CssModuleAssetOptions::new().minify(false));

#[component]
fn App() -> Element {
    rsx! {
        div {
            id: Styles::header,
            class: Styles::hello_class,
            p { "This div's id is {Styles::goodbye}" }
            p { "This div's class is {Styles::hello_class}" }
        }
    }
}
```
```css
/* mycss.module.css */

#hello { ... }

.hello { ... }

#header { ... }
```

The macro's docs explain more about how it works & features w/examples.

### Breaking
This is breaking as it alters the asset options enum creating a version requirement in the CLI.


DogeDark:
This is still more or less an experiment, but I created this PR to track it, as I believe it could be viable for 0.7 if desired.

jkelleyrtp:
Is it possible to support a syntax for declaring modules that uses a static or a struct as the source item?

```rust
static MyModule: CSSModule = css_module!("/blah.css");
```

or

```rust
#[derive(CssModule(file = "/blah.css")]
struct MyModule;
```

I know this might be challenging to achieve but it would be more "idiomatic" rust and likely more discoverable. We'd really want to provide good intellisense and I think providing inline items like this help with rust analyzer. That, and it's usually more transparent to the users of the macro what exactly is being generated. 

I'll see if I can tinker with getting #1 working but using #2 might be an easier migration for this PR.

DogeDark:
@jkelleyrtp I wouldn't know where to begin for 1. but 2. is likely something I can do.

jkelleyrtp:
We could try something with extension traits to get #1 working
```rust
struct CssModule {}

static MY_MODULE: CssModule = CssModule {};

trait LocalExt {
    fn id(&self) -> &'static str {
        "mymodule"
    }
}

impl LocalExt for CssModule {}

fn it_works() {
    MY_MODULE.id();
}
```

still brainstorming...

PR #3579, opened 2025-01-17
fix: window state preservation by soft-killing the process

### LINKED ISSUE #3355: Reopen window in same place
## Feature Request

It would be really cool if (when performing a full rebuild of desktop app), the CLI could open the window of the new app at the same size and in the same position as the window of the previous version of the app was when it was killed.

## Implement Suggestion

- Determine the size of the apps window (perhaps by asking it over devtools protocol?)
- Set window size of new window (perhaps with environment variable?)

jkelleyrtp:
In the big CLI rewrite I accidentally removed the code for doing soft-kills on the desktop window.

This fixes that, fixing window state preservation.

Fixes #3355

### DIFF
--- a/packages/cli/src/cli/run.rs
+++ b/packages/cli/src/cli/run.rs
@@ -46,7 +46,7 @@ impl RunArgs {
                     tracing::info!("[{platform}]: {msg}")
                 }
                 ServeUpdate::ProcessExited { platform, status } => {
-                    runner.kill(platform);
+                    runner.kill(platform).await;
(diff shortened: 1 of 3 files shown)


Godnoken:
Fyi, on my Windows 10, this only works if you first close down the app, and then reopen it with 'o'. Rebuilding off the bat will open it in its initial position.

jannik4:
Same for me (Windows 11).
Actually, it is enough to close the app once manually. After this all rebuilds will open at that position.

PR #3584, opened 2025-01-17
Fix: Only Resize WebView Associated With Window

CryZe:
When resizing a window, Dioxus previously accidentally resized every single webview to the dimensions of the window that just got resized. This fixes it so only the webview that is associated with the window is resized.

https://github.com/user-attachments/assets/515aaeb5-aa9c-4ab0-9d95-559689312e53

### DIFF
--- a/packages/desktop/src/app.rs
+++ b/packages/desktop/src/app.rs
@@ -226,14 +226,14 @@ impl App {
         }
     }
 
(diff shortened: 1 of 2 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Thank you! Good catch - I must've missed that change in a previous review. Wasn't always like that.

Will release today, hopefully.

PR #3499, opened 2025-01-04
Fixed the multiple window closing issue, where the two multiwindow examples weren't working.

sertschgi:
This was due to the tao window instance not being dropped. Furthermore the whole desktop context should have been dropped but wasn't due to the Rc that had 3 strong references too many because of copying in webview::new. I replaced everything with weak pointers so that the desktop context has 1 strong reference to the app, so that it is dropped when it is removed from the app.webviews hashmap.

Fixes #3542
Fixes #3113 
Fixes #3342

PR #3588, opened 2025-01-18
Fixed multiwindow closing issue again

sertschgi:
Like I explained in #3499 there is this issue where when creating a new window it cannot be closed again. This is due to references still pointing to the DesktopContext after being removed from webviews in app.rs. They themselves also cannot be dropped and this is not only because of the self referential structure. The problem has not been fixed with my previous pull request as half of it was simply ignored: If the DesktopContext Rc reference is upgraded again and added to other Structs it obviously cannot be dropped. Please test the examples. Also note that this fixes it for linux. For windows only partially.
Furthermore what I have noticed was that the issue was made worse because the main window cannot even be closed anymore because of the window decoration buttons not working - but this has probably nothing to do with this I'm just pointing it out...


jkelleyrtp:
We have to be careful not to break semver here unfortunately, so the interfaces need to remain the same but the internal details can change. IE we need to take Rc<T> even if we downgrade it internally

jkelleyrtp:
The multi-window example works for me on main - is there a particular repro you have?

freedomtowin:
I tested the example on Windows and can confirm that the multiwindow example does not work. 

But adding `async move` fixed the display/close issues for the new window:

```rust
    let onclick = move |_| async move {
        let dom = VirtualDom::new(popup);
        dioxus::desktop::window().new_window(dom, Config::new().with_close_behaviour(WindowCloseBehaviour::CloseWindow));
    };
```

cortesi:
I still see this problem on the main branch. I'm concerned that this PR is not associated with an issue against 0.7 and may have have gotten lost. Is this being tracked, or should I create a separate issue describing the problem?

jkelleyrtp:
Does anyone have a good repro for this? I tested the most recent main on windows + macOS and wasn't able to repro

cortesi:
Here's a minimal reproduction - macos, using main branch Dioxus. I'm experimenting with Dioxus for the first time, so this could totally be a "holding it wrong" issue - let me know. I'm on the Discord at @cortesi if you need me to try anything else on my system. 

```rust
use dioxus::desktop::{Config, WindowCloseBehaviour, use_window};
use dioxus::prelude::*;

fn main() {
    dioxus::LaunchBuilder::desktop()
        .with_cfg(
            Config::new()
                .with_close_behaviour(WindowCloseBehaviour::LastWindowHides)
        )
        .launch(WindowA);
}

#[component]
fn WindowA() -> Element {
    let open_window_b = move |_| {
        println!("Opening Window B");
        let dom = VirtualDom::new(WindowB);
        let config = Config::new()
            .with_close_behaviour(WindowCloseBehaviour::CloseWindow);
        dioxus::desktop::window().new_window(dom, config);
    };

    rsx! {
        div {
            h1 { "Window A" }
            button {
                onclick: open_window_b,
                "Open Window B"
            }
        }
    }
}

#[component]
fn WindowB() -> Element {
    let window = use_window();
    
    let close_window = move |_| {
        println!("Attempting to close Window B");
        window.close();
    };

    rsx! {
        div {
            h1 { "Window B" }
            p { "Try closing this window with the window decoration close button" }
            button {
                onclick: close_window,
                "Close Window B (button)"
            }
        }
    }
}
```

jkelleyrtp:
Thanks for the repro!!!

This is a separate issue entirely - we don't need to change desktop itself, but instead manually change the drop order of the VirtualDom itself.

The virtualdom and the runtime are separate objects, and the data for the `onclick` handler lives inside the runtime, not the virtualdom. When we call `webviews.remove(id)`, we don't call `root.destroy()`, and thus the event handlers and their state still persists. 

I just created a new PR to fix that, making the destruction of a VirtualDom equivalent to deleting the root node.

https://github.com/DioxusLabs/dioxus/pull/4351

Issue #3594, opened 2025-01-20
Option "--target" not passed to cargo when build platform = Server


marcobergamin-videam:
Option "--target" not passed to cargo when build platform = Server
This problem cause the impossibility to build the 'server' side of a Dioxus fullstack application for the correct architecture using dx build/dx bundle/...

**Steps To Reproduce**

- Create a docker container with a docker file like this one (the gcc version does not matter):
```docker
FROM rust:latest 

ARG GCC_VERSION=-11
 
RUN apt update && apt upgrade -y 
RUN apt install -y gcc$GCC_VERSION-aarch64-linux-gnu  g++$GCC_VERSION-aarch64-linux-gnu
 
RUN rustup target add aarch64-unknown-linux-gnu
RUN rustup toolchain install stable-aarch64-unknown-linux-gnu
RUN cargo install cargo-binstall && cargo binstall dioxus-cli --version 0.6.1

WORKDIR /app 

ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc$GCC_VERSION CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc$GCC_VERSION CXX_aarch64_unknown_linux_gnu=aarch64-linux-gnu-g++$GCC_VERSION
 
CMD ["dx", "bundle", "--target", "aarch64-unknown-linux-gnu", "--platform", "web", "--arch", "arm64"]
```
- Build the docker image: ``docker build . -t dioxus-cross-compile``
- Create a basic project: ``dx new example``
- Build the example: ``docker run --rm -v ./example:/app dioxus-cross-compile``

**Expected behavior**

The ``server`` binary should be cross compiled for the correct arch.

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version:    1.83
 - OS info:         Ubuntu 22.04
 - App platform:    Web

**Solution**
I have found the problem and I'm gonna share my fix with a pull request. Feedbacks are of course welcome.

ealmloff:
> Sorry I don't understand the use case. When do a desktop app could have the server and the client side build for different architectures?

Yes, for example I could build a desktop app that targets `aarch64-unknown-linux-gnu` with a server that runs on a server less platform like spin that requires `wasm32-wasip1`

jkelleyrtp:
[...]

PR #3595, opened 2025-01-20
Fix --target flag not passed to cargo when Platform == Server

marcobergamin-videam:
PR related to the Issue #3594


DogeDark:
I revised this a bit to add a new `--server-target` flag which is used if provided for server builds.

jkelleyrtp (approved):

PR #3602, opened 2025-01-20
Switch the default server function codec to json

### LINKED ISSUE #2628: structs containing Vec<T>, HashSet<T> and HashMap<T, T> are not working with Signal/ServerFn
**Problem**

**Steps To Reproduce**
[Reproducile repo](https://github.com/chungwong/bar-chart/tree/issue_2288)

```rs
#[component]
fn Home() -> Element {
    let global_filter: Signal<GlobalFilter> = use_signal(|| GlobalFilter::default());

    rsx! {
        button {
            onclick: move |_| {
                let filter: GlobalFilter = global_filter.read_unchecked().to_owned();
                async move {
                    get_server_data(filter).await;
                }
            },
            "Button"
        }
    }
}

#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
pub struct GlobalFilter {
    pub test_string: String,

    pub test_vec: Vec<String>,
    pub test_hashset: HashSet<String>,
    pub test_hashmap: HashMap<String, String>,

    // wrap them in Option to bypass the bug
    // pub test_vec: Option<Vec<String>>,
    // pub test_hashset: Option<HashSet<String>>,
    // pub test_hashmap: Option<HashMap<String, String>>,
}

#[server]
async fn get_server_data(global_filter: GlobalFilter) -> Result<(), ServerFnError> {
    Ok(())
}
```
After clicking the button, an error( can be observed in browser's dev tool)
```
Args|missing field `test_vec`
```
happened. This error happened both on 0.5 and the main branch.

After digging on discord, [the tmp solution](https://discord.com/channels/899851952891002890/943190605067079712/1227520285520167003) is to wrap those fields in an Option. [Evan mentioned](https://discord.com/channels/899851952891002890/943190605067079712/1227589708268306493) it should be fixed in https://github.com/DioxusLabs/dioxus/pull/2288 but looks like the above case is not covered.

**Expected behavior**
Those fields in the struct should work without the extra Option

**Environment:**
 - Dioxus version: 0.5 and main
 - Rust version:   1.79
 - OS info:  Debian 12
 - App platform:  Fullstack

ealmloff:
This PR switches from URL encoded post requests to Json as the default server function encoding. The JSON encoding handles more rust types which avoid some confusing errors with empty data structures. This isn't semvar breaking on the rust side, but it is a breaking change for any 3rd party server function consumers 

Fixes #2628


jkelleyrtp (approved):
this is really really gross - pre-parsing and then emitting - but I get why. Let's see if we can upstream a better interface for server-fn or just own it ourselves in the future. I think it's a rather thin crate?

PR #3617, opened 2025-01-21
fixed the bug when future restarts the state goes to ready instead of pending

### LINKED ISSUE #2812: `Resource` of `use_resource` state should go back to `Pending` when a dependency changes.
**Problem**

When using `use_resource` with reactive, dependencies, the state go from `Pending` to `Ready`, but when the future is restarted (when a dependency changes), the state remains `Ready`.

This should be pretty easy to fix, just change the implementation here: https://github.com/DioxusLabs/dioxus/blob/b97e607aca0bd8a12971fcf403908addf9a45dd7/packages/hooks/src/use_resource.rs#L31

Note that the behaviour of `SuspenseBoundary` would change. It would display the fallback each time the future is restarted. But that would make more sense in my opinion.

A related issue would be to update the docs of `use_future` to clarify that it will not re-run when the dependencies changes, and that one should use `use_resource` instead.

**Environment:**
 - Dioxus version:  `master`

avij1109:
This PR fixes the bug when future is restarted, the state remains ready instead of pending

Fixes #2812

### DIFF
--- a/packages/hooks/src/use_resource.rs
+++ b/packages/hooks/src/use_resource.rs
@@ -28,7 +28,7 @@ where
         (rc, Rc::new(Cell::new(Some(changed))))
     });
 
-    llet cb = use_callback(move |_| {
+    let cb = use_callback(move |_| {
         // Set the state to Pending when the task is restarted
         state.set(UseResourceState::Pending);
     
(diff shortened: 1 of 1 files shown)


packages/hooks/src/use_resource.rs
```diff
@@ -28,16 +28,19 @@ where
         (rc, Rc::new(Cell::new(Some(changed))))
     });
 
-    let cb = use_callback(move |_| {
+    llet cb = use_callback(move |_| {
```
candunaj, line 31: There is a typo. llet should be let.

jkelleyrtp (approved):

PR #3624, opened 2025-01-22
Make test_github_install only run on platforms with prebuilt binaries

hack3ric:
This allows tests to pass on all platforms (in my case riscv64gc-unknown-linux-gnu).


jkelleyrtp:
Thanks for the PR! I ended up just changing the tests to only run when a "full test suite" env var is set since they were drastically slowing down our CI.

Issue #3530, opened 2025-01-09
Examples for LaunchBuilder methods


ealmloff:
## Feature Request

The LaunchBuilder method is a very common API in dioxus. We should include some examples of how to use the API with a custom platform config and context

## Implement Suggestion

Include examples of launching apps with the `with_cfg`, `with_context`, and `with_context_provider` methods in the `LaunchBuilder` documentation

saleha-muzammil:
I would like to work on this.

ealmloff:
> I would like to work on this.

Great! Feel free to ask any questions here or on the discord if you run into any issues

ealmloff:
Closed by https://github.com/DioxusLabs/dioxus/pull/3629

PR #3629, opened 2025-01-22
Add examples for LaunchBuilder methods

vishwamartur:
Related to #3530

Add examples for `with_cfg`, `with_context`, and `with_context_provider` methods in the `LaunchBuilder` documentation.

* **with_context_provider**: Add an example demonstrating how to inject state into the root component's context created on the thread that the app is launched on.
* **with_context**: Add an example demonstrating how to inject state into the root component's context.
* **with_cfg**: Add an example demonstrating how to provide a platform-specific config to the builder.


jkelleyrtp (approved):
Thanks for the PR! Much nicer

PR #3630, opened 2025-01-22
Hash and compare templates by value in debug mode

### LINKED ISSUE #3235: On the 0.6.0-alpha.4 version, clicking Route:: Link to route does not refresh locally, but reloads the whole page
**On the 0.6.0-alpha.4 version, clicking Route:: Link to route does not refresh locally, but reloads the whole page**

ealmloff:
In debug mode, rust doesn't merge statics which makes all rsx generate unique template pointers. That causes state to be lost when switching between identical rsx blocks only in debug mode. We generate identical rsx blocks for layouts in the router. This PR fixes the issue by comparing by value in debug mode and by pointer in release mode

Fixes #3235


packages/core/src/nodes.rs
```diff
@@ -378,17 +378,39 @@ pub struct Template {
 
 impl std::hash::Hash for Template {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        std::ptr::hash(self.roots as *const _, state);
-        std::ptr::hash(self.node_paths as *const _, state);
-        std::ptr::hash(self.attr_paths as *const _, state);
+        // If debug assertions are enabled, rust is probably built with a lower
+        // optimization level which will not merge identical statics. If it is,
+        // then we hash by contents
+        if cfg!(debug_assertions) {
```
ealmloff, line 384: I think `debug_assertions` is the closest we can get to checking if that optimization is likely active, but it can fail in some cases like if you disable `debug_assertions` in debug mode for faster build times

ealmloff: This should be more reliable now. We can check if function pointers are merged in a way that hopefully gets resolved at compile time. If they are merged, this is a build that is optimized enough to merge statics

jkelleyrtp (approved):

PR #3634, opened 2025-01-23
Allow to connect to the dev server from a physical Android device

### LINKED ISSUE #3444: Make it possible to connect to the dev server on a physical Android device (`dx serve`)
## Feature Request

Currently, only a virtual Android device supports hot-reloading etc. but not a physical one.

## Implement Suggestion

I already have [a branch](https://github.com/Andrew15-5/dioxus/tree/fix-dev-server-android) with a working version, but I don't have time to make a PR. Should be free in a week or so.

Andrew15-5:
This is a complimentary fix for #3444 to allow effortlessly connecting and using the hot-reloading stuff etc. when there are some network setup limitations that prevent the current implementation from working.

As a QoL feature, I can highlight the fact that you don't have to explicitly open a local port on your host if you have an active firewall set up. As a result, you don't have to use `sudo` and write your password and then no need to remember to close that port again. And also starting developing would be just faster. 🚀

[Read more](https://github.com/DioxusLabs/dioxus/issues/3444#issuecomment-2610973993).

Fixes #3444.

### DIFF
--- a/packages/cli-config/src/lib.rs
+++ b/packages/cli-config/src/lib.rs
@@ -100,6 +100,12 @@ macro_rules! read_env_config {
 /// For reference, the devserver typically lives on `127.0.0.1:8080` and serves the devserver websocket
 /// on `127.0.0.1:8080/_dioxus`.
 pub fn devserver_raw_addr() -> Option<SocketAddr> {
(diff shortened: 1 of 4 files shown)


jkelleyrtp (approved):
Looks good, thank you!!

PR #3663, opened 2025-01-28
fix: Downgrade publish ci to ubuntu 22.04

marc2332:


marc2332:
This makes it possible to use the cli from binstall in ubuntu 22.04. It was not possible with `ubuntu-latest` as that points to 24.04, and it comes with a glibc bump.

jkelleyrtp (approved):

PR #3675, opened 2025-01-31
feat: inline dioxus-native from blitz

jkelleyrtp:
This PR inlines the commits from the blitz repo for the dioxus-native project.

- [x] Migrate package over
- [x] asset resolver
- [x] dx support
- [x] document support
- [x] history support

It also makes `dx` work native.

If we want to publish 0.6.3, we will need to disable this integration, but I'm assuming we're mostly done with the 0.6 patch cycle. Alternatively, we could publish blitz and then release dioxus-native under a minor release (0.6.3 / 0.6.4, etc). We would just need to commit to the "native" feature flag for the time being.

For feature-flag resolution, we use the named feature flags in the users cargo.toml for platform selection:
```toml
[dependencies]
dioxus = { features = ["fullstack"] }

[features]
default = []
desktop = ["dioxus/desktop"]
mobile = ["dioxus/mobile"]
web = ["dioxus/web"]
server = ["dioxus/server"]
```

This means to use the new native renderer, you would simply migrate the desktop and mobile renderers over:

```toml
[features]
default = []
desktop = ["dioxus/native"]
mobile = ["dioxus/native"]
web = ["dioxus/web"]
server = ["dioxus/server"]
```


jkelleyrtp:
Screenshot 2025-01-30 at 8 59 11 PM

asset resolver works!

jkelleyrtp:
Screenshot 2025-01-30 at 9 31 03 PM

dx serve also works

nicoburns:
> For feature-flag resolution, we use the named feature flags in the users cargo.toml for platform selection:
> This means to use the new native renderer, you would simply migrate the desktop and mobile renderers over:

With the current code, is it possible to setup feature flags so that you can have both "webview" and "blitz" versions of an app (with different feature flags)?

nicoburns:
Moving the crate here will mean that we wont be able to easily run `dioxus-native`-based examples from the Blitz repository. Is the idea is that we would move any relevant examples here? Most of them were based on dioxus examples in the first place, but we might want to check that the unmodified examples work with the native renderer, and port some of the blitz examples that are just testing rendering features to the html renderer?

nicoburns:
Hmm... so this is running for me (if I change feature from `"dioxus/desktop"` to `"dioxus/native"` in Cargo.toml. But I'm not getting assets (stylesheets) through. Seems like the asset isn't even being requested, so an issue with detecting the stylesheet in the first place rather than finding the correct file.

Issue #3689, opened 2025-02-04
Bundled fullstack desktop apps don't work (Linux, Windows)


phkoester:
**Problem**

I simply create a fullstack app with `dx new`, bundle it with `dx bundle --platform desktop` on Linux and Windows. Neither the Linux AppImage nor Windows MSI nor Windows Setup work.

**Steps To Reproduce**

- `dx new some_app`, choose fullstack.
- `dx bundle --platform desktop` (Linux, Windows)
- Start the AppImage on Linux, or install via MSI or Setup on Windows

**Expected behavior**

- I expect the assets to be found and the app to run, i.e. the echo demo should work

**Actual behavior:**

- Linux: Assets arent't found, echo demo panics with

```
called `Result::unwrap()` on an `Err` value: Request("error sending request for url (http://127.0.0.1:8080/api/echo_server10415857101433548226)")
```

- Windows: App doesn't even launch, panics with

```thread 'main' panicked at C:\Users\<ME>\.cargo\registry\src\index.crates.io-6f17d22bba15001f\dioxus-fullstack-0.6.2\src\server\launch.rs:104:73:
called `Result::unwrap()` on an `Err` value: Os { code: 10048, kind: AddrInUse, message: "Normalerweise darf jede Socketadresse (Protokoll, Netzwerkadresse oder Anschluss) nur jeweils einmal verwendet werden." }
```

(Sorry, my Windows is German, but `AddrInUse` should be clear.)

**Environment:**
 - Dioxus version:  0.6.2
 - Rust version:    1.83.0
 - OS info:         Ubuntu 24.04 / Windows 11
 - App platform:    `desktop`

There is probably a lot I can configure in `Dioxus.toml`, but the documentation is somewhat scattered, the `bundle` sections are unclear to me, also I'm not sure how to write my `main` function properly. Currently it looks like this, I'm still toying around:

```rust
fn main() {
  LaunchBuilder::new()
    .with_cfg(desktop! {
      println!("cfg desktop");
      if cfg!(target_os = "windows") {
        // Someone said this should work on Windows, but I haven't tested it yet because the app panics early
        let data_dir = std::env::var("LOCALAPPDATA").expect("env var LOCALAPPDATA not found");
        dioxus::desktop::Config::default().with_data_directory(data_dir)
      } else {
        dioxus::desktop::Config::default()
      }
    })
    .with_cfg(server_only! {
      println!("cfg server_only");
      ServeConfigBuilder::default()
    })
[...]

PR #3693, opened 2025-02-04
Use different release profiles for desktop and server builds, adhoc profiles

### LINKED ISSUE #3573: `dx bundle` maybe bundle the wrong desktop binary file
**Problem**

For a fullstack app, when bundling to desktop, dx build two binary files *in parallel*:
- the server binary (`--feature server`)
- the desktop binary (`--feature desktop`)

Both of them are output to the same directory with the same file name, and yes, one of them will be overwriten by another. So the finally bundled file maybe is the server binary instead of desktop binary.

Update: I'm using `CARGO_TARGET_DIR`.

**Steps To Reproduce**

- follows the HotDog tutorial, enable `fullstack` feature and add server functions
- run `dx bundle --platform desktop` (also follows the tutorial)

**Expected behavior**

the desktop (GUI) binary is always bundled, and the server binary should be bundled too.

**Screenshots**

n/a

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version: 1.83.0
 - OS info: Windows 10, 22H2, 19045
 - App platform: `fullstack` and `desktop`

**Questionnaire**

### LINKED ISSUE #3600: `dx bundle` does not respect `--profile`
## Feature Request

Allow users to specify the profile to use when building as part of a `dx bundle --profile my-release-profile`.

## Implement Suggestion

Turn [this](https://github.com/DioxusLabs/dioxus/blob/c2952a79f8fdd5ca1bdf687c05ba1407e7fe8cac/packages/cli/src/cli/bundle.rs#L47-L48)
```rust
// We always use `release` mode for bundling
self.build_arguments.release = true;
```

Into this
```rust
// Default to `release` mode when bundling
self.build_arguments.profile = self.build_arguments.profile.or(Some("release".to_string()));
```

I doubt that's everything that would need to happen, but it would be a start.

More generally, I'm very surprised to see that the `BuildArgs::release` and `BuildArgs::profile` clap fields aren't mutually exclusive, and that neither are accepted from CLI when bundling. I haven't walked the whole cli crate, but I'm guessing the former is used for more than just selecting the cargo profile -- which is why I don't think the naive change above will be sufficient for this -- but it's very surprising to see the profile clobbered by the always-true `release` field.

### LINKED ISSUE #3884: Bundled fullstack app for macOS crashes on server request
**Problem**

Default fullstack project crashes on server request when bundled for OS X.

**Steps To Reproduce**

Create a new project:

```sh
dx new test && cd test
```

Options:

- `? 🤷   Which sub-template should be expanded?` — anything
- `? 🤷   Do you want to use Dioxus Fullstack?` — `true`
- `? 🤷   Do you want to use Dioxus Router?` — anything
- `? 🤷   Do you want to use Tailwind CSS?` — anything
- `? 🤷   Which platform do you want DX to serve by default?` — `Desktop`

Fill `[bundle]` section in `Dioxus.toml`:

```toml
[bundle]
identifier = "com.mycompany"
publisher = "MyCompany"
```

Bundle application for macOS:

```sh
dx bundle --platform desktop \
    --package-types "macos" \
    --package-types "dmg"
```

Open application (directly or installed from the `.dmg`) and type anything in the `ServerFn Echo` input.

Application closes immediately.

**Expected behavior**

There should be an echo response from the server.

**Environment:**

- Dioxus version:  `0.6.3` (as resolved by Cargo and set in `Cargo.lock`)
- Dioxus CLI version: `0.6.3` (via `cargo-binstall`)
- Rust version:    `1.85.0`
- OS info: `macOS 15.2`
- App platform: `desktop`

ealmloff:
In release builds, both desktop and server builds were using the same target and profile so they output their binary to the same location. This was causing either the server or the client to be used for both bundles. This PR fixes the issue by introducing release variants of the desktop and server profiles. Now when the server and desktop targets are built they output to a separate profile folder inside of `/target`

Fixes https://github.com/DioxusLabs/dioxus/issues/3689
Fixes #3600
Fixes #3573
Fixes #3884
Fixes https://github.com/DioxusLabs/dioxus/issues/4168


jkelleyrtp:
I was chatting with the bevy folks about https://github.com/DioxusLabs/dioxus/issues/4168 and they pointed me to some code that lets you create adhoc profiles. They do it here:

https://github.com/TheBevyFlock/bevy_cli/blob/main/src%2Fweb%2Fprofiles.rs

Since this PR has sat here so long and I've neglected to merge it, I'll adjust it so it doesn't generate profiles anymore, but instead just uses the adhoc profile approach.

Sorry for refusing to merge it, but I guess it worked out since we found a better solution.

It's worth noting that we should document these profiles.

PR #3696, opened 2025-02-05
use main exe name without extension for bundler on windows

oletf:
fixes msi & nsis bundles having `.exe.exe` extension in installer created shortcuts [as mentioned here](https://github.com/DioxusLabs/dioxus/issues/3233#issuecomment-2637001313).

Haven't tested on other platforms.


jkelleyrtp (approved):
Thanks!

PR #3698, opened 2025-02-05
docs: improved issue templates

Andrew15-5:
The "Please add the corresponding labels to the issue." line started the chain reaction (issue OP can't add labels), and now I just overall polished the current templates.

From [the docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), the non-YAML templates are considered legacy templates, but the documentation for them is [still up](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository). I'm not user if we should switch. But Typst uses them and I didn't have any problems with them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve Dioxus
+labels: bug
 ---
 
 **Problem**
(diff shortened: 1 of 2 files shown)


Andrew15-5:
Actually, I requested this for Typst, and I think for Dioxus this also will be helpful. I want to have a separate docs template. This is what I very frequently wanted to use in Typst, but they didn't have it until I asked.

Andrew15-5:
Oh, and with the form style template you can choose the platform from the dropdown menu, and it should be able to automatically add the appropriate labels. Though I haven't tried/seen that. So I guess there is a good reason to switch to them.

jkelleyrtp (approved):
Looks good, thank you!

PR #3722, opened 2025-02-09
feat(html): add scroll method to MountedData

sehnryr:
Implement `Element`'s `scroll` method https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll.

*Note that this method only works for "scrollable" elements (i.e. overflowing elements) (See 10. https://drafts.csswg.org/cssom-view/#dom-element-scroll)*

Added 2 examples:
- `scroll_to_offset`: programmatic scrolling
- `router_restore_scroll`: restore scroll position with a contextually scoped scroll value


jkelleyrtp (approved):
LGTM, thank you!

PR #3734, opened 2025-02-13
fix(interpreter): missing scrollTo boolean return value

sehnryr:
`scrollTo` missed a boolean return value so upstream queries (like in `DesktopElement`) couldn't deserialize any bool, and calls to `scroll_to` would always result in an error.

https://github.com/DioxusLabs/dioxus/blob/fc1f1c26f1b5e732c958c83a1665a0a7e5efd205/packages/desktop/src/element.rs#L93


jkelleyrtp (approved):
Good find, thanks!

PR #3741, opened 2025-02-15
Fix global shortcut event handler being called twice

### LINKED ISSUE #3739: Global shortcuts trigger twice
**Problem**

When using `dioxus_desktop::hooks::use_global_shortcut`, the handler is called twice. Once when the key is pressed, and once when it's released.

This bug probably got introduced when the global-hotkey crate was updated to 0.4.0, as that version's change log mentions adding key state to the event: https://github.com/tauri-apps/global-hotkey/releases/tag/global-hotkey-v0.4.0

The source of the problem lies here: https://github.com/DioxusLabs/dioxus/blob/fc1f1c26f1b5e732c958c83a1665a0a7e5efd205/packages/desktop/src/shortcut.rs#L68

The `id: GlobalHotKeyEvent` parameter has a field that says if the shortcut was pressed or not. It is not used in the function or passed onto handlers.

**Steps To Reproduce**

- Add a global shortcut using `use_global_shortcut`
- When triggered, the event is called both on press and release

Running the `shortcut` example project also displays the incorrect behavior.

**Expected behavior**

The event should only be called when pressed, or for a future breaking change the pressed state should be passed to the event handler.

**Environment:**

- Dioxus version: 0.6.3
- Rust version: 1.84.1
- OS info:  Windows 11 23H2
- App platform: desktop

Skippeh:
Fixes #3739

### DIFF
--- a/packages/desktop/src/shortcut.rs
+++ b/packages/desktop/src/shortcut.rs
@@ -66,6 +66,10 @@ impl ShortcutRegistry {
 
     #[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
     pub(crate) fn call_handlers(&self, id: GlobalHotKeyEvent) {
+        if id.state != global_hotkey::HotKeyState::Pressed {
+            return;
+        }
+
         if let Some(ShortcutInner { callbacks, .. }) = self.shortcuts.borrow_mut().get_mut(&id.id) {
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Hey, thanks for the PR!

Another PR fixed this and I randomly chose to push that one along to add the state object instead.

https://github.com/DioxusLabs/dioxus/pull/3822

PR #3743, opened 2025-02-15
fix: fixed error of example video_stream create video file

osenberg-x:


jkelleyrtp:
I think this might depend on where you run `cargo run` or `dx run` from. When running from the root directory it works.

Issue #3596, opened 2025-01-20
Default icon


Klemen2:
## Feature Request

Currently, when bundling an app with an icon option in the Dioxus.toml bundle, the provided icon will not be set for the window or trayicon (at least not on windows 11) - this is because if no icon is explicitly set, the default dioxus icon will be set instead. Preferably setting the icon in the toml file would also set it for the window and trayicon and only if none is set it would use the default.

Referenced code where default icon is loaded:
https://github.com/DioxusLabs/dioxus/blob/eaf9f80abddd75a6963edf2f48e8e21765146122/packages/desktop/src/webview.rs#L195-L204
https://github.com/DioxusLabs/dioxus/blob/eaf9f80abddd75a6963edf2f48e8e21765146122/packages/desktop/src/trayicon.rs#L37-L45

PR #3753, opened 2025-02-16
feat: windows app icon

### LINKED ISSUE #3648: Icon option in Dioxus.toml does not work on Windows
**Problem**

On Windows, if you try to change the location of your icon image via editing dioxus.toml under [bundle] using the tag icon, your choice is not respected when bundling for Desktop Windows. It will automatically default to `project\\icons/icon.ico\`

**Steps To Reproduce**

- Create Windows app
- Set icon in Dioxus.toml
- Attempt to bundle

**Expected behavior**

The icon should be selected from the path given in the icon tag.

**Screenshots**

Image

**Environment:**
 - Dioxus version:   0.6.2
 - Rust version:     1.84.0
 - OS info:          Windows 11 Pro
 - App platform:     Desktop Windows

**Workaround**

This can be worked around by creating the file and folder that Dioxus tries to access `icons\icon.ico`.

Klemen2:
image

Closes: #3648
Implement #3596 for windows

Windows app resources will only be build with --release flag, otherwise it will be embedded (no app icon (thumbnail) or metadata when installed)

This pr also reduces the binary size since including icon is now deduplicated.

Added functions: default_icon, icon_from_memory, icon_from_path

Tested only on windows 11


photino:
Why didn't previous versions support Windows app icons?

packages/cli/src/build/request.rs
```diff
@@ -4575,4 +4580,110 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
 
         Ok(())
     }
+
+    fn winres_icon_path(&self) -> crate::winres::WindowsResourceLinker {
+        todo!()
```
rohitsangwan01: This is causing issue for me, any idea when this can be merged ?

Klemen2: You can use [6321f1f](https://github.com/DioxusLabs/dioxus/tree/6321f1fc72fc37bf6d928e283055b90f347c6cd1) or my main fork, while it's being refactored.

packages/cli/src/build/request.rs
```diff
@@ -4575,4 +4580,110 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
...
+    }
+
+    fn write_winres_icon(&self) -> Result<crate::winres::WindowsResourceLinker> {
```
Klemen2: This should probably be named write_winres or compile_winres since it also adds other metadata to the app, not just the icon

packages/cli/src/build/request.rs
rohitsangwan01: @Klemen2 can you bump your fork to alpha 3 ?

Klemen2: The [main](https://github.com/Klemen2/dioxus/tree/main) branch of the fork is synced with main and has this PR merged in it @rohitsangwan01 
You will have to define .ico icon in your bundle settings since something broke between 5 months ago and now on how the default icon is handled if you don't have .ico icon defined

Klemen2:
In the docs somewhere there should be a notice that windows is very bad at updating app icons of installed apps, so if you bundle the app, install it, change the icon and bundle again and reinstall - it will look buggy with old icon in some places and new one on other, with the only fix being restarting the pc or at least restarting the explorer didn't seem to fix it on my machine

pythoneer:
Anything left i can help with to bring this over the finish line? 0.7 is is nearing its release and i would love to not use a workaround for my current releases with 0.6 to have app icons.

Klemen2:
> Anything left i can help with to bring this over the finish line? 0.7 is is nearing its release and i would love to not use a workaround for my current releases with 0.6 to have app icons.

There are a few changes that needs to be done. 

- [ ] 1. testing bundling apps for windows on mac/linux
- [x] 2. It should probably not fail to build for windows if you do not have the tooling, it should only fail for bundling and release builds this requires changing how default icon is handled on windows (embedded with include_bytes instead of loaded from_resource) and if that's the case notify the user that there will be no icon for the app
- [x] 3. Make it work with workspaces, since currently it doesn't compile unless you cd to the app crate itself (e.g will not work with #4811) - there should probably be a test for this so you see that it fails
- [ ] 4. I noticed that the logs "Compiling Windows resource file with ..." gets outputted twice, it should just be once
- [x] 5. default icon handling for other platforms (related to 2. but ideally it would get it from the app itself and fallback  to include_bytes for non release/bundle) -- not required for this PR, but would be nice for consistency
- [ ] 6. caching the windows resource file so it only compiles when the associated files changes  -- not required for this PR

I think this is everything unless @jkelleyrtp has anything to add, feel free to work on this

Klemen2:
microsoft has since added windows-resource to crates, but before they release anything it will be another half a year or more or it may not even be related to the resource compiler based on their sense of naming things

Klemen2:
I think that this is pretty much finished now. From the comment above 1. is probably out of scope but should work if you have the tools setup correctly, 
4. is not a big issue and I don't really know why is happening
6. would be nice to have, but not really important

pythoneer:
I think the work here is relevant https://github.com/DioxusLabs/dioxus/pull/4958 idk.

pythoneer:
This PR https://github.com/DioxusLabs/dioxus/pull/5072 worked on the same code location and created a conflict.

jkelleyrtp:
Sorry this has sat so long. I'm just not a fan of the winres approach. 

Deno built a crate called sui which can embed resources into binaries after they've been built.

https://littledivy.com/sui.html
https://github.com/denoland/sui

It specifically handles embedding icons into PE (windows exe files).

I would prefer we use sui to embed the icon for windows instead of the build.rs / winres system.

Alternatively, we could perform the compilation of the icon in the CLI and then pass it in as a linker arg.

https://claude.ai/share/b0f8e502-2932-4734-b4b0-da1241a262c1

I just don't want icon configuration to be done from within the build.rs. The dioxus.toml should be the source of truth for these configuration options and `dx` should perform the linker flag modifications if needed.
```
// ❌
println!("cargo:rustc-env=DIOXUS_APP_ICON=./assets/default_icon.png")
```

I might also just be misunderstanding the PR. It seems that the inlined winres system has code for printing link statements and reading the cargo_manifest_dir, which I don't think is applicable here

https://github.com/klemen2/dioxus/blob/2322b03a74b1b13e53327d803cd20ea0255b3091/packages/cli/src/build/winres.rs#L782-L786

Klemen2:
winres was already doing compilation of the icon in the cli and was not using any cargo / build script since the first working version of this PR - basically what cloud suggested, the original was based on the tauri winres but this version is without embed_resource since it prints to the console, the tooling finder is from the script that zed is using [winresource](https://github.com/BenjaminRi/winresource), I removed a bunch of irrelevant comments and unused stuff and also changed the structure a little

I haven't tested the gnu toolkit or on other platforms, only mscv on windows only

`println!("cargo:rustc-env=DIOXUS_APP_ICON=./assets/default_icon.png")` was done so that #3596 could be closed - now it's only implemented for windows, there is probably a better way of doing that on other platforms instead of embedding the icon in the exe but sui could be used for quick implementation, but I would need to look into it how and where it can be implemented

pythoneer:
relevant PR https://github.com/DioxusLabs/dioxus/pull/5328

packages/cli/src/build/request.rs
```diff
@@ -2708,9 +2708,7 @@ impl BuildRequest {
                 }
                 Err(err) => {
                     if self.using_dioxus_explicitly {
-                        panic!("Windows resources couldn't be built. Err: {}", err)
-                    } else {
-                        tracing::warn!("Application may not have an icon.")
+                        tracing::warn!("Application may not have an icon: {err}");
```
Klemen2: I don't think that this is a good idea, since the app will crash in dioxus if the icon fails to compile on windows as it's only loaded from resources so I think it would be better to fail the build instead, especially for release builds

Klemen2: The default_icon can be change to stop the panic, but at the same time, if you are also using the tray icon, the tray will have a transparent icon that is very hard to even notice that is there - this probably could be fixed upstream.

pythoneer:
I think https://github.com/DioxusLabs/dioxus/pull/5412 and https://github.com/DioxusLabs/dioxus/pull/5328 caused conflicts again. Not sure how this affects this PR :(

jkelleyrtp:
Would love to merge this. What's the current state?

After digging into windows bundling recently, I have several thoughts:
- `dx bundle` shells out to wix / nsi which have options to include icons. For many users this could just be enough? You don't get icons during dev, but you do in bundle, which is really what matters.
- There is a crate called sui which lets you inject icons and files into binaries after they've been built (https://github.com/denoland/sui). This might be pretty simple.
- We can just add the icon via the linker - ie dx runs `cargo rustc -- -C link-arg=app.res` automatically. I think that's what's happening in this PR?

It seems that on many platforms we need some sort of pre-processing step for icon generation (android as well, and mac/ios). This adds extra overhead in dev (slower dev-server starts until we properly parallelize things), which is why we've deferred it to bundle. 

I'm happy to add this step in dev too if folks think it's important.

packages/desktop/src/default_icon.rs
```diff
@@ -0,0 +1,173 @@
...
+    where
+        Self: Sized;
+    fn path<P: AsRef<Path>>(path: P, size: Option<(u32, u32)>) -> Result<Self>
+    where
+        Self: Sized;
+}
+
+// preferably this would have platform specific implementations, not just for windows
+#[cfg(not(target_os = "windows"))]
+static DEFAULT_ICON: &[u8] = include_bytes!("./assets/default_icon.png");
```
Klemen2: as for other platforms than windows, the easy way would be to replace this with an user icon for the default icon from toml to work, so if thats fine I can do that

Klemen2 (commented):
As far as I know, wix / nsi settings only add icon to the installer, not the actual app.
the pr is basically doing cargo rustc -- -C link-arg=app.res where res needs to be compiled with rc.exe (or equivalent) because windows is special.

I don't think that dev builds necessarily need an icon, so atm you just get a warning if it can't be compiled.

I think that I only need to update a couple of things which I can probably do this weekend.

packages/desktop/src/default_icon.rs
Klemen2: there is probably a better way of doing that than this, but I don't think that I have time for that

Klemen2: as far as I can tell, the window icon matters only on windows, wayland and x11, for others it shouldn't have any effects for the current stable tao/winit

for tray icon it matters on all desktop platforms, so it probably would be a good idea to implement it for all desktop platforms, not just windows

I've changed it right now so it just returns an error which makes it so there is no icon

Klemen2:
I'd say this is ready, unless you want any changes in the windows scope.

jkelleyrtp:
Thanks so much!

jkelleyrtp (approved):

adamgreg:
The addition of `ApplicationConfig::name` in this PR is very interesting! It looks like it's not really used yet, but is it intended as a way to override the application name in bundles, so it doesn't have to be a PascalCase version of the crate name? That would be really useful for my app.

Is anyone working on that? I couldn't find an open PR. I'd be willing to give it a try myself if that would be welcome.

Klemen2:
right now it's only used for the windows product name (app -> right click -> properties -> details)
the app name is not yet affected but fell free to implement it

adamgreg:
> right now it's only used for the windows product name (app -> right click -> properties -> details) the app name is not yet affected but fell free to implement it

FYI I just raised #5533 for this

Issue #3688, opened 2025-02-04
bundle: asset folder gets installed to relative full source path as destination


oletf:
**Problem**

title seems self explanatory to me, although confusing, so details below

**Steps To Reproduce**

I got for assets a css file and a font folder with multiple files in it
```
assets
├── fonts
│   ├── font1.woff2
│   └── font2.woff2
└── main.css
```
```rust
const STYLESHEET: Asset = asset!("/assets/main.css");
const FONT_FOLDER: Asset = asset!("/assets/font/");
```
then bundle to NSIS

**Expected behavior**
Expecting to have the same relative tree (with manganis suffixes added).

**Current behavior**
but within the install folder I get : 
```
assets
└── main-b8a30b510d2e26ae.css
Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1
├── font1.woff2
└── font2.woff2
```

Here's the relevant part of the generated `installer.nsi`
```
  ; Copy resources
    CreateDirectory "$INSTDIR\assets"
    CreateDirectory "$INSTDIR\/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1"
    File /a "/oname=/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1\font1.woff2" "C:/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1\font1.woff2"
    File /a "/oname=/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1\font2.woff2" "C:/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\fonts-d0aa87f5140a3bd1\font2.woff2"
    File /a "/oname=assets\main-b8a30b510d2e26ae.css" "C:/Users\[...full project path...]\target\dx\projectname\release\windows\app\assets\main-b8a30b510d2e26ae.css"
```

All the full path is recreated under the install folder just for the asset folder, individual files don't have any problem.
`FONT_FOLDER.resolve()` gives the expected `assets\fonts-d0aa87f5140a3bd1`, but of course doesn't find them

**Environment:**
 - Dioxus version:  0.6.2
 - Rust version:    1.84.1
 - OS info:   Windows 10
 - App platform:    desktop

[...]

PR #3763, opened 2025-02-19
fix: wix assets path

Klemen2:
Now all files are passed to tauri_builder instead of a mix of files and directories.
Closes: #3688


jkelleyrtp (approved):

jkelleyrtp:
Thank yoU!

Issue #3756, opened 2025-02-17
Document Level Tags Are Applied In Reverse Order


mcmah309:
**Problem**
Document level tags do not respect order. E.g. for
```rust
    rsx! {
            document::Script { src: "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js" }
            document::Script { src: "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/languages/rust.min.js" }
            document::Script { {r#"
document.addEventListener("DOMContentLoaded", () => {
    hljs.highlightAll();
});
"#} }
        }
    }
```
In browser:
```html
<head>
<script>
document.addEventListener("DOMContentLoaded", () => {
    hljs.highlightAll();
});
</script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/languages/rust.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
</head>
```

Also worth noting that muli-line str `r#" "#` breaks `dx fmt`.

**Expected behavior**

```html
<head>
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/languages/rust.min.js"></script>
<script>
document.addEventListener("DOMContentLoaded", () => {
    hljs.highlightAll();
});
</script>
</head>
```

**Screenshots**

**Environment:**

- Dioxus version:  0.6.2
- Rust version:    1.86.0 nightly
- OS info:         NixOS 25.05
- App platform:    web

**Questionnaire**

I'm interested in fixing this myself but don't know where to start.

ealmloff:
Dioxus mounts nodes in a reverse order [here](https://github.com/DioxusLabs/dioxus/blob/main/packages/core/src/diff/node.rs#L746-L758). Outside of a single rsx block, I'm not sure there is a clear order for head elements. Head elements could be loaded asynchronously based on the result of a resource. A script that loads immediately will be inserted before a script that loads asynchronously regardless of the depth first order of the script elements in the tree.

You can use imports that don't rely on the order of script tags. es6 imports seem like a much more predictable way to handle dependencies:
```rust
use dioxus::prelude::*;

[...]

PR #3771, opened 2025-02-20
doc: Add head ordering doc part

mcmah309:
Document: https://github.com/DioxusLabs/dioxus/issues/3756

I chose not to leak the implementation information here and rather advise users to not depend on ordering. Otherwise users depending on ordering may have negative downstream effects - changes to the implementation may case downstream code to break without users knowing why (which may happen between non-breaking semver versions), this can easily become out of date, ordering is already not guaranteed with async and changing components.

### DIFF
--- a/packages/document/docs/head.md
+++ b/packages/document/docs/head.md
@@ -16,6 +16,7 @@ Components that render into the head of the page do have a few key limitations:
 
 - With the exception of the `Title` component, all components that render into the head cannot be modified after the first time they are rendered.
 - Components that render into the head will not be removed even after the component is removed from the tree.
+- Components that render into the head do not have a guaranteed ordering; thus, components should ideally not be order dependent, since they may not appear in the head in the order they are defined.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):
This is fine for now, but we should really make the ordering *stable*. I'm not exactly sure how other frameworks go about this (namely effect ordering) but the fact that components don't run effects in a sensible order will lead to more bugs and quirks. I'd rather not have to document work-arounds and instead make this work predictably. 

Documenting is fine as a stop-gap measure but we should try to fix the underlying issue. I've put this into our own bug backlog, but let's make sure the github issue stays around and tracks the bug properly.

mcmah309:
Do you want me to close https://github.com/DioxusLabs/dioxus/issues/3756 or is this the issue that was added to the backlog? If so, it should probably be renamed from "Document Level Tags Are Applied In Reverse Order" to "Stabilize Document Level Tag Ordering" and link back to this comment for future reference.

Issue #3774, opened 2025-02-20
Dioxus Logger Env Filtering


mcmah309:
## Feature Request

Currently `dioxus_logger` only filters on `Level`, this pollutes the logs with all the tracing logs, like from Dioxus itself.

<details>
    <summary>Example</summary>

```console
mplate { roots: [Dynamic { id: 0 }], node_paths: [[0]], attr_paths: [] }, dynamic_nodes: [Component(VComponent { name: "dioxus_core::suspense::component::SuspenseBoundary" })], dynamic_attrs: [] }, mount: Cell { value: MountId(18446744073709551615) } }) parent=None}:VirtualDom::create_scope{scope=ScopeId(2, "dioxus_core::error_boundary::ErrorBoundary") new_nodes=Ok(VNode { vnode: VNodeInner { key: None, template: Template { roots: [Dynamic { id: 0 }], node_paths: [[0]], attr_paths: [] }, dynamic_nodes: [Fragment([VNode { vnode: VNodeInner { key: None, template: Template { roots: [Dynamic { id: 0 }], node_paths: [[0]], attr_paths: [] }, dynamic_nodes: [Component(VComponent { name: "root" })], dynamic_attrs: [] }, mount: Cell { value: MountId(18446744073709551615) } }])], dynamic_attrs: [] }, mount: Cell { value: MountId(18446744073709551615) } }) parent=Some(ElementRef { path: ElementPath { path: [0] }, mount: MountId(1) })}:VirtualDom::run_scope{scope_id=ScopeId(3, "root")}:render{scope=root}: dioxus_core::scope_context: providing context dioxus_signals::global::GlobalLazyContext (TypeId(0x7ca54824a09ef4e988804ff4b802241f)) in ScopeId(0, "app")
[...]

PR #3775, opened 2025-02-20
Env logger feature

mcmah309:
Closes https://github.com/DioxusLabs/dioxus/issues/3774


jkelleyrtp (changes requested):
If we support the env filter, then we should support it for all our platforms:

- [x] mac
- [x] win
- [x] linux
- [ ] ios
- [ ] android
- [ ] web

The last three are the hardest but definitely quite important.

I believe for ios and android we just need to re-export the RUST_LOG env var into the spawned processes:

https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/serve/handle.rs#L78-L109

For these two platforms, env vars don't automatically inherit, so they have their own bespoke methods for receiving env.

For WASM, I'd like us to invent a way of passing env vars into wasm (allowing env::var *in* WASM) but that's outside the scope of this PR, so it's fine to leave WASM unimplemented.

I would also like to make this the default behavior of the logger. We probably don't need to feature-gate it and instead just make the dioxus_logger crate runtime configurable instead.

mcmah309:
I tried testing on android, but I actually could not get any logs to show, even for the published `dioxus-cli` and `dioxus` for version `0.6.3` (no changes from this branch applied). Are you observing this issue?

jkelleyrtp (approved):
Awesome! I think we might have a conflict with `dx serve` and the child colliding since both rely on env logging, but that shouldn't be a huge issue

Edit: can you fix the clippy error?

jkelleyrtp (changes requested):
clippy!

jkelleyrtp (approved):

PR #3776, opened 2025-02-20
Revision: Open Browser With Localhost

### LINKED ISSUE #3766: Serving a web app with `--open` or 'o' opens 10.0.0.32 instead of 127.0.0.1
Serving a web application with `--open` or pressing 'o' opens 10.0.0.32 instead of 127.0.0.1 (ie localhost), which is shown in the terminal. 

To reproduce: 
```
dx serve --example hello_world --platform web --open
```

Localhost is treated as a secure context by browsers, which is required for many web API's. This is how pages might fail. For instance, the fullstack-auth example fails to store secure cookies.

This bug is fairly easily worked around by either copying the address from the terminal or opening `http://127.0.0.1:8080` manually.

DogeDark:
Closes #3766

### DIFF
--- a/packages/cli/src/cli/run.rs
+++ b/packages/cli/src/cli/run.rs
@@ -25,14 +25,16 @@ impl RunArgs {
 
         let devserver_ip = "127.0.0.1:8081".parse().unwrap();
         let fullstack_ip = "127.0.0.1:8080".parse().unwrap();
+        let mut open_address = None;
 
         if self.build_args.platform() == Platform::Web || self.build_args.fullstack {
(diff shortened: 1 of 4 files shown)


jkelleyrtp (approved):
great! though we might end up winding back the 0.0.0.0 changes at some point since we get bugs like #3787 and people seem to report windows firewall getting mad.

In that case we'd still want to "open" the "displayed address" but make the 0.0.0.0 an explicit choice.

DogeDark:
The firewall is a good point!

PR #3782, opened 2025-02-21
reclaim el id for nested dyn nodes

### LINKED ISSUE #3760: Memory leak/Detached DOM nodes when component arg with type Signal<Vec<T>> changes lengths across mutations
**Problem**

Rendering a `Signal<Vec<T>>` with something like the following causes unbounded DOM node growth when the `Vec<T>` can change length up and down between renders. It creates a bunch of detached nodes as can be seen in the image. 

Image

```rust
rsx! { 
  for item in vec_signal { 
    Child { item }
  }
}
```

**Steps To Reproduce**

```rust
use dioxus::prelude::*;
use uuid::Uuid;
use wasm_bindgen::prelude::wasm_bindgen;
use web_sys::js_sys;

#[wasm_bindgen]
pub fn sleep(ms: i32) -> js_sys::Promise {
    js_sys::Promise::new(&mut |resolve, _| {
        web_sys::window()
            .unwrap()
            .set_timeout_with_callback_and_timeout_and_arguments_0(&resolve, ms)
            .unwrap();
    })
}

pub async fn shleep(ms: i32) {
    wasm_bindgen_futures::JsFuture::from(sleep(ms))
        .await
        .unwrap();
}

fn main() {
    dioxus::launch(app);
}

fn app() -> Element {
    let mut nums = use_signal(|| vec![]);

    let _tx = use_coroutine(move |_rx: UnboundedReceiver<()>| async move {
        let mut i = 0;
        loop {
            i += 1;
            shleep(100).await;
            if i % 2 == 0 {
                nums.set((i..i + 1).collect());
            } else {
                nums.set((i..i + 2).collect()); // switch this to i + 101 to make the detached nodes grow faster
            }
        }
    });

    rsx! {
        div {
            Child { nums }
        }
    }
}

#[component]
fn Child(nums: Signal<Vec<i32>>) -> Element {
    rsx! {
        for num in nums() {
            div { "Num: {num}" }
        }
    }
}
```

**Expected behavior**

Nodes should drop without hanging around in a detached state

**Screenshots**

**Environment:**

- Dioxus version:  0.6.3
- Rust version:    1.82.0
- OS info:         macOS
- App platform:    web

gammahead:
element ids associated with nodes removed via `fn remove_nested_dyn_nodes` are never reclaimed. This throws off the ordering of adding and removing elements from a list 

fixes #3760


gammahead:
@ealmloff can you give this a quick look? It's small and I think you'll easily be able to tell me if it's a step in the wrong direction.

packages/core/src/diff/node.rs
```diff
@@ -328,6 +328,9 @@ impl VNode {
                     to.remove_node(id);
                 }
                 dom.reclaim(id);
+            } else {
+                let id = dom.get_mounted_root_node(mount, idx);
```
ealmloff, line 332: Reclaiming the root mounted node here when mutations is None seems odd. If this is a suspended node, that id should never have been set [here](https://github.com/DioxusLabs/dioxus/blob/797e7492a1b9e188627ab7530a66607ba19376d6/packages/core/src/diff/node.rs#L607-L609)

packages/core/src/diff/node.rs
```diff
@@ -379,8 +382,8 @@ impl VNode {
                     } else {
                         to.remove_node(id);
                     }
-                    dom.reclaim(id)
                 }
+                dom.reclaim(id)
```
ealmloff, line 386: Same thing here. If the node was created under suspense [here](https://github.com/DioxusLabs/dioxus/blob/797e7492a1b9e188627ab7530a66607ba19376d6/packages/core/src/diff/node.rs#L672-L674). I don't think it would have an id

ealmloff:
I think the `None` mutations variant is only used in suspense (for example [here](https://github.com/DioxusLabs/dioxus/blob/main/packages/core/src/suspense/component.rs#L313)). It disables writing the mutations to the dom while the component is suspended. You could check if the bug is caused by suspense by removing any suspense components in your reproduction and disabling the [root suspense component](https://github.com/DioxusLabs/dioxus/blob/main/packages/core/src/root_wrapper.rs#L16-L20)

gammahead:
@ealmloff I'm not explicitly using suspense boundaries anywhere. The `None` mutations variant is also used in `remove_nested_dyn_nodes` which is where the problem arrises for me. My understanding is that, since this recurses through the entire subtree of nodes with `None` mutations, all children element ids fail to be reclaimed regardless of suspense context. 

If the `Option<&mut impl WriteMutations>` was intended to communicate a suspense context when it's `None`, then perhaps we need an enum with 3 variants ?

ealmloff:
> @ealmloff I'm not explicitly using suspense boundaries anywhere. The `None` mutations variant is also used in [remove_nested_dyn_nodes](https://github.com/DioxusLabs/dioxus/blob/797e7492a1b9e188627ab7530a66607ba19376d6/packages/core/src/diff/node.rs#L349C31-L349C34) which is where the problem arrises for me. My understanding is that, since this recurses through the entire subtree of nodes with `None` mutations, all children element ids fail to be reclaimed regardless of suspense context.
> 
> If the `Option<&mut impl WriteMutations>` was intended to communicate a suspense context when it's `None`, then perhaps we need an enum with 3 variants ?

Sorry it took so long to get back to you. Yes, I think that is correct there are three cases for `remove_dynamic_node`:
1) The nodes were suspended, so they were never actually allocated ids and we don't need to reclaim them
2) The nodes were created nested with ids. They do exist in the node slab and their ids need to be removed, but the dom nodes were already removed when the root elements were removed from the dom
3) The nodes are not suspended and are root nodes. They do exist in the node slab and we need to generate mutations to remove them from the dom.


For `reclaim_roots`, it could be called with None outside of suspense if it is called from `remove_dynamic_node` in case `2`


We could add a separate "remove_from_dom" bool that is true just in case 3 or switch to a three variant enum

gammahead:
@ealmloff picking this back up since a couple more people ran into #3760. You see a reason not to just always reclaim the ID, regardless of whether we're writing mutations?

jkelleyrtp:
I spent a few hours digging into this today and then spun out a branch that properly wires the "should_write" flag into the callsites:

https://github.com/DioxusLabs/dioxus/pull/4352

Suspense is a less-used feature and I'd rather have regular vdom code not leak than the extra logging in the suspense case. Seems like playwright and tests pass just fine even with suspense.

What I think is happening is that our call to remove_nested_dyn_nodes passes a `None` for `to` (the mutation writer) which then hits this case. In some cases those nodes *should* be reclaimed, but because we assume they will be somewhere else, we're missing the relevant case.

PR #3788, opened 2025-02-22
Expose Outlet Context

wheregmis:
AI Description

Here are the key reasons for exposing `OutletContext`:

- **Custom Route Rendering**
  - Implement custom outlet rendering logic
  - Control how nested routes are displayed
  - Override default outlet behavior when needed

- **Advanced Use Cases**
  - Build middleware that needs route depth information
  - Implement route-based caching strategies
  - Create nested loading states with different behaviors

- **Component Libraries**
  - Build reusable navigation components
  - Create layout systems that adapt to route depth
  - Implement smart containers that respond to routing context

These use cases justify exposing `OutletContext` as part of the public API, allowing developers to build more sophisticated routing-aware features.

My Description:
Outlet is handling nested routing and levels, and gives the level to the router to render with the level. So without OutletContext and having Somekind of animation between page route components will miss the Layout component during the animation period. 

Current Example Gif:
ezgif-8dca0cb88546e6

In the gif above, Blog Layout is disappearing, and i can't apply custom rendering based on any levels.

If i can increment level during the animation phase as well
```rust
    provide_context({
        OutletContext::<R> {
            current_level: current_level + 1,
            _marker: std::marker::PhantomData,
        }
    });
```
My custom component can mimic as an Outlet which have all the goodies that dioxus Outlet provides and i can do custom rendering with the level and the router.

New Example Gif:
ezgif-8f73855265bb9f

We still have super tight coupling of Rendering with Router. Eventually we might want to decouple things a bit, so other developers can use different rendering ways for their use case. Like i want to render just the layout without any child route, which is not easy for the time being. Maybe in future we can add some stuffs here

```rust
pub trait Routable: FromStr + Display + Clone + 'static {
    /// The error that can occur when parsing a route.
    const SITE_MAP: &'static [SiteMapSegment];

    /// Render the route at the given level
    fn render(&self, level: usize) -> Element;

    /// Render the layout only at the given level
    fn render_layout(&self, level: usize) -> Element;

    /// Render the child route only at the given level
    fn render_child_route(&self, level: usize) -> Element;
```

To have more flexibility for the devs to build more complex animations or rendering strategies

### DIFF
--- a/packages/router/src/contexts/outlet.rs
+++ b/packages/router/src/contexts/outlet.rs
@@ -10,37 +10,55 @@ use crate::{routable::Routable, utils::use_router_internal::use_router_internal}
 /// # Type Parameters
 ///
 /// * `R` - The routable type that implements the routing logic
-///
-/// # Fields
-///
-/// * `current_level` - The current nesting level of the route
-/// * `_marker` - Phantom data to hold the generic type parameter
-///
-/// # Examples
-///
-/// ```rust
-/// let outlet = OutletContext {
-///     current_level: 1,
-///     _marker: std::marker::PhantomData,
-/// };
-/// ```
+#[derive(Clone)]
 pub struct OutletContext<R> {
-    /// The current nesting level of the route in the outlet hierarchy.
-    /// Level 0 represents the root route, and each nested route increases the level by 1.
-    pub current_level: usize,
-    /// Phantom data marker to hold the generic type parameter `R`.
-    /// This field is not used at runtime and has zero size.
-    pub _marker: std::marker::PhantomData<R>,
+    current_level: usize,
+    _marker: std::marker::PhantomData<R>,
 }
 
-impl<R> Clone for OutletContext<R> {
-    fn clone(&self) -> Self {
-        OutletContext {
-            current_level: self.current_level,
+impl<R> OutletContext<R> {
+    /// Creates a new outlet context starting at level 1
+    pub fn new() -> Self {
+        Self {
+            current_level: 1,
             _marker: std::marker::PhantomData,
         }
     }
+
+    /// Creates a new outlet context for the next nesting level
+    pub fn next(&self) -> Self {
+        Self {
+            current_level: self.current_level + 1,
+            _marker: std::marker::PhantomData,
+        }
+    }
+
+    /// Returns the current nesting level of this outlet
+    pub fn level(&self) -> usize {
+        self.current_level
+    }
+
+    pub(crate) fn render() -> Element
+    where
+        R: Routable + Clone,
+    {
+        let router = use_router_internal().expect("Outlet must be inside of a router");
+        let outlet: OutletContext<R> = use_outlet_context();
+        let current_level = outlet.level();
+        provide_context(outlet.next());
+
+        if let Some(error) = router.render_error() {
+            return if current_level == 0 {
+                error
+            } else {
+                VNode::empty()
+            };
+        }
+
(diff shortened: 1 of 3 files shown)


packages/router/src/contexts/outlet.rs
```diff
@@ -2,8 +2,34 @@ use dioxus_lib::prelude::*;
...
+/// # Examples
+///
+/// ```rust
+/// let outlet = OutletContext {
+///     current_level: 1,
+///     _marker: std::marker::PhantomData,
+/// };
+/// ```
+pub struct OutletContext<R> {
+    /// The current nesting level of the route in the outlet hierarchy.
```
ealmloff, line 28: I think exposing the outlet context makes sense for your use case. You still need to keep track of what level you are rendering at, but you might render the child route at each outlet level or you might render the transition instead.

However, I don't think we need to expose the internals of the outlet context. Exposing the internal fields can make it more difficult to change things in a future minor release. Instead we could expose two methods:
- `OutletContext::new() -> Self` which creases a new outlet context at level 1
- `OutletContext::next(&self) -> Self` which creates a new outlet context at the next level

wheregmis: That would work as well. Appreciate the feedback will modify asap

wheregmis: I might also need access to the outlet level so i had to expose level with another addidional function than the proposed ones. Let me know if its okay, or not
```rust
pub fn level(&self) -> usize
```
Just to render the From and To route
 ```rust
{from.render(outlet.level())}
```

Comment (wheregmis) [reply packages/router/src/contexts/outlet.rs]: I have tested the my changes with https://github.com/DioxusLabs/dioxus/blob/main/examples/router.rs to verify i am not breaking anything.

**Review by jkelleyrtp — APPROVED** [review]
This looks awesome! Code looks good, thanks for the PR

**Comment by wheregmis** [comment]
Hi @mcmah309, thank you for pointing out the issues, 

The doc says the level will be 1 but it creates one with 0 <- This one is a typo from my end. I will create a pr asap to fix it.
I did tried with 1 as initialization, but router initialize the outlet context first. And it never initialized with 1 even before my PR, so i had to start with 0 to begin with. As router is the first one to initialize the outlet context. (First i did tried with 1 and router wasnt rendering properly)

Old Code.
```rust
 provide_context(OutletContext::<R> {
            current_level: 0,
            _marker: std::marker::PhantomData,
        });
```

Maybe, i might have to increase the counter here as well.

```rust
pub fn use_outlet_context<R: Clone + 'static>() -> OutletContext<R> {
  ##  I think this needs to be start from one. This went under my raidar as i didnt came under this use case issue.
    use_hook(|| try_consume_context().unwrap_or_else(OutletContext::new)). 
}
```

@ealmloff Can you confirm if there is something else that might be some potential issues?

ealmloff:
I think the new behavior is correct. Previously, the outlet context was always provided by the router [here](https://github.com/DioxusLabs/dioxus/blob/022e4ad203e62f94e5b4f44a38a12f1d55209868/packages/router/src/components/router.rs#L54-L57) at level 0. You cannot use the `Outlet` outside of the router, so I don't think the default case was ever getting hit. If it does get hit, then the 0 level is the right default. Every route has a `0` level rendering case, but only routes under a layout have a `1` level rendering case in the routable rendering match statement

wheregmis:
Great, then just fixing the typo in documentation should be good enough. Its not draft but i have marked it as draft just incase we might have some more discussion.

Issue #3791, opened 2025-02-23
[Bug] `onclick` event handlers broken within form component etc


caseykneale:
**Problem**

Certain handlers used on child components under a parent form component will brick an instance. This has been observed in desktop mode. The resulting state is a white window with all interactivity lost until execution of the binary is performed once again.

**Steps To Reproduce**

Clicking the button will result in a white screen for desktop mode, and it probably shouldn't.
```rust
#[component]
pub fn this_will_break_onclick() -> Element {
    rsx! {
        p { "I am text" }
        form {
            button { onclick: move |_| println!("Uhoh"), "Sync" }
        }
    }
}
````

This will not:
```rust
#[component]
pub fn this_wont_break_onclick() -> Element {
    rsx! {
        p { "I am text" }
        button { onclick: move |_| println!("Uhoh"), "Sync" } 
    }
}
````

This will not:
```rust
#[component]
pub fn this_wont_break_onclick() -> Element {
    rsx! {
        p { "I am text" }
        form {
            button { onclick: |_| println!("Uhoh"), "Sync" }
        }
    }
}
````

**Expected behavior**

Print the println and don't brick the application.

**Screenshots**
It's a white application frame.

**Environment:**

- Dioxus version:  0.63
- - Rust version:    1.84.1-1.85
- OS info:         Ubuntu
- - App platform:    Desktop

**Questionnaire**
I don't have time to fix this right now, but maybe later.

**Comment by ealmloff** [comment]
I can't reproduce the difference between the move and no-move closure, but I do see panic in the webview console when clicking the button inside of the form. I think there is a bug in our system in the interpreter that should prevent the default reload action for form submission events [here](https://github.com/DioxusLabs/dioxus/blob/main/packages/interpreter/src/ts/native.ts#L337-L340). The button causes the webview page to reload which breaks the desktop renderer. Adding prevent default to the form fixes the issue:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        p { "I am text" }
        form {
            onclick: move |evt| evt.prevent_default(),
            button { onclick: move |_| {}, "Sync" }
        }
    }
}
```

caseykneale:
[...]

PR #3804, opened 2025-02-26
Prevent default if an onclick event is triggered on a button under a form

ealmloff:
Instead of just preventing the default action for submit events, this PR prevents the default action for all onclick events on desktop under a form

Fixes https://github.com/DioxusLabs/dioxus/issues/3791


jkelleyrtp:
Not 100% sure if this is technically "right" though it might not matter.

Buttons in forms are given the type "submit" by default whereas they usually receive "button".

We *shouldn't* prevent default on buttons in forms with a type="button". However, a button with type="button" has no default behavior so there's nothing to prevent.

This implementation should be fine in that case, but it's something to keep in mind.

jkelleyrtp (approved):

PR #3812, opened 2025-02-28
Make GenericRouterContext pub

3lpsy:
`RouterConfig` accepts a `on_default` callback. To be able to create types for this callback, you need access to `GenericRouterContext`. This PR just makes that type pub. 

Here is the customer use case:
```rust
/// requires type 
pub fn on_update_route<R>(state: GenericRouterContext<R>) -> Option<NavigationTarget<R>>
where
    R: Routable,
{
    None
}

rsx! {
  Router::<router::Route> {
    config: || {
      RouterConfig::default().on_update(on_update_route)
    }
   }
}
```

### DIFF
--- a/packages/router/src/contexts/router.rs
+++ b/packages/router/src/contexts/router.rs
@@ -309,6 +309,7 @@ impl RouterContext {
     }
 }
 
+/// This context is set to the RouterConfig on_update method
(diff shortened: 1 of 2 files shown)


jkelleyrtp (approved):
Thanks!

PR #3814, opened 2025-03-01
debugger support for dx

jkelleyrtp:
Powered by the same tooling from http://github.com/jkelleyrtp/cargo-debugger

Lets you open the current app in vscode's debugger by pressing `d`.

todo: 
- [x] web (requires vscode extension)
- [x] desktop
- [x] server (shift + d)
- [x] ios
- [x] android

We should allow configuration of the debugger at some point since people seem to use gdb, dap, etc, but for now this works well with our vscode-oriented workflow.

https://github.com/user-attachments/assets/5481f388-fe7e-49d5-a15b-d1982b944fef


mcmah309:
Does this work inside a dev container? - https://github.com/jkelleyrtp/cargo-debugger/issues/2

jkelleyrtp:
> Does this work inside a dev container? - [jkelleyrtp/cargo-debugger#2](https://github.com/jkelleyrtp/cargo-debugger/issues/2) . Might be worth testing with https://github.com/DioxusLabs/dioxus/tree/main/.devcontainer if not already done

Unfortunately no, since we call `code --open-url "link"` to get code-lldb to open the debugger.

To add web support, I had to add an extra handler to our vscode extension, so in theory, the two *could* work together, but I wouldn't even know where to start to get it working.

mcmah309:
> Unfortunately no, since we call `code --open-url "link"` to get code-lldb to open the debugger.

That's unfortunate, we do all our coding inside dev containers.

N3xed:
I think this needs a new extension release. The current version available on the VSCode marketplace (`0.6.0`) doesn't support this yet.

jgardona:
It doesnt works for me. Breakepoints are going gray when I attach debug on server

image

PR #3820, opened 2025-03-03
Changes to support axum 0.8

### LINKED ISSUE #3479: Upgrade to axum 0.8.x breaks DioxusRouterExt
**Problem**

After upgrading to axum 0.8.x, the serve_dioxus_application method is no longer found on Router, breaking Dioxus fullstack applications.

**Steps To Reproduce**

- Update axum to 0.8 in Cargo.toml
- Run `cargo build` on a project using `serve_dioxus_application`
- Observe compiler error:
```
error[E0599]: no method named `serve_dioxus_application` found for type `Router` in the current scope
   |
   |     .serve_dioxus_application(ServeConfigBuilder::default(), App);
   |     -^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `Router<_>`
```

**Expected behavior**

`serve_dioxus_application` should remain available after upgrading axum, allowing seamless server setup of Dioxus applications.

**Environment:**
 - Dioxus version: 0.6.1
 - Rust version: 1.83.0
 - OS info: MacOS
 - App platform: web

[Announcing axum 0.8.0](https://tokio.rs/blog/2025-01-01-announcing-axum-0-8-0)

CristianCapsuna:
Hi guys. Very excited to make my first, small, contribution.

Before this PR can be closed, the `server_fn` and `server_fn_macro` dependencies need to be pointed to their released crates once those are available.

I was not able to do the make tests but I don't think it's because of this. See below. Any guidance on how I could make this work?
Screenshot from 2025-03-03 16-07-33

I plugged this is my wannabe website project and it worked fine.

Closes #3479


CristianCapsuna:
@ealmloff would you have any guidance on what the problem is with the playwright test?

ealmloff:
> @ealmloff would you have any guidance on what the problem is with the playwright test?

My guess is some of the CLI changes are causing it to not proxy the fullstack server. Try going into one of the playwright tests and running the [server command](https://github.com/DioxusLabs/dioxus/blob/main/packages/playwright-tests/playwright.config.js#L99) manually then opening the site

ealmloff:
I pulled in the latest git version of server_fn which includes https://github.com/leptos-rs/leptos/pull/3725. That makes our server function integration much simpler. The server fn crate has a 0.8 beta out, but it doesn't include that change yet. It sounds like a full 0.8 release is coming soon

jkelleyrtp (approved):
Would like to follow this up with the other PR that moves everything over to axum 0.8, but this is good for now!

I moved us to the 0.8 pre-release of server_fn though we might just end up forking it.

PR #3822, opened 2025-03-03
Provide `HotkeyState` on Global Shortcut Events 

CryZe:
This state is now provided as an argument to the event handler, such that the user can decide whether to react to the global shortcut being released or not.

### DIFF
--- a/examples/overlay.rs
+++ b/examples/overlay.rs
@@ -6,7 +6,7 @@
 //! We also add a global shortcut to toggle the overlay on and off, so you could build a raycast-type app with this.
 
 use dioxus::desktop::{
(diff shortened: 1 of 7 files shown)


jkelleyrtp:
I feel like the handlers should still be called and the selection is done in user code. Someone might want to handle released events?

CryZe:
That would require public API changes, but I can definitely implement it that way.

CryZe:
This should be ready to be reviewed again. It includes the breaking changes now.

jkelleyrtp (approved):
Thank you! Sorry this sat for so long.

PR #3829, opened 2025-03-05
fix(fullstack): Resolve suspense before the head chunk when streaming is disabled

hackartists:
Move suspense handling to the correct position in the SSR renderer to ensure proper rendering when streaming is disabled.

## Issues
- `Meta` tag does not be rendered after suspense.
- It can improve `SEO` by suspense `Meta`

## Before
- Server returned HTML with `Meta`s before suspense.

## Contribution
- Server returns HTML with `Meta`s after resolving suspense


ealmloff (approved):
Thanks, this looks right. We should wait for suspense to resolve when streaming is disabled before rendering the head chunk

jkelleyrtp (approved):

Issue #3772, opened 2025-02-20
Dioxus LLM Model


mcmah309:
## Feature Request

As Dioxus is new and rapidly evolving, it would be nice if there was an up to date llama like model for complex code and a [tabby](https://github.com/TabbyML/tabby) compatible model for code completion. This likely related to work presented at the [GOSIM 2024](https://www.youtube.com/watch?v=xGKs1UYRPOM) talk.

This could also take a lot of pressure off the Dioxus team answering questions and provide quicker onboarding.

## Implement Suggestion

Fine tune models on: 
- Open source projects using dioxus
- Dioxus documentation
- Dioxus test code
- Dioxus issues
- Dioxus discussions
- Dioxus discord

For things like issues, discussions, and discord, the date and dioxus version at that time should be provided so the model has that context. When prompting the model one should provide the current version and date.

mcmah309:
Closing as https://github.com/DioxusLabs/docsite/pull/492 suites my needs. Though, it would be nice if there was a way to expose discord discussions so commercial LLMs can be trained on them.

PR #3830, opened 2025-03-05
Typo Fix on Outlet Documentation

wheregmis:


wheregmis:
Eventhough its ready to merge, just making it draft incase we have more discussion.

mcmah309:
Thanks for creating this small fix, I'm all good with it! I would be interested in seeing a useful example of using `use_outlet_context` in https://github.com/DioxusLabs/dioxus/tree/main/examples if you are also interested in sharing.

wheregmis:
Thank you @mcmah309 Mostly for normal dioxus user, i am not sure how i can use use_outlet_context hook. This is edge case scenerio as i am trying to build page transitions on dioxus-motion library. https://github.com/wheregmis/animated_router here is an example for it. In near future i will try to come up with something that can help #3772

Issue #3783, opened 2025-02-21
login_form example doesn't work


webhype:
**Problem**

The example `login_form` fails to compile.

```
$ dx serve --example login_form --platform web --features="http"
08:45:47 [cargo] error[E0277]: the trait bound `FormValue: serde::ser::Serialize` is not satisfied
   --> examples/login_form.rs:19:19
    |
19  |               .form(&[
    |  ______________----_^
    | |              |
    | |              required by a bound introduced by this call
20  | |                 ("username", &evt.values()["username"]),
21  | |                 ("password", &evt.values()["password"]),
22  | |             ])
    | |_____________^ the trait `serde::ser::Serialize` is not implemented for `FormValue`, which is required by `[(&str, &FormValue); 2]: serde::ser::Serialize`
    |
    = note: for local types consider adding `#[derive(serde::Serialize)]` to your `FormValue` type
``` 

**Expected behavior**

That it compile and execute as expected. Lack of working and complete examples and tutorials is really Dioxus'  Achilles heel. Hard to estimate the number of devs that want to give this a try but quietly fail and give up. I know it's a huge complex project but I can't overemphasize how frustrating such things are. This is your calling card, your storefront, the place that's supposed to be welcoming to new adopters. So it can't ever be an afterthought, it has to be at the forefront. End of unsolicited advice... 🤣

**Environment:**

- Dioxus version:  0.6.3
- Rust version:    rustc 1.83.0
- OS info:         MacOS 15.2
- App platform:    desktop
 
**Questionnaire**

I'm interested in fixing this myself but don't know where to start.

webhype:
```rust
		let mut form_data = HashMap::new();
		form_data.insert("username", evt.values()["username"]);
		form_data.insert("password", evt.values()["password"]);
		let resp = reqwest::Client::new()
			.post("http://localhost:8080/login")
			.form(&form_data)
			.send()
			.await;
```
This seems to compile; haven't tested it but at least compiles.

created-by-varun:
Just adding as_value() to username and password should get it running.
`
("username", evt.values()["username"].as_value()),
("password", evt.values()["password"].as_value())
`

I've raised a PR for the same

PR #3835, opened 2025-03-06
fix: Use as_value() to properly serialize form values in login example

created-by-varun:
This fix is for issue: https://github.com/DioxusLabs/dioxus/issues/3783

The login form example was trying to directly use FormValue in reqwest's form method, which doesn't implement Serialize. This change extracts the actual string values using as_value() before sending them to the server.


created-by-varun:
Hi team,
Can we get some eyes on this?
It's a very small PR - fix for the login example project

PR #3837, opened 2025-03-06
Rename wgpu child window example for discoverability

carlosdp:
This might seem trivial, but it's referred to in blog posts and everywhere as "child window", and like 5 times now I've forgotten where this example was stored ("was it in the dixous repo, or the blitz repo?") and Googling for "dixous child window" doesn't help.

I think this simple name change might help discoverability, until this stuff is stable enough to add to the official docs!


carlosdp:
Ah yep, also grepped to make sure I didn't miss anything else like that. Should work this time!

jkelleyrtp (approved):

PR #3840, opened 2025-03-07
adds feature `axum_core` that compiles for wasm target exposing server functions and ssr

conectado:
The current `axum` feature is incompatible with wasm32-unknown-unknown target due to depending on `tower-http` for serving static files.

With this patch I add the `axum_core` feature which disables `tower-http` functions and enables targeting wasm32-unknown-unknown with server functions enabled and ssr.

Note that for this to work, `server_fn` depends on `inventory` and as stated in this docs depending on the context you might need to expose and call [`__wasm_call_ctors`](https://docs.rs/inventory/latest/inventory/#webassembly-and-constructors) (I specifically needed that for Cloudflare workers)

Right now the feature isn't exposed by the main crate, I'm using it including the fullstack crate explicitly, I could add it to the main crate too!


packages/fullstack/Cargo.toml
```diff
@@ -87,7 +87,34 @@ desktop = ["dep:dioxus-desktop", "server_fn/reqwest", "dioxus_server_macro/reqwe
...
+]
+axum = [
+    "dep:axum",
+    "dep:tower-http",
+    "server",
+    "server_fn/axum",
+    "axum_server_fn",
+    "dioxus_server_macro/axum",
+    "default-tls",
+    "axum/ws",
```
ealmloff, line 115: I don't think we use the ws or macros features any more. We could remove them entirely instead of moving them to the full axum feature

packages/fullstack/Cargo.toml
```diff
@@ -87,7 +87,34 @@ desktop = ["dep:dioxus-desktop", "server_fn/reqwest", "dioxus_server_macro/reqwe
...
+    "dep:async-trait",
+    "server_fn/axum-no-default",
+    "dioxus_server_macro/axum",
+    "default-tls",
+    "dep:parking_lot",
+    "dioxus-interpreter-js",
+]
+axum = [
+    "dep:axum",
+    "dep:tower-http",
```
ealmloff, line 109: Messing with dependencies, it looks like `tower-http` and `server_fn/axum` are the only dependencies that are causing compilation errors on wasm with the axum feature enabled. We only use `tower-http` in `serve_static_assets`. Instead of two duplicate traits that contain mostly duplicate methods, could we pull out that and `serve_dioxus_application` into a separate super trait behind an axum feature flag?

```rust
// WASM compatible integration
pub trait DioxusRouterServerFnExt<S> {
    fn register_server_functions(self) -> Self
    where
        Self: Sized,
    {
        self.register_server_functions_with_context(Default::default())
    }

    fn register_server_functions_with_context(self, context_providers: ContextProviders) -> Self;
}

// WASM incompatible integration
#[cfg(feature = "axum")]
pub trait DioxusRouterExt<S>: DioxusRouterServerFnExt<S> {
    fn serve_static_assets(self) -> Self
    where
        Self: Sized;

    fn serve_dioxus_application<Cfg, Error>(self, cfg: Cfg, app: fn() -> Element) -> Self
    where
        Cfg: TryInto<ServeConfig, Error = Error>,
        Error: std::error::Error,
        Self: Sized;
}
```

If we still expose [`render_handler`](https://docs.rs/dioxus-fullstack/latest/dioxus_fullstack/server/fn.render_handler.html) when compiling to wasm, you can pull in that and serve your static files to render SSR in serverless

conectado:
@ealmloff Thanks for the review! updated the PR for the feedback, lmk if there's something more to adjust

packages/fullstack/Cargo.toml
ealmloff: Yes, we have already started merging breaking PRs for the next release and this isn't breaking for most users so I don't think it is an issue.

ealmloff (commented):
Looks much better! The name `axum_wasm` might be slightly confusing if you need to import it in non-wasm targets. Maybe just `axum_core` for the wasm compatible axum integration and `axum` if you have access to both axum and a filesystem for native targets targets?

I expect tests will fail here again because the default features of axum were removed from the workspace. Adding `default-features = true` to the dev dependencies that rely on them should fix the issue

conectado:
> Looks much better! The name `axum_wasm` might be slightly confusing if you need to import it in non-wasm targets. Maybe just `axum_core` for the wasm compatible axum integration and `axum` if you have access to both axum and a filesystem for native targets targets?
> 
> I expect tests will fail here again because the default features of axum were removed from the workspace. Adding `default-features = true` to the dev dependencies that rely on them should fix the issue

ah nice! I struggled to find a better name. feedback applied and also cleaned up the dependencies of each feature a bit more

ealmloff (approved):
Looks great! Thanks for working on this!

conectado:
Hopefully CI should be passing now!

conectado:
my bad! now it should work

conectado:
rebased and conflicts resolved

conectado:
I think CI failure right now is unrelated

conectado:
rebased and conflicts solved! :)

conectado:
Just checking in on this PR—let me know if there's anything else I can do to help get it merged. No rush, just asking!

ealmloff:
> Just checking in on this PR—let me know if there's anything else I can do to help get it merged. No rush, just asking!

Looks great to me, just waiting for a review from Jon before this can be merged

jkelleyrtp (approved):
Thank you!

PR #3846, opened 2025-03-08
feat: Add logger to logger example

mcmah309:
initializing the logger was missing from the logging example so nothing was ever logged.

### DIFF
--- a/examples/logging.rs
+++ b/examples/logging.rs
@@ -11,10 +11,12 @@
 //!
 //! To use the dioxus logger in your app, simply call any of the tracing functions (info!(), warn!(), error!())
 
-use dioxus::logger::tracing::{debug, error, info, warn};
+#[allow(unused_imports)]
+use dioxus::logger::tracing::{debug, error, info, warn, Level};
 use dioxus::prelude::*;
 
 fn main() {
+    dioxus::logger::init(Level::INFO).expect("Failed to initialize logger");
     dioxus::launch(app);
 }
 
@@ -31,7 +33,10 @@ fn app() -> Element {
                 "Error!"
             }
             button {
-                onclick: move |_| debug!("Here's a debug"),
+                onclick: move |_| {
+                    debug!("Here's a debug");
+                    warn!("The log level is set to info so there should not be a debug message")
+                },
                 "Debug!"
             }
             button {


mcmah309:
Ah yes, you are correct. I had these small changes from a few weeks ago when I worked on https://github.com/DioxusLabs/dioxus/pull/3775 and misremembered the reason I made them. I was starting work on somthing else in dioxus so I decided to commit them in this small PR. Still, might be a good idea to include explict logging setup in the logging example. If not, feel free to close.

Side: In addition to https://github.com/DioxusLabs/dioxus/pull/3775 , I thought it would be a good idea to add programmatic filtering to the api e.g. [Targets](https://docs.rs/tracing-subscriber/0.3.19/tracing_subscriber/filter/targets/struct.Targets.html) . Which would allow web support for filtering as well. And adding https://crates.io/crates/tracing-log as default might be worthwhile as well (crates like [rfd](https://crates.io/crates/rfd) use the `log` crate so nothing gets logged without it) . But I was waiting on feedback on that PR before doing anything.

ousado:
At the very least there should be documentation what happens where and how to customize/override the defaults.

examples/logging.rs
```diff
@@ -11,10 +11,12 @@
 //!
 //! To use the dioxus logger in your app, simply call any of the tracing functions (info!(), warn!(), error!())
 
-use dioxus::logger::tracing::{debug, error, info, warn};
+#[allow(unused_imports)]
```
jkelleyrtp, line 14: don't need this

examples/logging.rs
```diff
@@ -11,10 +11,12 @@
...
+use dioxus::logger::tracing::{debug, error, info, warn, Level};
 use dioxus::prelude::*;
 
 fn main() {
+    dioxus::logger::init(Level::INFO).expect("Failed to initialize logger");
```
jkelleyrtp, line 19: can you put a comment here say that `dioxus::logger::init` is *optional* and called automatically by `dioxus::launch`, but you *can* call it yourself manually in the cases you:
- want to customize its behavior 
- aren't using `dioxus::launch` (ie custom fullstack setups) but want the integration.

jkelleyrtp (changes requested):

jkelleyrtp (approved):
Thanks!

jkelleyrtp (approved):

PR #3851, opened 2025-03-09
fix(server): Provide GoogleBot compatibility.

hackartists:
- Actually, GoogleBot will crawl as below command.
```
curl -A GoogleBot 'http://localhost:8080'
```
- However, dioxus `render_handler` rejects GoogleBot crawl.
  - It DOES mean dioxus webapp can't be indexed by GoogleBot properly.
- I removed accept checking instead of optional parameter.
  - Because all dioxus app based fullstack will be desired to be crawl and proper SEO.

- Before this PR, it only allows requests including `accept`, however it allows all requests

- I would say that its part doesn't need to provide backward compatibility
because SEO capability is a crucial issue.
  - It means it can make starters confusion why they can't check their page on
  google inspection.
  - So I propose that it SHOULD provide SEO feature by default.

### DIFF
--- a/packages/fullstack/src/server/mod.rs
+++ b/packages/fullstack/src/server/mod.rs
@@ -374,16 +374,7 @@ impl RenderHandleState {
 pub async fn render_handler(
     State(state): State<RenderHandleState>,
     request: Request<Body>,
-) -> impl IntoResponse {
-    // Only respond to requests for HTML
-    if let Some(mime) = request.headers().get("Accept") {
-        let mime = mime.to_str().map(|mime| mime.to_ascii_lowercase());
-        match mime {
-            Ok(accepts) if accepts.contains("text/html") => {}
-            _ => return Err(StatusCode::NOT_ACCEPTABLE),
-        }
-    }
-
+) -> Result<http::Response<axum::body::Body>, StatusCode> {
     let cfg = &state.config;
     let ssr_state = state.ssr_state();
     let build_virtual_dom = {


jkelleyrtp (approved):
I have been meaning to make this change for a while - it's annoying when trying to test endpoints locally that they deny requests without any content type. I've suspected that it hurts SEO.

LGTM but pinging @ealmloff since he's more familiar with this and *why* would we want this in the first place.

I *think* we just wanted better filtering for MIME types, but this code has always been suspect to me.

packages/fullstack/src/server/mod.rs
```diff
@@ -374,16 +374,7 @@ impl RenderHandleState {
...
-    // Only respond to requests for HTML
-    if let Some(mime) = request.headers().get("Accept") {
-        let mime = mime.to_str().map(|mime| mime.to_ascii_lowercase());
-        match mime {
-            Ok(accepts) if accepts.contains("text/html") => {}
-            _ => return Err(StatusCode::NOT_ACCEPTABLE),
-        }
-    }
-
+) -> Result<http::Response<axum::body::Body>, StatusCode> {
```
ealmloff, line 377: Could we keep this as `impl IntoResponse`? This is a public API and I would like to be able to change the error type without a breaking change

ealmloff (approved):
The code that was removed filters out requests if they have a content type, but it is not `text/html`. If there is no content type, then it should accept requests either way. However, that doesn't handle the case where content type is `*/*` which might be what some scrapers use. I don't think we need to handle the accept content type at all as long as we don't have multiple different handlers on one route (eg. one for json and one for html)

packages/fullstack/src/server/mod.rs
ealmloff: I moved the type information to `Ok` type we return which seems to fix the issue

Issue #2739, opened 2024-07-29
Allow returning HTTP STATUS in fullstack


chungwong:
**Problem**

[Discord reference
](https://discord.com/channels/899851952891002890/1266229182145368194)
**Steps To Reproduce**
Given a route
```rs
#[derive(Clone, Routable, PartialEq, Serialize, Deserialize)]
enum Route {
  #[route("/product/:url_slug")]
  Product { url_slug: String },
}
```
and a component `Product` where it takes a url and fetch a product by THAT url. It is normal for that to failed due to unrecognised urls. In those case, it would be HTTP 404 NOT FOUND. 
```rs
#[component]
fn Product(url_slug: String) -> Element {
    let product = use_resource(|| async {
        get_product().await
    });

    rsx! { }
}
```
The advised solution is to try `server_context`, however, that doesn't seem to work.
```rs
#[server]
async fn get_product() -> Result<String, ServerFnError> {
    if let Ok(mut resp) = server_context().response_parts_mut() {
        resp.status = StatusCode::NOT_FOUND;
        Err(ServerFnError::Response("404 NOT FOUND".into()))
    } else {
        Ok(
            "A dummy product".to_string()
        )
    }
}
```

**Expected behavior**
The first request in the below screenshot should return 404(or whatever that was returned from the function. For the second AJAX call, that should behave the same?

**Screenshots**

![Screenshot from 2024-07-29 10-28-05](https://github.com/user-attachments/assets/1d2e0773-5c3d-4a1d-bad9-7c674d951a0b)

**Environment:**
 - Dioxus version: 0.5
 - Rust version:   1.80
 - OS info:   Debian
 - App platform:    fullstack

ealmloff:
The fact that setting the status happens after some async work makes this complicated. Fullstack now supports streaming which means there are a few different stages of the response:

## Before any content is sent

Fullstack needs to choose chat response is sent to the client. Once the status is sent, it cannot be changed (mostly - we can send an incomplete streaming chunk to fail even after we send a 200 response)

## Before the body is started

Fullstack needs to insert any `<head>` content that should be available for SEO. `<head>` elements can be streamed in after this, but they may not be visible for SEO

## After the body is started

Fullstack can only modify the head and body through javascript. Any text content should still be visible for SEO, but it may not be in the right order

### How fullstack should handle this

I was talking with Jon about this, and we came up with this order:
1) Initially fullstack waits and accumulates contents without sending any response
[...]

PR #3860, opened 2025-03-11
Return a 404 when paring the route fails

ealmloff:
This PR changes the router to throw a url parsing error into the context instead of panicing when it fails to parse the current url. On the server, it changes fullstack to return a 404 status if the route is not found or a 500 status if there is another error in the root context

Partially addresses https://github.com/DioxusLabs/dioxus/issues/2739 with better default status codes

### DIFF
--- a/packages/fullstack/src/render.rs
+++ b/packages/fullstack/src/render.rs
@@ -6,7 +6,7 @@ use dioxus_cli_config::base_path;
 use dioxus_interpreter_js::INITIALIZE_STREAMING_JS;
 use dioxus_isrg::{CachedRender, IncrementalRendererError, RenderFreshness};
 use dioxus_lib::document::Document;
-use dioxus_router::prelude::RoutingFailed;
+use dioxus_router::prelude::ParseRouteError;
 use dioxus_ssr::Renderer;
 use futures_channel::mpsc::Sender;
 use futures_util::{Stream, StreamExt};
@@ -53,7 +53,7 @@ where
 
 pub enum SSRError {
     Incremental(IncrementalRendererError),
-    Routing(RoutingFailed),
+    Routing(ParseRouteError),
 }
 
 struct SsrRendererPool {
--- a/packages/fullstack/src/render.rs
+++ b/packages/fullstack/src/render.rs
@@ -6,6 +6,7 @@ use dioxus_cli_config::base_path;
 use dioxus_interpreter_js::INITIALIZE_STREAMING_JS;
 use dioxus_isrg::{CachedRender, IncrementalRendererError, RenderFreshness};
 use dioxus_lib::document::Document;
+use dioxus_router::prelude::RoutingFailed;
 use dioxus_ssr::Renderer;
 use futures_channel::mpsc::Sender;
 use futures_util::{Stream, StreamExt};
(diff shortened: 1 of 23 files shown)


packages/fullstack/src/render.rs
```diff
@@ -50,6 +51,11 @@ where
     }
 }
 
+pub enum SSRError {
+    Incremental(IncrementalRendererError),
+    Routing(ParseRouteError),
+}
+
```
jkelleyrtp, line 58: I believe these need docs if they're part of our public API?

jkelleyrtp (approved):
Awesome!! I wonder if (for spas) we should render just a blank page for 404 routes or perhaps lint that the router doesn't have a 404 route enabled? But definitely the default behavior should be an error.

Just the clippy test (windows might just be failing now) and some docs and it's good to go!

Also - this doesn't completely address throwing a `http::ErrorCode::MY_CODE` does it?

packages/fullstack/src/render.rs
ealmloff: It should be private because of the export. I marked it as pub(crate) to make it more clear

ealmloff:
> Also - this doesn't completely address throwing a `http::ErrorCode::MY_CODE` does it?

No, that will be in a follow up PR. It requires more integration with suspense to only send the first chunk down after the suspense boundary the router is in has been resolved

ealmloff:
> Awesome!! I wonder if (for spas) we should render just a blank page for 404 routes or perhaps lint that the router doesn't have a 404 route enabled? But definitely the default behavior should be an error.

Maybe in the CLI? Figuring out if we are running in a spa might be difficult. You could nest dioxus-web inside of an axum router that handles the 404

packages/router/src/hooks/use_route.rs
```diff
@@ -42,7 +42,10 @@ use crate::utils::use_router_internal::use_router_internal;
 /// ```
 #[doc(alias = "use_url")]
 #[must_use]
-pub fn use_route<R: Routable + Clone>() -> R {
+pub fn use_route<R: Routable + Clone>() -> R
+where
+    <R as std::str::FromStr>::Err: std::fmt::Display,
+{
     match use_router_internal() {
```
jkelleyrtp, line 49: the `<R as std::str::FromStr>...` stuff is infectious to all these APIs. Is there a way we can implement this without having to go around changing all our APIs?

The built-in Rust error type already has the display bound, why can't that just be on the Routable bound?

https://doc.rust-lang.org/std/error/trait.Error.html

ealmloff: > the `<R as std::str::FromStr>...` stuff is infectious to all these APIs. Is there a way we can implement this without having to go around changing all our APIs?

I tried this initially with `<Self as FromStr>::Err: Display`, but unfortunately that is also infectious. It looks like the new associated generic bound syntax in 1.79 does let us omit the boilerplate: `FromStr<Err: Display>`

> The built-in Rust error type already has the display bound, why can't that just be on the Routable bound implicitly?

I generally try to avoid the Error trait because it is not implemented for some error types including our own error type [CapturedError](https://docs.rs/dioxus/latest/dioxus/struct.CapturedError.html). We don't need the full version of the error type, so I would rather not require it

> We also don't even need to display the error - I imagine developers might want to emit specific error strings rather than let whatever implicit error they're cascading potentially leak sensitive information.

This now returns a less specific error message with the 404. The full error message to help debug the parsing logic is in a trace log.

The router structure is always leaked in SPA routers. We could try to obfuscate it by compiling a less useful error message in release builds, but applications should generally not assume their route structure is hidden

PR #3880, opened 2025-03-15
adhoc: Fix Clippy lints in const-serialize etc.

gbutler69:
Fix the clippy lints that were preventing building:

   * const-serialize/src/lib.rs
   * core/src/nodes.rs
   * fullstack/src/serve_config.rs


gbutler69:
OK, the version of the compiler being used in CI doesn't like the ```std::ptr::fn_addr_eq(a as fn(), b as fn())```. It seems that it was an unstable library feature for this version of the compiler. It has now been stabilized in the latest 1.85 compiler it seems and is needed to avoid the Clippy error on comparing pointers.

gbutler69:
I also added fixes for all the doc tests that were failing when you run ```cargo make test``` on the repo

PR #3895, opened 2025-03-20
Fix dx serve proxying of websocket connections established from the browser

### LINKED ISSUE #3894: Can't establish websocket connection between browser and server when using dx serve
**Problem**

When using `dx serve` with a custom Axum websocket endpoint (using `WebSocketUpgrade`), websocket connections between the browser and the server fail, even though the same code works when using `dx bundle`.

Also see #2155 which was previously closed, but this is still an issue for browser-to-server connections.

**Steps To Reproduce**

<details>
<summary><strong>Cargo.toml</strong></summary>

```toml
[dependencies]
dioxus = { version = "0.6.3", features = ["fullstack"] }
dioxus-cli-config = "0.6.3"
futures = "0.3.31"
tokio = { version = "1.43.0", optional = true }
axum = { version = "0.7.9", optional = true }
gloo = { version = "0.11.0", optional = true, features = ["net"] }

[features]
default = []
web = ["dioxus/web", "dep:gloo"]
server = ["dioxus/server", "dep:tokio", "dep:axum"]
```
</details>

<details>
<summary><strong>main.rs</strong></summary>

```rs
use dioxus::prelude::*;

fn main() {
    #[cfg(not(feature = "server"))]
    dioxus::launch(App);
    #[cfg(feature = "server")]
    server::server_main();
}

#[cfg(feature = "server")]
mod server {
    use axum::extract::WebSocketUpgrade;
    use axum::response::IntoResponse;
    use dioxus::logger::tracing::{info, warn};
    use dioxus::prelude::*;
    use futures::StreamExt;

    #[tokio::main]
    pub async fn server_main() {
        dioxus::logger::initialize_default();
        let socket_addr = dioxus_cli_config::fullstack_address_or_localhost();
        let router = axum::Router::new()
            .serve_dioxus_application(ServeConfigBuilder::new(), super::App)
            .route("/events", axum::routing::any(websocket_handler))
            .into_make_service();

        let listener = tokio::net::TcpListener::bind(socket_addr).await.unwrap();
        axum::serve(listener, router).await.unwrap();
    }

    async fn websocket_handler(ws: WebSocketUpgrade) -> impl IntoResponse {
        ws.on_upgrade(|mut ws| async move {
            while let Some(msg) = ws.next().await {
                match msg {
                    Ok(msg) => info!("Got message: {msg:?}"),
                    Err(e) => warn!("Error receiving message: {e}"),
                }
            }
        })
    }
}

#[component]
fn App() -> Element {
    #[cfg(feature = "web")]
    {
        _ = use_future(|| async {
            use dioxus::logger::tracing::error;
            use futures::SinkExt;
            use gloo::net::websocket::futures::WebSocket;

            let mut ws = match WebSocket::open("/events") {
                Ok(ws) => ws,
                Err(e) => {
                    error!("Could not open websocket connection: {e}");
                    return;
                }
            };

            if let Err(e) = ws
                .send(gloo::net::websocket::Message::Text(
                    "this is a websocket message".into(),
                ))
                .await
            {
                error!("Could not send websocket message: {e}");
            }
        });
    }

    rsx! {
        "hi"
    }
}
```
</details>

1. Create a new Dioxus project with the above files
2. Run `dx bundle --platform web && target/dx/$PROJECT_NAME/release/web/server`
3. Navigate to http://127.0.0.1:8080/
4. Run `dx serve`
5. Navigate to http://127.0.0.1:8080/

**Expected behavior**

The websocket message is successfully received by the server in both cases.

**Actual output**

The connection fails to be established when using `dx serve`:
```
> dx bundle --platform web
[...]
> target/dx/websocket_test/release/web/server
2025-03-20T23:44:26.893572Z  INFO websocket_test::server: Got message: Text("this is a websocket message")
2025-03-20T23:44:26.893600Z  INFO websocket_test::server: Got message: Close(None)
^C
> dx serve
[...]
23:48:27 [dev] [101] /events
23:48:27 [server]  WARN Error receiving message: WebSocket protocol error: Connection reset without closing handshake
```

**Environment:**

- Dioxus version: 0.6.3, main
- Rust version: 1.85.1
- OS info:  Arch Linux
- App platform: fullstack (web + server)

**Questionnaire**

- [x]  I would like to fix and I have a solution.

arvidfm:
This PR makes the following changes:
* Check for the `Upgrade` header to detect websocket connections where the URI doesn't include the scheme
* Replace redirect with explicit proxying of websocket connections

The explicit proxying is required as browsers (at least Firefox and Chrome) don't handle redirect responses. E.g. Chrome emits an error complaining about an invalid response.

This PR is somewhat of a proof of concept in its current state. I'm happy to make any necessary changes to conform to Dioxus's code style and general guidelines, and to discuss alternative approaches.

Fixes #3894


arvidfm:
I fixed the compilation errors following the Axum upgrade. Unfortunately even though Tungstenite and Axum now use the same `Utf8Bytes` type for its messages there doesn't seem to be an easy way to reuse the allocation since Axum doesn't expose the underlying Tungstenite `Utf8Bytes` value that it wraps. You can in principle do something like `Bytes::from(axum_string).try_into().unwrap()` but I was hesitant to introduce an `unwrap()` call even if it's guaranteed to succeed in this case.

Not that I expect it'll actually have a measurable impact on performance, it just feels a bit silly.

arvidfm:
It's not entirely clear to me what the cause of the Playwright issues is, but I don't *think* it's related to these changes? From what I can tell digging through the logs it looks like the `dioxus` crate is failing to build for certain tests due to the specific combination of feature flags passed

ealmloff:
The playwright test failures are not related. Playwright is failing on main. https://github.com/DioxusLabs/dioxus/pull/4042 fixes some of the issues

Issue #3470, opened 2024-12-30
Android App crash when the mobile rotate from portrait to landscape and 


GuillaumeSavin:
**Problem**

When using the app on a true android phone produce by dioxus serve --platform mobile and move the mobile to provoke a change of screen orientation from portrait to landscape or from portrait to landscape. The app immediately crash.

**Steps To Reproduce**

-  Create a dioxus 0.6.1 project
-  execute dioxus serve --platform mobile (with android phone connected by usb and detected by adb command)
-  move the mobile to provoke a change of screen orientation of the app

**Expected behavior**

not crash, for what I see the app work fine either in landscape and portrait initially but don't support change of screen orientation (going from portrait to landscape or from landscape to portrait)

**Screenshots**

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version:    rustc 1.83.0 (90b35a623 2024-11-26)
 - OS info:        builder of app: linux mint 22, mobile: android 14 
 - App platform:    mobile (android maybe ios too but can't test)
**Questionnaire**

GuillaumeSavin:
it seems that since dioxus 0.6.3 the app don't crash anymore but stay completely white/blank.

jjvn84:
There is a fix for this issue in Tauri's repository. Applying the changes described there to AndroidManifest.xml works for Dioxus. [https://github.com/tauri-apps/tauri/commit/947eb391ca41cebdb11abd9ffaec642baffbf44a](url)

PR #3905, opened 2025-03-24
Fixed android app crashing when changing phone orientation

jjvn84:
Modified the AndroidManifest.xml file according to [https://developer.android.com/guide/topics/manifest/activity-element#config](https://developer.android.com/guide/topics/manifest/activity-element#config) so Android apps created with Dioxus handle orientation changes properly.

Fixes [https://github.com/DioxusLabs/dioxus/issues/3470](https://github.com/DioxusLabs/dioxus/issues/3470)


jkelleyrtp:
Thank you!!

PR #3908, opened 2025-03-25
Fix later-attached wry event handlers not getting called

### LINKED ISSUE #3907: Only the earliest attached wry event handler ever gets called
When attaching event handlers to windows using `use_wry_event_handler` or `create_wry_event_handler`, only the one attached the earliest ever gets called. If the earliest attached event handler is removed, the other ones start getting called.
**Environment:**

- Dioxus version:  0.6.3
- Rust version: 1.85
- OS info:  Arch
- App platform: Desktop

**Questionnaire**
I would like to fix and I have a solution: #3908

MintSoup:
Fixes #3907.

The issue is caused by a simple logic error: instead of aborting the entire loop when we find a window that doesn't match our id, we should just skip the current iteration.


jkelleyrtp (approved):
Good catch, thank you!

PR #3911, opened 2025-03-25
Fix warning when calling VirtualDom new with props

### LINKED ISSUE #3872: An error gets printed when calling VirtualDom::new_with_props
**Problem**

Calling VirtualDom::new_with_props prints out the following error:
```
ERROR It looks like you called the component <the component> like a function instead of a component. Components should be called with braces like <the component> { prop: value } instead of as a function
```

Here's a code example:

```rust
use dioxus::{
    desktop::{wry::dpi::Size, Config, WindowBuilder},
    prelude::*,
};

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        Hero {}
    }
}

#[component]
pub fn Component(sig: Signal<Vec<i32>>) -> Element {
    rsx! {
        "Component"
    }
}

#[component]
pub fn Hero() -> Element {
    let sig = use_signal(Vec::<i32>::new);
    rsx! {
        button {
            onclick: move |_| {
                dioxus::desktop::window().new_window(
                    VirtualDom::new_with_props(Component, ComponentProps { sig } ),
                    Config::default(),
                );
            }
        }
    }
}
```

Strangely enough though the app itself seems to work fine.

**Environment:**

- Dioxus version:  0.6.3
- Rust version: 1.85
- OS info:  Arch
- App platform:  desktop

**Questionnaire**

ealmloff:
Fixes #3872

### DIFF
--- a/packages/core/src/virtual_dom.rs
+++ b/packages/core/src/virtual_dom.rs
@@ -16,6 +16,7 @@ use crate::{Task, VComponent};
 use futures_util::StreamExt;
 use slab::Slab;
 use std::collections::BTreeSet;
+use std::marker::PhantomData;
 use std::{any::Any, rc::Rc};
 use tracing::instrument;
 
@@ -239,14 +240,7 @@ impl VirtualDom {
     ///
     /// Note: the VirtualDom is not progressed, you must either "run_with_deadline" or use "rebuild" to progress it.
     pub fn new(app: fn() -> Element) -> Self {
-        Self::new_with_props(
-            move || {
-                use warnings::Warning;
-                // The root props don't come from a vcomponent so we need to manually rerun them sometimes
-                crate::properties::component_called_as_function::allow(app)
-            },
-            (),
-        )
+        Self::new_with_props(app, ())
     }
 
     /// Create a new VirtualDom with the given properties for the root component.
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I get many many warnings on random dioxus examples - I think our warning system is rather underbaked.

I think we should remove it (the warnings integration) for the time being. IMO when we add it back it maybe shouldn't be using stdout and instead log over the devtools or something.

This PR is a good example of a long list of workarounds we're going to need to add and we've already added quite a few.

DogeDark:
I mostly agree. I can see a warning system being useful in certain situations, but I often find myself chasing down a warning when the implementation works fine.

Most of the warnings I get are signal-based (write during run, write in component body). Maybe a different implementation that detects infinite write-read would be better instead of assuming all signal writes during these states are an issue.

DogeDark:
Yeah, 2 and 3 are the most common. I don't think I've ever seen 4.

> We now also just don't rerun if you write in a scope the signal is subscribed to.

Is that behavior in 0.6.3?

ealmloff:
> Is that behavior in 0.6.3?

Yes, I think it was added before the 0.6 release

Issue #3278, opened 2024-12-03
Web app resources are not cached


ochrons:
**Problem**

When Dioxus fullstack serves resources like the app.wasm and web.js files, these are not properly cached by the browser or any intermediate CDNs like CloudFlare.

**Expected behavior**

All immutable resources should be served with proper cache control headers and with unique hashed names to enable most efficient caching.

**How to improve**

Here's a network snapshot of a second load of a Dioxus web app.
![image](https://github.com/user-attachments/assets/30737ccf-c23c-4a6a-9efd-43bbe62e8a98)

Even though none of the resources are actually downloaded (beyond the initial index.html), it still takes over 500ms for the page to properly load. This is due to browser issuing requests for all resources with `if-modified-since: Tue, 03 Dec 2024 17:49:30 GMT` header, to which the server promptly responses `304 Not Modified`. But due to network latency, this takes 136ms.

Having an intermediate cache like CloudFlare would not help, since it would also need to check with the origin server if the resource has changed.

Solution involves two things:

1. Encode a content hash to resource file names to make them unique (instead of `web_bg.wasm` use `web_bg_12398718923.wasm`)
2. Provide proper cache control headers indicating immutability and a very long max age

Third thing is to bundle up all the JS files into a single JS file as suggested in #3277 

Example header
```
Cache-Control: public,max-age=31536000,immutable
```

This same strategy should also be extended to all public application assets coming through Manganis. But not to any (user) private assets, but these would probably not be managed with Manganis in the first place.

See also https://engineering.fb.com/2017/01/26/web/this-browser-tweak-saved-60-of-requests-to-facebook/

jkelleyrtp:
In [3531](https://github.com/DioxusLabs/dioxus/pull/3531) we tried to add this but then realized that assets that weren't handled via manganis (ie had permanent names) need to be handled differently than hashed assets.

This would require building a manifest/list of assets that the devserver reads when booting up to populate the asset cache.

ochrons:
Instead of a manifest file, the same could be achieved by a clever naming scheme for cacheable assets. This would also allow developers to control caching of non-manganis assets by applying the same naming scheme. Something like

```
image-71fe028312371ad7(cache-immutable).jpg
styles(cache-3600).css
```

PR #3928, opened 2025-03-29
Cache hashed assets forever in fullstack

ealmloff:
This PR adds a serve config json file that the CLI uses to tell the fullstack server what assets were hashed and should be cached forever. The JSON format was chosen so we can add fields to the file in the future without a breaking release.

A test was added to the fullstack playwright tests to assert the new cache behavior

Closes https://github.com/DioxusLabs/dioxus/issues/3278


jkelleyrtp (changes requested):
I'm not a huge fan of another additional runtime file (serve_config.json) that the server needs to decide if an asset is hashed.

I was brainstorming some alternatives and perhaps we could steal a trick from symbol mangling and add a prefix to all hashes?

```
-> /assets/image.png
-> /assets/image-hash123123123.png
-> /assets/image-H123123123.png // or something shorter?
```

that way when we split the filename, we check if hash starts with a marker symbol. 

Is there a reason we might want the json file other than for assets? I do think we'd like to include lots of metadata for the binaries to load up at runtime, but I'm not ready to commit to any particular approach until we've thought about it's design yet.

ealmloff:
I'm not opinionated on the way we communicate the hashed assets. If I remember correctly, it is currently implemented as a manifest because that was your original implementation suggestion in the linked issue. I can't think of any information we would need only at runtime from the CLI for the server. I can switch to looking for a `dxh` prefix before the hash in the filename instead

> Also, the ruby on rails guide caches hashed assets for just 1 year instead of "forever". Might be worth adding a max-age=31104000. However, it also generates an entry list similar to the one you implemented, so if we decide to ship the config file approach then this might be fine.

This PR sets `max-age=31104000` as the timeout for assets we are ok with caching "forever"

jkelleyrtp (approved):

PR #3933, opened 2025-03-31
Fix escaped text in ssr stylesheets and scripts

ealmloff:
In #3928, I noticed one of the fullstack playwright tests was pointing to the web port. The test for fullstack head elements was passing, but after switching it to the right port, it fails for two reasons:
1) The meta head element was not getting deduplicated
2) We try to escape text inside of script and style tags which breaks the tags

This PR fixes both issues and adds more tests

### DIFF
--- a/packages/ssr/src/cache.rs
+++ b/packages/ssr/src/cache.rs
@@ -78,14 +78,52 @@ impl AddAssign<Segment> for StringChain {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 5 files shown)


packages/ssr/src/cache.rs
```diff
@@ -243,18 +308,49 @@ fn from_template_recursive(
...
-        TemplateNode::Dynamic { id: idx } => *chain += Segment::Node(*idx),
+        TemplateNode::Dynamic { id: idx } => {
+            *chain += Segment::Node {
+                index: *idx,
+                escape_text,
+            }
+        }
     }
 
+    println!("this {:?}", chain.segments);
```
jkelleyrtp, line 352: no println

jkelleyrtp (approved):

jkelleyrtp (approved):

Issue #3878, opened 2025-03-15
Add Details For Tailwind V4.0 To Documentation


nickspurry:
## Feature Request
The [Tailwind page] in the documentation provides instructions to integrate Tailwind into a Dioxus app. The instructions include invoking the `npx tailwindcss init` command. This command was removed in Tailwind V4.0 released January 22 2025. The instructions can not be followed as written. Following the instructions to the letter will cause someone to install Tailwnd V4.0, which will then fail at a non-existent command.

My suggestion is to either: 
1. explicitly change these instructions so that a user downloads V3.x; or 
2. update/add to work with V4.0.

For bonus points, `dx new` should be updated to work with V4.0 as it currently produces a template only compatible with V3.x, but that might sit better in a separate issue.

PR #3943, opened 2025-04-01
feat: update examples to tailwind v4

danielkov:
## Context

- [Tailwind CSS V4 announcement](https://tailwindcss.com/blog/tailwindcss-v4)
- Tailwind V3 -> V4 [Upgrade Guide](https://tailwindcss.com/docs/upgrade-guide)
- Enhancement suggestion: DioxusLabs/dioxus#3878

## What changed?

- Updated `examples/tailwind` to use V4 Tailwind CSS CLI
  - Removed `tailwind.config.js` as it is unnecessary for V4 with a minimal setup 
  - Updated outdated utility classes
- Updated `example-projects/ecommerce-site` to use V4 Tailwind CSS CLI

## See also

- Updated docs: DioxusLabs/docsite#473


jkelleyrtp (changes requested):
Thanks for the PR! 

I see several `package.json` and `package-lock.json` files being generated.... 

I believe we don't need those - the tailwind CLI should be able to work without it.

Do you mind removing those and/or updating any docs to match?

danielkov:
Hi @jkelleyrtp  👋 

Thank you for the review!

> the tailwind CLI should be able to work without it

That has not been my experience. I had to run `npm install` with at minimum this `package.json` file present, before running `npx @tailwindcss/cli -i ./input.css -o ./assets/tailwind.css`, otherwise I got something along the lines of:

```
module "tailwindcss" not found
```

Do you mind sharing your setup / steps, just so I can confirm and update accordingly? 🙏

jlowry:
> Do you mind sharing your setup / steps, just so I can confirm and update accordingly? 🙏
I've tested it and confirm the The Tailwind CSS Standalone CLI does not need a package.json. See https://tailwindcss.com/blog/standalone-cli

ealmloff:
It depends how you install the tailwind CLI. If you run the CLI via npx which is the default in the tailwind docs then the new version does get installed in your node modules. If you use the standalone CLI, it does not.

The other two PRs that update tailwind (https://github.com/DioxusLabs/dioxus-template/pull/68 and https://github.com/DioxusLabs/docsite/pull/473) both use the npx CLI

PR #3944, opened 2025-04-02
Fix rsx evaluation order

### LINKED ISSUE #3737: RSX macro changes expression evaluation order
**Problem**

The rsx macro can include expressions in the dynamic nodes or dynamic attributes arrays. We evaluate all expressions in the dynamic nodes before expressions in the dynamic attributes. This can lead to counterintuitive compiler errors because the execution order is changed

**Steps To Reproduce**

This code looks like it would work in debug and release mode, but it fails to compile in release mode:
```rust
use dioxus::prelude::*;

#[derive(Clone, PartialEq, Eq)]
struct TestOuter;

impl TestOuter {
    fn borrow(&self) -> String {
        "".to_string()
    }
}

fn Parent() -> Element {
    let outer = TestOuter;

    rsx! {
        Fragment { key: outer.borrow(),
            Child { outer }
        }
    }
}

#[component]
fn Child(outer: TestOuter) -> Element {
    rsx! {
        div { "Hello" }
    }
}

fn main() {}
```

**Expected behavior**

The RSX macro should execute expressions in the order they were created in. We could pull out all expressions in the rsx macro and evaluate them in order then use the results as needed inside the macro expansion.

**Environment:**

- Dioxus version:  main
- Rust version:    nightly
- OS info:         macO
- App platform:    all

ealmloff:
We expand dynamic expressions, component, and attributes in different sections of the rsx macro. This can lead to very confusing borrow checker errors where rust complains that a item is borrowed before it is moved which should be valid, but because we shuffle expressions the item ends up being borrowed after it is moved.

This PR fixes that issue by introducing a dynamic expression pool and expanding items in depth first order in that pool before using the bindings in our dynamic node and dynamic attribute pools

Closes #3737

### DIFF
--- a/packages/core/src/global_context.rs
+++ b/packages/core/src/global_context.rs
@@ -250,7 +250,10 @@ pub fn remove_future(id: Task) {
 ///
 ///     rsx! {
 ///         button {
-///             onclick: move |_| *count.borrow_mut() += 1,
+///             onclick: {
+///                 let count = count.clone(); // Rc doesn't implement copy, so we need to clone
+///                 move |_| *count.borrow_mut() += 1
+///             },
 ///             "{count.borrow()}"
 ///         }
 ///     }
--- a/packages/core/src/scope_context.rs
+++ b/packages/core/src/scope_context.rs
@@ -389,7 +389,10 @@ impl Scope {
     ///
     ///     rsx! {
     ///         button {
-    ///             onclick: move |_| *count.borrow_mut() += 1,
+    ///             onclick: {
+    ///                 let count = count.clone(); // Rc doesn't implement copy, so we need to clone
+    ///                 move |_| *count.borrow_mut() += 1
+    ///             },
     ///             "{count.borrow()}"
     ///         }
     ///     }


jkelleyrtp:
Is there a way we can get away with keeping our current behavior but just fixing the places where it falls over?

Specifically, we want to expand attributes in this order:
1. Pass variables to formatting expressions *first*, since they usually just borrow the inner T
2. Pass variables to attribute values and component properties, since usually SuperFrom can work from &T
3. Finally, pass variables into callbacks, since those usually take the value as Owned.

I think this is similar to how we do it today, but our current ordering might have hiccups around attribute merging and keys.

ealmloff:
> Is there a way we can get away with keeping our current behavior but just fixing the places where it falls over?

The current expansion order is fairly arbitrary based on the position of the expression in the vnode struct, and doesn't necessarily result in less borrow checker issues. For example, this fails to compile today, but passes under a depth-first order:
```rust
#[derive(Clone, PartialEq, Eq, Debug)]
struct TestOuter;

#[component]
fn Child(outer: TestOuter) -> Element {
    rsx! {
        div { "Hello" }
    }
}

#[component]
fn Parent() -> Element {
    let outer = TestOuter;

    rsx! {
        div { width: "{outer:?}",
            Child { outer }
        }
    }
}
```

The main thing I want to avoid here is borrow checker errors that point you in the wrong direction. If you try to compile the above snippet on main you get this error which says you can't use the value after it is moved and then says you moved the value in the line after you use the value!

Screenshot 2025-08-07 at 6 20 06 PM

Usually the issue happens with attributes and properties, but you can also cause similar issues with formatted text expressions that take values. We don't have a good way to tell the type of an expression or what expressions borrow what values, so I think a clear and consistent evaluation order is more important than avoiding a few more borrow errors.


If we did want to sacrifice some consistency for less borrow errors while avoiding incorrect borrow before move errors we have today, we could resolve each of these in depth first order:
1) Idents in formatted expressions only. Idents will always borrow, unlike expressions which could move the value
2) All other expressions (attributes, props, child nodes, etc) which may borrow or move values

We can't expand closures last without causing the same errors we have today when a value is moved into a closure and another value at the same time:
```rust
#[component]
fn Parent() -> Element {
    let value = "hello".to_string();

    rsx! {
        button {
            // If we expanded this last, cargo would point out that we can't borrow the value
            // in the closure after it was moved
            onclick: move |_| {
                println!("Button clicked {value}!");
            },
            // But it is moved in what looks like after the closure?
            width: value,
        }
    }
}
```

Issue #3942, opened 2025-04-01
base_path does not work with dx bundle


Gisleburt:
**Problem**

`dx bundle` doesn't work with `base_path`

**Steps To Reproduce**

- Set up SSG following https://dioxuslabs.com/learn/0.6/guides/fullstack/static_site_generation
- Set a `base_path` in `Dioxus.toml`
- Attempt to run `dx bundle`

I've created a minimum repo that demonstrates this bug with a README here: https://github.com/Gisleburt/dioxus-base-path-issue

**Expected behavior**

Static Site Generation should create a bundle that opperates from `base_path`

**Build Log**

```text
 ERROR err=Other(Failed to parse static routes from the server. Make sure your server function returns Vec<String> with the (default) json encoding
```

**Environment:**

- Dioxus version: v0.6.0
- Rust version: v1.84.1
- OS info: macOS
- App platform: web

**Questionnaire**

I'm interested in fixing this myself but don't know where to start.

ealmloff:
If you are interested in fixing this, I think we just need to add the base path from the CLI config to the address for the server function [here](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/build/prerender.rs#L45-L48). You can get the base path from the web config in the builder's crate like we do [here](https://github.com/DioxusLabs/dioxus/blob/bdf87aadc748c7531041b1f89166ebd64d7c8c26/packages/cli/src/serve/handle.rs#L111)

PR #3958, opened 2025-04-05
Respect `base_path` when pre-rendering static routes

LeWimbes:
This pull request fixes https://github.com/DioxusLabs/dioxus/issues/3942 by passing the `base_path` from the `AppBundle` to `pre_render_static_routes` and using it to construct the URL accordingly.

I would like to add a test harness to verify that the `base_path` is always respected. However, I'd appreciate some guidance on how best to integrate such tests if that's something you'd like to see.

While investigating this issue and testing the example app from https://github.com/DioxusLabs/dioxus/issues/3942 (https://github.com/Gisleburt/dioxus-base-path-issue), I also encountered two additional problems that might be unrelated:

Sometimes content is duplicated or even triplicated.
image

When a non-existent page is requested, the bundle throws an error but then returns a 200 status and renders the homepage without duplicated content.
image

The error:
```
thread '<unnamed>' panicked at /home/leon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-router-0.6.3/src/contexts/router.rs:243:17:
route's display implementation must be parsable by FromStr
2025-04-05T13:11:19.269111Z ERROR dioxus_core::any_props: Panic while rendering component `dioxus_router::components::outlet::Outlet<base_path_issue::Route>`: Any { .. }
2025-04-05T13:11:19.269118Z ERROR dioxus_core::scope_arena: Error while rendering component `dioxus_router::components::outlet::Outlet<base_path_issue::Route>`:
Encountered error: CapturedPanic
In scope: ScopeId(5)
Backtrace: disabled backtrace
Context: 
```

Any feedback or guidance on my fix, the test harness for `base_path`, and whether you think the other two problems are related or if I should open new issues for them (I couldn't find any open Issues mentioning them) would be much appreciated.

### DIFF
--- a/packages/cli/src/cli/run.rs
+++ b/packages/cli/src/cli/run.rs
@@ -65,8 +65,12 @@ impl RunArgs {
                                 });
 
                             if bundle_format == BundleFormat::Web {
+                                let base_path = match builder.client.build.base_path() {
+                                    Some(base_path) => format!("/{}", base_path.trim_matches('/')),
+                                    None => "".to_owned(),
+                                };
                                 tracing::info!(
-                                    "Serving app at http://{}:{}",
+                                    "Serving app at http://{}:{}{base_path}",
                                     builder.devserver_bind_ip,
                                     builder.devserver_port
                                 );


LeWimbes:
I've added a Playwright test for serving an app with `base_path`.

However, I struggled adding a test for the bundling and SSG case. When I run the commented-out command in `playwright.config.js` manually, everything works as expected and I can visit the site at `127.0.0.1:8080/base-path`. But when Playwright executes it, I always encounter the error: "Failed to parse static routes from the server." Do you have any ideas on why that might be happening?

Additionally, is there a way to specify the port? I tried using `--port` as with `serve`, but that didn't seem to work; using `8080` might not be ideal.

I've also added `test-results` to the `.gitignore` (please let me know if the intent was to publish them) and rebased the changes upon the `main` branch.

ealmloff:
> I've added a Playwright test for serving an app with `base_path`.
> 
> However, I struggled adding a test for the bundling and SSG case. When I run the commented-out command in `playwright.config.js` manually, everything works as expected and I can visit the site at `127.0.0.1:8080/base-path`. But when Playwright executes it, I always encounter the error: "Failed to parse static routes from the server." Do you have any ideas on why that might be happening?

You might need to pull in the latest commits from main. https://github.com/DioxusLabs/dioxus/commit/12c6938346709a42079e23186d1d7e748cdddbbb#diff-62798fdae610a4cce6fb5d11419a6306a2381d67743e0ea2a50c1729470e8e6b fixes the post body the CLI sends to the server for SSG. 


> Additionally, is there a way to specify the port? I tried using `--port` as with `serve`, but that didn't seem to work; using `8080` might not be ideal.

`--port` should work for ssg. It seems like it is working for me with the existing ssg playwright test command. What error did you get?

LeWimbes:
> You might need to pull in the latest commits from main. https://github.com/DioxusLabs/dioxus/commit/12c6938346709a42079e23186d1d7e748cdddbbb#diff-62798fdae610a4cce6fb5d11419a6306a2381d67743e0ea2a50c1729470e8e6b fixes the post body the CLI sends to the server for SSG.

My branch is already up-to-date with `main`. I found that it actually works when I comment out the "web" test; therefore, it seems that port collisions are the problem.

> `--port` should work for ssg. It seems like it is working for me with the existing ssg playwright test command. What error did you get?

The `--port` option appears to only work with `serve`, not with `bundle`. Since the issue was specifically observed with `bundle`, I wanted to verify that scenario as well. I could assign the "web" test a different port to free up port `9999`.

I also noticed that running `serve` in SSG mode with a `base_path` fails ("Failed to parse static routes from the server.") when the `--release` flag is not set. So far I haven't been able to find the cause of this behavior in debug mode; any insight would be appreciated.

ealmloff:
> > You might need to pull in the latest commits from main. [12c6938#diff-62798fdae610a4cce6fb5d11419a6306a2381d67743e0ea2a50c1729470e8e6b](https://github.com/DioxusLabs/dioxus/commit/12c6938346709a42079e23186d1d7e748cdddbbb#diff-62798fdae610a4cce6fb5d11419a6306a2381d67743e0ea2a50c1729470e8e6b) fixes the post body the CLI sends to the server for SSG.
> 
> My branch is already up-to-date with `main`. I found that it actually works when I comment out the "web" test; therefore, it seems that port collisions are the problem.
> 
> > `--port` should work for ssg. It seems like it is working for me with the existing ssg playwright test command. What error did you get?
> 
> The `--port` option appears to only work with `serve`, not with `bundle`. Also, the server created by `bundle` doesn't seem to accept a `--port` option, right? Since the issue was specifically mentioned `bundle`, I wanted to create a test for that scenario. I could assign the "web" test a different port to free up port `9999`.

Oh, it looks like ssg does hardcode the port 9999 as the default [here](https://github.com/DioxusLabs/dioxus/blob/1001d78c7fbf717df50e9209513b03676e03635c/packages/cli/src/build/prerender.rs#L13). You can set the `PORT` environment variable to override the internal SSG port. Or we could change the CLI to look for a random free port like we do for the normal backend proxy with [this](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/serve/server.rs#L682) method if the default port is busy

> I also noticed that running `serve` in SSG mode with a `base_path` fails ("Failed to parse static routes from the server.") when the `--release` flag is not set. So far I haven't been able to find the cause of this behavior in debug mode; any insight would be appreciated.

I have seen that error with some older versions of the dioxus CLI. The default server function encoding changed which means if your CLI is older than the last couple of commits to main, it will try to send an empty post body. The new JSON server function encoding will reject that as invalid which causes the request to fail. On the latest commit in the main branch running ssg in debug mode with the playwright command works on my macbook: `cargo run --package dioxus-cli --release -- serve --bin nested-suspense-ssg --force-sequential --platform web --ssg --addr "127.0.0.1" --open`

LeWimbes:
> > I also noticed that running `serve` in SSG mode with a `base_path` fails ("Failed to parse static routes from the server.") when the `--release` flag is not set. So far I haven't been able to find the cause of this behavior in debug mode; any insight would be appreciated.
> 
> I have seen that error with some older versions of the dioxus CLI. The default server function encoding changed which means if your CLI is older than the last couple of commits to main, it will try to send an empty post body. The new JSON server function encoding will reject that as invalid which causes the request to fail. On the latest commit in the main branch running ssg in debug mode with the playwright command works on my macbook: `cargo run --package dioxus-cli --release -- serve --bin nested-suspense-ssg --force-sequential --platform web --ssg --addr "127.0.0.1" --open`

Running `nested-suspense-ssg` also works for me. So I guess there must be something about SSG and `base_path` together that breaks things in debug mode.
I'm a bit short on time right now, but I'll look into it further and improve the port situation as soon as possible. If it's super urgent, feel free to keep working on it, or close this PR and let someone else try.

LeWimbes:
With #4111, #4276, and several other `base_path`- and `bundle`-related PRs merged, I decided to take another look.

Although my changes to `pre_render.rs` make `base_path` work with SSG in some cases, there seem to be additional underlying issues that cause problems in other scenarios.

Using [this](https://github.com/LeWimbes/dioxus-base-path-issue) test project and Dioxus from this PR, I tried every combination of run/bundle, debug/release, and no-SSG/SSG (see the [README](https://github.com/LeWimbes/dioxus-base-path-issue/blob/main/README.md)). It appears that `run` with SSG does not work properly (issue with the Router?), and `bundle` in debug seems to just ignore the `base_path`.

EDIT: The issues remain the same with Dioxus 0.7.0

Issue #3920, opened 2025-03-26
Setting Default For Signal Props Is Verbose


DogeDark:
## Request

If you have a `ReadOnlySignal<bool>` or `Signal<bool>`, it can be verbose to  use the `#[props(default = x)]` syntax. For example:
```rust
#[derive(Props, Clone, PartialEq)]
pub struct SomeProps {

    // This sets bool to be false
    #[props(default)] 
    value: ReadOnlySignal<bool>,

    // This is what I'd like, except it wants a ReadOnlySignal
    #[props(default = true)] 
    value: ReadOnlySignal<bool>,

    // Instead you have to do this:
    #[props(default = ReadOnlySignal::new(Signal::new(true)))]
    value: ReadOnlySignal<bool>,

   // Same for signals:
   #[props(default = Signal::new(true))]
   value: Signal<bool>,
}
```

## Implement Suggestion
Detect if the prop type is a `ReadOnlySignal` or `Signal` and automatically create boilerplate code, injecting the provided expression as the input to the `::new` functions.

ealmloff:
> Seems like it would not be hard to extend the props macro to recognize Signal/ReadOnlySignal and adjust the default accordingly. Something sort of similar is done for [optionals](https://github.com/DioxusLabs/dioxus/blob/bdf87aadc748c7531041b1f89166ebd64d7c8c26/packages/core-macro/src/props/mod.rs#L230).
> 
> If this is a something the team wants handled in the macro I can send a PR for it.

That sounds like a good extension. If you want to work on a PR, keep in mind the signal we create in the props macro needs to be owned by the child component which means we need to call new inside of an owner in the builder. We have some logic for this already for converting `T` to `ReadOnlySignal<T>` [here](https://github.com/DioxusLabs/dioxus/blob/bdf87aadc748c7531041b1f89166ebd64d7c8c26/packages/core-macro/src/props/mod.rs#L1142-L1150)

PR #3959, opened 2025-04-06
Use SuperInto and correct Owner for props defaults

pandarrr:
This PR updates the generated props `build()` to use SuperInto and the props Owner for defaults of props that are child owned types (ReadOnlySignal/Callback), matching the behavior when the prop is set explicitly with its builder method.

Compile-time tests were added to assert that both existing default assignments and less-verbose defaults work.

Closes [3920](https://github.com/DioxusLabs/dioxus/issues/3920)


jkelleyrtp (approved):

jkelleyrtp:
Thank you!

Issue #3963, opened 2025-04-07
binary operation `==` cannot be applied to type `&mut Props<T>`


clouds56:
**Problem**

When have generics in `Props`, it would be required to be `Clone` and `PartialEq` even not needed.

**Steps To Reproduce**

```rust
use dioxus::prelude::*;

struct Box<T: Sized> {
  data: u32,
  phantom: std::marker::PhantomData<T>,
}

impl<T> PartialEq for Box<T> {
  fn eq(&self, other: &Self) -> bool {
    self.data == other.data
  }
}

impl<T> Clone for Box<T> {
  fn clone(&self) -> Self {
    Self {
      data: self.data,
      phantom: std::marker::PhantomData,
    }
  }
}

// C1 compiles as expected
#[component]
fn C1(a: Box<u32>) -> Element {
  unimplemented!()
}

// C2 complains about "binary operation `==` cannot be applied to type `&mut C2Props<T>`"
// t.rs(29, 17): consider further restricting type parameter `T` with trait `PartialEq`: ` + std::cmp::PartialEq`
#[component]
fn C2<T: 'static>(a: Box<T>) -> Element {
  unimplemented!()
}
```

**Expected behavior**

Should compile

**Screenshots**

**Environment:**

- Dioxus version:  0.6.3
- Rust version:    1.85.0
- OS info:         Windows 11
- App platform:    N/A

**Questionnaire**

I would like to fix and I have a solution.

https://github.com/DioxusLabs/dioxus/blob/23286ade2b032d7fd648f51c5f0fc794467c459d/packages/core-macro/src/component.rs#L231-L237
Now it would derive `Clone` and `PartialEq`, we shall implement these traits ourselves, and fill in with more correct where clause

PR #3968, opened 2025-04-08
fix: more accurate derive `Clone, PartialEq` for generic components

clouds56:
fixes https://github.com/DioxusLabs/dioxus/issues/3963


jkelleyrtp (approved):
Looks good to me - but not sure why checks aren't running. This feels similar to me as to why impl `Copy` doesn't work properly in Rust std

PR #3979, opened 2025-04-11
Dioxus-Native WGPU Canvas Integration

jkelleyrtp:


nicoburns:
Ultralight has a cool API (https://ultralig.ht/api/cpp/1_4_0/classultralight_1_1_image_source.html) where you can register a texture with a URL then you can use that URL anywhere an image URL is expected (e.g `src` attribute of `img` tag or `background-image` css property) and it will render the texture as the image. Aside from just being very cool functionality, this would sidestep the requirement for "getElementById" or similar functionality for implementing this feature.

nicoburns:
Bevy example of rendering to texture: https://github.com/bevyengine/bevy/blob/main/examples/3d/render_to_texture.rs

jkelleyrtp:
Implemented here instead

https://github.com/DioxusLabs/dioxus/pull/4286

PR #3988, opened 2025-04-14
Move asset hashing into the CLI

### LINKED ISSUE #3467: Regression: dioxus-cli fails to create multiple processed copies of a same image with different options

### LINKED ISSUE #3705: asset cache hashes don't take into account included scripts
**Problem**

If I have a script called `main.js` and it has a line:

```
import * from "./side.js"
```

then when I change the `side.js` file, the hash for the `main.js` file doesn't change after bundling.

This is especially odd when including `wasm_import_modules` since the `main.js` doesn't actually update even when the dynamically loaded `side.js` file *does* change.

https://rustwasm.github.io/docs/book/reference/js-ffi.html#importing-and-exporting-js-functions

The current workaround is to include some entropy in the `main.js` somehow to bust its cache. Ideally when we register assets, we run the bundling step *first* and then generate the hash *after* before inserting it into the manifest.

### LINKED ISSUE #3918: JS included by wasm_bindgen binding is not updated
**Problem**

Given the following binding:

```
#[wasm_bindgen(module = "assets/bundle.js")]
unsafe extern "C" {
    pub fn test1();
}
```

And the file `assets/bundle.js`
```
function test1() {
  console.log("test");
}
export {
  test1
};
```

Everything works as expected. Problems occur when for example `test1`is renamed to `test2` in `bundle.js` and the binding declaration, `dx serve` continues to use the old, cached, version for the javascript code, but still tries to call `test2`.

- Hot reloading does not help
- Manual rebuilding with `r` does not help
- Restarting `dx serve` does not help
- Removing the `target` directory works
- Changing the referenced script name in `#[wasm_bindgen(module = "assets/bundle.js")]` to something else and back again works

Error in browser:
```
Uncaught (in promise) LinkError: import object field 'test2' is not a Function 
```

When not changing the function name, but the content, also the old version is used (without the LinkError).

**Environment:**

- Dioxus version:  0.6.3
- Rust version:     1.85.1
- OS info:             macOS
- App platform:    web, fullstack

ealmloff:
TODO:
- [x] Test linked bugs are fixed on wasm
- [x] Test linked bugs on Macos desktop
- [x] Test linked bugs on Windows desktop
- [x] Test linked bugs on Linux desktop

Fixes #3705 
Fixes #3467 
Fixes #3918

Stacked on https://github.com/DioxusLabs/dioxus/pull/4042


ealmloff:
Because asset optimization is now a full linker step, it no longer works with the command added in https://github.com/DioxusLabs/dioxus/pull/3429. If you want to use the dx asset optimizer outside of the CLI, you need to use the CLI as a linker like this (copied from the new help section):

# Dioxus Asset Optimizer
When you serve dioxus with dx, it will automatically handle the build process for you. If you need to integrate with a larger build system,
you can use the `dx` asset optimizer separately to link to your assets. If you set the `DX_LINK_ASSETS_TARGET` environment variable, dx will
proxy your linker and copy the optimized assets it finds in your program into the specified directory.

## Usage with trunk

If you are using trunk, you need to create a temporary asset directory to store the output of the dx asset optimizer that will be copied by trunk into your dist directory:
```html
<html>
  <head>
    <link data-trunk rel="rust"/>
    <link data-trunk rel="copy-dir" href="./dist_assets/" data-target-path="./assets/"/>
  </head>
  <body>
    <div id="main"></div>
  </body>
</html>
```
Then when you build, you need to set the `DX_LINK_ASSETS_TARGET` environment variable to the path of the temporary asset directory and `dx` as your linker:
```sh
DX_LINK_ASSETS_TARGET="dist_assets" RUSTFLAGS="-Clinker=dx" trunk serve
```

## Usage with cargo

If you are using cargo, you need to set the `DX_LINK_ASSETS_TARGET` environment variable to the path where your optimize assets will be stored and `dx` as your linker:
```sh
DX_LINK_ASSETS_TARGET="dist_assets" RUSTFLAGS="-Clinker=dx" cargo run
```
## Custom linker path

DX will try to find the default linker for your system, but if you need to use a custom linker on top of the dx proxy, you can set the `DX_LINKER` environment variable to the path of your custom linker. For example, if you are using `lld` as your linker, you can set the `DX_LINKER` environment variable to the path of `lld`:
```sh
DX_LINKER="/path/to/lld" DX_LINK_ASSETS_TARGET="dist_assets" RUSTFLAGS="-Clinker=dx" cargo run
```

packages/cli/src/build/bundle.rs
```diff
@@ -354,38 +353,19 @@ impl AppBundle {
...
-        let mut keep_bundled_output_paths: HashSet<_> = self
+        let keep_bundled_output_paths: HashSet<_> = self
             .app
             .assets
-            .assets
-            .values()
+            .assets()
             .map(|a| asset_dir.join(a.bundled_path()))
             .collect();
-        // The CLI creates a .version file in the asset dir to keep track of what version of the optimizer
```
ealmloff, line 378: The optimizer's version is now included in the file hash so we don't need to keep track of an asset optimizer version file

packages/cli/src/cli/build_assets.rs
```diff
@@ -1,30 +1,35 @@
...
 use tracing::debug;
+use wasmparser::BinaryReader;
 
 #[derive(Clone, Debug, Parser)]
 pub struct BuildAssets {
     /// The source executable to build assets for.
     pub(crate) executable: PathBuf,
 
-    /// The source directory for the assets.
-    pub(crate) source: PathBuf,
```
ealmloff, line 14: The path this is joined against is always absolute, so setting this argument makes no difference

ealmloff:
This PR is now modifying the final executable and resigning it on MacOS. Windows and Ubuntu platforms worked without needing a signature when I tested them

jkelleyrtp:
Hotpatching is causing asset src paths to be set to `/assets/` for me on this branch, breaking assets altogether. Maybe a merge conflict issue?

https://github.com/user-attachments/assets/0e43353d-a3f2-466d-8256-e1c4947b3056

packages/cli/src/build/assets.rs
```diff
@@ -0,0 +1,320 @@
...
+    let pdb_file = find_pdb_file(path);
+
+    if file.format() == object::BinaryFormat::Wasm {
+        find_wasm_symbol_offsets(file_contents, file)
+    } else if let Some(pdb_file) = pdb_file {
+        find_pdb_symbol_offsets(&pdb_file)
+    } else {
+        find_native_symbol_offsets(file)
+    }
+}
```
jkelleyrtp, line 50: maybe match on the binary format / os like we do in patch.rs?

jkelleyrtp (approved):

packages/cli/src/build/assets.rs
ealmloff: I don't think we have the OS info here if you run the asset command directly, and you might be cross-compiling, so we shouldn't rely on the host os

PR #4014, opened 2025-04-19
fix: dx bundle progress output off-by-one in log lines

chrivers:
Before:

```
...
   1.811s  INFO dx::build::builder: Compiling [270/272]: dioxus
   1.811s  INFO dx::build::builder: Compiling [271/272]: bifrost_api
  25.410s DEBUG dx::build::request: Build completed successfully - output location: "..."
...
  26.94s  INFO dx::build::builder: Copying asset (0/7): ...
  26.95s  INFO dx::build::builder: Copying asset (1/7): ...
...
```

After:

```
...
   1.811s  INFO dx::build::builder: Compiling [271/272]: dioxus
   1.811s  INFO dx::build::builder: Compiling [272/272]: bifrost_api
  25.410s DEBUG dx::build::request: Build completed successfully - output location: "..."
...
  26.94s  INFO dx::build::builder: Copying asset (1/7): ...
  26.95s  INFO dx::build::builder: Copying asset (2/7): ...
...
```


DogeDark (changes requested):
It looks like the crate count isn't off-by-one, but rather we don't emit the last log. If you want to fix that, I think we need to change:

https://github.com/DioxusLabs/dioxus/blob/a9468429d7181499f920be89b747a58f32727d08/packages/cli/src/build/request.rs#L201-L208

To be something along the lines of:
```rust
unites_compiled += 1;
self.status_build_progress(units_compiled, crate_count, artifact.target.name);
output_location = artifact.executable.map(|e| e.into());
```

The rest looks good!

chrivers:
@DogeDark thanks for the hint - fixed, and tested. Indeed, one more unit turns up now.

Also, I changed the text from `Compiling` to `Compiled`, since the log lines refer to what is completed, not actually what is in progress.

This makes the log output much less confusing. Previously, if something was taking a long time to compile, the wrong crate was getting the blame!

DogeDark (approved):

chrivers:
Wow, that's a heavy CI job.. is that normal?

@DogeDark I guess the CI needs to be poked? :)

PR #4029, opened 2025-04-22
Fix: CSS Module Processing

### LINKED ISSUE #4028: CSS modules can't handle prefixes
**Problem**

If I have a CSS module like so:

```css
.sidebar { 
    background: red;
    padding: 20px;
}
.sidebarButton {
    background: blue;
}
```

then the `Classes::sidebar_button` class doesn't actually work. The reason is that because it has the same `sidebar` prefix as an already mangled class, it mangles the name twice on accident.

```css
.sidebarfb811ee670c8c644 {
    background: red;
    padding: 20px
}

.sidebarfb811ee670c8c644Button {
    background: #00f
}
```

in the DOM:

```html
<div class="sidebarfb811ee670c8c644">
    <button class="sidebarButtonfb811ee670c8c644">Hi</button>
</div>
```

**Steps To Reproduce**

```rust
css_module!(Classes = "/assets/sidebar.css");

#[component]
pub fn Sidebar() -> Element {
    rsx! {
        div { class: "{Classes::sidebar}",
                button { class: "{Classes::sidebar_button}", "Hi" }
            }
    }
}
```

**Expected behavior**

The button should be blue.

**Screenshots**

Image

**Environment:**

- Dioxus version:  main@1001d78c7fbf717df50e9209513b03676e03635c
- Rust version:    1.86
- OS info:        Windows 11
- App platform:    desktop

**Questionnaire**

DogeDark:
Fixes #4028 

The problem was that we replaced any CSS ident with `Str::replace`, which, if given `foo` would replace `foo` and the foo in `fooBar`. To fix this, we now check for the full matching ident line by line instead of replacing haphazardly.

### DIFF
--- a/packages/cli-opt/src/css.rs
+++ b/packages/cli-opt/src/css.rs
@@ -49,7 +49,7 @@ pub(crate) fn process_css_module(
     final_path: &Path,
     output_path: &Path,
 ) -> anyhow::Result<()> {
-    let mut css = std::fs::read_to_string(source)?;
+    let css = std::fs::read_to_string(source)?;
 
     // Collect the file hash name.
     let mut src_name = source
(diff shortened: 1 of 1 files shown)


nicoburns (commented):
Hmm. Checking whether a line starts with `.foo` or `#foo` doesn't seem like it's going to be particularly reliable. What if I have:

```css
.width { width: 100px }
```
or
```css
div > .foo {
     width: 100px;
}
```

I think you ought to be actually parsing the CSS to implement this. You ought to be able to use the `cssparser` (and perhaps `selectors`) crate(s) for this.

DogeDark:
Yeah, that will be a problem. I tried to get away with a small hand-rolled parser, but it's not going to work well.

Issue #3945, opened 2025-04-02
Braced expressions followed by a method don't parse


ealmloff:
**Problem**

The rsx macro doesn't handle braced expressions correctly

**Reproduction**

This code fails to compile

```rust
// [dependencies]
// dioxus = { version = "0.6" }
use dioxus::prelude::*;

fn main() {
    rsx! {
        div {
            width: {"100%"}.to_string()
        }
    };
}
```

**Expected behavior**

The code should compile and run

**Environment:**

- Dioxus version:  0.6, and main
- Rust version:    nightly
- OS info:         macOS
- App platform:    all

jjvn84:
Changing it to:

```
// [dependencies]
// dioxus = { version = "0.6" }
use dioxus::prelude::*;

fn main() {
    rsx! {
        div {
            width: {{"100%"}.to_string()}
        }
    };
}
```

makes the code compile but gives an "unnecessary braces" warning. This could be a workaround for people facing this issue.

ealmloff:
Fixed by #4035

PR #4035, opened 2025-04-23
Fix parsing of braced expressions followed by a method

jjvn84:
This is an attempt at solving issue [https://github.com/DioxusLabs/dioxus/issues/3945](https://github.com/DioxusLabs/dioxus/issues/3945) when parsing attributes.
I added code that checks if there is a TokenTree::Group followed by a dot.  If it finds it, it parses the whole attribute as an expression. It stops if it finds a comma, signaling the end of the attribute.
I added the example in the original issue to the "complex_kitchen_sink" test, and everything passed.

### DIFF
--- a/packages/rsx/src/raw_expr.rs
+++ b/packages/rsx/src/raw_expr.rs
@@ -1,4 +1,4 @@
-use proc_macro2::{TokenStream as TokenStream2, TokenTree};
+use proc_macro2::{Delimiter, TokenStream as TokenStream2, TokenTree};
 use quote::ToTokens;
 use std::hash;
 use syn::{parse::Parse, spanned::Spanned, token::Brace, Expr};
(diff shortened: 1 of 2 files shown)


ealmloff (changes requested):
I think the issue is slightly broader than this. If we find anything except a comma after a braced expression we should try to parse it as a normal expression. The check you added fixes this expression `width: {"100%"}.to_string(),`, but this expression still fails `width: {|| "100%"}(),`

jjvn84:
@ealmloff These new changes should fix the issue. Could you please check again if there is another case I'm missing?
And thank you for the review.

ealmloff:
I found a few more cases that are broken. This test fails on the PR, but passes on the main branch:
```rust
#[test]
fn invalid_braced_expression() {
    let item = quote::quote! {
        // Partial expressions in braces rsx should be allowed and output as-is
        // for autocomplete
        {invalid.}
        div {}
        {invalid.}
        // Comments should be ignored
        div {}
        {invalid.}
        "hello world"
        {invalid.}
        if true {}
    };

    let cb: CallBody = syn::parse2(item).unwrap();
    println!("{}", cb.to_token_stream());
}
```

These all pass now which is great!
```rust
#[test]
fn expressions_after_braced() {
    let item = quote::quote! {
        div {
            width: {100} - 50,
            width: {"100%"}.to_string(),
            width: {|| "100%"}(),
        }
    };

    let cb: CallBody = syn::parse2(item).unwrap();
    println!("{}", cb.to_token_stream());
}
```

I think these two cases should cover everything?
1) Braced expression followed by a symbol other than a brace (`-`, `*`, `(`, etc) should be parsed as a `syn::Expr`
2) Braced expression followed by a term which is either another braced expression, a raw ident or a literal should be parsed as a partial expression

jjvn84:
This new change fixes the missing cases. Please let me know if there are more test cases I should include.

ealmloff (approved):
Looks great, thanks for fixing this! Confirmed autocomplete still works with the base raw brace expressions

PR #4038, opened 2025-04-24
fix: dx check now respects files to ignore (e.g. .gitignore)

### LINKED ISSUE #3751: dx check fail because it check every crate at the .cargo folder
**Problem**

My `$CARGO_HOME` is set to the project dir + `.cargo`.
Dx check also this folder and fail.
**Steps To Reproduce**

- set $CARGO_HOME` to prosect dir + `.cargo`
- depend at `thiserror-impl-1.0.69`
- execute `dx check`

**Expected behavior**
Do only check the file which are included in the rust src and ignore dependencies.

**Screenshots**

**Environment:**

- Dioxus version:  dioxus 0.6.3 (was built without git repository)
- Rust version:    rustc 1.84.1 (e71f9a9a9 2025-01-27)
- OS info:         arch linux
- App platform:    web

**Questionnaire**

SilentVoid13:
this should fix #3751. basically does what @ealmloff suggested, using the same function to collect `.rs` files for both `dx check` and `dx fmt`

### DIFF
--- a/packages/cli/src/cli/autoformat.rs
+++ b/packages/cli/src/cli/autoformat.rs
@@ -1,5 +1,5 @@
 use super::*;
-use crate::DioxusCrate;
+use crate::{metadata::collect_rs_files, DioxusCrate};
 use anyhow::Context;
 use dioxus_autofmt::{IndentOptions, IndentType};
 use rayon::prelude::*;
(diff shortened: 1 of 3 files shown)


ealmloff (approved):
Thanks! Confirmed this fixes the git ignore issue and it is nice to unify some of the file collection logic. The CI failure is unrelated

PR #4042, opened 2025-04-25
Fix playwright tests, liveview and feature selection

ealmloff:
Liveview broke with the 0.8 axum upgrade which is now causing the playwright tests to fail. This PR fixes the issue by updating the router definition

This PR also fixes a bug in the CLI that is causing the wrong features to activate which breaks the playwright test

Fixes https://github.com/DioxusLabs/dioxus/issues/4046

### DIFF
--- a/packages/cli/src/cli/build.rs
+++ b/packages/cli/src/cli/build.rs
@@ -101,7 +101,7 @@ impl BuildArgs {
         self.fullstack = Some(
             self.fullstack()
                 || self.fullstack.is_none()
-                    && (default_server || krate.has_dioxus_feature("fullstack")),
+                    && (default_server || krate.directly_depends_on_dioxus_feature("fullstack")),
         );
 
         // If the current build is a fullstack build which includes either the client or the server in the default features,
--- a/packages/cli/src/dioxus_crate.rs
+++ b/packages/cli/src/dioxus_crate.rs
@@ -3,6 +3,7 @@ use crate::{Arch, CliSettings};
 use crate::{Platform, Result};
 use anyhow::Context;
 use itertools::Itertools;
+use krates::Node;
 use krates::{cm::Target, KrateDetails};
 use krates::{cm::TargetKind, Cmd, Krates, NodeId};
 use once_cell::sync::OnceCell;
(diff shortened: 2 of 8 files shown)


nicoburns (approved):

ealmloff:
The wasm split feature is failing with this error after fixing the dependency issues:
```
11:52:35 [dev] Emitting main bundle split module
11:52:36 [dev] Build failed: Other(Failed to write main executable

Caused by:
    0: Failed to read module
    1: [parse exception: invalid code after misc prefix: 17 (at 0:441310)])
```

ealmloff:
> gh pr checkout 4042
> 
> This does not fix #4046 for me.

This PR changes the CLI to fix the issue so you will need to either install the CLI on the branch or run the CLI package. This command works on for me on macos with this branch:
```sh
cargo run --package dioxus-cli -- serve --example dynamic_asset --features desktop
```

davehorner:
I can verify that the PR does fix the problem and produce a working build.  The same command does not work on the latest dioxus repo. I'm confused on how the axum playwright affect the hello_world dynamic_asset examples.  

Thanks for fixing this.  I'd like to understand what happened all the examples were building OK and then all my machines give me errors about unresolved module or unlinked crate `dioxus_fullstack`

ealmloff:
> `cargo run --package dioxus-cli -- serve --example dynamic_asset --features desktop` I can verify that the PR does fix the problem and produce a working build. The same command does not work on the latest dioxus repo. I'm confused on how the axum playwright affect the hello_world dynamic_asset examples.
> 
> Thanks for fixing this. I'd like to understand what happened all the examples were building OK and then all my machines give me errors about unresolved module or unlinked crate `dioxus_fullstack`

The same issue that was causing playwright tests to fail also broke running the example. When you run an app with dioxus, we try to detect if it is a fullstack app. That logic was broken, so the CLI included fullstack even when fullstack was not enabled

davehorner:
Thanks very much for the fix and explanation.
https://crates.io/crates/cargo-e/
I'm working on tool that helps people run dioxus examples; when this occurs in a random user's project, I guess the only solution is to upgrade the dx cli when this fix becomes available.

jkelleyrtp:
> For me, this PR is fixing just regular `dx serve` for any Dioxus example which is broken on `main`.

Another issue with the CLI currently is that we don't set `workspace=true` when running cargo metadata so a number of our examples / packages don't show up in the list even when they should.

davehorner:
tried the latest main today.  The auto-platform detection still returns unknown on many.  More concerning now is that calculator and other examples double launch windows.
```
dx.exe serve --example calculator --features desktop
```
image
I have no idea.  I can send more log; just wanted to report calculator and `dx.exe serve --example custom_menu --features desktop` both popp'd double windows.

ealmloff:
> tried the latest main today. The auto-platform detection still returns unknown on many. More concerning now is that calculator and other examples double launch windows.
> 
> ```
> dx.exe serve --example calculator --features desktop
> ```

I can't reproduce that issue on [this](https://github.com/DioxusLabs/dioxus/commit/e47d0aa86838ec757405801fc75bf05646205a03) commit with the matching version of the CLI. Try running with the dioxus-cli package from the same commit:
```sh
cargo run --package dioxus-cli -- serve --example calculator --features desktop
```

davehorner:
just tested on 18477a037 working on win11 and win 10.

PR #4061, opened 2025-04-29
Save logical position when restoring on macos

clouds56:
On macOS, it behaves strange with multiple monitor physical location. That is to say

Consider we have 2 monitor main and extra which physical size is `2560x1440` (scale: 2.0) and `1920x1080` (scale: 1.0)
```
+-----------+---------+
|     .<-A  |  .<-B   |
|   main    |  extra  |
| 2560x1440 |1920x1080|
|           |---------+
+-----------+
```
The physical position of A is `(1200, 100)`, and the logical position of it would be `(600, 50)` due to scale factor.
The code before this PR would save `(1200, 100)` to "window-state.json" and everything just works.

But for point B, which "relative" physical position is `(100, 100)` (offset from left-top of "extra" monitor),
In this case, winit would tell the physical position is `(1380, 100)`, which is the same as logical position `(1380, 100)`.
If we save `(1380, 100)` to state file, and when restore, the winit would think we are at main screen and position to `(640, 50)` in logical position.

| Point name | p1 = `window.outer_position()` | p2 = `p1.to_logical(window.scale_factor())` |  restored p1 (in logical) | restore p2 |
| - | --- | --- | --- | --- |
| A | PhysicalPosition(1200, 100) | LogicalPosition (600, 50) | LogicalPosition (600, 50) | LogicalPosition (600, 50) |
| B | PhysicalPosition(1380, 100) | LogicalPosition (1380, 100) | LogicalPosition (640, 50) | LogicalPosition (1380, 100) |

TL;DR, in macOS the position system is consist in "logical" coordinates.

Edit:
This seems only macOS related, so I restrict patch only on `cfg!(target_os = "macos")`

There're several related issues you could reference
* https://github.com/tauri-apps/tao/issues/816
* https://github.com/rust-windowing/winit/issues/2645
* https://github.com/tauri-apps/plugins-workspace/issues/2620


clouds56:
I found `let adjustment = 56` failed on extra monitor, investigating.

clouds56:
What's blocking here?

jkelleyrtp:
Thanks for the PR!

Should we implement LogicalPosition for all platforms, not just macos? Does it make sense to maintain both?

I believe the 56 offset is meant to handle the thickness of the topbar, otherwise it marches down on a reload.

Also CI not passing is blocking this PR, might be worth merging main to see if the failures go away.

clouds56:
On macos, there's 1-1 mapping from points on screen to points in Logical Position, (and there's 2 points on screen have the same expression in Physical Position, like (1380, 100))
While on windows, there's 1-1 mapping from points on screen to points in **Physical** Position.

I once tried use logical position on windows but it failed to restore to the same location if 2 monitors have different scale. So I revert it in https://github.com/DioxusLabs/dioxus/pull/4061/commits/0fbd886a950cc16a6d53cfa73e0d14486e0f890a

For CI failure, I think there's some network issue, can re rerun it?

Issue #3685, opened 2025-02-04
Android App Icon Configuration Not Working


ksmit323:
**Problem**

Unable to configure custom app icon for Android builds using Dioxus.toml file. The default Android photo icon is always used regardless of configuration attempts.

**Steps To Reproduce**

1. Create a Dioxus project with Android target
2. Configure Dioxus.toml with icon paths:
```toml
[bundle]
icon = [
    "assets/icons/icon-32x32.png",
    "assets/icons/icon-128x128.png",
    "assets/icons/icon-128x128@2x.png",
    "assets/icons/icon-256x256.png",
    "assets/icons/icon-512x512.png"
]
```
3. Create PNG icons in specified sizes
4. Run `dx build --platform android`
5. Install with `adb install target/.../app-debug.apk`
6. Observe that app still uses default Android photo icon

**Expected behavior**

The app should display the custom icon specified in Dioxus.toml when installed on an Android device. Instead, it always shows the default Android photo placeholder icon.

**Environment:**
 - Dioxus version:  0.6.2
 - Rust version:    1.84
 - OS info:         Linux (Ubuntu 24.04)
 - App platform:    Android

ksmit323:
For anyone who is looking for a fix, I was able to get this to work.  Not through the `Dioxus.toml` but through changing the default "droid" image in the `mipmap` files in the target directory that `dx build` creates. 

This is for **Linux**, fyi, and assumes you already used `dx build --platform android` to create the target directory.

#### The short answer is:
1. I created five square PNG images, just the default Android size (48x48, 72x72, 96x96, 144x144, 192x192)
2. I replaced the default images in the `mipmap` folder, starting with the smallest image in the `mdpi` folder.  There will be five of these image folders and after running `dx build` you should find them under your `target/` directory like this:
```bash
target/dx/your-app-name/debug/android/app/app/src/main/res
```

3. Now the weird part. My `ic_launcher.xml` file was referencing a file called `ic_launcher_foreground.xml` but that file did not exist. So I think that was causing the Android to default back to the green droid icon. So I just deleted the `ic_launcher.xml` file:
```bash
rm target/dx/your-app-name/debug/android/app/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```
and that disabled "adaptive icons". I think this forces Android to use the regular mipmap icons instead.

4. After that, I cleaned, rebuilt and reinstalled and it worked! However, I rebuilt with the `./gradlew` executable instead:
[...]

PR #4063, opened 2025-04-30
feat(cli): Generate icons for Android when building APK

fontlos:
When building an APK, if an icon is specified in the configuration, it will generate `*dpi` variants to override the default Android icon. For simple SVG files, it additionally generates Android VectorDrawable XML for `anydpi`. Refs #3685 

The icon file is specified in the following configuration:

```toml
# Dioxus.toml
[bundle]
icon = ["path/to/icon"]
```

However, there are two issues:

First, I'm not sure why the icon is defined as `Vec<String>`. I couldn't find any usage of this field in the project as a reference, so I default to using the first element as the icon.

Second, Android appears to apply some scaling to the foreground of Android VectorDrawable XML, causing the actual icon to be enlarged and cropped. This requires adding some margin when creating the SVG. Additionally, currently the same icon file is used for both foreground and background, and only SVGs composed of simple paths are supported.


brisk0:
Hi Fontlos, regarding your `Vec<string>` query:

> icon - Paths to icon files to be used in the bundle. Icon files must be square and 16, 24, 32, 64, or 256 pixels in size. PNG icons must have a 32 bit depth in the RGBA format. If you use a .icns file is must fit [this](https://github.com/tauri-apps/tauri/blob/d8db5042a28635259f646c329c3ec5ccf23eac9e/tooling/cli/src/helpers/icns.json) format. *The icons must include a .icns icon for macOS, .ico for Windows and .png for Linux.*

https://dioxuslabs.com/learn/0.6/CLI/configure#bundle

I read this part of the docs as implying that `icon` is defined as a `Vec<String>` in part to accommodate different format requirements for different platforms.

Based on this the behaviour I would expect from Dioxus would be to search for the first svg in the list, or to have an order of preferences (e.g. .xml > .svg > big png > small png).

I think this describes the scaling, essentially the final size is 75% of the total size

https://developer.android.com/distribute/google-play/resources/icon-design-specifications#legacy_mode

fontlos:
Thank you for your answer @brisk0 . I made some change

Now we will select an icon from the icon list to generate the icon.
The XML file (Android VectorDrawable XML) has the highest priority, and it is assumed that the user has completed all necessary work. It will be copied directly without any processing.
The SVG file has the next highest priority, and we will attempt to generate an Android VectorDrawable XML file from it. According to https://developer.android.com/distribute/google-play/resources/icon-design-specifications, the icon file needs to be scaled down by 75%. I think that this is also an issue that the user should handle, not dioxus-cli.

pythonpoet:
Whats the state on this?

Issue #4037, opened 2025-04-24
Event handling broken after Dioxus 0.6 migration


wiseaidev:
Hiya (ᵔ⩊ᵔ),

After upgrading from Dioxus `0.5` to `0.6` in [this PR](https://github.com/opensass/opensass/pull/22), most interactions and event handling stopped working.

I'm also trying to implement ISG by adding [a comment section](https://opensass.org/blogs/beyond-typescript#final-word-for-now) to statically generated blogs using a combination of [layout](https://github.com/opensass/opensass/blob/92a4f3b038ac95e39605d3951e3d830b3abf777e/src/router.rs#L83-L86) and [`Outlet`](https://github.com/opensass/opensass/blob/92a4f3b038ac95e39605d3951e3d830b3abf777e/src/pages/blog.rs#L103). This worked in 0.5 (without SSG), but in `0.6` with SSG, events no longer fire and nothing sent the server.

I'll check the release notes, but if you're aware of breaking changes around this, I'd appreciate any guidance.

Thanks!

ealmloff:
Make sure there are no errors in the console after triggering events. That could be a [hydration issue](https://dioxuslabs.com/learn/0.6/guides/fullstack/hydration#)

wiseaidev:
> Make sure there are no errors in the console after triggering events. That could be a [hydration issue](https://dioxuslabs.com/learn/0.6/guides/fullstack/hydration#)

Nice tutorial! But the thing is that there is no hydration errors in the console, nor error messages in the terminal. Any advice on how to debug this further?

Thanks!

ealmloff:
> Nice tutorial! But the thing is that there is no hydration errors in the console, nor error messages in the terminal. Any advice on how to debug this further?

Here is what I would expect to see for an SSG app:
1) When you run `dx bundle --fullstack --ssg --release` it generates a bunch of static routes and assets including as wasm file in the public directory
2) You serve that public directory with some static file server
3) In the console you should see the wasm asset loading. The file size should be larger than 100kb
4) You should see attributes with `hydration` somewhere in the name in some of your html elements initially rendered on the server

If all of that looks correct, I would recommend trying to minimize the issue by eliminating parts of your app. If this is a bug with dioxus, a standalone repo that reproduces the bug makes it a lot easier to identify and fix the issue

wiseaidev:
[...]

PR #4073, opened 2025-05-04
fix: add missing `img` attrs

wiseaidev:
While adding support for Dioxus in [`image-rs`](https://github.com/opensass/image-rs), i noticed some of the `img` tag attrs are missing. So, i added them.

Also, I have a question: how can we attach a node reference to a tag? I don't see any tags that accept the `ref` attr. Why tho?

### DIFF
--- a/packages/html/src/elements.rs
+++ b/packages/html/src/elements.rs
@@ -1103,7 +1103,10 @@ builder_constructors! {
         usemap: String DEFAULT, // FIXME should be a fragment starting with '#'
         width: usize DEFAULT,
         referrerpolicy: String DEFAULT,
-        // sizes: SpacedList<String>, // FIXME it's not really just a string
+        sizes: String DEFAULT, // FIXME
+        elementtiming: String DEFAULT,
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):

wiseaidev:
> The way we suggest using node refs right now is using getElementById. We'd like to update this interface but need a way to do it cross-platform.

Ah, ok, that makes sense!

> You can also use onmounted which will give you the MountedData which is not exactly the same as a NodeRef but is similar in spirit.

I see. I've used this attr before in [opensass/opensass](https://github.com/opensass/opensass/blob/360e7680c50e0928abc64865394757365f7fbede/src/pages/home.rs#L40). but it's not working rn because all events stopped working due to #4037 . I'm trying to implement dynamic data handling alongside static content (e.g. comments for SSG blogs) something like [ISG](http://patterns.dev/react/incremental-static-rendering), but I'm not sure how to implement it in dioxus atm. If you can help with this issue, that would be much appreciated.

Thanks for the response!

PR #4095, opened 2025-05-08
CLI self-updater, warn on version mismatch, direct install.sh option, androidmanifest and info.plist customization, bump to 0.7-alpha.0

jkelleyrtp:
This PR adds a few things
- [x] the ability to customize androidmanifest.xml and info.plist 
- [x] an install.sh script to save the hassle of installing binstall
- [x] wipe the android asset tmp dir when starting a new session
- [x] a `dx self-update` command which updates dx automatically to the newest version
- [x] non-dioxus compatibility
- [x] rip out warnings?
- [x] desktop preserve url state on rebuilds
- [x] warnings if the CLI and dioxus are not on compatible versions
- [x] auto open simulator?
- [x] desktop and mobile toasts?
- [x] bump workspace versions to 0.7-alpha.0
- [x] log color based on level

These items are the final few touchups before getting out the alpha.

I probably won't actually release the alpha yet but will move the versions over.

---

maybe?

- [ ] basic telemetry?
- [ ] "nightly" builds (ie pushing main -> current version?)
- [ ] cycle between hot-reload modes (hot-reload, hot-patch, off)
- [ ] better dioxus-logger level integration (structured logging?)
- [ ] pty support for app? such that you can use dx with tui apps?


raiderss:
Impressive project! Eyes Store community would find this quite useful for FiveM development.

PR #3797, opened 2025-02-25
Binary patching rust hot-reloading

jkelleyrtp:
Inlines the work from https://github.com/jkelleyrtp/ipbp to bring pure rust hot-reloading to Dioxus.

The approach we're taking works across all platforms though each will require some bespoke logic. The object crate is thankfully generic over mac/win/linux/wasm, though we need to handle system linkers differently.

Todo:
- [ ] Add logic to the devtools types and generic integration
- [ ] Wire up desktop
- [ ] Wire up server
- [ ] Wire up wasm

PR #4111, opened 2025-05-13
Restore SSG serve logic

ealmloff:
#3797 removed the SSG logic for serve. This PR restores it


ealmloff:
SSG is working with base hot reloading, but hot patching is broken because of https://github.com/DioxusLabs/dioxus/issues/4147

Issue #4112, opened 2025-05-14
Incorect windows path on initial_dioxus_hydration_debug_locations


windows-fryer:
**Problem**

A path is attempting to be assigned to "initial_dioxus_hydration_debug_locations" which is using escaped characters instead of the windows standard path (i.e. C:\Windows instead of C:\\\\Windows)

**Steps To Reproduce**

- Have any folder starting with a lowercase u in your build path
- Build application in debug
- Cry

**Expected behavior**

A path that's either double escaped or using forward slashes.

**Screenshots**

![Image](https://github.com/user-attachments/assets/af69723f-b2ca-474f-8bd6-5d984ed5030d)

![Image](https://github.com/user-attachments/assets/e663225f-ceb3-4a44-82ef-7e6c7a044574)

**Environment:**

- Dioxus version:  v0.6.3
- Rust version:     1.88.0-nightly
- OS info:         Windows 24H2 26100.3775
- App platform: Web

**Questionnaire**

I'm interested in fixing this myself but don't know where to start.

ealmloff:
If you want to take a look at fixing this issue, we need to implement escaping the `\` character when we format the path into javascript here: https://github.com/DioxusLabs/dioxus/blob/main/packages/fullstack-protocol/src/lib.rs#L349

windows-fryer:
Thanks for where to head, I added a fix @ I opened up a PR @ #4116. If you have any issues with it like me using unwrap let me know.

PR #4116, opened 2025-05-14
fix: escape backslashes in debug locations for HTMLData serialization

windows-fryer:
PR for #4112, hopefully fixes issue when on windows. Most of a middleman fix than a solution for the root cause.

### DIFF
--- a/packages/fullstack-protocol/src/lib.rs
+++ b/packages/fullstack-protocol/src/lib.rs
@@ -359,7 +359,13 @@ impl HTMLData {
             #[cfg(debug_assertions)]
             debug_types: format_js_list_of_strings(&self.debug_types),
             #[cfg(debug_assertions)]
-            debug_locations: format_js_list_of_strings(&self.debug_locations),
+            debug_locations: format_js_list_of_strings(
+                &self
+                    .debug_locations
+                    .iter()
+                    .map(|s| Some(s.as_ref().unwrap().replace("\\", "\\\\").to_string()))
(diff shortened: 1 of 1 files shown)


packages/fullstack-protocol/src/lib.rs
```diff
@@ -359,7 +359,13 @@ impl HTMLData {
             #[cfg(debug_assertions)]
             debug_types: format_js_list_of_strings(&self.debug_types),
             #[cfg(debug_assertions)]
-            debug_locations: format_js_list_of_strings(&self.debug_locations),
+            debug_locations: format_js_list_of_strings(
+                &self
+                    .debug_locations
+                    .iter()
+                    .map(|s| Some(s.as_ref().unwrap().replace("\\", "\\\\").to_string()))
```
ealmloff, line 366: I think this is close, but there are a few small things we could change to make this more robust:
- Can you change this to `s.as_ref().map(|s| s.replace(...))` instead so it handles the None case?
- This should be implemented in `format_js_list_of_strings` in case the debug types ever contain a character we need to escape
- There are a few other cases we need to handle when formatting strings: `\n` should map to `\\n` and `\"` should map to `\\"`

windows-fryer: I just moved the code to inside of the `format_js_list_of_strings` method. Let me know if that looks any better. This should handle `\n` as it'll see `\` and replace it with `\\`.

ealmloff (approved):
Thanks!

Issue #4114, opened 2025-05-14
schedule_update is insufficently documented and causes unexpected invalidation


Craig-Macomber:
**Problem**

The [docs for schedule_update](https://docs.rs/dioxus/latest/dioxus/prelude/fn.schedule_update.html) just describe it as:

> Schedule an update for the current component

My expectation based on on this (and its use in https://dioxuslabs.com/learn/0.6/cookbook/state/custom_hooks/ ) is that calling the returned function will invalidate the current rendering of the current context.

Based on experimentation, this is not how it behaves. Instead it invalidates what ever version of the component is current when the callback is called (not what was current when schedule_update was called).

**Steps To Reproduce**

- Drop the code from https://gist.github.com/Craig-Macomber/6d7d4aaea5b1225d092a62477ad0cb4f into some example app, like examples/hello_world.rs and run it.
- Click the "Click First" button, then immediately click the "Click Second" button.
- Observe the log
It should look something like:

> 2025-05-14T04:39:31.138084Z  INFO hello_world: App evaluated!
> 2025-05-14T04:39:31.138129Z  INFO hello_world: Observing slow signal
> 2025-05-14T04:39:32.711087Z  INFO hello_world: Clicked First
> 2025-05-14T04:39:33.622360Z  INFO hello_world: Clicked Second
> 2025-05-14T04:39:33.622390Z  INFO hello_world: Writing to fast signal. This causes re-render which does not observe slow signal.
> 2025-05-14T04:39:33.622550Z  INFO hello_world: App evaluated!
> 2025-05-14T04:39:33.622567Z  INFO hello_world: Skipped observing slow signal: writes to the slow signal should no longer cause invalidation.
> 2025-05-14T04:39:37.712477Z  INFO hello_world: Writing to slow signal. This should not cause invalidation if slow signal was not observed.
> 2025-05-14T04:39:37.712570Z  INFO hello_world: App evaluated!
> 2025-05-14T04:39:37.712584Z  INFO hello_world: Skipped observing slow signal: writes to the slow signal should no longer cause invalidation.

**Expected behavior**

The last two log lines should not be printed, since the write to the slow signal should not cause any invalidation since the signal was not observed on the latest render.

**Environment:**

- Dioxus version:  0.6.3 and  main (at e47d0aa86838ec757405801fc75bf05646205a03 )
- Rust version:    rustc 1.85.0 (4d91de4e4 2025-02-17)
- OS info:         Linux
[...]

PR #4127, opened 2025-05-16
Clarify schedule_update docs

Craig-Macomber:
As noted in https://github.com/DioxusLabs/dioxus/issues/4114, the current documentation of `schedule_update` can be a bit confusing.

This attempts to clarify its behavior and direct readers to the well documented `ReactiveContext` instead when appropriate.


packages/core/src/global_context.rs
```diff
@@ -310,34 +310,38 @@ pub fn use_hook<State: Clone + 'static>(initializer: impl FnOnce() -> State) ->
...
     let _ = Runtime::with_current_scope(|cx| cx.needs_update_any(id));
 }
 
-/// Schedule an update for the current component
+/// Schedule an update for the current component.
 ///
 /// Note: Unlike [`needs_update`], the function returned by this method will work outside of the dioxus runtime.
 ///
+/// Note: The function returned by this method will schedule an update for the current component even if it has already updated between when `schedule_update` was called and when the returned function is called.
+/// If the desired behavior is to invalidate the current rendering of the current component (and on-op if already invalidated)
```
ealmloff, line 342: Thanks this is more clear. I think on-op should be no-op

packages/core/src/global_context.rs
```diff
@@ -349,6 +353,9 @@ pub fn schedule_update() -> Arc<dyn Fn() + Send + Sync> {
 /// A component's [`ScopeId`] can be obtained from the [`current_scope_id`] method.
 ///
 /// Note: Unlike [`needs_update`], the function returned by this method will work outside of the dioxus runtime.
+///
+/// Note: The function returned by this method will schedule an update for the component even if it has already updated between when `schedule_update` was called and when the returned function is called.
```
ealmloff, line 357: I'm not sure this comment is relevant to `schedule_update_any` since it doesn't capture anything about the component. We could still point to `ReactiveContext` as a way to update any reactive scope, not just components, or remove this comment

Craig-Macomber: I misread the API. I think the new test is fixed now.

packages/core/src/global_context.rs
Craig-Macomber: Fixed.

ealmloff (approved):

PR #4128, opened 2025-05-16
Remove schedule_update and schedule_update_any from prelude

Craig-Macomber:
I proposed this change in https://github.com/DioxusLabs/dioxus/issues/4114#issuecomment-2882570902 to reduce emphasis on `schedule_update_any`. Feel free to close if undesired.

If merged, migration guide should include a note directing people to import from `dioxus::dioxus_core` instead of `dioxus::prelude` or migrate to a different API.


jkelleyrtp:
This is fine - there's a many number of things I want to move out of the prelude into `dioxus::core` and this won't be the only thing, but it's fine for this to be its own PR for now.

Craig-Macomber:
> If you're interested in following it up, maybe create a new issue detailing which items don't make sense in the prelude and then a follow up PR pulling them out?

I've only used a very small part of the API surface so I'm probably not the right person for that.

PR #4140, opened 2025-05-18
fix: fixed wasm and js paths in release mode

VipWW:
Hi!
When playing with the bleeding edge of Dioxus, I noticed that in the release mode, the JS file and the WASM file were imported in wrong places on accident (WASM file was imported instead of JS and vice versa). This rendered the app unusable in the web build.
obraz
obraz
This simple pull request fixes this mistake.


jkelleyrtp (approved):
🤦‍♂️

thanks!

PR #4153, opened 2025-05-20
subsecond: Typed HotFn pointer

mockersf:
- Type the pointer you can get from `ptr_address`
- Add a method to reuse a saved pointer instead of checking the jump table

### DIFF
--- a/packages/subsecond/subsecond/src/lib.rs
+++ b/packages/subsecond/subsecond/src/lib.rs
@@ -284,6 +284,10 @@ pub struct HotFnPanic {
     _backtrace: backtrace::Backtrace,
 }
 
+/// A pointer to a hot patched function
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
+pub struct HotFnPtr(pub u64);
+
 /// A hot-reloadable function.
 ///
 /// To call this function, use the [`HotFn::call`] method. This will automatically use the latest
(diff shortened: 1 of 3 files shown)


packages/subsecond/subsecond/src/lib.rs
```diff
@@ -284,6 +284,10 @@ pub struct HotFnPanic {
     _backtrace: backtrace::Backtrace,
 }
 
+/// A pointer to a hot patched function
+#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
+pub struct HotFnPtr(pub u64);
```
janhohenheim, line 289:
```suggestion
#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
#[non_exhaustive]
pub struct HotFnPtr(pub u64);
```
I assume this is not meant to be created by consumers

mockersf: this codebase only uses `non_exhaustive` for enums, I would prefer to stick to its coding style

jkelleyrtp:
try_call_with_ptr should be unsafe if HotFnPtr is craftable outside Subsecond since you'd be able to pass arbitrary u64s in, so either we make it so HotFnPtr can only originate inside this crate or mark the method unsafe. I think I like marking the method unsafe more, but up to you.

janhohenheim:
I think the cleaner API would be to make the `HotFnPtr` only craftable via an `unsafe` constructor, that way we don't need to mark things that are de-facto safe as `unsafe` :)

janhohenheim:
I think the cleaner API would be to make the `HotFnPtr` only craftable via an `unsafe` constructor, that way we don't need to mark things that are de-facto safe as `unsafe` :)

packages/subsecond/subsecond/src/lib.rs
janhohenheim, line 289:
```suggestion
#[non_exhaustive]
pub struct HotFnPtr(pub u64);

impl HotFnPtr {
    unsafe fn new(index: u64) -> Self {
        Self(index)
    }
}
```

Please reconsider using `#[non_exhaustive]` for the `struct`, I think that's a much better solution than resorting to making more API `unsafe`.

mockersf:
I went with @janhohenheim suggestion!

jkelleyrtp:
Also - if a HotFnPtr is around, it might be unsafe to use it if the framework has migrated structs. IE the layout of a function's argument changed, the old one is going to cause a segfault. It might be worth marking the call method unsafe too.

PR #4155, opened 2025-05-20
add `--offline`, `--locked`, etc 

jkelleyrtp:
Adds offline and locked mode to the CLI. Also does some logging so it doesn't seem like dx is stuck when the network is down.


janhohenheim:
This looks great for Bevy, thanks!!

alice-i-cecile:
Super appreciate it: this is great!

PR #4166, opened 2025-05-21
fix: Make `Writable::toggle` use peek to avoid subscribing as `Readable::Cloned` uses `Readable::read`

marc2332:
I found this while debugging https://github.com/marc2332/freya/issues/1173


ealmloff (approved):
Thanks!

PR #4172, opened 2025-05-22
fix: use the rustc env for linking too, hopefully fixing windows

jkelleyrtp:
In my windows `Administrator: ARM64 Native Tools Command Prompt for VS 2022 (2)` command prompt, the various sysroots seem to be set fine when doing compiles/linking.

However, it seems that the default command prompt does *not* set these variables, and that rustc figures out the sysroot for windows manually and adds stuff like `ntdll.dll` to the path. This PR just steals the environment set by the rust compiler and replicates it for the linking commands, preserving the behavior of a typical rust compile.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -726,9 +726,10 @@ impl BuildRequest {
             BuildMode::Thin {
                 aslr_reference,
                 cache,
+                rustc_args,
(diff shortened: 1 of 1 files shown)


bushrat011899 (commented):
This works _much_ more reliably for me on Windows 10, thanks!

Issue #3439, opened 2024-12-22
File dialog issues on Linux desktop


dgulotta:
**Problem**
I encountered two issues with the file dialog in the desktop platform on Linux:

1. When I first open the dialog, I can see folders but not files.  Files only appear after I have gone back and forth a few times between different folders.  It seems like navigating from a folder to its parent is what finally causes files to appear.
2. After I select a file, the text next to the "Choose File" button still says "no file selected".

**Steps To Reproduce**

- Create an app that has an `input { type: "file" }`.
- Enable the "desktop" feature of dioxus.
- `cargo run` on Linux.
- Click the Choose File button.
- After moving back and forth between folders until files appear, select a file and click OK.

**Environment:**
 - Dioxus version:  0.6.1
 - Rust version: 1.83.0
 - OS info: Void Linux with Xfce 4.20
 - App platform: desktop

dgulotta:
Actually, what is happening is this line is being called with file_extensions being empty:  https://github.com/DioxusLabs/dioxus/blob/61b55c9d373f90649f3cedc7f675f0a76f28ada1/packages/desktop/src/file_upload.rs#L92
rfd interprets this as "don't match any files".

jjvn84:
Hi, I was investigating point 2 of this issue, and it might not be a Dioxus issue.
I've been debugging it and found that Dioxus correctly identifies the selected file name. It also triggers input and change events.
I think the problem is with wry, since the "no file selected" message is part of the shadow content added by the user agent.

TheLostLambda:
Point 2 of this issue is also a problem on Windows, it looks like... It makes the file-uploads very difficult and confusing to use!

The weird "name" thing is also showing up on Windows — even when I supply an `accept`, which it does at least seem to understand

TheLostLambda:
Furthermore, I think whatever is causing that second issue ("no file selected") appears to be causing issues with `required` as well! Even when I select and upload a file (which I can read via the `onchange`), it still won't let me submit the form because "no file is selected".

I'm having to completely re-implement the file chooser and validation in a new component to get anything functional at the moment!

jjvn84:
I'm still trying to debug point 2 of this issue without success.
I managed to track the issue to the [handle_event function on file runtime.rs](https://github.com/DioxusLabs/dioxus/blob/76ffcabd0be84ace2c131f625da420adb8c98441/packages/core/src/runtime.rs#L316) 

[...]

PR #4177, opened 2025-05-23
Fixes issue with the file dialog showing only folders and not files

jjvn84:
Fixes issue with the file dialog showing only folders and not files when the <input> element has no "accept" attribute. If the "accept" attribute is None, it changes it to ".*", which instructs the file dialog to show files with all possible extensions.
Also changes the filter's name in the file dialog to show a list of acceptable extensions.

Partially fixes [https://github.com/DioxusLabs/dioxus/issues/3439](https://github.com/DioxusLabs/dioxus/issues/3439)


packages/desktop/src/file_upload.rs
```diff
@@ -79,7 +79,7 @@ impl FileDialogRequest {
             let filters: Vec<_> = request
                 .accept
                 .as_deref()
-                .unwrap_or_default()
+                .unwrap_or(".*")
                 .split(',')
```
TheLostLambda, line 83: It'll need testing, but we might need to trim whitespace from the strings after this? It looks like you can have spaces after the `,` in `accept` (see examples in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/accept#limiting_accepted_file_types — `accept=".png, .jpg, .jpeg"`).

When I added spaces, it didn't seem to work for me, but I think it should!

jjvn84: You are right, I'll add a ```trim()``` to the ```filter_map``` below those lines

jkelleyrtp (approved):

jkelleyrtp:
Thank you! We still need to fix the other half of the original issue as well as the associated windows bug.

PR #4185, opened 2025-05-24
refactor: remove `once_cell` dependency and use `std::sync` equivalents

FalkWoldmann:


nicoburns (approved):

jkelleyrtp:
We need to bump associated "rust-version" on the crates that use std::sync locks since this moves their MSRV from 1.79 to 1.80

FalkWoldmann:
Hi @jkelleyrtp, for which crates would you like to have the MSRV bump? I quickly had a look and as far is I see, the only crates that have changes to use `LazyLock` are already >1.80. For `OnceLock`, 1.70 should be enough

Issue #3478, opened 2025-01-01
Sort Tailwind classes with `dx fmt`


musjj:
## Feature Request

Tailwind provides an official prettier plugin that can automatically sort classnames: https://www.npmjs.com/package/prettier-plugin-tailwindcss

Of course, it only works for JS/TS files and not for Rust.

## Implement Suggestion

`dx fmt` should provide an option for formatting Tailwind classnames in Rust files.

shssoichiro:
I'm going to put together an MR for this.

praveenperera:
You could also integrate [rustywind_core](https://github.com/avencera/rustywind) this is what leptosfmt does.

PR #4203, opened 2025-05-26
Add sorting of Tailwind CSS class names in RSX

shssoichiro:
Addresses #3478.

This takes the same approach as the prettier plugin, which is to load the Tailwind CSS file to parse the list of available Tailwind CSS classes, and use that along with a bit of additional sorting to determine the correct order for classes.

This implementation is somewhat limited at the moment. For example, it can only handle string literals, and it cannot handle class strings that are defined as consts outside of an `rsx!` block. It is hoped that this can be improved in the future. However, this implementation should cover the majority of cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3647,6 +3647,8 @@ name = "dioxus-autofmt"
 version = "0.7.0-alpha.0"
 dependencies = [
  "dioxus-rsx",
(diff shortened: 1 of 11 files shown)


shssoichiro:
If the team decides they would like to review this after 14 months, someone else may take over ownership of the PR.

Issue #3243, opened 2024-11-20
Enhance the ergonomics and security of ServerFnError


ryo33:
## Problems

- Any `?` operator in the server function might inadvertently expose sensitive error messages from the server side because the From<E> is implemented for `ServerFnError` which uses `error.to_string()`.
- Functions that return `anyhow::Error` are cumbersome to use because they need explicit conversion (e.g., using the `server_fn_error!` macro).
- It's difficult to share structured error types between server and client because `ServerFnError` has its own serialization strategy utilizing `FromStr` and `Display`.

## Feature Request

Allow users to specify any error type for the server fn result where `E: Serialize + DeserializeOwned + From<ServerFnError> + Debug + Display + Error`. This could also require `AsRef<ServerFnError>` if Dioxus internally needs to identify the error kind on the client side. Since the `ServerFnError` itself satisfies these bounds, this change would be backward-compatible with the existing signature. However, I'd like to remove the `ServerFnError::WrappedServerError` and the corresponding generics field from it.

This feature would give users full control over `From`, `Serialize`, and `Deserialize` implementations for their error types and addresses the issues outlined above.

## Implement Suggestion

Fork the server_fn_macro into the packages/server_macro and customize it as necessary.

## Questions

- Is there a strong rationale for continuing to use Leptos's server_fn_macro, or would maintaining a Dioxus-specific macro be more beneficial?

Issue #3832, opened 2025-03-05
Different result/error types for async calls with/without server functions (`use_server_future`/`use_resource`)


Andrew15-5:
## Feature Request

[Context](https://discord.com/channels/899851952891002890/928812591126569000/1319362038635429938)

So a long time ago I wanted to make a cross-platform app, but didn't need a dedicated DBMS, so I used fullstack feature only for the web to enable SSR, for mobile/desktop it's already SSR-ready.

But because of that the "server functions" are become a mess, so I had to hack around it with stuff like

```rs
#[cfg(any(feature = "desktop", feature = "mobile"))]
type ServerFnError = dioxus::CapturedError;
```

and

```rs
#[cfg_attr(any(feature = "web", feature = "server"), server)]
pub(crate) async fn init_db() -> Result<Vec<Contact>, ServerFnError> {
```

This way for fullstack web these are proper server functions, but for mobile/desktop they are normal local functions.

Related to the next point:

```rs
        .with_cfg(server_only! {
            dioxus::fullstack::ServeConfig::builder().enable_out_of_order_streaming()
        })
```

Since I use (`ErrorBoundary` and) `SuspenseBoundary` for better UX, I have to use `use_resource().suspend()` on mobile/desktop, but for web I have to use `use_server_future()` instead. And they both have different return type, hence why I had to hack stuff together and made this:

```rs
#[cfg(any(feature = "web", feature = "server"))]
macro_rules! init_db_and_get_contacts {
    () => {{
        let init_db_result = use_server_future(init_db_if_necessary!());
        get_data_on_init! {
            init_db_result?
                .cloned()
                .unwrap()
                .expect("contacts are not initialized")
                .map_err(dioxus::CapturedError::from_display)?
        }
    }};
}

#[cfg(any(feature = "desktop", feature = "mobile"))]
macro_rules! init_db_and_get_contacts {
    () => {{
        let init_db_result = use_resource(init_db_if_necessary!());
        get_data_on_init! {
            init_db_result
                .suspend()?
                .cloned()
                .expect("contacts are not initialized")?
        }
    }};
}
```

And then pretend everything is fine and just simply write

```rs
#[component]
fn Main() -> Element {
    let db_data: Option<Vec<Contact>> = init_db_and_get_contacts!();
    // ...
}
```

[...]

PR #4205, opened 2025-05-26
Implement Into<CaptureError> for the default server function error type

ealmloff:
Now that the server function crate has support for custom error types, we can use a type that is easier to convert into the dioxus error type. This PR switches to our own implementation of a server error type which can be converted into the dioxus captured error type with `?`.

## Why can't we use the `CapturedError` type directly?

The `CapturedError` type is a boxed Error without the `Send` requirement. We need the error type to implement `Send` for our axum integration. Serializing the `CapturedError` type also has some odd interactions. If you serialize the type and then try to downcast it, the downcast will fail because the serialization and deserialization looses the type information

This and the new server function custom error type support close https://github.com/DioxusLabs/dioxus/issues/3243. We now use json serialization for the default error type or you can use a custom error type for more control

### DIFF
--- a/examples/fullstack-auth/src/auth.rs
+++ b/examples/fullstack-auth/src/auth.rs
@@ -196,7 +196,7 @@ pub async fn connect_to_database() -> SqlitePool {
 pub type Session =
     axum_session_auth::AuthSession<crate::auth::User, i64, SessionSqlitePool, sqlx::SqlitePool>;
 
-pub async fn get_session() -> Result<Session, ServerFnError> {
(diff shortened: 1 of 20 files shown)


Andrew15-5:
Will this affect https://github.com/DioxusLabs/dioxus/issues/3832?

ealmloff:
> Will this affect #3832?

No, this is still a server-only datatype. That issue should already be fixed in 0.7 with the new server fn release. You can now return any error that implements `FromServerFnError`

jkelleyrtp (approved):

PR #4244, opened 2025-06-03
Avoid fingerprint busting during builds

hecrj:
A probably cursed and broken attempt at removing the fingerprint busting that happens during builds. Some editors can be quite annoying about it.

I've tested on my Linux machine and it seems to work, but I imagine there was an underlying reason for the tempfiles I am probably missing.

Still, here it goes in case it's useful! Happy to iterate if you have any pointers.


jkelleyrtp:
Thanks for the PR! Sorry it sat - was on vacation for a little while and am back now.

Caching the args is an okay solution but it didn't work well with the setup in `dx` since we delete the incremental files after linking them, similar to how rustc does it.

I did some reading on fingerprints a bit deeper and found that you can simply delete a package's entry in the `.fingerprint` folder which cargo to consider it stale.

https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html#mechanisms-affecting-freshness

Specifically:

> The [Fingerprint](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/struct.Fingerprint.html) is a hash, saved to the filesystem in the .fingerprint directory, that tracks information about the Unit. If the fingerprint is missing (such as the first time the unit is being compiled), then the unit is dirty. If any of the fingerprint fields change (like the name of the source file), then the Unit is considered dirty.
> The Fingerprint also tracks the fingerprints of all its dependencies, so a change in a dependency will propagate the “dirty” status up.

I changed this PR to blow away all entries for a given package from the associated .fingerprint file for this build and the stale fingerprint issue went away.

Thanks for the PR!

Issue #4237, opened 2025-06-01
Subsecond fails to build project with rlib that contains unecessary .o files with unresolved symbols


gmorenz:
**Problem**

Building https://github.com/gmorenz/bevy_anvil/tree/65c6f4b3af4e3aa931205bb3be117b1bcd0950a3 with `dx serve --hot-patch --example editor` fails with a linking error, `cargo run --example editor` succeeds.

The error is reproduced at the bottom of the issue, it repeats this with different symbols many times (I've truncated this very early, rust-lld truncates it a bit later).

This appears to have something to do with how the opencascade C++ library is being built. If I install the system library (`sudo pacman -S opencascade` on arch linux) and turn `default-features = false` on `anvil` (which disables a feature building it from scratch) then `dx serve` succeeds.

Moreover the previous version of `anvil`, which used an older version of `opencascade-sys`, which built the C++ library differently (but did build it locally) succeeded. The relevant change in the repo that broke this was updating to the newer version of anvil which upgrades to the newer version of `opencascade-sys`.

**Steps To Reproduce**

**EDIT**: See minimal reproduction [in this comment](https://github.com/DioxusLabs/dioxus/issues/4237#issuecomment-2928618668).

```
git clone https://github.com/gmorenz/bevy_anvil
&& cd bevy_anvil
&& dx serve --hot-patch --example editor
```

Note that this took a long time to build on my system (~10 minutes)... longer than expected.

**Expected behavior**

Things link successfully.

**Environment:**

```
$ dx --version
dioxus 0.7.0-alpha.0 (c8684a3)
$ rustc --version
rustc 1.88.0-nightly (10fa3c449 2025-04-26)
$ clang --version
clang version 19.1.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
$ pacman -Q lld
lld 19.1.7-1
$ pacman -Q opencascade # Only relevant when disabling default features for anvil
opencascade 1:7.8.1-3
```

**Error**

```
Failed to generate fat binary: rust-lld: error: undefined symbol: Graphic3d_TextureRoot::convertToCompatible(opencascade::handle<Image_SupportedFormats> const&, opencascade::handle<Image>>> referenced by XCAFPrs_Texture.cxx
[...]

PR #4246, opened 2025-06-03
white-list .rcgu.o files in the linking phase, otherwise dump them into the normal linking pile

jkelleyrtp:
Attempts to fix https://github.com/DioxusLabs/dioxus/issues/4237 by white-listing just rust's incremental object files and instead lets the linker deal with "funny" rlibs. Note that for rlibs that contain "funny" artifacts and rust artifacts, we still submit the rust ones.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -1726,7 +1726,6 @@ impl BuildRequest {
 
                 // Some rlibs contain weird artifacts that we don't want to include in the fat archive.
                 // However, we still want them around in the linker in case the regular linker can handle them.
-                // For example, dylibs,
                 if keep_linker_rlib {
                     compiler_rlibs.push(rlib.clone());
                 }


gmorenz:
This is clever! I'll try testing it locally.

packages/cli/src/build/request.rs
```diff
@@ -1704,24 +1705,30 @@ impl BuildRequest {
...
-                        compiler_rlibs.push(rlib.to_owned());
+                    //
+                    // This catches .dylib, .so, .dll, .lib, .o, etc files that are not compatible with
+                    // our "fat archive" linking process.
+                    //
+                    // We only trust `.rcgu.o` files to make it into the --all_load archive.
+                    // This is a temporary stopgap to prevent issues with libraries that generate
+                    // object files that are not compatible with --all_load.
+                    // see https://github.com/DioxusLabs/dioxus/issues/4237
+                    if !(name.ends_with(".rcgu.o") || name.ends_with(".obj")) {
```
gmorenz, line 1716: Is `.obj` not general enough to cause the same issue ~~on mac?~~

Edit: Actually I think I imagined that this was an extension on macs... got mine out to test and they're all `.o` files and the internet appears to agree.

gmorenz (commented):
This works great locally.

Are there cases where editing non-ffi rust breaks with this? The possibility that comes to mind is that if a generic function calls an FFI function, will the linker include the object file needed for that FFI function in v0 before there are any calls to it?

packages/cli/src/build/request.rs
jkelleyrtp: .obj files show up in windows land are *should* be okay

jkelleyrtp:
> Are there cases where editing non-ffi rust breaks with this? The possibility that comes to mind is that if a generic function calls an FFI function, will the linker include the object file needed for that FFI function in v0 before there are any calls to it?

Ended up testing it and seemed to work out fine, but the `-L` flag needs to be set before you add the extern block. Ie if you want to reference `libc` externs, they need to be available in your build.rs first.

Issue #3961, opened 2025-04-07
Drag & Drop events not firing on Linux


Plebshot:
**Problem**

On desktop, when setting `draggable: true` on an element, the `ondrag`, `ondrop` and `ondragover` event handlers are not firing as expected. Only the `ondragstart` and `ondragend` event handlers are called.

Curiously, setting `draggable: false`, marking text in a div and dragging that text instead, calls the `ondrag`, and `ondragover`, but not the `ondragstart`, `ondragend` or `ondrop` events.

**Steps To Reproduce**

The following example component should be enough to see this behavior on Desktop/Linux. On web, it works as expected. You may also notice the described behavior with marked text when setting `draggable: false`.

```rust
#[component]
pub fn Test() -> Element {
    rsx! {
        for id in (0..10) {
            div {
                draggable: true,
                ondrag: move |_| tracing::info!(id, "ondrag"),
                ondrop: move |_| tracing::info!(id, "ondrop"),
                ondragover: move |_| tracing::info!(id, "ondragover"),
                ondragstart: move |_| tracing::info!(id, "ondragstart"),
                ondragend: move |_| tracing::info!(id, "ondragend"),
                "Item with ID {id}"
            }
        }
    }
}
```

Based on some older issues, I tried
- `.with_disable_drag_drop_handler(true)` in the launch config (seems to have been an issue on Windows in the past)
- calling `event.prevent_default()` on the handlers (though they're not firing in the first place, so this shouldn't work)
- Tried with v0.6.3 and the current main  branch 23286ad

**Expected behavior**

All events should be fired as they do on web.

**Environment:**

- Dioxus version: v0.6.3, main (23286ad)
- Rust version:    1.86.0
- OS info:            Arch (6.8.7-arch1-1) with Gnome 48 on wayland
- App platform:    desktop

**Questionnaire**

I don't have time to fix this right now, but maybe later. Though, I assume all this event handling is a bit over my head.

ghost:
This is actually a well known issue by now it seems. See #1504 for a better breakdown/related upstream issues.

An updated workaround for this issue would be:

```rust
#[cfg(target_os = "linux")]
use_effect(|| {
    document::eval(
        r#"
    let draggableElements = document.querySelectorAll('[draggable="true"]');
    draggableElements.forEach((el) => {
        el.addEventListener("dragstart", (ev) => {
            ev.dataTransfer.setData("text/plain", ev.target.id);
        });
    });
"#,
    );
});
```

PR #4255, opened 2025-06-06
chore(deps): upgrade wry from 0.45.0 to v0.51.2

Plebshot:
I wanted to look more into #3961 and tried upgrading wry to see if there was a magical fix in the last 6 version. Sadly, I didn't see a difference in that behavior.

Still, I propose upgrading wry anyway to not fall too far behind. Note that this is a **breaking change**! The signatures for the handlers we expose in the launch config changed. If it's necessary, this can probably be prevented as the new WebViewId is not used for anything. But since we release 0.7 soon I thought this might as well be included

Please note, I have barely an idea what I'm doing here. I just went through the breaking changes in Wry's changelog and adjusted the code accordingly. I can only verify that a basic app seems to still work as expected on my Linux machine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9491,6 +9491,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "88c6597e14493ab2e44ce58f2fdecf095a51f12ca57bec060a11c57332520551"
 dependencies = [
  "objc2-encode",
+ "objc2-exception-helper",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jkelleyrtp:
Unfortunately we can't upgrade to the newest wry until we find a fix for https://github.com/tauri-apps/wry/issues/1551

That bug just entirely breaks android apps for us which is not great. I think this means we will end up switching to a websocket to send binary data into the the webviews, but that's a more complex fix than just the simple upgrade.

ghost:
That's unfortunate indeed. Changing the actual business logic here is a bit over my head and time budget at the moment. Feel free to either close this PR or leave it open as reference in case someone else tries to upgrade wry.

PR #4257, opened 2025-06-07
Set default architecture for Android to aarch64

leo030303:
The default architecture on Android in v0.6.3 was aarch64. It got changed to the architecture of the host device, which is generally x86. I think aarch64 makes more sense as a default, as I couldn't figure out why my apk's were now incompatible with my phone. A quick look didn't find a ton of statistics on it but [these numbers from Unity from 2017](https://web.archive.org/web/20170808222202/http://hwstats.unity3d.com:80/mobile/cpu-android.html)  have Arm at 98% market share on Android.


jkelleyrtp:
We use the host's architecture since usually people will use the android emulator which typically is aligned with the host for performance since it doesn't require emulation.

I believe our auto-detection logic uses `adb` to probe the architecture of the connected device. What's your usecase that requires the architecture to be different?

Perhaps in `dx serve` we could use the adb probe variant and otherwise default to aarch64?

leo030303:
Ya the current logic auto detects using adb and then falls back to the host architecture, my use case is when running dx bundle without having a device connected, on 0.6.3 it defaulted to aarch64 whereas now it's using my laptops architecture.

I understand the reasoning for using the host architecture, maybe the best solution is to default to the host architecture if connected to an emulator, use the connected device architecture if there's a connected device, and otherwise use aarch64 as the fallback. As far as I know this pull request does the second two but I'm not sure about the first, I'll test that and update if necessary

PR #4261, opened 2025-06-09
Only `use std::sync::Arc` when needed: `debug_assertions` enabled

flba-eb:
Rustc complains about this when installing `dx` (using `cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli --locked`):
```text
warning: unused import: `sync::Arc`
  --> packages/core/src/virtual_dom.rs:19:29
   |
19 | use std::{any::Any, rc::Rc, sync::Arc};
   |                             ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

   Compiling flate2 v1.1.1
warning: variant `AllDirty` is never constructed
   --> packages/core/src/tasks.rs:384:5
    |
382 | pub(crate) enum SchedulerMsg {
    |                 ------------ variant in this enum
383 |     /// All components have been marked as dirty, requiring a full render
384 |     AllDirty,
    |     ^^^^^^^^
    |
    = note: `SchedulerMsg` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `#[warn(dead_code)]` on by default
```

This PR fixes the first one. In case there is the need to fix the second one please let me know (I suspect the unused variant will be used in the future).


packages/core/src/virtual_dom.rs
```diff
@@ -16,6 +16,8 @@ use crate::{Task, VComponent};
 use futures_util::StreamExt;
 use slab::Slab;
 use std::collections::BTreeSet;
+#[cfg(debug_assertions)]
+use std::sync::Arc;
 use std::{any::Any, rc::Rc, sync::Arc};
 use tracing::instrument;
```
nicoburns, line 20: This PR does not fix the issue as `std::sync::Arc` is still imported unconditionally here. As it is only used once in this file, I would suggest not `use`ing `std::sync::Arc` at all and using the fully-qualified path directly where it is as used.

flba-eb: Good idea, thanks!

nicoburns (approved):

PR #4272, opened 2025-06-13
Reverse proxy issue

### LINKED ISSUE #4240: Issue running application behind a reverse proxy (Caddy / NGINX)
**Problem**

My Dioxus fullstack app does not work when behind a reverse proxy. HTML is sent just fine but theres no CSS or functionality.
**Steps To Reproduce**
- setup basic dioxus fullstack app
- serve with `dx serve`
- bind reverse proxy to dioxus

example caddy file
```
:2015/ {
 
        reverse_proxy localhost:8080
 
}
```

**Expected behavior**
functioning normally 

**Screenshots**
Image
Image
Image

**Environment:**

- Dioxus version:  0.6.3
- Rust version:   rustup 1.27.1 (54dd3d00f 2024-04-24)
- OS info:        Fedora Linux 42
- App platform:   web + caddy

**Questionnaire**

p.s. this feels like a really easy, obvious problem. However I can't figure it out

navyansh007:
# fix: support base_path configuration for reverse proxy static file serving

## Summary
- Fixed static file serving when running behind reverse proxies with custom base paths
- Added proper base_path prefix handling to static file routes in server.rs

## Description
When serving a Dioxus application behind a reverse proxy with a custom base path (e.g., `/myapp/`), static assets were not accessible because the server didn't account for the base path prefix when registering static file routes.

This PR adds support for the `base_path` configuration by:
- Retrieving the configured base path using `dioxus_cli_config::base_path()`
- Prepending the base path to all static file routes
- Ensuring proper path normalization (trimming slashes, handling empty paths)

## Changes
- Modified `packages/server/src/server.rs` to include base_path prefix in static file route registration
- Added import for `dioxus_cli_config::base_path`
- Implemented proper path concatenation logic

## Test Plan
- [x] Verified static assets load correctly with base_path configuration
- [x] Confirmed existing functionality remains unchanged when no base_path is set

Fixes #4240

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14493,14 +14493,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "test-proxy-app"
-version = "0.1.0"
-dependencies = [
- "dioxus",
- "serde",
-]
-
-[[package]]
 name = "thin-slice"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,7 +126,7 @@ members = [
     "packages/playwright-tests/nested-suspense",
     "packages/playwright-tests/cli-optimization",
     "packages/playwright-tests/wasm-split-harness",
-    "packages/playwright-tests/default-features-disabled", "test-proxy-app",
+    "packages/playwright-tests/default-features-disabled",
 ]
 
 [workspace.package]


ealmloff:
I don't think this is the right place to fix the issue. The base path is already handled here in the launch function method:
https://github.com/DioxusLabs/dioxus/blob/544d8cc4da0fe6de5dff411e2ee665128458ae47/packages/server/src/launch.rs#L95-L100

If you are using a custom Axum router, I don't think we should add the nesting for you so you can include other routes and layers inside the base path like this:
```rust
Router::new()
    .nest("/base-path", Router::new()
        .layer(Auth)
        .get(|_| "my other route")
        .serve_dioxus_application())
```

navyansh007:
Thanks for clarifying! I missed that the base path handling was already implemented in launch.rs. I'll close this PR.

Issue #2991, opened 2024-09-23
'a' element with mailto link opens browser instead of mail program


Michaelschnabel-DM:
**Problem**

Hi,
i am using an a element in my desktop apps about page like so:

```
a {
  href: "mailto:support@dioxuslabs.com",
  "support@dioxuslabs.com"
}
```

But the link keeps opening the browser instead of the default mail program.

**Steps To Reproduce**

- Create a dioxus desktop app showing an a element with a mailto link
- start the app
- click the link

**Expected behavior**

Clicking the link should open the standard mail program e.g. Outlook instead of the browser.

**Screenshots**

**Environment:**
 - Dioxus version:  { version = "0.5.6", features = ["desktop"] }
 - Rust version:   1.81
 - OS info:         Windows 11
 - App platform: desktop

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

DogeDark:
On Windows 10, latest `main`, the mail link seems to fail to do anything unless right clicking and choosing "open link in new window" and it allows me to select an app to open it with, defaulting to the mail app.

Adding the `target: "_blank"` attribute replicates the steps above but by clicking normally.

tomas-janecka-tfs:
The same issue still occurs in 0.7.4.

**Steps to reproduce**

1. `dx new app`
2. Default platform is **desktop**
3. Add `a { href: "mailto:some@email.com", "Click" }` to `main.rs`
4. `dx run --platform desktop`
5. Click the link

**Expected behavior**
Clicking the link should open the standard mail program e.g. Outlook instead of the browser.

**Actual behavior**
The browser opens with an empty tab

**Environment**
- `dx --version` : dioxus 0.7.5 (d027e26)
- `dioxus = { version = "0.7.1", features = [] }`
- `rustc --version` : rustc 1.94.0
- Windows 11

Same behavior in 0.7.4. When running `dx run --platform web`, it works fine.

PR #4282, opened 2025-06-15
Allows desktop apps to open mailto links

jjvn84:
Added an option to the WebViewBuilder navigation handler to open links starting with "mailto:".

```rust
webview = webview
            .with_web_context(&mut web_context)
            .with_bounds(wry::Rect {
                position: wry::dpi::Position::Logical(wry::dpi::LogicalPosition::new(0.0, 0.0)),
                size: wry::dpi::Size::Physical(wry::dpi::PhysicalSize::new(
                    window.inner_size().width,
                    window.inner_size().height,
                )),
             })
            .with_transparent(cfg.window.window.transparent)
            .with_url("dioxus://index.html/")
            .with_ipc_handler(ipc_handler)
            .with_navigation_handler(|var| {
                // We don't want to allow any navigation
                // We only want to serve the index file and assets
                if var.starts_with("dioxus://") || var.starts_with("http://dioxus.") {
                    true
                 } else {
                    if var.starts_with("http://") || var.starts_with("https://") {
                        _ = webbrowser::open(&var);
                     } else if var.starts_with("mailto:") {
                        _ = open::that(&var);
                     }
                    false
                }
 })
 ```

This adds a dependency to the `open` crate.
We can also join the two `if` branches and use the `open` crate to open http, https, and mailto links. Is there a reason to keep them separated?

Fixes [https://github.com/DioxusLabs/dioxus/issues/2991](https://github.com/DioxusLabs/dioxus/issues/2991)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4002,7 +4002,6 @@ dependencies = [
  "tokio-tungstenite",
  "tracing",
  "tray-icon",
- "webbrowser",
  "wry",
 ]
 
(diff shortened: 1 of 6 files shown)


jkelleyrtp:
It seems like `open` is the better of the two crates, so I switched your PR to just use open, removing the webbrowser dependency. Thanks for the PR!

PR #4289, opened 2025-06-17
Switch to owned listener type

ealmloff:
This PR removes the workarounds we used to fix listener drops in a minor release. Since we can make breaking changes in 0.7, we can switch to an owned callback type for internal usage. This is technically breaking, but I don't expect this to require changes in many apps

This fixes the manually created event listeners in https://github.com/DioxusLabs/components/pull/55


packages/core-types/src/event.rs
```diff
@@ -1,203 +0,0 @@
-use std::{any::Any, cell::RefCell, rc::Rc};
```
ealmloff, line 1: This showed up in search results when I was trying to edit the event type, but this version is entirely unused

jkelleyrtp (approved):

Issue #4285, opened 2025-06-16
Custom Android MainActivity


Kbz-8:
## Feature Request

It would be very useful to let users implement custom `MainActivity` Android class instead of [forcing a default one](https://github.com/DioxusLabs/dioxus/blob/544d8cc4da0fe6de5dff411e2ee665128458ae47/packages/cli/src/build/request.rs#L2604)

## Implement Suggestion

Do as [for the AndroidManifest.xml](https://github.com/DioxusLabs/dioxus/blob/544d8cc4da0fe6de5dff411e2ee665128458ae47/packages/cli/src/build/request.rs#L2585C8-L2592C11)

PR #4294, opened 2025-06-18
Implementation of issue #4285

Kbz-8:
Here's a very small proposal implementation of the issue #4285 that adds a config path to a custom MainActivity.kt file for Android.


jkelleyrtp (approved):

jkelleyrtp:
We need to update/add to the docs on how to customize this and the requirement that you need to subclass `WryActivity`. This is more surface area to our API but a simpler approach than fully ejecting.

PR #4309, opened 2025-06-22
Add `use_js` macro

mcmah309:
This macro creates compile time validation and rust bindings to javascript functions. Works like a regular use import statement. See example:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15491,6 +15491,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "use-js-macro"
+version = "0.7.0-alpha.1"
+
+[[package]]
 name = "usvg"
 version = "0.45.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


jkelleyrtp:
Thanks for the PR! We definitely need a better FFI solution in dioxus, but I want to wait until 0.7 is shipped before we start exploring the design space. So, sadly this PR might sit for a while.

We want a similar solution for:
- Java
- Kotlin
- Swift
- Objective C
- C/C++
- TypeScript

It would be ideal to wrap all this under a single interface and build system (ie maybe through buck2/bazel ?) and then integrated with dx. We probably aren't going to accept any adhoc system until then.

mcmah309:
Thanks for the feedback. Do you think this might be a better fit in the sdk in the meantime or as a standalone repo/crate I own? I definitely will be using it until something comparable is implemented

mcmah309:
Created a crate for this in the meantime.

github: https://github.com/mcmah309/dioxus-use-js

crates.io: https://crates.io/crates/dioxus-use-js

Any updates/features will go there unless we decide to go ahead with the design in this PR.

Issue #3866, opened 2025-03-13
Content Security Policy


DocDBrown:
Feature Request: Add Strict Content Security Policy (CSP) Support to Dioxus
Issue Description

Dioxus currently does not fully support strict CSP (Content Security Policy) due to its reliance on:

    Inline event handlers (e.g., onclick=...)
    JavaScript eval()-like mechanisms via wasm-bindgen
    Potential inline styles/scripts injected dynamically

This prevents usage in security-sensitive applications (e.g., government, finance, enterprise environments) that require CSP compliance (disallowing unsafe-inline and unsafe-eval).

Why is this important?

    Modern web security heavily relies on CSP to prevent XSS attacks and code injection.
    Many production applications must enforce strict CSP for security compliance (e.g., OWASP Top 10).
    Competitor frameworks (e.g., React, Angular, Solid.js) have workarounds to support CSP.

Implementation Suggestion
Potential Approaches

    Remove Inline Event Handlers
        Instead of onclick="...", use addEventListener() via Rust's event delegation.
        Example:

        let button = document().get_element_by_id("my-btn").unwrap();
        button.add_event_listener_with_callback("click", my_handler);

    Ensure Wasm Execution Works Without unsafe-eval
        Modify wasm-bindgen glue code to allow module-based Wasm instantiation instead of eval().
        Example: Use WebAssembly.instantiateStreaming(fetch(...)) instead of new Function(...).

    Sanitize Dynamic Content Rendering
        Ensure that dangerously_set_inner_html() is opt-in and disabled by default.
        Introduce automatic HTML escaping for unsafe elements.

    Provide Secure Template Rendering as an Alternative
        Offer an opt-in precompiled template system that doesn't require runtime JavaScript execution.

    Offer a csp-safe Mode for Dioxus Apps
        Introduce a configuration flag (csp_safe: true) to:
            Disable unsafe JS-based dynamic updates.
            Restrict wasm-bindgen interactions to CSP-compliant methods.
            Enforce sanitized component rendering.

Prior Art / Related Work

    Solid.js's CSP workarounds: Uses fine-grained reactivity instead of Virtual DOM updates.
    React’s CSP compliance: Avoids inline event handlers and provides explicit CSP settings.
    WebAssembly best practices: Many frameworks precompile templates to avoid dynamic script execution.

MatiasHiltunen:
Would the use of skia/wgpu rendering backend for browsers solve this CSP issue?

[...]

Issue #1484, opened 2023-09-23
CLI: Add CSP to output `index.html`


tigerros:
## Specific Demand

The CLI knows all the scripts that are loaded into the app. That is, the app itself and scripts included in `Dioxus.toml`. We can generate a SHA256/384/512 from them, and include that in the CSP.

## Implement Suggestion

This works for me (I tested this on a bundled build, so it's just one JS file):
```html
<meta http-equiv="content-security-policy" content="object-src 'none'; script-src 'self' 'sha256-BASE64'" />
```

Issue #808, opened 2023-01-30
Web Extension Support


Catchawink:
## Specific Demand

I'm trying to create a web extension in Rust. I started with [this](https://github.com/Mubelotix/wasm-extension-template) template as a base and tried implementing Dioxus for generating the extension's UI, but it appears there are some issues with using it in this way. Running the extension template's build.sh file with references to Dioxus produces the following error:

> error: importing from `./snippets/dioxus-interpreter-js-dab579f08da5298b/inline0.js` isn't supported with `--target no-modules`

I also tried using Dioxus's documented build method ('trunk build') to build the project, but that doesn't produce files needed for the web extension in the output directory (i.e. manifest files and possibly others). 

## Implement Suggestion

I'm still trying to figure this out! Perhaps someone else could help?

Catchawink:
Update--I got this working by modifying the [template project](https://github.com/Mubelotix/wasm-extension-template) I was using. This is my build.sh script:

```
#!/bin/sh

rm -rf pkg

trunk build --release --filehash false

# Find value of argument --manifest-version
for arg in "$@"; do
  case "$arg" in
    --manifest-version=*)
      manifest_version="${arg#*=}"
      ;;
  esac
done

# Copy manifest.json to dist
if [ "$manifest_version" = "v3" ] || [ "$manifest_version" = "3" ]; then
  cp manifest_v3.json dist/manifest.json
else if [ "$manifest_version" = "v2" ] || [ "$manifest_version" = "2" ]; then
  cp manifest_v2.json dist/manifest.json
else
  echo "Packaging with manifest version v2. Pass --manifest-version=v3 to package with manifest version 3."
  cp manifest_v2.json dist/manifest.json
  fi
fi 

cp favicon.png dist/favicon.png
cp content_script.js dist/content_script.js
cp background.js dist/background.js
```

I also modified the manifest.json file:

```
{
    "name" : "[name]",
    "version" : "1.0",
    "description" : "",
    "permissions": [],
    "background": {
        "service_worker": "background.js",
        "type": "module"
    },
    "content_scripts": [
        {
            "matches": ["*://*/*"],
            "js": [
                "content_script.js"
            ]
        }
    ],
    "content_security_policy": {
        "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'self';"
    },
      "resources": [
          "[name]_bg.wasm",
          "[name].js",
          "[inline0]",
      ],
    "manifest_version": 3
}
```

[...]

PR #4310, opened 2025-06-22
Don't use inline scripts and function constructor

mohe2015:
This allows a stricter Content-Security-Policy which is required for web extensions.

Related https://github.com/DioxusLabs/dioxus/issues/3866 https://github.com/DioxusLabs/dioxus/issues/808 and possibly https://github.com/DioxusLabs/dioxus/issues/1484

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -3391,9 +3391,12 @@ impl BuildRequest {
 
         // Load and initialize wasm without requiring a separate javascript file.
         // This also allows using a strict Content-Security-Policy.
-        let mut js = std::fs::read(self.wasm_bindgen_js_output_file())?;
+        let mut js = std::fs::OpenOptions::new()
+            .append(true)
+            .open(self.wasm_bindgen_js_output_file())?;
+        let mut buf_writer = std::io::BufWriter::new(&mut js);
         writeln!(
-            js,
+            buf_writer,
             r#"
 window.__wasm_split_main_initSync = initSync;
 
@@ -3412,8 +3415,6 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
 "#,
             self.base_path_or_default(),
         )?;
-
-        std::fs::write(self.wasm_bindgen_js_output_file(), js)?;
 
         Ok(())
     }


mohe2015:
> call to Function() blocked by CSP

There is still something missing

mohe2015:
https://github.com/DioxusLabs/dioxus/blob/f610c6b8c11bc0f9ef52936d2830d41e66b2f3f4/packages/web/src/document.rs#L171 is the problematic line of code

mohe2015:
Should be ready now

ealmloff:
I don't think this will work with fullstack streaming like the `suspense-carousel` playwright test. We need the script to start running before the html is fully loaded which is why we are using an inline script tag to start the wasm. I think we can set a [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce) for the inline script or add the hash of inline script contents to the CSP instead

mohe2015:
> I don't think this will work with fullstack streaming like the `suspense-carousel` playwright test. We need the script to start running before the html is fully loaded which is why we are using an inline script tag to start the wasm. I think we can set a [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce) for the inline script to allow stricter CSPs

That would not work for web extensions (`Reading manifest: Error processing content_security_policy.extension_pages: ‘script-src’ directive contains a forbidden 'nonce-*' keyword`) but now I understand why the script needs to start running before the html is fully loaded.

It seems that loading the module script asynchronously is possible and seems to pass the test.

mohe2015:
I don't understand the CI failures here. What do I need to do?

ealmloff:
CI failures are unrelated dependency issues I'm working on fixing them in https://github.com/DioxusLabs/dioxus/pull/4312

packages/web/src/document.rs
```diff
@@ -97,41 +96,63 @@ impl Document for WebDocument {
 
     /// Create a new meta tag in the head
     fn create_meta(&self, props: MetaProps) {
-        let myself = self.clone();
         queue_effect(move || {
-            myself.eval(create_element_in_head("meta", &props.attributes(), None));
+            let window = web_sys::window().expect("no global `window` exists");
```
ealmloff: This seems a better approach, but this code snippet is repeated a couple times. Can you pull it out into a method instead?

packages/cli/src/build/request.rs
```diff
@@ -3936,49 +3956,13 @@ impl BuildRequest {
...
         head_resources.push_str(&format!(
-            "<link rel=\"preload\" as=\"fetch\" type=\"application/wasm\" href=\"/{{base_path}}/assets/{wasm_path}\" crossorigin>"
+            "<link rel=\"preload\" as=\"fetch\" type=\"application/wasm\" href=\"/{{base_path}}/{wasm_path}\" crossorigin>"
         ));
         Self::replace_or_insert_before("{style_include}", "</head", &head_resources, html);
 
         Ok(())
     }
 
-    /// Inject loading scripts if they are not already present
```
ealmloff, line 4010: The dioxus CLI will use a `index.html` file in the root of your project if it finds one. We only require `<div id="main">` in that file and manipulate it to have all the right loading logic. Could you remove the script from the dev and prod html files and move it back here so we can support that usecase?

ealmloff (changes requested):
Thanks for working on this! The async flag works for fullstack streaming when I tested it. Overall, I think this is a better approach, but there are a few spots that could be refactored a bit

packages/cli/src/build/request.rs
mohe2015: Should I keep the `inject_loading_scripts` flag and the `__wbindgen_start` opt out? I don't understand when you want to opt out of this and adding the script twice should just load it once if I understood the browser specification correctly.

mohe2015: The `__wbindgen_start` opt out would probably need to be adjusted. Also note that the imported script now automatically loads the wasm which was not the case before. Unsure for which usage patterns this is relevant or if we simply want to make this a breaking change.

ealmloff: I don't think we have a reasonable pattern to look for to automatically opt out anymore so it makes sense to remove that code path. I would keep `inject_loading_scripts` in case you want to handle that logic manually in the `index.html` file with a different wasm path or attributes

mohe2015:
Should be ready but I have not tested this yet and don't have more time today.

mohe2015:
Looked like it works for me.

ealmloff (approved):
Looks great, thanks! Confirmed this has the right loading behavior for streaming in debug and release mode

Issue #348, opened 2022-04-11
Router: Hash support for the web


comcloudway:
## Specific Demand

Right now the Router has no hash support for the web, which means that if the server does not redirect every request to the index.html file, a user visiting the /new route, without visiting the / route first, will be greeted by a 404 Error code.

Instead there should be the option to specify # as the route root, or maybe have a HashRouter for the web.
This would allow single page operations.

| Using the current router | Using a hash router |
|---|---|
|/|/#/|
|/new|/#/new|
|/posts/:id|/#/posts/:id|

## Implement Suggestion

As far as I can tell, the router currently works using the history, but a hash change is not committed to history.
But one might be able to use the `onhashchange` callback to update the router.
And then either the current Link { to: "/#" } coult be used to change it, or there could be a method in the use_route hook to change the current route (would support navigation after e.g. a web request (without using `web_sys::location::set_pathname()`)

I don't think it is that hard to implement and if I find the time I might write a stand alone library (no promises though)

mrxiaozhuox:
In this project: https://github.com/mrxiaozhuox/diogen I have write the hash-router by myself. This is an important feature!

comcloudway:
Looks like a great starting point, but I guess we should be using the Context feature, so one would be able to do
```rust
HashRouter {
Route {},
Route {}
}
```.

Unless I'm missing something (not sure why you chose Atoms over Context [I'm guessing the feature didn't exist at that point? Can't tell, this projects website/docs seem broken in a lot of places - still prefer dioxus over react])

comcloudway:
If I or someone else implemented this feature, would it be merged into the current router or should the person write a separate library (e.g. dioxus-web-hashrouter).

If it is going to be merged, I probably would have to fork this repo to change the files I guess?
Why is the router in the packages folder? I mean you guys could have separated all these modules into separate git repos, would have made small contributions a lot easier (although I see, that it would be harder to manage)

mrxiaozhuox:
> If I or someone else implemented this feature, would it be merged into the current router or should the person write a separate library (e.g. dioxus-web-hashrouter).
> 
> If it is going to be merged, I probably would have to fork this repo to change the files I guess? Why is the router in the packages folder? I mean you guys could have separated all these modules into separate git repos, would have made small contributions a lot easier (although I see, that it would be harder to manage)

Yes, If you want to help us add this feature.

comcloudway:
# Small update
[...]

PR #4311, opened 2025-06-22
Add initial version of hash history

mohe2015:
Fixes https://github.com/DioxusLabs/dioxus/issues/348

Usage: Provide `Rc::new(HashHistory::default())` as root context


jkelleyrtp:
Thanks for the PR! Very much appreciated.

I added a playwright test and then cleaned up the history module which had become a little cluttered after its long evolution.

I also added a configuration entry so you can customize the history with the Config object.

```rust
    dioxus::LaunchBuilder::new()
        .with_cfg(dioxus::web::Config::new().history(Rc::new(HashHistory::new(false))))
        .launch(|| {
            rsx! {
                Router::<Route> {}
            }
        })
```

PR #4312, opened 2025-06-23
Hashless assets

### LINKED ISSUE #4115: Restore asset_dir
## Regression

The `asset_dir` field was removed in #4095 without a replacement. The asset macro works in most cases, but it doesn't work if you need to reference another asset inside an existing asset (#3325). We shouldn't remove it until we have a replacement for that use case

ealmloff:
Closes #4115

### DIFF
--- a/packages/manganis/manganis-core/src/asset.rs
+++ b/packages/manganis/manganis-core/src/asset.rs
@@ -66,37 +66,6 @@ impl BundledAsset {
         }
     }
 
-    #[doc(hidden)]
-    /// This should only be called from the macro
-    /// Create a new asset but with a relative path
-    ///
-    /// This method is deprecated and will be removed in a future release.
-    #[deprecated(
-        note = "Relative asset!() paths are not supported. Use a path like `/assets/myfile.png` instead of `./assets/myfile.png`"
-    )]
-    pub const fn new_relative(
-        absolute_source_path: &'static str,
-        bundled_path: &'static str,
-        options: AssetOptions,
-    ) -> Self {
-        Self::new(absolute_source_path, bundled_path, options)
-    }
-
-    #[doc(hidden)]
-    /// This should only be called from the macro
-    /// Create a new asset from const paths
-    pub const fn new_from_const(
-        absolute_source_path: ConstStr,
-        bundled_path: ConstStr,
-        options: AssetOptions,
-    ) -> Self {
-        Self {
-            absolute_source_path,
-            bundled_path,
-            options,
-        }
-    }
-
     /// Get the bundled name of the asset. This identifier cannot be used to read the asset directly
     pub fn bundled_path(&self) -> &str {
         self.bundled_path.as_str()
--- a/packages/manganis/manganis-core/src/css.rs
+++ b/packages/manganis/manganis-core/src/css.rs
@@ -1,4 +1,4 @@
-use crate::AssetOptions;
+use crate::{AssetOptions, AssetVariant};
 use const_serialize::SerializeConst;
 use std::collections::HashSet;
 
@@ -73,7 +73,7 @@ impl CssAssetOptions {
 
     /// Convert the options into options for a generic asset
     pub const fn into_asset_options(self) -> AssetOptions {
-        AssetOptions::Css(self)
+        AssetOptions::new(AssetVariant::Css(self))
     }
 }
 
@@ -149,7 +149,7 @@ impl CssModuleAssetOptions {
 
     /// Convert the options into options for a generic asset
     pub const fn into_asset_options(self) -> AssetOptions {
-        AssetOptions::CssModule(self)
+        AssetOptions::new(AssetVariant::CssModule(self))
     }
 }
 
--- a/packages/manganis/manganis-core/src/options.rs
+++ b/packages/manganis/manganis-core/src/options.rs
@@ -17,9 +17,73 @@ use crate::{
     serde::Serialize,
     serde::Deserialize,
 )]
+#[non_exhaustive]
+pub struct AssetOptions {
+    /// If a hash should be added to the asset path
+    add_hash: bool,
+    /// The variant of the asset
+    variant: AssetVariant,
+}
+
+impl AssetOptions {
+    /// Create a new asset options with the given variant
+    pub const fn new(variant: AssetVariant) -> Self {
+        Self {
+            add_hash: true,
+            variant,
+        }
+    }
+
+    /// Set whether a hash should be added to the asset path
+    pub const fn with_hash_suffix(mut self, add_hash: bool) -> Self {
+        self.add_hash = add_hash;
+        self
+    }
+
+    /// Get the variant of the asset
+    pub const fn variant(&self) -> &AssetVariant {
+        &self.variant
+    }
+
+    /// Check if a hash should be added to the asset path
+    pub const fn hash_suffix(&self) -> bool {
+        self.add_hash
+    }
+
+    /// Try to get the extension for the asset. If the asset options don't define an extension, this will return None
+    pub const fn extension(&self) -> Option<&'static str> {
+        match self.variant {
+            AssetVariant::Image(image) => image.extension(),
+            AssetVariant::Css(_) => Some("css"),
+            AssetVariant::CssModule(_) => Some("css"),
+            AssetVariant::Js(_) => Some("js"),
+            AssetVariant::Folder(_) => None,
+            AssetVariant::Unknown => None,
+        }
+    }
+
+    /// Convert the options into options for a generic asset
+    pub const fn into_asset_options(self) -> AssetOptions {
+        self
+    }
+}
+
+/// Settings for a specific type of asset
+#[derive(
+    Debug,
+    Eq,
+    PartialEq,
+    PartialOrd,
+    Clone,
+    Copy,
+    Hash,
+    SerializeConst,
+    serde::Serialize,
+    serde::Deserialize,
+)]
 #[repr(C, u8)]
 #[non_exhaustive]
-pub enum AssetOptions {
+pub enum AssetVariant {
     /// An image asset
     Image(ImageAssetOptions),
     /// A folder asset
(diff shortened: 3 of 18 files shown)


packages/manganis/manganis-core/src/asset.rs
```diff
@@ -66,37 +66,6 @@ impl BundledAsset {
...
 
-    #[doc(hidden)]
-    /// This should only be called from the macro
-    /// Create a new asset but with a relative path
-    ///
-    /// This method is deprecated and will be removed in a future release.
-    #[deprecated(
-        note = "Relative asset!() paths are not supported. Use a path like `/assets/myfile.png` instead of `./assets/myfile.png`"
-    )]
-    pub const fn new_relative(
```
ealmloff, line 77: These are unused methods left in in the minor releases because different minor releases of the manganis macro used them. We don't use these in 0.7

packages/manganis/manganis-core/src/options.rs
```diff
@@ -17,9 +17,73 @@ use crate::{
     serde::Serialize,
     serde::Deserialize,
 )]
+#[non_exhaustive]
+pub struct AssetOptions {
```
ealmloff, line 21: Changing this from an enum to a struct is somewhat breaking, but most users are probably using the variants directly like this instead of manually using the enum, so I don't expect this to break much user code:
```rust
const _: Asset = asset!("/assets/image.png", ImageAssetOptions::new().with_size(ImageSize::Manual { width: 52, height: 52 }));
```

jkelleyrtp:
Haven't looked at the code, but do you think it's possible to have `external_asset` be just a flag on the regular asset builder options?

ealmloff:
> Haven't looked at the code, but do you think it's possible to have `external_asset` be just a flag on the regular asset builder options?

No,  there is a flag that controls if the hash is added, but it can't modify if the symbol is kept. `external_asset` just enables that flag and adds a `#[used]` attribute to the asset static so that it is kept even if the rust symbol is never used (which will be the case for most hashless assets). We can't read the contents of the asset options inside the macro because it is an arbitrary const expression which could be an ident. We could accept a third argument of literal-only values we parse in the macro like this, but without other options we need there I'm not sure it is a better api: `asset!("path", ImageOptions::new(), { hashless = true })`

ealmloff:
Another API we could support is this with the `#[used]` in user code:
```rust
#[used]
static _EXTERNAL_ASSET: Asset = asset!(
    "/66571940.jpeg",
    ImageAssetOptions::new()
        .into_asset_options()
        .with_hash_suffix(false)
);
```

jkelleyrtp:
> Another API we could support is this with the `#[used]` in user code:
> 
> ```rust
> #[used]
> static _EXTERNAL_ASSET: Asset = asset!(
>     "/66571940.jpeg",
>     ImageAssetOptions::new()
>         .into_asset_options()
>         .with_hash_suffix(false)
> );
> ```

From what I remember this might have issues with LTO since the static won't escape the module. Only `extern "C"` symbols make it out if they're private, but that also has some complications with duplicate symbols being defined during some edge cases in version unification.

Distortedlogic:
"From what I remember this might have issues with LTO since the static won't escape the module." - J.K.

Note: this isnt an "unused" image but a used one

I beleive this happens with `dx bundle --release` on 0.7.0-alpha.1

image

ealmloff:
> I beleive this happens with `dx bundle --release` on 0.7.0-alpha.1
> 
> image

That placeholder is only generated by the git version of dioxus web. If you are using the `0.7.0-alpha.1` version of the CLI with the git version of dioxus web that could be the issue

ealmloff:
> From what I remember this might have issues with LTO since the static won't escape the module. Only extern "C" symbols make it out if they're private, but that also has some complications with duplicate symbols being defined during some edge cases in version unification.

It looks like that is a general caveat to the used attribute [in the reference](https://doc.rust-lang.org/nightly/reference/abi.html#the-used-attribute). Since we now do the asset intercept after the linker, the linker can modify/remove symbols.

In practice, this doesn't seem to be the case on macos with the default linker. If you mark the asset symbol as used, it is kept even if it is private and unused in a library as long as that library is built. `extern "C"` functions can't use the used attribute and don't show up in the symbol table after lto if you objdump the final binary. This is the setup I used to test: https://github.com/ealmloff/linker-test-kept

> Note: this isnt an "unused" image but a used one
> 
> I beleive this happens with `dx bundle --release` on 0.7.0-alpha.1
> 
> image

I can reproduce this issue in the alpha, but not with the git version of the CLI + dioxus. It might have already been fixed?

Building on web with this profile: 

```toml
[profile.release]
opt-level = "z"
debug = false
lto = true
codegen-units = 1
panic = "abort"
strip = true
incremental = false
```

Distortedlogic:
@ealmloff got some very bad news in this situation. Idk that it possible to force __LINK_SECTION static to persist in `dx bundle --release` in a desirable way

I cloned dioxus repo, in asset macro, I added `#[used]` .. I added `pub` .. seems the only way to get it to persist is to have __LINK_SECTION actually used in the code somewhere outside the scoping of the macro expansion

I beat the various LLMs but they all telling me there isnt another way

Here is my dioxus repo with the used n pub
https://github.com/Distortedlogic/dioxus/tree/asset-link-section-preservation

Here is the repo I am bundling
https://github.com/Summit-Sailors/developer-portfolio

I did fork dioxus-free-icons and bumped the dioxus version in it as I was using it in the site but thats the only dep that lso deps on dioxus
https://github.com/Distortedlogic/dioxus-free-icons/tree/forked

> I can reproduce this issue in the alpha, but not with the git version of the CLI + dioxus. It might have already been fixed?

Did u use the --release flag with dx bundle? --release isnt default now for dx bundle .. without release works for me but also has the dev server trying to run .. with --release does not work for me even with latest git commit of dioxus

here is it working without --release
image

ealmloff:
> @ealmloff got some very bad news in this situation. Idk that it possible to force __LINK_SECTION static to persist in `dx bundle --release` in a desirable way
> 
> I cloned dioxus repo, in asset macro, I added `#[used]` .. I added `pub` .. seems the only way to get it to persist is to have __LINK_SECTION actually used in the code somewhere outside the scoping of the macro expansion
> 
> I beat the various LLMs but they all telling me there isnt another way
> 
> Here is my dioxus repo with the used n pub https://github.com/Distortedlogic/dioxus/tree/asset-link-section-preservation
> 
> Here is the repo I am bundling https://github.com/Summit-Sailors/developer-portfolio
> 
> I did fork dioxus-free-icons and bumped the dioxus version in it as I was using it in the site but thats the only dep that lso deps on dioxus https://github.com/Distortedlogic/dioxus-free-icons
> 
> > I can reproduce this issue in the alpha, but not with the git version of the CLI + dioxus. It might have already been fixed?
> 
> Did u use the --release flag with dx bundle? --release isnt default now for dx bundle .. without release works for me but also has the dev server trying to run .. with --release does not work for me even with latest git commit of dioxus
> 
> here is it working without --release image

You have `strip = true` defined in your cargo.toml which is not compatible with the symbol based asset system when building native binaries like the server. Try removing it or only enabling it in web builds:

https://github.com/Summit-Sailors/developer-portfolio/blob/main/Cargo.toml#L26

Distortedlogic:
yes, I tried without strip and i tried without lto

with strip = false
```
44.629s DEBUG dx::build::request: Build completed successfully in 44349857us: "/home/entropybender/repos/developer-portfolio/target/wasm32-unknown-unknown/release/app.wasm"
  44.629s  INFO dx::build::builder: Bundling app...
  44.629s DEBUG dx::build::request: Running wasm-bindgen dx_src=bundle
  44.629s  INFO dx::build::builder: Running wasm-bindgen...
  44.629s DEBUG dx::wasm_bindgen: wasm-bindgen: [
    "--target",
    "web",
    "--debug",
    "--no-demangle",
    "--keep-debug",
    "--keep-lld-exports",
    "--out-name",
    "app",
    "--no-typescript",
    "--out-dir",
    "/home/entropybender/repos/developer-portfolio/target/dx/app/release/web/public/wasm",
    "/home/entropybender/repos/developer-portfolio/target/wasm32-unknown-unknown/release/app.wasm",
]
  44.931s DEBUG dx::build::request: wasm-bindgen complete in 301.972793ms dx_src=bundle
  44.982s DEBUG dx::build::builder: Setting builder to failed state
  44.982s ERROR dx::build::builder: err=Other(Failed to write main executable

Caused by:
    0: Failed to read module
    1: [parse exception: invalid code after misc prefix: 17 (at 0:559040)]) 
Failed to write main executable
```

with lto=false it builds but issue persists

with strip=false and lto=false

```
39.731s  INFO dx::build::builder: Bundling app...
  39.731s DEBUG dx::build::request: Running wasm-bindgen dx_src=bundle
  39.731s  INFO dx::build::builder: Running wasm-bindgen...
  39.731s DEBUG dx::wasm_bindgen: wasm-bindgen: [
    "--target",
    "web",
    "--debug",
    "--no-demangle",
    "--keep-debug",
    "--keep-lld-exports",
    "--out-name",
    "app",
    "--no-typescript",
    "--out-dir",
    "/home/entropybender/repos/developer-portfolio/target/dx/app/release/web/public/wasm",
    "/home/entropybender/repos/developer-portfolio/target/wasm32-unknown-unknown/release/app.wasm",
]
  40.226s DEBUG dx::build::request: wasm-bindgen complete in 495.435904ms dx_src=bundle
  40.283s DEBUG dx::build::builder: Setting builder to failed state
  40.283s ERROR dx::build::builder: err=Other(Failed to write main executable

Caused by:
    0: Failed to read module
    1: [parse exception: invalid code after misc prefix: 17 (at 0:662914)]) 
Failed to write main executable
```

I tried a few other various combos and the doixosu toml wasm opt settings :/

packages/manganis/manganis-core/src/css.rs
```diff
@@ -21,35 +21,60 @@ pub struct CssAssetOptions {
...
-        Self::new()
+        Self::default()
     }
 }
 
 impl CssAssetOptions {
     /// Create a new css asset using the builder
-    pub const fn new() -> Self {
+    #[deprecated(note = "Use `AssetOptions::css()` instead", since = "0.7.0")]
+    pub const fn new() -> AssetOptionsBuilder<CssAssetOptions> {
```
ealmloff, line 30: On the fence about deprecating this. It no longer returns Self which is odd, but this is probably what you want if you try to build `CssAssetOptions`. Keeping the new method makes this change much less breaking

PR #4315, opened 2025-06-23
use `dx run` for playwright tests, properly abort if error occurs, don't ddos github releases, split apt-cache by os/target in matrix

jkelleyrtp:
Right now, our CI swallows errors during builds, leading to playwright running for hours if it fails. This switches our playwright to use `dx run` which is just a version of `dx serve` that disables hot-reloading.

To make it useful for CI, we also need to adjust `dx run` to properly fail if it encounters an error, which we should've been doing the whole time.


jkelleyrtp:
github is ddossing itself, and wasm-opt is not downloading after the changes to make wasm-opt always download.

will think of a way around this....

```
[WebServer]   63.612s DEBUG dx::wasm_opt: Response from GitHub: Object {
[WebServer]     "documentation_url": String("https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"),
[WebServer]     "message": String("API rate limit exceeded for 13.105.117.185. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"),
[WebServer] }

```

Issue #4287, opened 2025-06-16
Compiling dioxus-native apps with `dx` doesn't work unless the `desktop` feature is enabled


nicoburns:
**Problem**

Building `dioxus-native` apps with the CLI (`dx`) fails unless the `desktop` feature is enabled.

Error message: "No platform specified and no platform marked as default in Cargo.toml. Try specifying a platform with `--platform`"

(Note: compiling with `cargo` works fine)

**Steps To Reproduce**

- Edit `examples/wgpu-texture/Cargo.toml` to remove the desktop feature (or make it non-default)
- Run `dx serve -p wgpu-texture --features native

**Expected behavior**

App compiles and runs.

**Environment:**

- Dioxus version:  `00f2a322d88c0e5bdb9441492bdad48f70089fd4`

PR #4320, opened 2025-06-24
CLI: add `--renderer` argument

nicoburns:
# Objectives

- [x] Make it possible to build native backend without enabling desktop
- [x] Make it easier to switch between webview and native backends
- [x] Fixes https://github.com/DioxusLabs/dioxus/issues/4287
- [x] Add `Web`, `Liveview`, and `Server` renderers
- [x] Shorthand flags (`--ios`, `--native`, etc)


nicoburns:
Not sure if the playwright test failure is spurious or caused by my CLI changes

nicoburns:
It *was* a problem with my changes. But now that's fixed I'm getting `failed: Cargo build failed - no output location. Toggle tracing mode (press `t`) for more information.` which seems like it probably isn't?

nicoburns:
Ok, the recursive feature resolution was causing issues (build failures), so I've stripped that out for now. We can land that separately (probably not urgent?).

Cargo.toml
```diff
@@ -441,7 +441,6 @@ openssl = { version = "0.10", features = ["vendored"] }
 # To make most examples faster to compile, we split out assets and http-related stuff
 # This trims off like 270 dependencies, leading to a significant speedup in compilation time
 [features]
```
jkelleyrtp: I'd rather the default behavior of `cargo run --example todomvc` to not be a runtime error - is there any reason why we can't keep "desktop" as the defualt?

nicoburns: No reason except that IIRC it was me that added it recently. And so I was removing it again to put it back how I found it...

packages/cli/src/build/request.rs
```diff
@@ -552,24 +552,36 @@ impl BuildRequest {
...
             },
-            None if !using_dioxus_explicitly => Platform::autodetect_from_cargo_feature("desktop").unwrap(),
+            None if !using_dioxus_explicitly => Platform::TARGET_PLATFORM.unwrap(),
             None => match enabled_platforms.len() {
-                0 => return Err(anyhow::anyhow!("No platform specified and no platform marked as default in Cargo.toml. Try specifying a platform with `--platform`").into()),
+                0 => match renderer {
+                    Some(_) => Platform::TARGET_PLATFORM.unwrap(),
+                    None => Platform::TARGET_PLATFORM.unwrap(),
+                    // TODO: should we always have a default
+                    // None => return Err(anyhow::anyhow!("No platform specified and no platform marked as default in Cargo.toml. Try specifying a platform with `--platform`").into()),
```
nicoburns: @jkelleyrtp This was the only thing I wasn't sure about. Should we pick a default platform if none is specified at all? Either way I ought to either remove or reinstate the commented out code.

ealmloff:
Here are the changes implemented since the PR was originally opened:
- The old platform flag is now just a shorthand for rust's `--target` flag. `--web` -> `--target wasm32-unknown-unknown`, `--ios` -> currently installed ios simulator or device, etc.
- There is a new `--bundle` flag that controls how applications are bundled without enabling a dioxus renderer. Generally this is derived from the target triple and renderer (if present)
- The renderer flag is expanded to include web, liveview and the server. The renderer flag is also inferred from the target triple if we can't find it in the default features or args

Some examples of the new CLI interface:
```sh
# Serve with the web renderer (which enables --bundle web and --target wasm32-unknown-unknown
dx serve --web

# Serve with the web renderer if dioxus is found in the repo and bundle as web
dx serve --target wasm32-unknown-unknown

# Serve with the native renderer and the android target triple. This will use the
# android bundle from the resolved triple
dx serve --native --android

# Serve the readme example with the default webview renderer on a connected android device
dx serve --android --device --example readme
```

nicoburns:
I can't formally review as I opened the PR, but I approve @ealmloff's latest version. I still think that an abstract `platform` parameter would be nice. But I'm happy to go with this `target`-based approach. It solves my main problems with the status quo. And the short options `--ios`, `--native`, etc are really nice and mean that most users won't need to care about the underlying data model.

ealmloff:
All of the old platform arguments are now supported as aliases for the newer renderer, target alias and bundle format args

PR #4325, opened 2025-06-26
wip: fixes to get docsite working again

jkelleyrtp:
Not sure what happened, but the docsite wasn't hot-patching properly anymore. This fixes that.

We weren't have these import issues previously. My hypothesis is that we added a new feature to the docsite that stresses out the hotpatch engine in a way it wasn't previously.


jkelleyrtp:
Rolling up into #4326

PR #4328, opened 2025-06-28
Handle panics in main on android

mohe2015:
It seems that currently panicking in `main` on android will silently fail.

I assume there was a reason this was not done this way, you can probably tell me why?


jkelleyrtp:
Interesting... I remember this approach *not* working, but in my testing it seems to behave fine. 

Note that the Rust `main` generates its own runtime code that catches panics. I don't think you need to stop_unwind main.

https://doc.rust-lang.org/beta/src/std/rt.rs.html

Anyways, thanks for the PR! I think there's some folks in the discord that can give this a better test drive than I can, but seems to work fine to merge.

jkelleyrtp (approved):

Issue #3904, opened 2025-03-24
Submit buttons don't work on windows


MintSoup:
On Windows, buttons with `type=submit` sometimes do not submit the form they are in. From debugging the dioxus source code I have narrowed it down to the following lines:

https://github.com/DioxusLabs/dioxus/blob/9e09bcf3f79aa7b349da23ca80cb2a214230562b/packages/interpreter/src/ts/native.ts#L340-L342

This just blatantly cancels any onclick events from submit buttons inside all forms. However, if there are no elements on the page with an onclick handler attached, the toplevel main div that dioxus implicitly inserts does not get an onclick event handler attached to it, which causes `event.type` to be `click` so `handleClickNavigate` does not get called in the first place:

https://github.com/DioxusLabs/dioxus/blob/9e09bcf3f79aa7b349da23ca80cb2a214230562b/packages/interpreter/src/ts/native.ts#L294-L296

I haven't bothered to check why this issue is not present on other platforms but to me this seems like a logic error. From my (somewhat limited) understanding the code in question is used to cancel onsubmit events from navigation buttons in case they happen to be inside a form (very weird/rare edge case, btw), but that whole issue can be sidestepped if those buttons just get their `type` set to `button` so onsubmit events won't be emitted in the first place. 

- Dioxus version:  0.6.3
- Rust version: 1.85
- OS info: Windows 10
- App platform: Desktop

mzdk100:
Perhaps other platforms also have this issue.

MintSoup:
Actually, open further inspection, it seems that the navigator buttons cancel their default events anyway, so we should be good? https://github.com/DioxusLabs/dioxus/blob/9e09bcf3f79aa7b349da23ca80cb2a214230562b/packages/router/src/components/history_buttons.rs#L78-L81

Does this mean we can just remove 

https://github.com/DioxusLabs/dioxus/blob/6bdb69e23529e0fd9813f535b7e4d957cc1d9b28/packages/interpreter/src/ts/native.ts#L337-L342

and be good?

ealmloff:
> Does this mean we can just remove
> 
> [dioxus/packages/interpreter/src/ts/native.ts](https://github.com/DioxusLabs/dioxus/blob/6bdb69e23529e0fd9813f535b7e4d957cc1d9b28/packages/interpreter/src/ts/native.ts#L337-L342)
> 
[...]

PR #4329, opened 2025-06-28
Fix form submission on native if there is an onclick handler somewhere

mohe2015:
Closes https://github.com/DioxusLabs/dioxus/issues/3904
See also https://github.com/DioxusLabs/dioxus/issues/3791#issuecomment-3015486477
Related PR https://github.com/DioxusLabs/dioxus/pull/3804


packages/interpreter/src/ts/native.ts
```diff
@@ -277,11 +277,14 @@ export class NativeInterpreter extends JSChannel_ {
...
-    // now that we expose preventDefault to the virtualdom on desktop
-    // Liveview will still need to use this
-    this.preventDefaults(event);
+    // This should:
+    // - prevent form submissions from navigating
+    // - prevent anchor tags from navigating
+    // - prevent buttons from submitting forms
+    // - let the virtualdom attempt to prevent the event
+    if (event.type === "submit") {
+      event.preventDefault();
```
mohe2015, line 286: to my understanding all submit events should be prevented here so the code below is not needed

jkelleyrtp:
~~From what I can tell this PR doesn't actually do anything - just inlines the preventsDefault function and then deletes the other line?~~

Ahhhhh read through the linked issue and then realized what's happening. 

There was a logic bug here:

```
    let form = target.closest("form");
    if (target.tagName === "BUTTON" && (event.type == "submit" || form)) {
      event.preventDefault();
    }
```

The event handling happens in two phases:
- A PointerEvent (click)
- A SubmitEvent

We were cancelling the PointerEvent which then cancelled the submit event.

We should *only* be cancelling submit events. Click events should still fire.

The PR looks good, thanks for the fix.

jkelleyrtp (approved):

Issue #4331, opened 2025-06-29
Routable doesn't work with component in a `core` module.


davidB:
**Problem**

After a refactor where some components have been moved to new modules, I got the following errors:

```txt
error[E0433]: failed to resolve: could not find `convert` in `core`
 --> src/main.rs:5:17
  |
5 | #[derive(Clone, Routable, Debug, PartialEq)]
  |                 ^^^^^^^^ could not find `convert` in `core`
  |
  = note: this error originates in the derive macro `Routable` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this module
  |
1 + use std::convert;
  |
...
error[E0599]: no variant named `Error` found for enum `Route`
  --> src/main.rs:15:17
   |
15 | #[derive(Clone, Routable, Debug, PartialEq)]
   |                 ^^^^^^^^ variant not found in `Route`
16 | pub(crate) enum Route {
   | --------------------- variant `Error` not found here
   |
   = note: this error originates in the derive macro `Routable` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0433, E0599.
```

I found the cause, I introduced a new module named **core**.

Workaround: renamed the module `core` (e.g., `base`)

**Steps To Reproduce**

- Create a module "core" with a component
- Use the component in a Routable
- `cargo build`

A minimal example (raise only the error related to `convert`):

```rust
use dioxus::prelude::*;

use crate::core::Home;

#[derive(Clone, Routable, Debug, PartialEq)]
pub(crate) enum Route {
    #[route("/")]
    Home,
}

fn main() {
    launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        Router::<Route> {}
    }
}

mod core {
    use dioxus::prelude::*;

    #[component]
    pub fn Home() -> Element {
        rsx! {
            h1 { "Welcome" }
        }
    }

}

```

**Expected behavior**

Being able to name a module `core` without issue.

**Environment:**

- Dioxus version:  0.6.3
- Rust version:    1.88
- OS info:         Linux 6.12.34-1-MANJARO
- App platform:    web

**Questionnaire**

I'm interested in fixing this myself but I don't have experience in derive/proc macro, and so it could take time to propose a PR.

ealmloff:
[...]

PR #4350, opened 2025-07-02
fix issue: Routable doesn't work with component in a core module. #4331

zhiyanzhaijie:
link to issue: https://github.com/DioxusLabs/dioxus/issues/4331

what happens:
  in route_macro/lib.rs, there is a core::convert method call, in user's case, he/she has a self `core` module

fix:
  clarify `core::convert` to `::core::convert`


jkelleyrtp:
Thanks!

Issue #4216, opened 2025-05-28
`dioxus-ssr` performs worse than other templating engines


FalkWoldmann:
Hi,

based on the information from https://github.com/askama-rs/template-benchmark, `dioxus` has the lowest performance compared to other templating engines. Although this isn't the main purpose of `dioxus`, the performance gap is quite noticeable.

I really enjoy working with `rsx`, so it would be fantastic if its performance could be improved to match the level of other engines. However, if this is beyond the scope, feel free to close this issue!

PR #4352, opened 2025-07-02
refactor to core/diff to fix some leaks

jkelleyrtp:
- **wip...**
- **compiling, tests not passing**


wheregmis:
I think if we are tapping into core.
We should also try to enhance it on for performance. I think the benchmark comparing dioxus-ssr is directly toed into dioxus-core. I see the relevant issue on #4216 

https://github.com/askama-rs/template-benchmark

Issue #4359, opened 2025-07-03
Add an example integrating Bevy rendering with the native backend


jerome-caucat:
## Feature Request

In order to integrate Bevy with Dioxus, I adapted the `wgpu-texture` example to use Bevy for the 3D rendering.

I added a new `bevy-texture` example where:
- The rendering is done by a headless Bevy app.
- Each frame is captured as a screenshot and used to fill a wgpu texture.
  - This part could probably be improved.
- The texture is displayed using `dioxus_native` custom paint source.

https://github.com/user-attachments/assets/0f501a37-b436-4551-8e3f-7edbbcd1d203

## Implement Suggestion

I submitted a PR: https://github.com/DioxusLabs/dioxus/pull/4360

PR #4360, opened 2025-07-03
Add a bevy-texture example based on wgpu-texture

jerome-caucat:
In order to integrate Bevy with Dioxus, I adapted the `wgpu-texture` example to use Bevy for the 3D rendering.

I added a new `bevy-texture` example where:
- The rendering is done by a headless Bevy app.
- Each frame is captured as a screenshot and used to fill a wgpu texture.
  - This part could probably be improved.
- The texture is displayed using `dioxus_native` custom paint source.

[Screencast from 2025-07-03 15-22-42.webm](https://github.com/user-attachments/assets/ba0581cc-266d-4002-a240-322d8408a511)

closes https://github.com/DioxusLabs/dioxus/issues/4359


nicoburns:
Thank you for submitting this! However, it look like this is copying from the bevy `Image` into a cpu-side `Vec<u8>` and then re-uploading that back into a wgpu Texture, which is going to be terrible for performance. I think it ought to be possible to either get Bevy to render directly into an appropriate texture, or at least do a gpu-side texture-to-texture copy.

nicoburns:
I think we might want to use a https://docs.rs/bevy/latest/bevy/render/camera/struct.ManualTextureView.html as the render target (or an `Image`, but access the texture directly if that's possible).

jerome-caucat:
> I think we might want to use a https://docs.rs/bevy/latest/bevy/render/camera/struct.ManualTextureView.html as the render target (or an `Image`, but access the texture directly if that's possible).

@nicoburns Yes, I am going to fix that. Thanks for the suggestion!

I also would like to add a simple camera controller, to showcase how to pass mouse and key events to Bevy. Not sure how I am going to transfer the events yet.

jerome-caucat:
I am struggling to use [ManualTextureView](https://docs.rs/bevy/latest/bevy/render/camera/struct.ManualTextureView.html).

With [this code](https://github.com/jerome-caucat/dioxus/blob/879ffe11c2127a8ea92e576bc7e23dc3a53960cd/examples/bevy-texture/src/bevy_renderer.rs#L129) I am getting the following error:

```
thread '<unnamed>' panicked at /home/jeromecaucat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-core-24.0.5/src/storage.rs:130:9:
assertion `left == right` failed: TextureView[Id(1,6)] is no longer alive
  left: 6
 right: 1
```

I've tried storing the Texture and TextureView in the BevyRenderer. Perhaps the problem is that I am using the device from Dioxus to create the texture.

I'll see if I can render in a Bevy texture and get its wgpu texture.

nicoburns:
> Perhaps the problem is that I am using the device from Dioxus to create the texture.

Yes, I had the same problem with the `wgpu-texture` example when I was using two separate devices. I think that one way or the other you'll need Bevy and Blitz to use the same `wgpu::Device`

nicoburns:
If you initialise Bevy RenderPlugin with [`RenderCreation::Manual`](https://docs.rs/bevy/latest/bevy/render/settings/enum.RenderCreation.html) then it looks like you can pass in your own `Device`.

jerome-caucat:
I had a quick look at [RenderCreation::Manual](https://docs.rs/bevy/latest/bevy/render/settings/enum.RenderCreation.html) but it seems hard to setup so that the Bevy app works normally. I'll give it another go.

nicoburns:
> I had a quick look at [RenderCreation::Manual](https://docs.rs/bevy/latest/bevy/render/settings/enum.RenderCreation.html) but it seems hard to setup so that the Bevy app works normally. I'll give it another go.

Does this help? https://github.com/DioxusLabs/blitz/pull/232 (you would need to point `dioxus-native` to the git version of Blitz)

I think that gives you access to all the types you need. `Instance` is passed directly, `DeviceHandle` contains `Device`, `Queue`, and `Adapter` (you can clone them if you need owned versions), and `AdapterInfo` can be obtained from `Adapter`.

nicoburns:
Hang on, I'll see if I can hook it up.

nicoburns:
@jerome-caucat Ok, I have [RenderCreation::Manual](https://docs.rs/bevy/latest/bevy/render/settings/enum.RenderCreation.html) setup. So you should be able to try the texture code again now.

nicoburns:
Ok, and I also ported the code you [posted above](https://github.com/DioxusLabs/dioxus/pull/4360#issuecomment-3036206658) (sorry - I got excited!).

Seems to be working! I'm not sure if it is correctly responding to resizes: the `wgpu-texture` cube seems to resize to match the window size whereas bevy code does not (not sure if this is bug where the resize isn't being picked up, or just a difference in how the Bevy scene is being rendered)

examples/bevy-texture/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "bevy-texture"
```
nicoburns: Minor thing, but can we rename the crate to `bevy-example` and the directory to just `bevy` please/

jerome-caucat:
@nicoburns Thanks a lot for your help. I was also updating the code on my side. I'll clean the code and fix the remaining issues.

nicoburns:
> @nicoburns Thanks a lot for your help. I was also updating the code on my side. I'll clean the code and fix the remaining issues.

Awesome. Now I've unblocked you, I'll stop interfering and let you get on with it :)

nicoburns:
I've published a new versions of the `anyrender` and `blitz` crates so this PR doesn't need to depend on git versions. Anyrender crates should be set to `0.4`. Blitz crates to `=0.1.0-alpha.5`

jerome-caucat:
@nicoburns I cleaned the code, added some comments, fixed the resize problem and renamed the example.

I noted there is a segfault when we close the window, but the `wgpu-texture` example has the same issue.

I wanted to add mouse and keyboard events to this demo, but perhaps I should do this in a separate issue.

examples/bevy-texture/Cargo.toml
jerome-caucat: @nicoburns Done.

nicoburns (approved):
@jerome-caucat All looks good to me. Just need to fix the issues with CI and then we can land this. Long-term we may want to move this into a separate repo (as this one is already a very large workspace, and pulling in the whole of Bevy just for an example is quite a lot). But lets land it here first, and we move it and improve on it later.

> I noted there is a segfault when we close the window, but the wgpu-texture example has the same issue.

Not ideal, but not disastrous in an example. If the `wgpu-texture` example has the same problem let's not block this PR on it.

> I wanted to add mouse and keyboard events to this demo, but perhaps I should do this in a separate issue.

Yeah, let's do a new issue/PR. This will likely require further changes in the Blitz layer.

PR #4365, opened 2025-07-04
Create example of rendering a dioxus-native app to a texture

nicoburns:


nicoburns:
@jerome-caucat I've updated this to use the new `dioxus-native-dom` crate. This crate is split out of `dioxus-native` and only includes the bits you need for "headless" / "embedded" setup like this (doesn't pull in things like winit and tokio).

jerome-caucat:
@nicoburns Thank you for this example.

I've started implementing an example rendering dioxus-native to a texture in bevy: [native-headless-in-bevy](https://github.com/jerome-caucat/dioxus/blob/05373395cceffa5e6dd48893c247cb0aa1a16f8a/examples/native-headless-in-bevy/src/main.rs)

However, I'm running into an issue: the content of the Dioxus app (`rsx!`) doesn't appear in the texture (see the [FIXME](https://github.com/jerome-caucat/dioxus/blob/05373395cceffa5e6dd48893c247cb0aa1a16f8a/examples/native-headless-in-bevy/src/main.rs#L244)).

The VelloRenderer is rendering to the texture — I can confirm this because:
- I see the [white background](https://github.com/jerome-caucat/dioxus/blob/05373395cceffa5e6dd48893c247cb0aa1a16f8a/examples/native-headless-in-bevy/src/main.rs#L268) of the Vello renderer, instead of the [red background](https://github.com/jerome-caucat/dioxus/blob/05373395cceffa5e6dd48893c247cb0aa1a16f8a/examples/native-headless-in-bevy/src/main.rs#L186) of the initial texture.
- I can see a [test Scene](https://github.com/jerome-caucat/dioxus/blob/05373395cceffa5e6dd48893c247cb0aa1a16f8a/examples/native-headless-in-bevy/src/main.rs#L248) correctly.

Would you happen to have an idea of what is going on, or how I could debug this?

EDIT: More info:
- The waker callback is never called.
- `dioxus_doc.poll` retruns false

nicoburns:
On vacation this week. But I think you need to call .resolve() on the inner BaseDocument before rendering (this was missing from my example). This should be called every frame.

jerome-caucat:
> On vacation this week. But I think you need to call .resolve() on the inner BaseDocument before rendering (this was missing from my example). This should be called every frame.

@nicoburns Thanks, this helped, but I also had to add a call to `initial_build`:

```rs
        let mut dioxus_doc = DioxusDocument::new(vdom, None);
        dioxus_doc.initial_build();
        dioxus_doc.set_viewport(Viewport::new(WIDTH, HEIGHT, SCALE_FACTOR, COLOR_SCHEME));
```

BTW the call to `initial_build` is commented in [`DioxusDocument::new`](https://github.com/DioxusLabs/dioxus/blob/1a494d1f5e549383f0bba8f5975ffe5ef0475592/packages/native-dom/src/dioxus_document.rs#L102).

I'll handle some event and submit a PR once I have a working example.

nicoburns:
Glad you've got it working!

@jkelleyrtp Can you remember why you removed the call to "initial_build" from "new", and made it a separate method? Perhaps there is something that needs to be done in between in some use cases?

jkelleyrtp:
> initial_build



> Glad you've got it working!
> 
> @jkelleyrtp Can you remember why you removed the call to "initial_build" from "new", and made it a separate method? Perhaps there is something that needs to be done in between in some use cases?

This is because we need to set up the virtualdom and inject the window context into it before the first render.

See this code:
```
        if let Some(config) = self.pending_window.take() {
            let mut window = View::init(config, event_loop, &self.proxy);
            let renderer = window.renderer.clone();
            let window_id = window.window_id();
            let doc = window.downcast_doc_mut::<DioxusDocument>();

            doc.vdom.in_runtime(|| {
                let shared: Rc<dyn dioxus_document::Document> =
                    Rc::new(DioxusNativeDocument::new(self.proxy.clone(), window_id));
                ScopeId::ROOT.provide_context(shared);
            });

            // Add history
            let history_provider: Rc<dyn History> = Rc::new(MemoryHistory::default());
            doc.vdom
                .in_runtime(move || ScopeId::ROOT.provide_context(history_provider));

            // Add renderer
            doc.vdom
                .in_runtime(move || ScopeId::ROOT.provide_context(renderer));

            // Queue rebuild
            doc.initial_build();

            // And then request redraw
            window.request_redraw();

            // todo(jon): we should actually mess with the pending windows instead of passing along the contexts
            self.inner.windows.insert(window_id, window);
        }
```

PR #4367, opened 2025-07-05
Add generic `create_waker` function to `dioxus-core`

nicoburns:
`dioxus-core`'s `VirtualDom` expects you to pass it a `Context` which you create from a `Waker`. But it does not provide an easy way to create a `Waker`. This function will create a waker that will call `Box<dyn Fn()>` callback that you can do anything in (e.g. close over a Winit/Tao proxy and send an event).

### DIFF
--- a/packages/core/src/lib.rs
+++ b/packages/core/src/lib.rs
@@ -77,12 +77,13 @@ pub(crate) mod innerlude {
 }
 
 pub use crate::innerlude::{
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
The create_waker function doesn't seem to rely on any dioxus core types so I'm not sure it's worth being a part of core.

Also, I don't believe either Waker or Context is part of our public API anywhere in core.

PR #4386, opened 2025-07-07
prevent_default must be called synchronously or it will have no effect

LilahTovMoon:
When dropping js or txt files, the browser would still do its default behavior of showing the contents of the file (either in a new tab or navigating away from the current Dioxus app page, depending on the browser/user preferences). The docs for prevent_default say, "Note: This must be called synchronously when handling the event. Calling it after the event has been handled will have no effect."

This makes the `ondrop` handler synchronous and uses `spawn` to handle the file reading.

I'm not an expert on Dioxus or Rust so there might be something wrong/bad with this solution.

### DIFF
--- a/examples/file_upload.rs
+++ b/examples/file_upload.rs
@@ -79,11 +79,13 @@ fn app() -> Element {
                 hovered.set(true)
             },
             ondragleave: move |_| hovered.set(false),
-            ondrop: move |evt| async move {
+            ondrop: move |evt| {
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Good catch! I think we might poll futures once before putting them into the runtime to help the usecases where people prevent_default in an async block on accident.

PR #4394, opened 2025-07-08
Adds `spawn_in_scope`

jprider63:
This adds the function `spawn_in_scope` which spawns a future with the scope provided by the user. I needed this since I have a `Signal` at a given scope (created with `Signal::new_maybe_sync_in_scope_with_caller`) that is periodically updated by an associated future. `spawn_forever` did not work since the root scope in the future would attempt to update the subscope's Signal. 

Note: This PR is on the `v0.6` branch so you might want to upstream it to `main` too.

### DIFF
--- a/packages/core/src/global_context.rs
+++ b/packages/core/src/global_context.rs
@@ -204,7 +204,14 @@ pub fn queue_effect(f: impl FnOnce() + 'static) {
 ///
 #[doc = include_str!("../docs/common_spawn_errors.md")]
 pub fn spawn_forever(fut: impl Future<Output = ()> + 'static) -> Option<Task> {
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
There is already a function on `Runtime` called `spawn` that takes any ScopeId. We don't want to add more items to the prelude/global scope.

jprider63:
Thanks. I wasn't aware of that public API. If anyone needs this in the future, the following worked for me:

```
fn spawn_in_scope(scope: ScopeId, fut: impl Future<Output = ()> + 'static) -> Option<Task> {
    let rt = Runtime::current().ok()?;
    Some(rt.spawn(scope, fut))
}
```

PR #4400, opened 2025-07-10
Warn about assets being stripped if strip is true and the current package depends on manganis

ealmloff:
The symbol based version of manganis will break if cargo is build with symbols stripped. This PR adds a warning when we find manganis as a dependency and the current build includes strip in the profile.

Note `strip = "debuginfo"` seems to work fine in my testing so we don't warn in that case

## Why is this not a build script

We get the profile name through environment variables in the build script, but we don't have access to the cargo.toml of the main package. We could potentially look for it based on the out dir, but we would need to pull in cargo_toml as a build dependency for manganis which would slow down builds.


debanjanbasu:
Awesome... I've had identified this the hard way 😅. Thanks for this PR.

jkelleyrtp (approved):

Issue #4307, opened 2025-06-21
ontouchend returns a map and expects a sequence


bananabit-dev:
**Problem**

When using ontouchstart, ontouchmove, ontouchend i get this error in ontouchend:

This is only a excerpt of the error.
`RustStdoutStderr: ERROR Error parsing user_event: Error("Failed to deserialize event data for event touchend:  Error(\"invalid type: map, expected a sequence\", line: 0, column: 0)\n'Object {\n    \"alt_key\": Bool(false),\n    \"changed_touches\": Object {\n        \"0\": Object {},\n    },\n    \"ctrl_key\": Bool(false),\n    \"meta_key\": Bool(false),\n    \"shift_key\": Bool(false),\n    \"target_touches\": Object {},\n    \"touches\": Object {},\n}'", line: 0, column: 0).Contents: Ok(""), raw: Request { ...`

if i uncomment the other ones i get errors for the others aswell:

`06-21 20:46:17.922 15728 15765 I RustStdoutStderr: ERROR Error parsing user_event: Error("Failed to deserialize event data for event touchmove:  Error(\"invalid type: map, expected a sequence\", line: 0, column: 0)\n'Object {\n    \"alt_key\": Bool(false),\n    \"changed_touches\": Object {\n        \"0\": Object {},\n    },\n    \"ctrl_key\": Bool(false),\n    \"meta_key\": Bool(false),\n    \"shift_key\": Bool(false),\n    \"target_touches\": Object {\n        \"0\": Object {},\n    },\n    \"touches\": Object {\n        \"0\": Object {},\n    },\n}'", line: 0, column: 0).Contents: Ok("")`

**Steps To Reproduce**

- Touch on the display
- The event returns a wrong datatype a map instead of the expected type.

**Expected behavior**

I touch the display it registers the touch while i drag and displays the info.

**Environment:**

- Dioxus version: 6
- Rust version:   1.86.0
- OS info: Android
- App platform: Android

PR #4406, opened 2025-07-11
Fixed touch input it will now serialize the touchlist

bananabit-dev:
fixed the invalid type map expected sequence

#4307 its related to this issue.


bananabit-dev:
i shouldnt cast to any that needs a fix.

jkelleyrtp (approved):

Issue #2524, opened 2024-06-15
FMT tab issues and comment removal inside event handlers


Coder2195Text:
**Problem**

In 

**Steps To Reproduce**

Make some rsx with non 4 tabs
```
div {
   class: "w-1/4",
   onclick: move |_| {
     todo!();
     todo!();
     todo!();
     // some comment
   },
   "Name"
}
```

You will get this

```
div {
   class: "w-1/4",
   onclick: move |_| {
       todo!();
       todo!();
       todo!();
   },
   "Name"
}
```

**Expected behavior**

Consistent tab spacing, and the retaining of comments.

**Environment:**
 - Dioxus version: 0.5.4
 - Rust version: 1.79.0
 - OS info: Ubuntu 24.04
 - App platform: any

**Questionnaire**

- [ ] I'm interested in fixing this myself but don't know where to start
- [ ] I would like to fix and I have a solution
- [ ] I don't have time to fix this right now, but maybe later

Enduriel:
Well I'm gonna blame my not seeing this issue on the fact that you didn't mention indent/indentation anywhere in your issue so my single search didn't hit it 😛.

I dug into this issue in a bit more detail in #2537, the problem basically boils down to the fact that `prettyplease` forces 4 space formatting and dtolnay is [uninterested](https://github.com/dtolnay/prettyplease/issues/75) in supporting alternative indentation styles.

I see 2 potential solutions, either have Dioxus or someone else maintain a fork of prettyplease which adds support for alternative indentation styles, or some relatively ugly hacks around [`unparse_expr`](https://github.com/DioxusLabs/dioxus/blob/487570d89751b34bbfd5e9b5ff1e0fd3780bf332/packages/autofmt/src/prettier_please.rs#L97), specifically with regards to handling multiline strings and potentially other indentation edge cases.

I'm not sure what direction Dioxus would prefer to go with this, I'd be willing to take a look at writing some hacky conversion code which takes the output from `prettyplease` and tries to reformat it correctly, I think it would be possible to have accurate results with enough tests. From a relatively quick look, `prettyplease` shouldn't be *too* bad to adjust, but it would require some API changes and the maintenance of that fork (which I am personally uninterested in).

ealmloff:
There is an issue about dx fmt eating comments here: https://github.com/DioxusLabs/dioxus/issues/2341
And an older issue about using the rustfmt config here: https://github.com/DioxusLabs/dioxus/issues/1492

[...]

PR #4410, opened 2025-07-13
Avoid the accidental removal of comments inside expressions by dx fmt

jjvn84:
This PR modifies the behaviour of ```dx fmt``` to add back comments removed from expressions when parsing the code to reformat it. It should only affect expressions used in attributes for rsx elements.

Fixes [https://github.com/DioxusLabs/dioxus/issues/2524](https://github.com/DioxusLabs/dioxus/issues/2524)


jkelleyrtp:
Thanks for the PR!

I just opened a similar PR to fix whitespace removal from the end of rsx blocks here:

https://github.com/DioxusLabs/dioxus/pull/4423

Hopefully with these two PRs all whitespace should be handled properly!

jkelleyrtp:
Ended up adding a few more test cases and inverting the line pull logic to pull whitespace on top of pretty lines instead of pretty lines under whitespace.

Thanks for the really clever solution here - been something I've been meaning to fix for a long time and it was nice to see a PR popup fixing it :)

PR #4413, opened 2025-07-14
Boxed and mapped signals preparing for stores

### LINKED ISSUE #3610: Add `Writable::map` for Signal Mapping
### Description

Currently, Dioxus Signals lacks a `Writable::map` feature, making it difficult to derive a `Signal<T>` from a parent `Signal<Vec<T>>`. This is particularly important in scenarios like dynamic form generation or nested component state management, where child components require fine-grained reactivity while maintaining synchronization with the parent state.

For example, a parent component may receive a `Signal<Vec<T>>` and need to pass individual `Signal<T>` to its children. The lack of `Writable::map` forces developers to pass the `Signal<Vec<T>>` and the index separately and manually index into the vector on every access.

---

## Implement Suggestion

Introduce a `Writable::map` function to create a mutable view into a subset of a signalized structure. With this, developers can map a `Signal<Vec<T>>` to individual `Signal<T>` for child components in an ergonomic and efficient manner.

Example:

```rust
forms_signal.map_mut(move |forms| &mut forms[index])
```

This approach would simplify the following common use case:

```rust
#[component]
fn ParentComponent(props: ParentProps) -> Element {
    let forms_signal = props.forms_signal;

    rsx! {
        div {
            forms_signal.read().iter().enumerate().map(|(index, form)| rsx!(
                ChildComponent {
                    form_signal: ???, // Derive Signal<T> here
                }
            ))
        }
    }
}
```

With `Writable::map`, it becomes easy to pass individual signals to children while ensuring modifications in children sync back to the parent signalized vector.

---

## Why This Feature Matters

This feature is crucial for:
1. Dynamic form generation (e.g., JSON-driven forms).
2. Nested component state management.
3. Any pattern requiring signalized collections with fine-grained reactivity.

It improves code ergonomics, reduces verbosity, and minimizes potential errors from manual indexing.

---

## Workaround

Currently, developers can pass the `Signal<Vec<T>>` and index separately to child components and manually index into the vector. However, this approach is less ergonomic and more verbose compared to having a dedicated `Writable::map`.

---

## Related Discussions and Issues

- [Discussion #3570](https://github.com/dioxuslabs/dioxus/discussions/3570)
- Related Issue: #1805

--- 

You can now submit this issue on the project’s GitHub!

ealmloff:
This PR contains just the breaking changes required for the upcoming [store implementation](https://github.com/DioxusLabs/dioxus/pull/4483). It adds two new primitives:
1) Mutable mapped signals (without a different subscriber list). This lets you scope a writable value and pass it to a child component. It requires two closures to map the immutable and mutable references to the inner signal. With stores, these closures will be derived
2) Boxed readable and writable values (`ReadSignal`, and `WriteSignal`). Any readable or writable types can be converted to a read signal or write signal with the `boxed` method or built in types are automatically converted into the boxed variants when passing into props

## Design: Why isn't everything just ReadSignal and WriteSignal?

I still think exposing the underlying types in hooks are useful for three reasons:
1) Boxed signals are copy which means they cannot be created on the fly unlike mapped signal. `signal.map(|foo| &foo.bar)` can easily be called in the middle of a loop in rsx and doesn't allocate in the current component. `signal.map(|foo| &foo.bar).boxed()` should only be called in hooks because it allocates the box in the scope
2) Different types have different methods. The `Resource` type implements Readable, but it also has additional methods like `restart` that can be used to control the running future
3) Most state is local and avoiding the box is better for performance and backtraces. It also lets us more easily specialize the documentation of methods for each readable/writable type

TODO:
- [x] Make Readable and Writable dyn compatible. Most of the complex logic is split out into extension traits which makes this PR breaking
- [x] Implement `map_mut`. This requires two different closures to map the immutable and mutable borrows since closures cannot be generic over mutability
- [x] Make BoxedReadable Copy
- [x] Merge ReadOnlySignal and BoxedReadable (and keep ReadOnlySignal as a type alias)

Closes #3610


packages/signals/src/read.rs
```diff
@@ -41,64 +42,27 @@ pub trait Readable {
...
+    fn try_peek_unchecked(
+        &self,
+    ) -> Result<ReadableRef<'static, Self>, generational_box::BorrowError>;
 
+    /// Get the underlying subscriber list for this readable. This is used to track when the value changes and notify subscribers.
+    fn subscribers(&self) -> Option<Subscribers>;
+}
+
+/// An extension trait for `Readable` types that provides some convenience methods.
+pub trait ReadableExt: Readable {
```
ealmloff, line 65: Splitting out the ReadableExt and WritableExt traits make the Readable and Writable traits dyn compatible

packages/signals/src/write.rs
```diff
@@ -80,6 +64,255 @@ pub trait Writable: Readable {
     fn write_unchecked(&self) -> WritableRef<'static, Self> {
         self.try_write_unchecked().unwrap()
     }
+}
+
+/// A mutable reference to a writable value. This reference acts similarly to [`std::cell::RefMut`], but it has extra debug information
```
ealmloff, line 69: All writables return the same Writelock with a different drop guard if needed. We used to require the map methods on the Writable trait but they are not dyn compatible and  we only used them to track write drops for signals

packages/core/src/reactive_context.rs
```diff
@@ -315,9 +316,88 @@ impl Drop for Inner {
...
-                subscriber.remove(&self_);
-            }
+            subscriber.0.remove(&self_);
+        }
+    }
+}
+
+/// A list of [ReactiveContext]s that are subscribed. This is used to notify subscribers when the value changes.
+#[derive(Clone)]
+pub struct Subscribers {
```
ealmloff, line 326: This isn't directly related to boxed signals, but it is a breaking change required for stores. Stores maintain a tree of reactive contexts. Moving this to a boxed trait lets stores keep track of when that tree is empty to remove dead nodes. This lets us clean up the selector memory when adding a bunch of items to a Vec and then later removing those nodes. Stores let you `write` to the whole vec at once if you need to pass it to an external library so we can't easily track what subscriptions are removed in the store itself

PR #4416, opened 2025-07-15
clean up prelude

jkelleyrtp:


nicoburns:
Removing internal uses of the `dioxus_core::prelude` seems like a nice win for the readability of the Dioxus code itself :)

jkelleyrtp:
> Removing internal uses of the `dioxus_core::prelude` seems like a nice win for the readability of the Dioxus code itself :)

IMO we should only have preludes on dioxus and dioxus-lib. Everything else is a library and top-level imports are fine. Lots of churn though!

PR #4419, opened 2025-07-15
Always prevent form submission even if the form is missing onsubmit

### LINKED ISSUE #4417: Form navigate in webview bricks page (0.7 alpha 2)
I was writing an experimental app using Dioxus to get a feel for the tech.
I started using 0.7-alpha1, but tried to upgrade to 0.7-alpha2 after some multi windows issues on Windows.

Problem is that 0.7-alpha2 seems to have regressed.
Clicking on some buttons in my test app will now all the time cause the webview to turns full blank, and the application log will contain: "Webview 0 was already connected. Rejecting new connection.".

Using webview devtools, we can see following messages:
```
WebSocket connection to 'ws://127.0.0.1:36693/0/TJ9IanyV__4XAeIpYfz14xswGcS-cwhfkMkmgCwnblwxbcTHsWerhB0tU4LmXrvxTBTvUFZiul-usjJA7KSBZF5-wXdS4A4W38oUcsX3zCo0119HduH9Zp4wwLird08zzC9YwIL8XKuRTbCDkLoG_g4yZQmUx3N1gJfpASFtYAKZNRyK-bCPSwAMISgZ3PCoxXbLv8A9zgu7eIK6J3P8cEFL5Gt2RRWQOvE6egt2CI05Oy9HKCK-k1i0_DDku8cjP4NWdPjGAyWfauF4OVwbceqyljP6G27RAnyjzMnvNq6dLov4wxLZuV5KHU-rFedoNQ67rlrCrYvHcvfWtYcVIw==' failed: WebSocket is closed due to suspension.
```

As well as 
```
[Error] Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'this.templates[this.u16buf[this.u16bufp++]][this.u16buf[this.u16bufp++]]')
	run (index.html:364:411)
	run_from_bytes (index.html:405)
	rafEdits (index.html:410:12483)
	(anonymous function) (index.html:410:13010)
```

I've yet to understood why that specific action now causes issues (it doesn't do anything special as far as I can tell). 

This happens on both Windows and Linux so doesn't seems to be OS specifics.

Unfortunately, I cannot share application code as its mostly a test integrating company internal library to build a UI on top. I might try to get it to reproduce differently, but for now no success.

**Environment:**

- Dioxus version:  0.7-alpha2
- Rust version:   1.88 stables
- OS info:         Ubuntu
- App platform:    Desktop

ealmloff:
Currently our form submission logic is inside the event handling logic which only runs if you already have a dioxus event handler attached to your form. If you don't and you try to submit a form, the desktop application breaks. This PR moves the logic to a mutation observer which will work for any forms. 

Currently this application breaks if you click the button in main but not in this PR:
```rust
use dioxus::prelude::*;

fn main() {
    dioxus::launch(app);
}

fn app() -> Element {
    rsx! {
        form {
            button {}
        }
    }
}
```

The existing behavior is technically documented in the form example, but I don't think most people are aware of this behavior

Fixes #4417


jkelleyrtp:
Can we get this working without a MutationObserver? Is there some way we can connect to events that predicate submit events and tag form elements? We also do special case some elements in the interpreter - could we just do the same with forms?

A mutation observer is quite a hammer and likely has performance (memory/cpu) impact that we want to avoid.

mohe2015:
Maybe we should try adding some tests around this behavior and my changes in my PR? So this does not regress again?

ealmloff:
Closed in favor of removing this behavior and fixing reloads on desktop

PR #4422, opened 2025-07-15
Prevent desktop reloads at the wry level

### LINKED ISSUE #4417: Form navigate in webview bricks page (0.7 alpha 2)

ealmloff:
We currently prevent form submissions if you have a `onsubmit` handler in all renderers and don't prevent them if you call `prevent_default`. This behavior can be confusing.

This PR removes that behavior and just prevents the webview from reloading instead

Closes #4417


packages/desktop/Cargo.toml
```diff
@@ -51,6 +51,7 @@ libc = "0.2.174"
 rand = { workspace = true, features = ["std_rng"] }
 subtle = { version = "2.6", features = ["const-generics"] }
 webbrowser = { version = "1.0" }
+url = { workspace = true }
```
ealmloff: This is already a dependency of wry itself, so this shouldn't add to the total dependency count

jkelleyrtp (approved):

jkelleyrtp (approved):

PR #4427, opened 2025-07-16
Example of rendering a dioxus-native app to a texture in a Bevy application

jerome-caucat:
Created in answer to https://github.com/DioxusLabs/dioxus/discussions/4362, and following https://github.com/DioxusLabs/dioxus/pull/4365

This example demonstrates rendering a Dioxus application onto a texture within a Bevy application:
- Render a headless Dioxus-native app to a texture.
- Share and display the texture on a quad in the Bevy app.
- Transmit mouse and keyboard events from Bevy to Dioxus.
  - Events are captured when hovering over elements with the `catch-events` class, ensuring only Dioxus receives them.
- Manage application state through channel messages between Dioxus and Bevy.

Screenshot from 2025-07-16 15-14-55

Notes:
- I did not setup a `NetProvider` for the `DioxusDocument`, as I was not sure how to do it without the event loop proxy.
  - As a result I could not move the CSS to an asset.
- The inputs of type `number` in the UI do not seem to be rendered as spinboxes, I am not sure if this is expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10645,6 +10645,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "native-headless-in-bevy"
+version = "0.0.0"
+dependencies = [
+ "anyrender_vello",
+ "async-std",
+ "bevy",
+ "blitz-dom",
+ "blitz-paint",
+ "blitz-traits",
+ "crossbeam-channel",
+ "dioxus",
+ "dioxus-native-dom",
+ "paste",
+ "rustc-hash 1.1.0",
+ "tracing-subscriber",
+ "vello",
+ "wgpu",
+]
+
+[[package]]
 name = "native-tls"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,6 +113,7 @@ members = [
     "examples/fullstack-websockets",
     "examples/wgpu-texture",
     "examples/native-headless",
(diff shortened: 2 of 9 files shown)


jerome-caucat:
>>>
Notes:
- I did not setup a `NetProvider` for the `DioxusDocument`, as I was not sure how to do it without the event loop proxy.
  - As a result I could not move the CSS to an asset.
- The inputs of type `number` in the UI do not seem to be rendered as spinboxes, I am not sure if this is expected.
>>>

@nicoburns Do you see a simple way to add a `NetProvider` to the `DioxusDocument`? Am I missing some component to render the inputs of type `number` correctly?

nicoburns:
@jerome-caucat https://github.com/rectalogic/bevy_blitz has an example of a bevy NetProvider

nicoburns:
Number inputs simply aren't implemented in Blitz

jerome-caucat:
I did not succeed in adding a working NetProvider capable of loading the assets, but I did not spend too much time on it, as it is not the main goal of this example.

I fixed the CI.

@nicoburns Let me know if you would prefer I add a working NetProvider. Otherwise, this MR is ready to be reviewed.

doxle:
Gents, would this fast as pure bevy app as now we have 2 renderers [Dioxus , Bevy]? Am i right with dx-native, there will not be any dom so it should be fast? Does this run at 60fps?

nicoburns:
If you disabled vsync it would no doubt be slower than without Blitz (of course this is also true if you add things to your bevy scene), but you should be able to hit 60fps pretty easily.

PR #4428, opened 2025-07-16
Add secure context on android

OlivierLemoine:
Hi ! This PR adds the secure context on android.

I needed to access the GPS for an android app, and the best way I found was using the geolocation API on js. However, the webview blocked the location request because the webview interpreted the internal url as `http://dioxus.index.html/`. With a lot of digging, I found this bit of doc : https://docs.rs/wry/latest/wry/struct.WebViewBuilder.html#method.with_custom_protocol.

In the future with dioxus-sdk, I have no doubt that a better solution (without JS) will exist, but for the time being, I think it works well enough. And I don't see any downside to use this function.

I have also added 2 rules in the AndroidManifest. I see that it is a handlebars template, so maybe in the future, this should be a cli option ?

### DIFF
--- a/packages/cli/assets/android/gen/app/src/main/AndroidManifest.xml.hbs
+++ b/packages/cli/assets/android/gen/app/src/main/AndroidManifest.xml.hbs
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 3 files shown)


OlivierLemoine:
I've seen that the AndroidManifest.xml can be modified via Dioxus.toml, so I revert my changes to AndroidManifest.xml.hbs

jkelleyrtp (approved):

jkelleyrtp:
> I've seen that the AndroidManifest.xml can be modified via Dioxus.toml, so I revert my changes to AndroidManifest.xml.hbs

Yep, was going to comment on that. Thanks for the PR!

PR #4433, opened 2025-07-17
Fix Tailwind race condition in fullstack serve mode

### LINKED ISSUE #4431: First time run of dx serve with tailwind results in error
**Problem**

Running `dx serve` for the first time in a workspace after updating to `v0.7.0-alpha.3` with tailwind errors out with an error of `[dev] Build failed: failed to write tailwindcss binary
               1: Text file busy (os error 26) `
Subsequent runs of `dx serve` does not error out and tailwind generates just fine.

**Steps To Reproduce**

- Install  `v0.7.0-alpha.3` dx cli
- `dx serve --fullstack true --package web --platform web`

**Expected behavior**

To generate the `tailwind.css` with no errors.

**Environment:**

- Dioxus version: main (using git dependency from https://github.com/DioxusLabs/dioxus.git)
- Rust version: 1.86.0
- OS info: Linux (Ubuntu 6.8.0-64-generic)
- App platform: web (fullstack web application)

**Questionnaire**

Himmelschmidt:
## Summary
Fixes race condition between Tailwind watcher and prebuild() when running in fullstack serve mode.

## Problem
When running `dx serve --fullstack true --platform web`, both the Tailwind watcher and client build's prebuild() try to install/run Tailwind simultaneously, causing a "Text file busy" error due to concurrent access to the same binary file.

## Solution
Skip starting the Tailwind watcher in fullstack mode since the client build's prebuild() already handles Tailwind generation properly. This eliminates the race condition while maintaining correct functionality.

## Changes
- Modified serve runner to skip Tailwind watcher when `!fullstack` 
- In fullstack mode, prebuild() is the single source of truth for Tailwind generation
- Non-fullstack builds continue to use the Tailwind watcher normally

Fixes #4431

### DIFF
--- a/packages/cli/src/serve/runner.rs
+++ b/packages/cli/src/serve/runner.rs
@@ -166,11 +166,18 @@ impl AppServer {
         let client = AppBuilder::new(&client)?;
         let server = server.map(|server| AppBuilder::new(&server)).transpose()?;
 
-        let tw_watcher = TailwindCli::serve(
-            client.build.package_manifest_dir(),
-            client.build.config.application.tailwind_input.clone(),
-            client.build.config.application.tailwind_output.clone(),
-        );
+        // Only start Tailwind watcher for client builds that serve assets (not server builds or fullstack mode)
+        // In fullstack mode, the client build's prebuild() handles Tailwind generation to avoid race conditions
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Thank you!

PR #4436, opened 2025-07-18
Fix assets with odd extensions and race condition with the same asset at different paths

### LINKED ISSUE #4432: Javascript assets with atypical endings aren't properly invalidated from the asset cache
**Problem**

`dx serve` is not properly recognizing and invalidating changed javascript assets if they have atypical endings like `.out.js`.

**Steps To Reproduce**

- Create a new dioxus project using the jumpstart template
- Set up a minimal `main.rs` like this:
```rust
use dioxus::prelude::*;

const FOO_JS: Asset = asset!("/assets/foo.out.js");

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        document::Script { src: FOO_JS, r#type: "module" }
    }
}
```
- Write something into `/assets/foo.out.js`, e.g.:
```js
export class Foo {}
```
- Start and stop `dx serve`
- Change the contents of `/assets/foo.out.js`, e.g. to:
```js
export class Bar {}
```
- Start `dx serve` again
- Try to retrieve the asset and observe the old version:
```
$ curl localhost:8081/assets/foo.js
class Foo{}export{Foo as Foo};
```

**Expected behavior**

I'd expect the asset to be updated with the new contents automatically, without requiring `cargo clean`.
I would also expect the asset to contain a hash in its filename, but as can be seen in the reproduction steps its just `foo.js`.
If I change the name of `foo.out.js` to `foo.js`, the resulting asset contains a hash and the issue does not occur.

**Environment:**

- Dioxus version:  0.7.0-alpha.3
- Rust version:    1.88.0
- OS info:         NixOS 25.05 + https://github.com/NixOS/nixpkgs/pull/407060
- App platform:    web

**Additional context**
This issue was previously discussed on Discord: https://discord.com/channels/899851952891002890/1395476566468985073

ealmloff:
This fixes two bugs with the asset system:
1) Assets with multiple extensions like `foo.tar.gz` lost the hash because the second extension  with the hash (`tar-hash`) was replaced instead of pushing a the final extension
2) If the asset system includes two files with an identical content hash, but different source paths they were not deduplicated and one of the threads will fail to write the duplicate asset to the output path

Fixes #4432

### DIFF
--- a/packages/cli-opt/src/file.rs
+++ b/packages/cli-opt/src/file.rs
@@ -97,7 +97,8 @@ pub(crate) fn process_file_to_with_options(
     }
 
     // If everything was successful, rename the temp file to the final output path
-    std::fs::rename(temp_path, output_path).context("Failed to rename output file")?;
+    std::fs::rename(temp_path, output_path)
+        .with_context(|| format!("Failed to rename output file to: {}", output_path.display()))?;
 
     Ok(())
 }
--- a/packages/cli-opt/src/hash.rs
+++ b/packages/cli-opt/src/hash.rs
@@ -152,7 +152,10 @@ pub fn add_hash_to_asset(asset: &mut BundledAsset) {
             };
 
             if let Some(ext) = ext {
-                bundled_path.set_extension(ext);
+                // Push the extension to the bundled path. There may be multiple extensions (e.g. .js.map)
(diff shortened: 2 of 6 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Works on the docsite 👍

PR #4438, opened 2025-07-18
Remove dioxus-lib, fix doc imports

### LINKED ISSUE #4338: Make dioxus-lib more usable
Dioxus lib currently doesn't re-export some of the crates required to use dioxus in a library. For example, the component library currently relies on [dioxus](https://github.com/DioxusLabs/components/blob/main/primitives/Cargo.toml#L17) directly because dioxus-lib doesn't re-export the dioxus-fullstack hooks.

At some point dioxus-lib had everything we needed, but it has slowly gotten out of date as the api has changed. To make this easier to maintain we could:
- Change dioxus to depend on and re-export dioxus-lib and the renderers instead of duplicating the prelude
- Remove dioxus-lib and recommend libraries depend on dioxus without any renderer features. We could potentially warn about this by looking for a bin-only cargo environment variable in the build.rs?

ealmloff:
Dioxus lib is currently not very useful because it doesn't include all of the exports required for rsx usage. It has also been difficult to maintain because most of the code is duplicated with the dioxus crate. This PR removes the crate

CI was also not set up to test inline documentation so a bunch of the imports were incorrect with the changes in #4416. This PR also fixes those issues and adds the test to CI

Closes #4338

### DIFF
--- a/packages/fullstack/docs/request_origin.md
+++ b/packages/fullstack/docs/request_origin.md
@@ -1,6 +1,6 @@
 This method interacts with information from the current request. The request may come from:
 
-1. The initial SSR render if this method called from a [`Component`](dioxus::prelude::component) or a [`server`](dioxus_server) function that is called during the initial render
+1. The initial SSR render if this method called from a [`Component`](dioxus_core_macro::component) or a [`server`](dioxus_server) function that is called during the initial render
 
 ```rust
 # use dioxus::prelude::*;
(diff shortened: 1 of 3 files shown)


nicoburns:
This seems reasonable. Can we add a meta-feature to the dioxus crate to replace it (should not enable `launch` or bring in any of the renderers)?

ealmloff:
> This seems reasonable. Can we add a meta-feature to the dioxus crate to replace it (should not enable `launch` or bring in any of the renderers)?

Pulled out a `lib` feature without `launch`, `devtools`, or `logger` which should all be configured from the top level crate

jkelleyrtp (approved):

Issue #3785, opened 2025-02-21
Questions on [bundle.macos]


nigeleke:
I'm currently bundling a macos desktop app, using as many default settings as possible.

I had the following in my `Dioxus.toml`:

```toml
[bundle.macos]
icon = "platform/macos/appicon.png"
```

The bundle failed with "missing field" `files` then subsequently `hardened_resource`. 

The dioxus `bundle.rs` contains:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, Default)]
pub(crate) struct MacOsSettings {
    pub(crate) frameworks: Option<Vec<String>>,
    pub(crate) minimum_system_version: Option<String>,
    pub(crate) license: Option<String>,
    pub(crate) exception_domain: Option<String>,
    pub(crate) signing_identity: Option<String>,
    pub(crate) provider_short_name: Option<String>,
    pub(crate) entitlements: Option<String>,
    pub(crate) info_plist_path: Option<PathBuf>,
    /// List of custom files to add to the application bundle.
    /// Maps the path in the Contents directory in the app to the path of the file to include (relative to the current working directory).
    pub files: HashMap<PathBuf, PathBuf>,
    /// Preserve the hardened runtime version flag, see <https://developer.apple.com/documentation/security/hardened_runtime>
    ///
    /// Settings this to `false` is useful when using an ad-hoc signature, making it less strict.
    pub hardened_runtime: bool,
}
```

Questions:

1. Why aren't `files` and `hardened_runtime` optional?

2. Looking at the code, the same question applies to:
    - DebianSettings::files
    - WixSettings::<various>
    - WindowsSettings::tsp & allow_downgrades
    - NsisSettings::display_language_selector

3. I only really want to provide the icon, thou it's `png` for linux & macos and an `ico` for windows.  Is there a common to specify the icon that is suitable for all platforms?

4. ~~A Discord conversation last night implied these weren't documented, but I can see this morning that they are~~

ealmloff:
We should have defaults with `#[serde(default = "...")]` for the bundle fields where possible. The struct format is copied from the tauri bundler crate to make converting between them easier, but for deserialization we can use a default value if the key isn't present

pythoneer:
[...]

PR #4442, opened 2025-07-20
fix(bundle): add serde default for all fields that impl Default

wdjcodes:
- Adds #[serde(default)] for all fields of all bundle settings structs where the types impl Default
 - WindowsSettings: derive Default since all types impl Default

Default values for more [bundle] settings #3785


wdjcodes:
If there is a reasonable Default behavior for NSISInstallerMode, now might be a good time to `impl Default` for that struct. If that is done I believe all fields for all the different setting type should `impl Default`

Issue #4449, opened 2025-07-21
Move off swc bundler


ealmloff:
The swc bundler is [deprecated](https://swc.rs/docs/usage/bundling) and will be removed in a future release. It also doesn't support some important features like external assets (#4440).

Of the alternatives listed:
- parcel and turbopack don't have crates.io releases
- rspack and farm use nightly rustc

nicoburns:
https://rolldown.rs/ and https://oxc.rs/ look promising.  Might be a bit early though.

PR #4450, opened 2025-07-21
Fix JS bundling fallback

### LINKED ISSUE #4440: `dx serve --release` fails to include dioxus-internal JS scripts in certain cases where wasm-bindgen is used
**Problem**

When compiling a project using wasm-bindgen's `raw_module` attribute in release mode, the minification step seemingly fails and certain scripts end up missing from the assets, leading to the page not working as intended.

If the code is changed to make wasm-bindgen include the snippet itself through `module` and the dioxus asset management of the script is removed, the page works as expected and the scripts that failed to load previously aren't even being requested.

**Steps To Reproduce**

- Create a minimal dioxus project like this:
```rust
// src/main.rs
use dioxus::prelude::*;
use wasm_bindgen::prelude::*;

const BINDINGS_JS: Asset = asset!(
    "/assets/bindings.js",
    AssetOptions::js().with_hash_suffix(false)
);

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    use_effect(|| {
        Foo::new();
    });
    rsx! {}
}

#[wasm_bindgen(raw_module = "/assets/bindings.js")]
extern "C" {
    #[wasm_bindgen]
    pub type Foo;
    #[wasm_bindgen(constructor)]
    pub fn new() -> Foo;
}
```
```toml
# Cargo.toml
[package]
# etc.

[dependencies]
dioxus = { version = "0.7.0-alpha.3", features = ["router", "fullstack"] }
wasm-bindgen = "=0.2.100"

[features]
default = ["web"]
# etc., as generated by the jumpstart template
```
```js
export class Foo {
  constructor() {}
}
```
- Try serving it using `dx serve --release`
- Observe 'Failed to minify js. Falling back to non-minified: failed to bundle javascript with swc' errors in the CLI console and failures when loading scripts in the browser (see screenshots below). Relevant section of the logs with verbose+trace: https://paste.debian.net/hidden/2bc6f13f/ (the CLI outputs _a lot_, please let me know if I should provide the full logs)

Interestingly, the requests in this reproduction all return 200, but with the main pages HTML, while in my actual project where I first encountered the issue the same paths return 404.
Either way they don't contain the necessary scripts and the page fails to load.

**Expected behavior**

I'd expect similar behaviour to running the page in debug mode, which just works and correctly servers the included scripts.

**Screenshots**

Image

Image

Image

(^note the 0B size due to the request being aborted. if the path is requested through `curl` I get the `index.html` instead)

Image

(^screenshot from my own project)

**Environment:**

- Dioxus version: 0.7.0-alpha.3
- Rust version:    1.88.0
- OS info:         NixOS 25.05
- App platform:    web

**Additional context:**
This issue was first discussed on Discord: https://discord.com/channels/899851952891002890/1395874448606757035

The issue affects my project here: https://codeberg.org/bikeshedding-pad/pad as of `f3d1f0da63`

ealmloff:
If any wasm bindgen assets include links to external assets, the swc bundler fails to process them. This PR adds a CLI argument to skip the SWC bundler with wasm bindgen assets and instead hash and optimize the whole folder. The regression test added in the playwright folder has an example of a project that requires this flag. If you use a raw module link in wasm-bindgen and bundle your js asset separately, the swc bundler fails to read the link as a file:
```rust
#[used]
static BINDINGS_JS: Asset = asset!(
    "/assets/bindings.js",
    AssetOptions::js().with_hash_suffix(false)
);

#[wasm_bindgen(raw_module = "/assets/bindings.js")]
extern "C" {
    #[wasm_bindgen]
    pub type Foo;
    #[wasm_bindgen(constructor)]
    pub fn new() -> Foo;
}
```

Most bundlers include an option to exclude specific files from the bundling process for external assets, but it looks like swcpack does not support that feature. The SWC bundler is also deprecated and will be removed in the next release so we should start thinking about a migration plan. There is a separate issue discussing other bundlers here: #4449

Fixes #4440


jkelleyrtp:
Is SWC just failing here because we're not copying the referenced assets over? Maybe we could customize the path loader config argument to use the crate's root as a fallback.

I'm not a huge fan of adding a CLI arg to work around a bug rather than fixing the underlying issue itself, if we can.

ealmloff:
This is an issue with the swc bundler not being flexible enough for some use cases which isn't easy to fix without changing swc itself. SWC is deprecated so I don't think we should spend too much time working around issues like this. I made the fallback automatic which doesn't require a new flag, but it does mean the js optimization happens serially before instead of parelell with the other asset optimization

jkelleyrtp:
I think the switchover to esbuild might either solve this, or let us exclude files from the bundling/packing to ignore them. 

Will take a look

----

edit: raw_module is just not great for us. we would need to add explicit support for its imports, potentially by walking the emitted bindgen js? The problem here is that the raw_module script leaves the path resolution to the user, but since we assemble the app in a different folder than the root directory, we'll end up with different "root" locations.

jkelleyrtp:
Now that we have esbuild we can handle this automatically.

https://github.com/DioxusLabs/dioxus/pull/5515

PR #4459, opened 2025-07-23
Update README.md

benatkin:
This gives it a link to the latest alpha – I think updating the link with each alpha until 0.7 is released might be worth it, but maybe after that it would make sense to give it a link that doesn't change

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@
 </div>
 <br>
 <p align="center">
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I'd like to keep the link in readme.md pointing to the first post since that's the most detailed post with the feature overview. 

I've updated that post to point towards the most recent alpha in the top disclaimer, so at least people know where to start looking

PR #4471, opened 2025-07-25
fix: Use Rc for lifecycle management

yydcnjjw:


ealmloff:
This fixes https://github.com/DioxusLabs/dioxus/issues/4512. The implementation of `Clone` and `Drop` for `LifeCycle` in use_drop were poorly behaved such that `lifecycle` is not equal to `lifecycle.clone()`. We used to return a cloned version of the value and keep the original version to drop later. That changed in https://github.com/DioxusLabs/dioxus/pull/4381 to return the original version of the value passed into use_hook and store a cloned version. Changing to Rc fixes the behavior so we don't rely on which value is stored and which one is returned

PR #4473, opened 2025-07-26
Downgrade chrono to 0.4.39

DanielWarloch:
Certain packages do not yet support chrono 0.4.40
Examples:
arrow-arith which is used by influxdb3

Arrow-rs will have to update chrono, but as this is deep in dependency chain it will take time to do so. So as we do not require features from chrono >0.4.39, I propose to keep it at 0.4.39.


nicoburns:
Do you have more information on why .40 isn't supported? It is suprising that a patch release requires explicit support...

mohe2015:
See https://github.com/apache/arrow-rs/issues/7196 and https://github.com/chronotope/chrono/pull/1666

nicoburns (approved):

PR #4483, opened 2025-07-29
Stores: Externally tracked state and reactive collections

ealmloff:
This is stacked on top of #4413 with the store implementation. I expect the design for stores to take longer to resolve so this is pulled out into a separate non-breaking PR which could be released in a patch version in 0.7

We could publish stores as an external crate outside of the prelude with a decoupled version for the first few releases. Stores don't integrate deeply with any other crates in the workspace so maintaining a separate version seems relatively simple

TODO:
- [x] Generic store selectors
- [x] Vec/slice/array selectors
- [x] Hashmap selector
- [x] Result selector
- [x] Option selector
- [x] Box selector
- [x] derive(Store) on structs
- [x] derive(Store) on enums
- [x] derive(Store) on unit types
- [x] implement readable on intermediate stores
- [x] implement writable on intermediate stores
- [x] Documentation


examples/todomvc_store.rs
```diff
@@ -0,0 +1,274 @@
...
+                    onclick: move |_| todos.todos().retain(|_, todo| !todo.checked),
+                    "Clear completed"
+                }
+            }
+        }
+    }
+}
+
+#[derive(Store, PartialEq, Clone, Debug)]
+struct TodoState {
```
jkelleyrtp: Let's move this to the top of the example with some explanation on what the `Store` macro expands to.

packages/core-macro/src/props/mod.rs
```diff
@@ -1769,6 +1769,17 @@ fn looks_like_write_type(ty: &Type) -> bool {
...
+                || path_without_generics == parse_quote!(prelude::Store)
+                || path_without_generics == parse_quote!(Store)
+        }
+        None => false,
+    }
+}
+
 fn looks_like_callback_type(ty: &Type) -> bool {
     let type_without_option = remove_option_wrapper(ty.clone());
     match extract_base_type_without_generics(&type_without_option) {
```
jkelleyrtp: same comment from the base PR here - maybe it'd make sense to add a check if the final path segment is `Store`?

packages/stores/src/impls/hashmap.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// ```rust, no_run
+    /// use dioxus_stores::*;
+    /// use dioxus::prelude::*;
+    /// use std::collections::HashMap;
+    /// let mut store = use_store(|| HashMap::new());
+    /// assert_eq!(store.len(), 0);
+    /// store.insert(0, "value".to_string());
+    /// assert_eq!(store.len(), 1);
+    /// ```
+    pub fn len(self) -> usize {
```
jkelleyrtp: these methods probably should take &self so we still get scope-level borrow checking for signals

packages/stores/src/impls/hashmap.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// use dioxus::prelude::*;
+    /// use std::collections::HashMap;
+    /// let mut store = use_store(|| HashMap::new());
+    /// store.insert(0, "value1".to_string());
+    /// store.insert(1, "value2".to_string());
+    /// for (key, value_store) in store.iter() {
+    ///     println!("{}: {}", key, value_store.read());
+    /// }
+    /// ```
+    pub fn iter(
```
jkelleyrtp, line 70: for the iter methods, might be nice to add a "VecIter" iterator type to enable all the double-ended, exact-size, etc extensions

packages/stores/src/impls/hashmap.rs
```diff
@@ -0,0 +1,289 @@
...
+                W,
+                impl Fn(&HashMap<K, V, St>) -> &V + Copy + 'static,
+                impl Fn(&mut HashMap<K, V, St>) -> &mut V + Copy + 'static,
+            >,
+        >,
+    >
+    where
+        Q: Hash + Eq + Copy + 'static,
+        K: Borrow<Q> + Eq + Hash,
+        St: BuildHasher,
```
jkelleyrtp: the Copy bound might be over-restrictive especially for String keys (clone)

packages/stores/src/impls/option.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// match transposed {
+    ///     Some(inner_store) => assert_eq!(inner_store(), 42),
+    ///     None => panic!("Expected Some"),
+    /// }
+    /// ```
+    pub fn transpose(
+        self,
+    ) -> Option<
+        Store<
+            T,
```
jkelleyrtp: maybe expand the names here for readability

packages/stores/src/subscriptions.rs
```diff
@@ -0,0 +1,286 @@
...
+                node.remove(rest);
+            }
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+pub(crate) struct TinyVec {
+    length: usize,
+    path: [u32; 32],
```
jkelleyrtp: this is pretty big (128 bytes)

packages/stores/src/store.rs
```diff
@@ -0,0 +1,324 @@
...
+///             for value in children.iter() {
+///                 li {
+///                     Tree { value }
+///                 }
+///             }
+///         }
+///     }
+/// }
+/// ```
+pub struct Store<T: ?Sized, W = WriteSignal<T>> {
```
jkelleyrtp: mentioned in the other PR - but the `From<T>` for `ReadSignal<T>` decay logic doesn't include stores. maybe a "Decay" trait could be used to allow stores to implement the from logic without sharing the store impl in signals

packages/stores/src/impls/hashmap.rs
```diff
@@ -0,0 +1,289 @@
+use std::{
```
jkelleyrtp, line 1: maybe a btreemap too? a sorted set with `.top()` or `.last()` would be nice. `VecDeque` potentially?

packages/stores-macro/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+        syn::Data::Struct(data_struct) => derive_store_struct(&input, data_struct),
+        syn::Data::Enum(data_enum) => derive_store_enum(&input, data_enum),
+        syn::Data::Union(_) => Err(syn::Error::new(
+            input.span(),
+            "Store macro does not support unions",
+        )),
+    }
+}
+
+fn derive_store_struct(input: &DeriveInput, structure: &DataStruct) -> syn::Result<TokenStream2> {
```
jkelleyrtp: please add some docs here about the steps we're taking to do the transformation, and maybe some docs inline for the weird stuff

packages/stores/src/subscriptions.rs
```diff
@@ -0,0 +1,286 @@
...
+    sync::Arc,
+};
+
+#[derive(Clone, Default)]
+pub(crate) struct SelectorNode {
+    subscribers: HashSet<ReactiveContext>,
+    root: HashMap<u32, SelectorNode>,
+}
+
+impl SelectorNode {
```
jkelleyrtp, line 21: some comments here about the use of these methods could be nice?

examples/todomvc_store.rs
```diff
@@ -0,0 +1,274 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[component]
+fn ListFooter(mut todos: Store<TodoState>, active_todo_count: ReadSignal<usize>) -> Element {
```
jkelleyrtp, line 266: ReadStore/WriteStore could be a useful delineation?

jkelleyrtp:
I'm messing around with stores and noting some observations here:
- would be ideal if you were forced to make the lenses carry the mut somehow to force you to declare the `use_store` return as mutable
- I'm missing string methods  (ie clear) on lenses
- Missing `.expect()` on  result/option (and probably more methods like is_err)
- The store macro impl is so good that renaming the `.title` field renames the lens too(!!)
- I would like `.iter()` on `Signal<HashMap>`, didn't realize we didn't have that already
- `.boxed_mut()` works on Store, but `.boxed() -> ReadSignal` does not

---

Also, am I misunderstanding how recursive stores work? I want to get the `title` of the `inner` headerstate

```rust

#[derive(Store)]
struct HeaderState {
    title: String,
    subtitle: String,
    inner: Option<Box<HeaderState>>,
}

fn app2() -> Element {
    let header = use_store(|| HeaderState {
        title: "Hello, ".to_string(),
        subtitle: "world!".to_string(),
        inner: Some(Box::new(HeaderState {
            title: "Hello, ".to_string(),
            subtitle: "world!".to_string(),
            inner: None,
        })),
    });

    let r: Store<Box<HeaderState>, _> = header.inner().unwrap();
```


---

Seems like we need superinto/superfrom for stores into ReadSignal:

```rust
pub struct SuperFromSignalMarker;
impl<T, W> SuperInto<ReadSignal<T>, SuperFromSignalMarker> for Store<T, W>
where
    T: 'static,
    W: Readable<Target = T, Storage = UnsyncStorage> + 'static,
{
    fn super_into(self) -> ReadSignal<T> {
        ReadSignal::new(self)
    }
}
```

ealmloff:
Yes, we need more readable helpers. I think the string methods would fit better on the normal writable extension traits because the key type isn't clear with strings. It could be either chars or bytes.

> Also, am I misunderstanding how recursive stores work? I want to get the `title` of the `inner` headerstate
> 
> ```rust
> #[derive(Store)]
> struct HeaderState {
>     title: String,
>     subtitle: String,
>     inner: Option<Box<HeaderState>>,
> }
> 
> fn app2() -> Element {
>     let header = use_store(|| HeaderState {
>         title: "Hello, ".to_string(),
>         subtitle: "world!".to_string(),
>         inner: Some(Box::new(HeaderState {
>             title: "Hello, ".to_string(),
>             subtitle: "world!".to_string(),
>             inner: None,
>         })),
>     });
> 
>     let r: Store<Box<HeaderState>, _> = header.inner().unwrap();
> ```

You can get to the title by calling deref and then title like this:
```rust

#[derive(Store)]
struct HeaderState {
    title: String,
    subtitle: String,
    inner: Option<Box<HeaderState>>,
}

fn app() -> Element {
    let header = use_store(|| HeaderState {
        title: "Hello, ".to_string(),
        subtitle: "world!".to_string(),
        inner: Some(Box::new(HeaderState {
            title: "Hello, ".to_string(),
            subtitle: "world!".to_string(),
            inner: None,
        })),
    });

    let title: Store<String, _> = header.inner().unwrap().deref().title();

    rsx! {
        div {
            "{title}"
        }
    }
}
```

We need an explicit deref method both because we don't have a reference to the version of the deref mapped store object to return a reference to and we already have a deref implementation for the callable api

packages/stores/src/impls/hashmap.rs
ealmloff: If the only thing we want out of the custom type is the extra bounds, we could just make the return bounds more strict. Then we don't need to manually forward the implementations for all 4 traits for all of the iterators we expose

ealmloff:
I added a bunch more methods to result and option for stores. There are still some methods we don't support on Result, Option, and data structures like String that won't benefit from an optimized Store version. I would rather add those in a followup PR as this PR is already large and those changes are unrelated

PR #4487, opened 2025-07-30
fix: Handle undefined modifier keys in browser autofill events

### LINKED ISSUE #4486: WASM client crash on keyboard event being undefined
**Problem**

Fullstack web app client side crashes with a panicked at `already borrowed: BorrowMutError` when a user triggers browser autofill on a form.

I think the cause is that the browser autofill can dispatch synthetic keyboard events where modifier keys (altKey, ctrlKey, etc.) have a value of undefined. The Dioxus WASM bindings expect these properties to always be `booleans`. When the code attempts to access these properties (e.g., via `event.data.modifiers()`), the wasm-bindgen glue code fails with an `expected a boolean argument, found undefined` error.

**Steps To Reproduce**

- Create a Dioxus web application with a form containing `<input>` fields
- Add a keyboard event handler (e.g., `onkeydown`) that accesses the event's modifier keys using `event.data.modifiers()`
- Use the browser's autofill feature to complete the form fields
- The application crashes with a `BorrowMutError` when autofill triggers synthetic keyboard events

Minimal reproduction example:
```
  rsx! {
      div {
          // This handler will be triggered by autofill's synthetic events
          onkeydown: move |event| {
              // This line panics because autofill events have undefined modifier keys
              let modifiers = event.data.modifiers();
              println!("Modifiers: {:?}", modifiers);
          },
          input { name: "username", placeholder: "Username" }
          input { name: "password", r#type: "password", placeholder: "Password" }
      }
  }
```

I know that many browsers all handle autofill in different ways, this error happened on microsoft edge (chromium based), although I think it would be a good thing to handle regardless.

**Expected behavior**

The application should handle synthetic keyboard events from browser autofill gracefully without crashing. When modifier key values are undefined, they should be safely coerced to false, allowing the event handler to execute without error.

**Environment:**

- Dioxus version: main
- Rust version: 1.88
- OS info: Windows (but I think this applies to any)
- App platform: web (fullstack)

**Questionnaire**

Add `#[serde(default)]` attributes to the modifier key fields in `SerializedKeyboardData` struct (`packages/html/src/events/keyboard.rs`):

  ```#[serde(default)]
  alt_key: bool,
  #[serde(default)]
  ctrl_key: bool,
  #[serde(default)]
  meta_key: bool,
  #[serde(default)]
  shift_key: bool,```

Himmelschmidt:
Browser autofill can dispatch synthetic keyboard events where modifier keys (`altKey`, `ctrlKey`, etc.) have undefined values instead of `booleans`. This causes WASM binding errors and crashes the application with `BorrowMutError`.

Added `serde(default)` attributes to all modifier key fields in `SerializedKeyboardData` to gracefully handle undefined values by defaulting them to false.

Fixes #4486

### DIFF
--- a/packages/html/src/events/keyboard.rs
+++ b/packages/html/src/events/keyboard.rs
@@ -105,9 +105,13 @@ pub struct SerializedKeyboardData {
     key_code: KeyCode,
     #[serde(deserialize_with = "resilient_deserialize_code")]
     code: Code,
+    #[serde(default)]
     alt_key: bool,
+    #[serde(default)]
     ctrl_key: bool,
+    #[serde(default)]
     meta_key: bool,
+    #[serde(default)]
     shift_key: bool,
     location: usize,
     repeat: bool,


Himmelschmidt:
I did not test this enough, this did not fix the issue

PR #4492, opened 2025-07-30
Fix/browser autofill crash

### LINKED ISSUE #4486: WASM client crash on keyboard event being undefined

Himmelschmidt:
1. Core Runtime Fix (packages/core)

- Fixed `BorrowMutError` during rapid event firing (like browser autofill)
- Restructured event handling to collect data first, then drop `RefCell` borrows before calling handlers
- Added defensive `try_borrow_mut()` in `ListenerCallback` for edge cases  

2. TypeScript/JavaScript (packages/interpreter)

- Added undefined checks for modifier keys at the start of `serializeEvent()`
- Ensures all modifier keys default to false if undefined

2. Rust - Web Event Handling (packages/web)

- Added safe event conversion with `dyn_into` and fallback to default values
- Removed unsafe `unchecked_into` conversions
- Made `ClipboardEvent` use specific type instead of generic `Event`

3. Rust - HTML Event Types (packages/html)

- Added Default implementations for all event data types
- Created `EmptyEvent` type as fallback for malformed events
- Added `#[serde(default)]` to modifier keys in `SerializedPointInteraction`

Per my testing this seems to fix #4486


packages/core/src/events.rs
```diff
@@ -654,7 +654,26 @@ impl<T> ListenerCallback<T> {
...
+                    let callback = self.callback.clone();
+                    let event = event.clone();
+
+                    crate::spawn(async move {
+                        if let Ok(mut cb) = callback.try_borrow_mut() {
+                            cb(event);
+                        }
+                    });
+                }
+            }
```
jkelleyrtp: Haven't looked through the whole PR, but the BorrowMut error happens only in WASM after the app has panicked once. The runtime refcell is left in an open state. Since panics don't unwind in WASM, the Ref is not dropped and the refcell is stuck open.

When the next event is handled on the javascript side, we call a WASM callback regardless even if the Rust side has crashed.

I'd like to fix this in WASM, but we have to work around the fact that there aren't unwind tables and we won't know how to bring the app back into a recovered state.

We shouldn't need to modify this handler code. We also don't want to re-queue the event handler since it breaks synchronous event handling.

Himmelschmidt: Ok, I dealt with that issue first because I had figured it was easier to deal with, but based on your comment I was probably seeing it due to the panic of the undefined instead of a bool in the WASM. I've reverted that part of the PR and will test to ensure the fix still works

Himmelschmidt: yep this was unnecessary

PR #4508, opened 2025-08-05
feat: support openharmony

richerfu:
This is an experiential adaption for openharmony, and i made an example to test it https://github.com/richerfu/dioxus-in-openharmony.

There are some issues need to fix:

1. Adapt the `AssetManager`
2. lifecycle for handle_start_cause_init  `Resume` will crash when reopen the application.
3. JS crash

### DIFF
--- a/packages/native/src/lib.rs
+++ b/packages/native/src/lib.rs
@@ -139,11 +139,6 @@ pub fn launch_cfg_with_props<P: Clone + 'static, M: 'static>(
 
     let navigation_provider = Some(Arc::new(DioxusNativeNavigationProvider) as _);
 
-    #[cfg(feature = "html")]
-    let html_parser_provider = Some(std::sync::Arc::new(blitz_html::HtmlProvider) as _);
-    #[cfg(not(feature = "html"))]
-    let html_parser_provider = None;
-
     // Create document + window from the baked virtualdom
     let doc = DioxusDocument::new(
         vdom,


jkelleyrtp:
Thanks for working on this!! Would love to eventually add ohos as a `dx` target as well to get full hot-reload, bundle, deploy, etc support

richerfu:
> Thanks for working on this!! Would love to eventually add ohos as a `dx` target as well to get full hot-reload, bundle, deploy, etc support

Sure, but we need to make sure that can run normally at first. And then we can do more works for `dx`. :)

jkelleyrtp:
> > Thanks for working on this!! Would love to eventually add ohos as a `dx` target as well to get full hot-reload, bundle, deploy, etc support
> 
> Sure, but we need to make sure that can run normally at first. And then we can do more works for `dx`. :)

of course :)

I assume wry/tao will need changes to support dioxus and then those will need to be upstreamed before we can merge any of this work, unless we want to bake some of the work-arounds into dioxus-desktop itself?

It would also be interesting to see what changes need to be made to support dioxus-native. Does ohos provide a native UI kit? Dioxus-native can paint directly to WGPU textures as well.


---

Also, I recently merged a PR that vendors prebuilt openssl binaries for use in android. Seems like your ohos-rs tool does something similar, or at least requires custom openssl binaries?

https://github.com/DioxusLabs/dioxus/pull/4511

richerfu:
Yeah, i think we need to merge the upstream of wry and tao at first and then we can merge this work. I'll try to finish it as soon as possible.

> Does ohos provide a native UI kit? 

Sure, we can use WGPU in ohos which is powered by OpenGL.  Here is an example: https://github.com/richerfu/wgpu-demo-with-winit

> Also, I recently merged a PR that vendors prebuilt openssl binaries for use in android. Seems like your ohos-rs tool does something similar, or at least requires custom openssl binaries?

We can also use openssl verdor mode. I'll support it later.

Issue #4514, opened 2025-08-06
WASM splitter errors with Option::unwrap() on a None value


Himmelschmidt:
**Problem**

When running `dx build --verbose --fullstack true --package app-web --web --wasm-split --profile wasm-split-release` on a fullstack workspace app, the build errors out during the splitting process with: `Option::unwrap() on a None value`.

**Steps To Reproduce**

- Enable `wasm-split` feature in the dioxus and dioxus-router deps
- Run `dx build --fullstack true --package your-package --web --wasm-split`

Here is the backtrace:
```
thread 'tokio-runtime-worker' panicked at packages/wasm-split/wasm-split-cli/src/lib.rs:1331:22:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: wasm_split_cli::ModuleWithRelocations::get_symbol_dep_node
   5: wasm_split_cli::Splitter::new
   6: dx::build::request::BuildRequest::bundle_web::{{closure}}
   7: dx::build::request::BuildRequest::build::{{closure}}
   8: dx::build::builder::AppBuilder::start::{{closure}}
   9: tokio::runtime::task::core::Core<T,S>::poll
  10: tokio::runtime::task::harness::Harness<T,S>::poll
  11: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  12: tokio::runtime::scheduler::multi_thread::worker::Context::run
  13: tokio::runtime::context::scoped::Scoped<T>::set
  14: tokio::runtime::context::runtime::enter_runtime
  15: tokio::runtime::scheduler::multi_thread::worker::run
  16: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  17: tokio::runtime::task::core::Core<T,S>::poll
  18: tokio::runtime::task::harness::Harness<T,S>::poll
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
  27.164s DEBUG dx::build::builder: Setting builder to failed state
  27.165s ERROR dx::build::builder: err=Build panicked! JoinError::Panic(Id(19), "called `Option::unwrap()` on a `None` value", ...)

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
   1: anyhow::__private::format_err
[...]

PR #4515, opened 2025-08-06
fix: resolve wasm-split panic on missing function names (#4514)

Himmelschmidt:
- Replace unwrap() with get().map() in get_symbol_dep_node() to handle missing function names gracefully
- Add bounds checking for symbol index
- Disable function deletion in prune_main_symbols(), let Walrus GC handle cleanup
- Improve unused symbol detection to preserve shared symbols needed for indirect function table

### DIFF
--- a/packages/wasm-split/wasm-split-cli/src/lib.rs
+++ b/packages/wasm-split/wasm-split-cli/src/lib.rs
@@ -167,7 +167,8 @@ impl<'a> Splitter<'a> {
         //    When the side modules load, they will initialize functions into the table where the "holes" are.
         self.replace_segments_with_holes(&mut out, &unused_symbols);
 
-        // 2. Wipe away the unused functions and data symbols
+        // 2. Wipe away the split point exports and clean up data symbols
+        // Note: Function cleanup is left to the GC which handles it safely
         self.prune_main_symbols(&mut out, &unused_symbols)?;
 
         // 3. Change the functions called from split modules to be local functions that call the indirect function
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
do you have a test case that's tripping up the wasm-split engine?

We can't just rip out the function deletion logic since that's what's driving the splitter. It's likely symbol names aren't flowing through to the splitting step, likely because of your release profile, or another issue.

WASM parsing is pretty sensitive to the Rust version since LLVM changes its code-gen occasionally, so it might be an issue with rust version, wasm-opt version, and/or wasm-bindgen version along with cargo profiles.


---

You might need to enable `debug = true` when using wasm-split, otherwise we don't get enough symbol information to build the callgraph.

```toml
# crank up the opt level for wasm-split-cli in dev mode
# important here that lto is on and the debug symbols are present (since they're used by wasm-opt)
[profile.wasm-split-release]
inherits = "release"
opt-level = 'z'
lto = true
debug = true
```

Himmelschmidt:
> do you have a test case that's tripping up the wasm-split engine?

- `dx new` and follow the prompts to make a workspace
- enable the `wasm-split` and `dioxus-router` deps in the crates that have the router feature enabled
- `dx build --fullstack true --package web --web --wasm-split` or `dx build --fullstack true --package web --web --wasm-split --profile wasm-split-release`
- `[dev] Build failed: Build panicked! JoinError::Panic(Id(29), "called Option::unwrap() on a None value", ...) `

> We can't just rip out the function deletion logic since that's what's driving the splitter. It's likely symbol names aren't flowing through to the splitting step, likely because of your release profile, or another issue.

This is confusing to me because with my PR, it seemed to actually split them, i.e. I saw much smaller app sizes and they seemed to be split via route, so I figured it's just one piece of the `wasm-split` puzzle

Himmelschmidt:
Closing this as setting 
```
lto = true
debug = true
```
in the release profile fixed it as suggested above

jkelleyrtp:
I will follow this PR up with a PR that sets the appropriate flags for you automatically.

PR #4522, opened 2025-08-08
Split docs-rs and normal readmes

ealmloff:
We use the readme.md file as the homepage for most crates on docs.rs, but the rendering is slightly different. This PR adds a tiny new CLI that converts the docs.rs readme to the normal readme for crates.io and github and points to a README-docs-rs.md file for all crates.

The diff is quite large, but the actual changes here are minimal. Mostly copying over the files to the docs-rs variant


jkelleyrtp:
Sorry this has sat for so long! Will probably come through with claude and just do the various changes by hand instead of maintaining/automating it with the CLI

Issue #4517, opened 2025-08-07
Rename `read_unchecked` methods


mcmah309:
`read_unchecked` ([e.g](https://docs.rs/dioxus/latest/dioxus/prelude/trait.Readable.html#method.read_unchecked)) is poorly named. Whenever I use it I feel like I am doing something unsafe like with the unchecked functions in the std lib. There really is nothing unchecked here - the lifetime is not unchecked, it is extended. A better name would be e.g. `read_static` or `read_runtime_checked` or `read_extended` since it gives the read `Ref` a static lifetime and relies on runtime checking instead.

PR #4540, opened 2025-08-13
Rename unchecked methods to extended

mcmah309:
This renames all "unchecked" methods to better define their behavior. "unchecked" methods in Rust are traditionally used to signify methods that can have undefined behavior if used incorrectly. These methods do not have undefined behavior. These methods are completely safe and what is really happening is the lifetime is extended to `'static` and runtime checks are used to enforce safety instead. Thus `unchecked` is renamed to `extended`.

closes https://github.com/DioxusLabs/dioxus/issues/4517

### DIFF
--- a/examples/dog_app.rs
+++ b/examples/dog_app.rs
@@ -75,7 +75,7 @@ fn BreedPic(breed: WriteSignal<String>) -> Element {
             .await
     });
 
-    match fut.read_unchecked().as_ref() {
+    match fut.read_extended().as_ref() {
         Some(Ok(resp)) => rsx! {
             div {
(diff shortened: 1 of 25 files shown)


mcmah309:
@jkelleyrtp I added back the `unchecked` methods as deprecated default implemented members of the traits as per our discussion https://discord.com/channels/899851952891002890/928812591126569000/1407446072875417681

jkelleyrtp:
I think now that rust has better lifetime management in edition 2024, users will run into this less, and I don't necessarily think it's worth the rename anymore. Thanks for the PR anyways!

mcmah309:
If that case, keeping the methods around may still cause confusion/bloat. Should they removed then?

PR #4552, opened 2025-08-15
Deprecate ReadOnlySignal

mcmah309:
Removes `ReadOnlySignal` as it was renamed to `ReadSignal`. Keeping `ReadOnlySignal` will only lead to more code debt and confusion in the future.


jkelleyrtp:
I'd rather not remove ReadOnlySignal this release. I assume libraries in the ecosystem use it on their boundaries.

Instead we should just deprecate its use with a `#[deprecate()]` annotation.

mcmah309:
I added it back and marked it as deprecated

PR #4556, opened 2025-08-15
Don't output escape codes for the cursor when the tui isn't active

### LINKED ISSUE #4535: `dx fmt` adding terminal codes to output in rc.0
**Problem**

After updating from alpha.3 to rc.0, `dx fmt --file -` started adding the characters `[?2h` to the end of the output causing it to be appended visibly to the end of every edited file when I have it setup for formatting in neovim. I think it's the code for showing the cursor according to https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#common-private-modes

Note it also has been adding an extra newline `\n` to the output even in alpha.3 and I had just been stripping that, might be nice to stop adding that too.

It might be a good idea to change behavior based on [atty](https://crates.io/crates/atty) or something if either are needed to make it function nicely in the terminal.

**Steps To Reproduce**

- `cargo binstall dioxus-cli --version=0.7.0-rc.0 --force`
- `dx fmt --file - <<< 'rsx! { div { "test" } }' | od -c`
You should see:
```
0000000   r   s   x   !       {  \n                   d   i   v       {
0000020       "   t   e   s   t   "       }  \n   }  \n 033   [   ?   2
0000040   5   h
0000042
```
Note the `033   [   ?   2 5 h` at the end

**Expected behavior**

Here's the behavior of alpha.3:

- `cargo binstall dioxus-cli --version=0.7.0-alpha.3 --force -y`
- `dx fmt --file - <<< 'rsx! { div { "test" } }' | od -c`
You should see:
```
0000000   r   s   x   !       {  \n                   d   i   v       {
0000020       "   t   e   s   t   "       }  \n   }  \n
0000034
```

**Environment:**

- Dioxus version:  0.7.0-rc.0
- Rust version:    1.89
- OS info: EndeavorOS (Arch based rolling distro)
- App platform: web

**Questionnaire**

ealmloff:
Fixes #4535

### DIFF
--- a/packages/cli/src/logging.rs
+++ b/packages/cli/src/logging.rs
@@ -273,7 +273,7 @@ impl TraceController {
             tui_rx: Arc::new(tokio::sync::Mutex::new(tui_rx)),
             telemetry_rx: Arc::new(tokio::sync::Mutex::new(telemetry_rx)),
             http_client,
-            tui_active: tui_active.clone(),
+            tui_active,
         };
 
         // Spawn the telemetry uploader in the background
--- a/packages/cli/src/logging.rs
+++ b/packages/cli/src/logging.rs
@@ -263,7 +263,7 @@ impl TraceController {
             tui_rx: Arc::new(tokio::sync::Mutex::new(tui_rx)),
             telemetry_rx: Arc::new(tokio::sync::Mutex::new(telemetry_rx)),
             http_client,
-            tui_active,
+            tui_active: tui_active.clone(),
         };
 
         // Spawn the telemetry uploader in the background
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Technically the show_cursor code is meant for the dx new and dx init commands in the event cargo generate fails. The TUI being active is not set there, so we wouldn't be catching that case.

We can extract the tui_active bool back to a static and then set it in the relevant places. Should be nicer to not pass it around as an arg.

I think there might be a way to check if "alternate mode" is enabled without our own book. IIRC there's a method on stdout that lets you check

ealmloff:
Ah, tui_active will also silence logs which I don't think we want for the dx new and dx init commands. I switched to checking if stdout is a terminal

jkelleyrtp (approved):

PR #4557, opened 2025-08-15
Improve use_server_cached docs

mcmah309:
Improve `use_server_cached` docs

### DIFF
--- a/packages/fullstack-hooks/src/hooks/server_cached.rs
+++ b/packages/fullstack-hooks/src/hooks/server_cached.rs
@@ -2,9 +2,12 @@ use dioxus_core::use_hook;
 use dioxus_fullstack_protocol::SerializeContextEntry;
 use serde::{de::DeserializeOwned, Serialize};
 
-/// This allows you to send data from the server to the client. The data is serialized into the HTML on the server and hydrated on the client.
+/// This allows you to send data from the server to the client *during hydration*.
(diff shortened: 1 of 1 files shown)


packages/fullstack-hooks/src/hooks/server_cached.rs
```diff
@@ -2,9 +2,12 @@ use dioxus_core::use_hook;
 use dioxus_fullstack_protocol::SerializeContextEntry;
 use serde::{de::DeserializeOwned, Serialize};
 
-/// This allows you to send data from the server to the client. The data is serialized into the HTML on the server and hydrated on the client.
+/// This allows you to send data from the server to hydrate the client.
+/// - When compiled as server, the closure is ran and the resulting data is serialized on the server and sent to the client.
+/// - When compiled as web client, the data is deserialized from the server.
```
ealmloff: The closure can also be run on the client if the hook is run for the first time on the client. For example, if you have a component that calls a endpoint inside of `use_server_cached` on a certain route, then the hook will be run on the client if you navigate to the route on the client

mcmah309: I see. I think this hook makes a lot more sense to me now. How does it look now?

ealmloff: Much more clear

ealmloff (approved):
Thanks, this is much more clear!

PR #4563, opened 2025-08-16
Fix integration of SSL libs into non-ARM64 Android builds.

priezz:
When building for Android target `libssl.so` and `libcrypto.so` were always put to `app/src/main/jniLibs/arm64-v8a` regardless the target architecture.

This PR fixes that and selects the proper destination directory (`arm64-v81`, `armeabi-v7a`, `x86` or `x86_64`).


jkelleyrtp (approved):

jkelleyrtp:
Thank you!

PR #4584, opened 2025-08-20
ensure wasm-split has lto and debug enabled

jkelleyrtp:
In a previous PR (https://github.com/DioxusLabs/dioxus/pull/4515) a user was confused that wasm-split didn't work unless LTO and debug symbols were enabled. 

This PR makes lto and debug automatic if --wasm-split is passed. If the user is customizing their profile for release, then we throw a warning instead.

This also fixes a secondary issue where we weren't setting the proper release profile setting for wasm.


197g:
There's another related setting that must _not_ be active. I was quite confused:

```
[profile.release]
strip = "symbols"
```

`strip = "debuginfo"` works but note that `strip = true` is equivalent to `"symbols"`.

PR #4588, opened 2025-08-20
feat: add subsecond serve

jkelleyrtp:
This PR adds a `dioxus_devtools::subsecond_serve` which makes it easier to run the subsecond hot-reload engine outside of dioxus.

This should handle cases like TUIs, Axum, Burn, and other hot-reload systems that want a "hot" async integration.

For example, an axum app that gets hot-reload looks like so:
```rust
#[tokio::main]
async fn main() {
    dioxus_devtools::serve_subsecond(router_main).await;
}

async fn router_main() {
    use axum::{Router, routing::get};

    let app = Router::new().route("/", get(test_route));

    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
    println!("Server running on http://localhost:3000");

    axum::serve(listener, app.clone()).await.unwrap()
}

async fn test_route() -> axum::response::Html<&'static str> {
    "axum works!!!!!".into()
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5398,6 +5398,8 @@ dependencies = [
  "dioxus-core",
  "dioxus-devtools-types",
  "dioxus-signals",
+ "futures-channel",
+ "futures-util",
  "serde",
  "serde_json",
  "subsecond",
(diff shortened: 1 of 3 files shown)


Phosphorus-M:
Omg it's happening ❤️‍🔥

PR #4592, opened 2025-08-20
Check target compatibility before using the default renderer

ealmloff:
Currently if you have a cargo.toml set up like this:
```toml
[dependencies]
dioxus = { version = "0.7.0-rc.0", features = ["router"] }

[features]
default = ["web"]
web = ["dioxus/web"]
desktop = ["dioxus/desktop"]
```
and serve with this command `dx serve --platform desktop`, dx will disable the `web` feature and enable the `desktop` feature flag because the renderer you selected was different than the default

However, that doesn't work if you serve with `dx serve --destkop` which only sets the target, not the renderer. This PR makes `dx serve --desktop` work by checking compatibility between the currently set target and the renderer. The implicit web renderer is not compatible with the explicitly set desktop target so dx disables the web renderer just like it does with `--platform desktop`

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -593,7 +593,18 @@ impl BuildRequest {
             },
             None if !using_dioxus_explicitly => None,
             None => match enabled_renderers.as_slice() {
-                [] => None, // Wait until we resolve everything else first then we will resolve it from the triple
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
I changed all the logic here in #4596 , the user will see different errors now

PR #4620, opened 2025-08-29
Remove transitions from todomvc stylesheet

nicoburns:
Because Blitz doesn't implement transitions/animations yet, the presence of these styles causes the `color` style to not apply at all. Obviously we need to fix this in Blitz eventually, but for now let's just forgo the transition.

### DIFF
--- a/examples/assets/todomvc.css
+++ b/examples/assets/todomvc.css
@@ -211,7 +211,6 @@ body {
     padding: 15px 15px 15px 60px;
     display: block;
     line-height: 1.2;
-    transition: color 0.4s;
 }
 
 .todo-list li.completed label {
@@ -231,7 +230,6 @@ body {
     font-size: 30px;
     color: #cc9a9a;
     margin-bottom: 11px;
-    transition: color 0.2s ease-out;
 }
 
 .todo-list li .destroy:hover {


examples/assets/todomvc.css
```diff
@@ -231,7 +230,6 @@ body {
     font-size: 30px;
     color: #cc9a9a;
     margin-bottom: 11px;
-    transition: color 0.2s ease-out;
```
ealmloff, line 234: This style applies to the delete button in the todomvc example right? It works correctly on macos before and after the blitz upgrade. It doesn't transition, but this seems like reasonable fallback behavior:

https://github.com/user-attachments/assets/25c3151e-e835-4f05-a909-cec010afb77d

nicoburns: I believe it applies to the text of the todo ("what" here). With this PR that text correctly changes color when you toggle the todo from "done" to "not done" and back again. Without this PR that color doesn't change.

ealmloff: Thanks, I can see that fix now

ealmloff (approved):
Looks good, might be worth making a list of examples that are expected to work on native at some point

PR #4645, opened 2025-09-05
Look for strip settings in inherited profiles and global configs

### LINKED ISSUE #4637: Build with <code>strip = "symbols"</code>
**Problem**

Building with `strip = "symbols"` messes with the style sheets and probably any other asset, so they become unavailable in the final build.

**Steps To Reproduce**

- Compile a program that loads an asset (e.g., a stylesheet) with `strip = "symbols"`

**Expected behavior**

The assets are found.

**Environment:**

- Dioxus version:  0.7.0-rc.0
- Rust version:    1.91.0-nightly (cdb45c87e 2025-08-27)
- OS info:         Gentoo
- App platform:    native (desktop, android), works fine in web

ealmloff:
We use the `cargo_toml` crate to parse profiles and check for strip settings. It handles some workspace inheritance but it doesn't look for any `.cargo/config.toml` files. This PR adds logic to look for the local and global config.toml file and merge it into the manifest

This partially fixes #4637. We should now at least throw a warning if you have strip enabled globally

### DIFF
--- a/packages/cli/src/cargo_toml.rs
+++ b/packages/cli/src/cargo_toml.rs
@@ -1,0 +1,147 @@
+//! The cargo_toml crate contains some logic for resolving Cargo.toml files with workspace inheritance, but it
+//! doesn't handle global configs like ~/.cargo/config.toml. This module handles extending the manifest with those
+//! settings if they exist.
+
+use std::path::{Path, PathBuf};
+
+use cargo_toml::{Manifest, Profile, Profiles};
+
+/// Load the manifest from a path inheriting from the global config where needed
+pub fn load_manifest_from_path(path: &Path) -> Result<Manifest, cargo_toml::Error> {
+    let mut original = Manifest::from_path(path)?;
+    // Merge the .cargo/config.toml if it exists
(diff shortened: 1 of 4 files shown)


packages/cli/src/cargo_toml.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Get the global cargo config path if it exists
+fn global_cargo_config_path() -> Option<PathBuf> {
+    if cfg!(windows) {
+        std::env::var_os("USERPROFILE").map(|user_profile| {
+            PathBuf::from(user_profile)
+                .join(".cargo")
+                .join("config.toml")
+        })
+    } else if cfg!(unix) {
+        dirs::home_dir().map(|home_dir| home_dir.join(".cargo").join("config.toml"))
```
omar-khallaf: Not necessarily in home directory. I think `$CARGO_HOME` needs to be checked first and if unset default to home directory.

ealmloff: Thanks, should be fixed now

PR #4650, opened 2025-09-08
Fixed client config macro

snatvb:
Added mobile feature target to config macro


snatvb:
@ealmloff  it seems that ready to merge :)

PR #4651, opened 2025-09-09
fix: feature gate tokio and tracing

### LINKED ISSUE #4641: dioxus-native fails to compile with tracing or net feature turned off
**Problem**

Turning off tracing or net causes dioxus-native to fail to compile even though its marked as a feature flag that can be disabled.

```rust
dioxus-d8abd2ecf6e8b5b4/5ebde96/packages/native/src/assets.rs:38:21
   |
38 |                     tracing::trace!("fetching asset  from file system success {request:#?}");
   |                     ^^^^^^^ use of unresolved module or unlinked crate `tracing`
```
**Steps To Reproduce**

disable default features for dioxus-native
attempt to compile dioxus native

**Environment:**

- Dioxus version:  commit 5ebde96
- Rust version: 1.88
- OS info:  ubuntu 24
- App platform:  desktop native

omar-mohamed-khallaf:
Fixes #4641.

### DIFF
--- a/packages/native/src/lib.rs
+++ b/packages/native/src/lib.rs
@@ -82,10 +82,12 @@ pub fn launch_cfg_with_props<P: Clone + 'static, M: 'static>(
     let event_loop = create_default_event_loop::<BlitzShellEvent>();
 
     // Turn on the runtime and enter it
+    #[cfg(feature = "net")]
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()
         .unwrap();
+    #[cfg(feature = "net")]
     let _guard = rt.enter();
 
     // Setup hot-reloading if enabled.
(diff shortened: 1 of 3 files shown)


packages/native/src/lib.rs
```diff
@@ -82,10 +82,12 @@ pub fn launch_cfg_with_props<P: Clone + 'static, M: 'static>(
     let event_loop = create_default_event_loop::<BlitzShellEvent>();
 
     // Turn on the runtime and enter it
+    #[cfg(feature = "net")]
     let rt = tokio::runtime::Builder::new_multi_thread()
```
omar-khallaf, line 86: Is `tokio` required?

nicoburns: For blitz-net, yes it is currently as it's using `reqwest`. We want to add `ureq` backend at some point (which would probably use a `rayon` threadpool instead).

nicoburns: I also want to add a "data-uri and assets only" (but not http(s)) option, which wouldn't need tokio.

omar-khallaf: Ok, good thing I didn't delete it.

nicoburns (approved):

Issue #4644, opened 2025-09-05
Load scripts and stylesheets instead of preloading them


omar-mohamed-khallaf:
## Feature Request

The current implementation dynamically adds scripts and stylesheets at runtime, thus suffering from FOUC.
With desktop and mobile, one is able to add content to head with `with_custom_head`, but there's no such api for web and I couldn't figure a way to do the same.
What I found is that assets can take a second argument (e.g. `AssetOptions::css().with_preload(true)`) which preloads the asset but doesn't apply the stylesheets or execute scripts, so FOUC is still present.

Even trying to solve this problem with a JS script at the start won't work, since it requires listening for events like: 'load` and `DOMContentLoaded` and as far as I understand dynamically added scripts won't block these events.

## Implement Suggestion

I modified the code as follows to apply style sheets and execute JS files:
```diff
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -4547,7 +4547,10 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
                     if css_options.preloaded() {
                         head_resources.push_str(&format!(
                             "<link rel=\"preload\" as=\"style\" href=\"/{{base_path}}/assets/{asset_path}\" crossorigin>"
-                        ))
+                        ));
+                        head_resources.push_str(&format!(
+                            "<link rel=\"stylesheet\" href=\"/{{base_path}}/assets/{asset_path}\">"
+                        ));
                     }
                 }
                 AssetVariant::Image(image_options) => {
@@ -4561,7 +4564,10 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
                     if js_options.preloaded() {
                         head_resources.push_str(&format!(
                             "<link rel=\"preload\" as=\"script\" href=\"/{{base_path}}/assets/{asset_path}\" crossorigin>"
-                        ))
+                        ));
+                        head_resources.push_str(&format!(
+                            "<script src=\"/{{base_path}}/assets/{asset_path}\"></script>"
+                        ));
                     }
                 }
                 _ => {}
```
Although, I'm not sure if this is the right way. Should it be under a different config name? Is there another way?

And by the way, preloaded assets don't seem to be applied to desktop and mobile builds.

remysedlak:
I've been trying to investigate the Dioxus web FOUC myself, so annoying, great investigation!

PR #4653, opened 2025-09-09
Feat: eager loading of assets

omar-mohamed-khallaf:
The first commit implements a solution to #4644.
The other two are lints applied by `clippy`.


omar-khallaf:
Yeah, I have a different config showing 200 warnings for `request.rs` alone, so I nuked the whole crate with `cargo clippy --fix -- -Wclippy::all -Dclippy::correctness -Wclippy::pedantic`.

PR #4658, opened 2025-09-12
Allow relative assets in rust > 1.88

### LINKED ISSUE #4348: Relative asset import paths
## Feature Request

Relative asset imports `asset!("./styles.css")`. These should resolve relative to the current file the same as `include_str!()`. Absolute imports (`asset!("/styles.css")`) should continue to work as they currently do.

ealmloff:
Rust 1.88 added support for reading the path of a span. This lets us check if a relative paths exists inside the asset macro. This feature is very useful for the component library where we need to link to a style.css file for every component in a way that is easy to move around

Closes #4348

### DIFF
--- a/packages/manganis/manganis-macro/src/lib.rs
+++ b/packages/manganis/manganis-macro/src/lib.rs
@@ -211,13 +211,12 @@ fn resolve_path(raw: &str, span: Span) -> Result<PathBuf, AssetParseError> {
         });
     };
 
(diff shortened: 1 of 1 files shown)


ealmloff:
> Looks good. One question: is this new syntax compatible with `include_str!`?

Yes, both can accept the same syntax, although `asset!` allows more arguments. Both resolve files relative to the current folder and allow some built-in macros like `env!` and `concat!`

PR #4665, opened 2025-09-16
feat: Add use_coroutine_with_sender

theduke:
Adds a new `use_coroutine_with_sender`, which is equivalent to
`use_coroutine`, but also receives a sender in the init function.

Motivation: More complex coroutines often need to send messages to themselves,
for example inside futures to update the state.

### DIFF
--- a/packages/hooks/src/use_coroutine.rs
+++ b/packages/hooks/src/use_coroutine.rs
@@ -89,6 +89,27 @@ where
     })
 }
 
+/// Same as [`use_coroutine`] but also provides the sender to the init function.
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Sorry this has sat for so long. I'd rather we get drop of use_coroutine, but if we were to keep it, then I think it's fine to make a breaking change to provide both the sink and stream as a single object, letting you split it. It would be a breaking change. Going to close this PR while I have a think and then implement the appropriate fix.

theduke:
Why get rid of `use_coroutine`?

I use it a lot, it's very useful to structure logic for complex components.

PR #4669, opened 2025-09-17
DX components command

ealmloff:
Adds a few commands to manage dioxus components in your project:
```sh
# Adds the tabs component from the default registry to the components module
# along with any rust and component dependencies it may have.
dx component add tabs
# Removes the tabs component from the components module
dx component remove tabs
# Lists all components in the default registry
dx component list
# Update the registry cache. The next time a component is added, it will use the latest version
dx component update
# Print the json schema for the component.json file required for all components in registries
dx component schema
```

This pulls from the [manifest](https://github.com/DioxusLabs/components/blob/main/component.json) in the components repo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5500,7 +5500,7 @@ dependencies = [
 
 [[package]]
 name = "dioxus-component-manifest"
-version = "0.7.0-rc.1"
+version = "0.7.0-rc.2"
 dependencies = [
  "schemars 1.0.4",
  "serde",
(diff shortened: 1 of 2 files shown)


Jasper-Bekkers:
After discussing about this in Discord a bit a few smaller things came up so far in my testing;

1. It would be really nice if this picked up the assets-path and module-path from `Dioxus.toml` to it places the files and folders in the right directory
2. Having a shorthand notation for adding all components would be really helpful to set up an initial project

ealmloff:
> After discussing about this in Discord a bit a few smaller things came up so far in my testing;
> 
> 1. It would be really nice if this picked up the assets-path and module-path from `Dioxus.toml` to it places the files and folders in the right directory

You can now set the `application.asset_dir` for assets to be placed into, `component.registry` for the default registry and `component.component_dir` for where components will be copied to

> 2. Having a shorthand notation for adding all components would be really helpful to set up an initial project

You can now add multiple components separated by commas or add all components in a registry: `dx component add toast,tabs` or `dx component add --all`

Issue #4685, opened 2025-09-23
Integration tests may be passing incorrectly


martin-hughes:
**Problem**

When testing, using `assert_eq!` (and friends) within a component can fail without triggering a test failure. Possible false-positive tests in the Dioxus source.

**Steps To Reproduce**

- Checkout dioxus code
- cd to `packages/signals`
- `cargo test deref_signal`
- Test passes - and should pass
- Edit `tests/create.rs` to change line 46 from `assert_eq!(&*signal(), "hello world");` to `assert_eq!(&*signal(), "bye cruel world");`
- `cargo test deref_signal`
- Test *still passes*

**Expected behavior**

Test should fail.

Actually, if you run the test from within RustRover you can see the assertion failure error text being printed, but the test is still marked as passed.

The assertion failure text is *not* printed if you just invoke `cargo test`, so there may be existing false positive tests that haven't been noticed.

**Screenshots**

n/a

**Environment:**

- Dioxus version:  v0.6.3
- Rust version:    1.89 stable
- OS info:         Windows 10 (not WSL)
- App platform:    n/a

**Questionnaire**

- [x] I'm interested in fixing this myself but don't know where to start.
- [ ] I would like to fix and I have a solution.
- [ ] I don't have time to fix this right now, but maybe later.

Actually it would be helpful to have some guidance - I would like to write some tests in my project for some stuff I'm doing with signals, and I'll happily implement any changes here too.

I assume the assertion needs hoisting to the main body of the test, maybe with the component providing a failure response via a synchronization channel. But if you have a better idea please say and I'll implement that.

martin-hughes:
I've been looking more at this. I can see that the `assert!` or `panic!` calls in tested components are resulting in errors that get captured inside the `RootScopeWrapper`'s `ErrorBoundary` - thus not making it out to the test function itself. I think this was an unintentional change as a result of adding or improving Dioxus's basic error handling.

My question is, is there a way to access the error context using methods on the `VirtualDom` object. For example, something like the snippet below. Then I (or whoever) could update affected tests to assert that the `error_context` (or similar) doesn't contain any stored errors after the test completes. That way it wouldn't be necessary to change the test logic very much.

```
let dom = VirtualDom::new(/* something */);
/* run test */

 // Clearly the following isn't correct code, but hopefully it shows what I mean
let error_context = dom.get_scope(ScopeId(2 /* or whichever */)).get_context::<ErrorContext>();
assert!(error_context.empty());
```

[...]

PR #4686, opened 2025-09-23
Minor fixes for tests

martin-hughes:
I questioned in issue #4685 whether the tests could be fixed using a sync channel to move the assertion to the main function body.

This PR demonstrates that.

It's not intended to be merged just yet. Instead I wanted to see if this was an approach you'd support for the other tests, and if so I'll go looking for similar problems and update this PR.

Any comments welcome of course! Especially if you know of an easier way to achieve the same thing.

Just FYI, I haven't run the full set of tests from https://dioxuslabs.com/learn/0.6/contributing/ - I will do so if you get to the point of considering a merge.

Notes:

- You can check the logic by changing one of the "hello world" strings - the test will then fail (run `cargo test` to see this)
- I changed the body text to "arbitrary text" to emphasise that it is not one of the test strings.
- I've removed the multiplication of the Child elements, it didn't seem necessary for the purpose of the test - but I don't know why it was there to start with.

### DIFF
--- a/packages/core/tests/error_boundary.rs
+++ b/packages/core/tests/error_boundary.rs
@@ -86,7 +86,7 @@ fn clear_error_boundary() {
     assert_eq!(out, "We should see this");
 
     // There should be no errors left in the error boundary
-    dom.in_runtime(|| {
-        ScopeId::APP.in_runtime(|| assert!(consume_context::<ErrorContext>().errors().is_empty()))
+    dom.in_scope(ScopeId::APP, || {
+        assert!(consume_context::<ErrorContext>().error().is_none())
     })
 }
--- a/packages/core/tests/fuzzing.rs
+++ b/packages/core/tests/fuzzing.rs
@@ -288,9 +288,8 @@ fn create() {
             VirtualDom::new_with_props(create_random_element, DepthProps { depth: 0, root: true });
         vdom.rebuild(&mut NoOpMutations);
 
-        vdom.in_runtime(|| {
-            ScopeId::APP
-                .in_runtime(|| assert!(consume_context::<ErrorContext>().errors().is_empty()))
+        vdom.in_scope(ScopeId::APP, || {
+            assert!(consume_context::<ErrorContext>().error().is_none())
         })
     }
 }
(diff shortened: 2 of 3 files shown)


martin-hughes:
I realised overnight the test immediately below the one I changed used a static atomic variable to get results out of the component - which looks much cleaner and simpler than what I wrote.

I've updated the PR to match. I'll go looking for other tests that could do with updating.

martin-hughes:
See also my comment https://github.com/DioxusLabs/dioxus/issues/4685#issuecomment-3327757331 which asks about another possible solution

packages/core/tests/bubble_error.rs
```diff
@@ -16,6 +16,7 @@ fn app() -> Element {
 }
 
 #[test]
+#[ignore] // Test doesn't do anything.
```
martin-hughes: I might have misunderstood but I don't think this test generates any errors to bubble? I've marked it as ignore to flag it, but I'm open to changing or removing it.

packages/core/tests/suspense.rs
```diff
@@ -217,22 +217,6 @@ fn suspended_nodes_dont_trigger_effects() {
         }
     }
 
-    #[component]
```
martin-hughes, line 220: This component was unused

packages/core/tests/suspense.rs
```diff
@@ -474,22 +458,6 @@ fn toggle_suspense() {
 fn nested_suspense_resolves_client() {
     use Mutation::*;
 
-    async fn poll_three_times() {
```
martin-hughes, line 477: This exactly duplicated a function at the top of the file so removed for DRYness

packages/signals/tests/create.rs
```diff
@@ -22,35 +22,57 @@ fn create_signals_global() {
         }
     }
 
-    dom.rebuild_in_place();
-
     fn create_without_cx() -> Signal<String> {
         Signal::new("hello world".to_string())
     }
+
```
martin-hughes, line 28: I wasn't sure if this level of checking was overkill, but it checks that the signals are working...

martin-hughes:
I looked through all of the `tests` directories and found a handful of tests that didn't seem to be working how I'd expect. I've also added a few comments and clarifications where I thought that would be useful for the next newcomer who looks at them.

Ready for review, happy to do any updates you want.

jkelleyrtp (approved):

PR #4697, opened 2025-09-26
Don't rely on `window`

mohe2015:
This allows using `dx` for web workers.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -3875,15 +3875,15 @@ impl BuildRequest {
         writeln!(
             buf_writer,
             r#"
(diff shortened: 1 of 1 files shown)


mohe2015:
We could, browser support is extremely good though (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) apart from IE

mohe2015:
As we use webassembly https://caniuse.com/wasm I would expect it doesn't matter and I would prefer simplicity

nicoburns:
Yeah, that's a good point. Probably nothing that support WASM well, but not `globalThis`.

PR #4698, opened 2025-09-26
Honor subdir for hashless assets

mohe2015:
When combining multiple dioxus runs (for example web worker output with web application) it is helpful when the subdirectories are also honored for hashless assets so that the wasm imports from the `wasm/` folder of the web worker is correctly found.

I don't know if this is the best solution but maybe you have a better idea. Also at some places the correct relative path is not calculated.

### DIFF
--- a/packages/cli-opt/src/hash.rs
+++ b/packages/cli-opt/src/hash.rs
@@ -148,7 +148,7 @@ pub fn add_hash_to_asset(asset: &mut BundledAsset) {
             let mut bundled_path = if asset.options().hash_suffix() {
                 PathBuf::from(format!("{}-dxh{hash}", file_stem.to_string_lossy()))
             } else {
-                PathBuf::from(file_stem)
+                PathBuf::from(asset.relative_source_path().trim_start_matches("/assets/")).with_extension("")
             };
 
             if let Some(ext) = ext {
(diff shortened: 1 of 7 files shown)


packages/cli-opt/src/hash.rs
```diff
@@ -148,7 +148,7 @@ pub fn add_hash_to_asset(asset: &mut BundledAsset) {
             let mut bundled_path = if asset.options().hash_suffix() {
                 PathBuf::from(format!("{}-dxh{hash}", file_stem.to_string_lossy()))
             } else {
-                PathBuf::from(file_stem)
+                PathBuf::from(asset.relative_source_path().trim_start_matches("/assets/")).with_extension("")
```
mohe2015, line 151: This is the main functional change

ealmloff:
This design could have some odd behavior for relative assets. If I include `asset!("../../asset.png")` (which should be valid if you don't publish your app on crates.io), it shouldn't be copied outside of the asset directory.

You can control where assets are copied today by copying a folder into the asset dir, but it is more work to set up if you need to copy it into a deeply nested folder.

```rust
// / wasm
//   / myapp.wasm  
asset!("/wasm").with_hash(false); // this will copy the wasm folder and all of its contents into the asset folder
```

What do you think about a separate destination path option on the asset builder like this instead?
```rust
asset!("../../asset.png").with_bundled_at("/my/wasm/folder/asset.png")
```

Issue #4087, opened 2025-05-07
Default features usage in `dioxus` precludes disabling `tokio_runtime`


Jaltaire:
When adding `dioxus` as a dependency, `dioxus-desktop` is configured in `dioxus/Cargo.toml` as:

```
dioxus-desktop = { workspace = true, default-features = true, optional = true }
```

and because Cargo features are additive, this precludes `tokio_runtime` from being later disabled.

(As such, I am currently using a fork of Dioxus with `default-features = false` for `dioxus-desktop`, and this resolves the tokio runtime problem as expected: https://github.com/Jaltaire/dioxus/commit/6d3d3e99bf3c60bc452301eb37fd59a2005dfa09.)

Not sure what the preferred/idiomatic way would be to better handle this, but a couple ideas:

- We can set `default-features = false` (or remove `tokio_runtime` as a default feature) in `dioxus/Cargo.toml`, though this may be less than preferable for those who ordinarily do not need to set up a separate tokio runtime (as seems to be the expectation for most Dioxus users).
- Could add setup of tokio runtime to the `LaunchBuilder` instead.

Happy to put together a PR once there's some consensus around a preferred solution.

jkelleyrtp:
if you configure your cargo.toml as such:

```toml
[package]
name = "desktop-no-tokio"
version = "0.1.0"
edition = "2024"

[dependencies]
dioxus = { version = "0.7.0-alpha.0" }
dioxus-desktop = { version = "0.7.0-alpha.0", default-features = false }
```

Then you can launch the desktop app with the `dioxus::desktop`:

```
use dioxus::prelude::*;

fn main() {
    dioxus_desktop::launch::launch(app, Default::default(), Default::default())
}

fn app() -> Element {
    rsx! {
        div { style: "display: flex; flex-direction: column; align-items: center; justify-content: center; height: 100vh;",
            h1 { "Hello, Dioxus!" }
            p { "This is a simple Dioxus application." }
            button { onclick: |_| println!("Button clicked!"), "Click Me" }
        }
    }
}
```

However, we *should* adjust this so `dioxus` enables the default features of its child crates and `default-features=false` is set for desktop.

mohe2015:
If we want to allow disabling the tokio runtime we should somehow try to ensure that spawning still works properly. E.g. I just accidentially used a single threaded tokio runtime and then suddenly async tasks don't work because they can't be run anywhere.

PR #4699, opened 2025-09-26
Don't try launching a tokio runtime if we are already within one

mohe2015:
Closes https://github.com/DioxusLabs/dioxus/issues/4087


mohe2015:
Are the test failures spurious?

ealmloff:
> Are the test failures spurious?

Yes, those are caused by a hot patching failure tracked in https://github.com/DioxusLabs/dioxus/issues/4712

jkelleyrtp (approved):

PR #4726, opened 2025-10-05
Update gradle wrapper

mohe2015:
`#!/usr/bin/env sh` doesn't work in [Nix](https://nixos.org/) builds but `#!/bin/sh` which is used in newer wrapper versions works.
As you probably shouldn't accept jar binaries from outside contributors you may want to run this yourself or check the hash at https://gradle.org/release-checksums/


jkelleyrtp (approved):

jkelleyrtp:
Thanks for doing the upgrade!

PR #4727, opened 2025-10-05
Enable logs for tauri-bundler

mohe2015:
This hides logs from tauri-bundler which may be critical to find out what the issue is. I'm unsure about the performance impact but I think it's important to not hide the logs.


mohe2015:
I also have the feeling that this slows it down (didn't measure more accurately yet). Maybe we could instead get walrus to add some feature flag to disable their logging or something like that.

jkelleyrtp:
Having log enabled was a huge slowdown for wasm processing (2-3x). I am planing on just migrating away from tauri-bundle in #4958

PR #4736, opened 2025-10-07
FIX: web.https configuration - error: no process-level CryptoProvider available

### LINKED ISSUE #3815: web.https configuration - error: no process-level CryptoProvider available
**Problem**
I cannot get web.https to work.
I do not know if this is a bug or some kind of error from my side, as I am very new to Dioxus.

When I run:
`dx serve --platform web --target wasm32`

I get the following error:
```text
Created a new certificate valid for the following names 📜│
- "localhost"is already installed in the system trust store! 👍
- "::1"  The local CA is already installed in the Firefox and/or Chrome/Chromium trust store! 👍
- "127.0.0.1" The certificate is at "../certs/cert.pem" and the key at "../certs/key.pem" ✅
It will expire on 1 June 2027 🗓

thread 'tokio-runtime-worker' panicked at /home/marco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.23/src/crypto/mod.rs:249:14:
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
thread 'main' panicked at /home/marco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-cli-0.6.3/src/serve/server.rs:163:21:
Could not receive a socket - the devtools could not boot - the port is likely already in use
```

(it is not an address error)

**Steps To Reproduce**

- I have these configuration settings in `Dioxus.toml`
  ```toml
  [web.https]

  enabled = true
  mkcert = true
  cert_path = "../certs/cert.pem"
  key_path = "../certs/key.pem"
  ```

- I have read the following [issue](https://github.com/DioxusLabs/dioxus/pull/3619) (which seems to have fixed in 0.6.3) and tried to apply the suggestions found there and in the axum issue:
`rustls = { version = "0.23.12", default-features = false, features = ["logging","std","tls12","ring"]}`
`axum-server = { version = "0.7.1", features = ["tls-rustls-no-provider"] }`

- I have added at the beginning of my backend `main.rs` (which compiles successfully) the following:
- 
  ```rust
  use rustls::crypto::ring::default_provider as rustls_provider;

    rustls_provider()
        .install_default()
        .expect("Failed to install rustls crypto provider");
  ```

**Expected behavior**

I would expect to be able to run the application over https

**Screenshots**

**Environment:**

- Dioxus version: 0.6.3 
- Rust version: 1.8.5   
- OS info:   Linux arch 6.13.4-arch1-1  x86_64 GNU/Linux      
- App platform:  web  

**Questionnaire**

tgrushka:
This PR fixes #3815 because `rustls` requires installation of a `CryptoProvider` before `axum` (or any server) can use it.

Said installation is accomplished by calling:
```rust
rustls::crypto::CryptoProvider::install_default(
    rustls::crypto::aws_lc_rs::default_provider()
)
```

before calling `axum_server::from_tcp_rustls().serve()`.

`install_default` can fail, but I think it is safe to call `bail!` on failure here because:
- it should only fail if a `CryptoProvider` is not already installed;
- `start` and `devserver_mainloop` are not public API.

By submitting this PR, I agree that my contributions become part of the `dioxus` project as per the MIT and Apache licenses.

### DIFF
--- a/packages/cli/src/serve/server.rs
+++ b/packages/cli/src/serve/server.rs
@@ -26,6 +26,7 @@ use futures_util::{
     StreamExt,
 };
 use hyper::HeaderMap;
+use rustls::crypto::{aws_lc_rs::default_provider, CryptoProvider};
 use serde::{Deserialize, Serialize};
 use std::{
     convert::Infallible,
@@ -416,7 +417,10 @@ async fn devserver_mainloop(
         return Ok(());
     }
 
-    // If we're using rustls, we need to get the cert/key paths and then set up rustls
+    // If we're using rustls, we need to install the provider, get the cert/key paths, and then set up rustls
+    if let Err(provider) = CryptoProvider::install_default(default_provider()) {
+        bail!("Failed to install default CryptoProvider: {provider:?}");
+    }
     let (cert_path, key_path) = get_rustls(&https_cfg).await?;
     let rustls = axum_server::tls_rustls::RustlsConfig::from_pem_file(cert_path, key_path).await?;
 


jkelleyrtp:
Thanks for the PR! I was recently running into https issues too and didn't have enough time to figure it out.

I'd like to follow this up at some point with us bundling some pre-signed certs to make https work out of the box like some other devtools (ngrok, cloudflare tunnels). I'm not super knowledgable on the security concerns, otherwise I would've looked at it earlier.

jkelleyrtp (approved):

PR #4757, opened 2025-10-10
Don't rely on `window` Part 2

mohe2015:
Now I can use one wasm bundle for the worker and frontend


jkelleyrtp:
Awesome! We should get an example up of creating a new worker on the fly and entering from a different entrypoint.

With dataurls, you can create a virtual `worker.js` that loads the current wasm bundle and then runs an exported function as its main.

None of that is too hard, we just don't have a great example for it.

mohe2015:
Yeah I'm already doing that. Maybe I will find the time to add an example sometime. Currently I still want to upstream some stuff so I can switch back to upstream directly

rsaccon:
I have an interesting use case for an example using webworker in Dioxus: [Turso DB](https://github.com/tursodatabase/turso) (sqlite rust rewrite beta) in dioxus web.
Turso db can be compiled to wasm, they ship Javascript bindings and they just published an [Article](https://turso.tech/blog/introducing-turso-in-the-browser) on the inner workings of running the DB in the browser, which involves a webworker. I started myself to dig into this, but I am still at the very beginning, studying the examples at the `wasm-bindgen` guide.

mohe2015:
@rsaccon I'm actually already doing this with `sqlite_wasm_rs`, see https://github.com/tucan-plus/tucan-plus/blob/7d301ebae08d0e3940710077ef59db6aa7d5b45b/crates/tucan-plus-dioxus/src/main.rs

Edit: Also relevant is probably https://github.com/tucan-plus/tucan-plus/blob/7d301ebae08d0e3940710077ef59db6aa7d5b45b/crates/tucan-plus-worker/src/lib.rs#L576

Edit 2: Note that https://github.com/DioxusLabs/dioxus/issues/4758 is currently blocking this with `--release`

Edit 3: Note that my approach does not use the `SharedArrayBuffer` so I do not need to set the headers etc. Also I don't use WASI.

rsaccon:
@mohe2015 thanks a lot for the explanation and links.

Issue #4778, opened 2025-10-13
Assets should be able to live ABOVE the assets dir (two implementations attached)


tekacs:
## Feature Request

The current implementation of assets in Dioxus doesn't provide for being able to put assets in the root of the output public directory.

When targeting web, this is a _very_ common need - in particular [.well-known](https://en.wikipedia.org/wiki/Well-known_URI) paths are super commonly necessary, for everything from Apple site associations to security. They _must_ live in `/.well-known`, not `/assets/.well-known`, and yet the assets dir is hard-coded in manganis and Dioxus.

## Implement Suggestion

One possible option is to restore and improve static/asset dir functionality. This (fully implemented, tested, working) branch does that. It also correctly handles cleanup of assets that are moved/removed.

https://github.com/DioxusLabs/dioxus/compare/main...tekacs:dioxus:add-back-asset-dir

Another option would be to teach manganis and Dioxus to allow assets to be mounted higher in the output tree - i.e. to live above `assets/`. In combination with the folder functionality, we could probably just `external_asset!(".well-known)"` with `AssetOptions` directing it to live in the root instead of `/assets`. That's implemented here:

Pull Request https://github.com/DioxusLabs/dioxus/pull/4779

I figured that this version is more in line with the direction you seem to have been heading with the removal of static_dir, so that one's a (complete, working, tested) PR.

jongiddy:
Can this be closed now that https://github.com/DioxusLabs/dioxus/pull/4783 is merged?

PR #4779, opened 2025-10-13
Support custom asset mount paths

tekacs:
See issue #4778 for more details and motivation!

- add optional mount_path to manganis AssetOptions + builders
- have CLI honor mount paths when copying assets and generating URLs
- update HTML preload injection, docs, and hotreload notifications
- honor AssetOptions mount_path across every bundle type
- serve native assets from mount directories via resolver fallback
- ensure Android hotreload pushes files into mount-aware locations

Tests: cargo test -p manganis-core; cargo check -p dioxus-cli

### DIFF
--- a/packages/asset-resolver/src/native.rs
+++ b/packages/asset-resolver/src/native.rs
@@ -40,7 +40,7 @@ pub(crate) fn resolve_native_asset_path(path: &str) -> Result<PathBuf, AssetPath
 fn resolve_asset_path_from_filesystem(path: &str) -> Option<PathBuf> {
     // If the user provided a custom asset handler, then call it and return the response if the request was handled.
     // The path is the first part of the URI, so we need to trim the leading slash.
-    let mut uri_path = PathBuf::from(
+    let uri_path = PathBuf::from(
         percent_encoding::percent_decode_str(path)
             .decode_utf8()
             .expect("expected URL to be UTF-8 encoded")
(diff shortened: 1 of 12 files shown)


tekacs:
I spent some time trying to get the asset pipeline to correctly delete files that had been removed (with a manifest), but it proved fiddly with too many edge cases, so I've separated out that work for now.

Compared to the simpler [asset dir implementation](https://github.com/DioxusLabs/dioxus/compare/main...tekacs:dioxus:add-back-asset-dir), this is fiddlier (in particular hot patching deserves a close look), but feels clean and correct in use.

jkelleyrtp:
I was looking at Vite for inspiration and they have a `public` dir where the dir gets copied directly into the output public dir. 

https://vite.dev/guide/assets.html#the-public-directory

I think we can support something like this instead of an "assets" directory. The old assets directory concept conflicts with `asset!()` calls since assets get duplicated between the two and its unclear when assets need to be hashed/optimized. It also had issues with desktop/mobile since you might not want to ship all the assets from that folder along with each platform, so auto-including it would be suboptimal. The public dir would only apply to web apps here, making it a bit clearer / more narrow.

It seems you have this mostly implemented here - maybe it's worth reframing the PR to match the vite public dir?

That way the final structure of the bundle looks like:

```
server.exe
public/
   /_wasm/
   /assets/
   index.html
   index.js
   /* rest of merged public folder */
```

When we merge the public folders together, we could allow files to end up anywhere in the output tree (even in the `_wasm` or `asset` folders), which should solve your original issue.

If you read into the vite docs a bit more, they have an `/_immutable` folder for infinitely hashed assets, and all other assets end up in `/assets` (both in the final public folder) - could be worth just stealing that entirely

tekacs:
Ah yes, I had implemented a simpler static dir approach AS WELL in a parallel branch (as I'd mentioned in #4778), which felt smooth and behaved well (and deleted files properly). I hadn't opened a PR for it because it looked like there was a move away from it, but I opened it in #4783 now.

tekacs:
Closing in favor of #4783

PR #4783, opened 2025-10-14
Support `public` dir

tekacs:
- add optional application.static_dir field defaulting to none
- copy configured static directory into web output during build/serve
- document config knob and add unit tests
- force full rebuild when static_dir contents change
- sync removes stale outputs and logs copy activity
- track static manifest next to platform dir

Tests: cargo fmt

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -4430,12 +4430,13 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
                 .join("main")
                 .join("assets"),
 
+            // We put assets in public/assets for server apps
+            BundleFormat::Server => self.root_dir().join("public").join("assets"),
+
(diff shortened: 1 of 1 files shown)


tekacs:
Haha I didn't default it to `public` just so that it wouldn't implicitly change people's setups, but making it public by default is nice. :)

jkelleyrtp:
The asset system is already quite sophisticated so I changed this PR to just use that instead. Every file the public dir gets submitted into the asset system with hashing disabled. Our existing memoization system takes care of the rest.

I also adjusted things to merge the manifest together with the .cli-version fle we generate so we now have a `.manifest.json` file, similar to the one in ruby-on-rails.

Now, you can load that file and read its contents at runtime if you so choose. We'll add more to it eventually for deploy-related features.

NextJS supports a number of things like sitemap generation - we can probably re-use the public dir as a place to dump the output.

PR #4784, opened 2025-10-14
Prevent hot patch from leaving devserver in BuildError state

tekacs:
Scenario: without this patch,  if you fail a build, fix the code, let the thin build succeed, and then refresh the browser... then after reloading the browser page, it gets stuck on the "One sec" loader because the devserver status never clears the earlier BuildError. Prior to this patch, only a full rebuild fixes it.

- mark build status ready after sending hot patch messages so refreshes succeed
- reuse the same helper from reload command to avoid stale loading screens
- add set_ready() helper to avoid duplicate status broadcasts

This PR is tested, working correctly in other exercises including error, and found to fix the issue.

### DIFF
--- a/packages/cli/src/serve/server.rs
+++ b/packages/cli/src/serve/server.rs
@@ -316,6 +316,7 @@ impl WebServer {
             for_build_id: Some(build.0 as _),
         });
         self.send_devserver_message_to_all(msg).await;
+        self.set_ready().await;
     }
 
     /// Tells all clients that a hot patch has started.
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):

jkelleyrtp:
Thank you for helping with the bugs!!

tekacs:
@jkelleyrtp Thank _you_ for subsecond! I've been waiting so long to switch out frontend to Rust, had no idea when someone would finally get to building something like subsecond and it's a joy, even if there's teething to work through. :)

PR #4789, opened 2025-10-15
datatransfer object, drag_and_drop example, fix drag-and-drop 

jkelleyrtp:
When adding support for native file uploads, I accidentally broke file drag/drop.

In spirit of making our native apis more equivalent to dom APIs, I decided to just implement the DataTransfer API and add an example for a trello-like app.


jkelleyrtp:
It's late here - this PR currently works fine on web but is not fully implemented on desktop. I'm going to put out a new pre-release with this particular issue fixed and then we can finish implementing the data transfer stuff properly.

PR #4791, opened 2025-10-15
Add option_asset! to match asset!

tekacs:
As a user of the jujutsu VCS, manganis's `asset!` plays a little awkwardly with `/assets/tailwind.css`.

Since `tailwind.css` is generated and shouldn't really be checked in, it needs to be ignored from our Git repo. That said... the `asset!` macro statically fails to compile if the file is missing, so `cargo check` in a monorepo with a Dioxus workspace fails on a fresh checkout, until `dx` has been used to build that repo at least once.

With Git, an option is to put an empty `tailwind.css` there and use `--assume-unchanged` (although that's ultimately local), but... [jujutsu doesn't have this](https://github.com/jj-vcs/jj/issues/7737), and it'd be preferable if manganis could instead be encouraged to not fail hard when the file is missing.

### DIFF
--- a/packages/manganis/manganis-macro/src/asset.rs
+++ b/packages/manganis/manganis-macro/src/asset.rs
@@ -1,10 +1,10 @@
 use crate::{resolve_path, AssetParseError};
 use macro_string::MacroString;
 use proc_macro2::TokenStream as TokenStream2;
-use quote::{quote, ToTokens, TokenStreamExt};
+use quote::{quote, ToTokens};
 use std::{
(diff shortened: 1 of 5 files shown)


jkelleyrtp:
I'm not a *huge* fan of encouraging this pattern for generated assets, but an alternative system would likely end up closer as a "plugin" which is definitely out of scope for now. 

We should add your notes here to our guide on assets in the docs. Otherwise, the PR looks good, thanks for the contribution!

PR #4801, opened 2025-10-16
Fix: Lifetime unsafety in `ReadableExt::deref_impl`

bushrat011899:
# Objective

The current implementation of `dioxus_signals::read::ReadableExt::deref_impl` is unsafe with an undocumented invariant: it's possible to use this method to extend the lifetime of a borrow, allowing undefined behavior. This is because the "trick" used to cast `&'_ Self` into a `&'a dyn Fn() -> Self::Target` involves a non-annotated call to `core::mem::transmute`, with the lifetime `'a` disconnected from the implicit lifetime of `&self`. This _currently_ isn't abused as all calls to `deref_impl` then restrict the lifetime of the closure back down to that of the original borrow on `Self`.

## Solution

This unsafety can be entirely avoided using precise capturing to inform the compiler that the returned closure captures `Self`. However, since explicit precise capturing via `use<...>` requires an MSRV of 1.85.0, I've opted to instead move `deref_impl` out of the `ReadableExt` trait into its own standalone function. This reduces the scope of captured types in a 1.79.0 compatible way.

To further limit the possibility of UB, I've also added a full `Layout` check to the method which is run at compile time. This ensures that `Self` and a closure only capturing `Self` have the same alignment and size. Since those parameters match, and a closure containing `Self` does not alter the in-memory representation of `Self`, the two types can be safely transmuted between.

Finally, I've updated the comments within the method to more precisely describe what's happening and why it is a safe operation.


packages/fullstack-core/src/loader.rs
```diff
@@ -280,7 +280,7 @@ where
     type Target = dyn Fn() -> T;
 
     fn deref(&self) -> &Self::Target {
-        unsafe { ReadableExt::deref_impl(self) }
+        ReadableExt::deref_impl(self)
```
bushrat011899: Making `ReadableExt::deref_impl` itself safe solves the lack of `// SAFETY: ...` comments in the multitude of calls to it across Dioxus.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
         <Self::Storage as AnyStorage>::map(self.read(), |v| v.index(index))
     }
 
-    /// SAFETY: You must call this function directly with `self` as the argument.
-    /// This function relies on the size of the object you return from the deref
-    /// being the same as the object you pass in
```
bushrat011899, line 247: This safety comment did not previously capture the fact that you must also ensure the lifetime `'a` is no greater than that of the provided `&self` borrow. Additionally, I'm unsure it's possible in safe rust to invoke `deref_impl` with a value that _isn't_ `&self`, since that violates the function signature.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
...
+    /// Casts this [`Readable`] into a closure which defers to [read](ReadableExt::read).
     #[doc(hidden)]
-    unsafe fn deref_impl<'a>(&self) -> &'a dyn Fn() -> Self::Target
+    fn deref_impl(&self) -> &(impl Fn() -> Self::Target + use<Self>)
     where
-        Self: Sized + 'a,
+        Self: Sized,
```
bushrat011899: By returning `impl Fn` instead of `dyn Fn`, we can use precise capturing syntax, `+ use<Self>` to communicate to the compiler that the returned closure cannot outlive the _type_ `Self`. We can also remove the `unsafe` annotation from this function, as size and alignment requirements are now compile-time verified within the method itself.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
...
+        ///
+        /// **All** safety invariants of [transmute](core::mem::transmute) **must** be upheld by the
+        /// caller of this function.
+        const unsafe fn transmute_ref_by_value<'a, Source, Donor>(
+            src: &'a Source,
+            _: &Donor,
+        ) -> &'a Donor {
+            // SAFETY: Caller is responsible for upholding all invariants of transmute.
+            unsafe { ::core::mem::transmute::<&'a Source, &Donor>(src) }
         }
```
bushrat011899: This replaces the `cast_lifetime` method and better articulates that its real purpose is for allowing `transmute` to infer the type of an unnameable type (`uninit_closure`). Since we no longer violate lifetime safety, we don't need to use transmute to change the lifetime of the produced closure.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
...
-        let reference_to_closure = cast_lifetime(
+
+        // The real closure that we will never use.
+        let uninit_closure = const {
```
bushrat011899: Using a `const { ... }` block to ensure as much of this method as possible is evaluated at compile time. It's entirely possible the compiler was already optimizing this, but being explicit ensures this is the case.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
...
+            // First we create a closure that captures something with the same
+            // in memory layout as Self (MaybeUninit<Self>).
+            let uninit_callable = ::core::mem::MaybeUninit::<Self>::uninit();
+
+            // Then move that value into the closure.
+            let uninit_closure = move || {
+                // SAFETY: Initialization comes from transposing `self` into the position of `uninit_callable`.
+                let this: &Self = unsafe { uninit_callable.assume_init_ref() };
```
bushrat011899: `MaybeUninit::assume_init_ref` is more idiomatic than getting a `*const Self` pointer and casting it to a `&Self`.

packages/signals/src/read.rs
```diff
@@ -239,44 +239,75 @@ pub trait ReadableExt: Readable {
...
+                let layout_closure = ::core::alloc::Layout::for_value(&uninit_closure);
+
+                if layout_self.align() != layout_closure.align()
+                    || layout_self.size() != layout_closure.size()
+                {
+                    // This panic will be cause evaluation of the const block to fail at compile time
+                    // if the assumption above is not valid.
+                    panic!("assumed layout of closures capturing a single value proven false!");
+                }
+            }
```
bushrat011899: Ideally, there would be a `const_assert_eq!(::core::alloc::Layout::new::<Self>(), ::core::alloc::Layout::for_value(&uninit_closure))` call here, but that's not currently possible in stable Rust. Instead, I've explicitly checked the size and alignment within the `const` block where this closure is created. This translates the possible `panic` into a compile-time error.

bushrat011899 (commented):
Some more targeted comments to assist with understanding the changes I'm proposing.

bushrat011899:
While this isn't the point of this change, I just want to make clear that the proposed changes seem to improve the final generated assembly as well. Comparing [before](https://godbolt.org/z/Mjj6vdr6e) and [after](https://godbolt.org/z/T7j351dhf) in a simplified example, this PR produces around 109 lines of assembly, vs 140 currently. With `-C opt-level=2` optimizations enabled, the difference reduces to 15 after and 31 before. Worth noting though is that in both cases (with optimizations), the actual code path taken in `self.read().clone()` and `self()` are the same, so the final performance should be identical. It just appears that this PR allows the deref implementation and the VTable to be entirely omitted, whereas the current technique still requires them in a vestigial sense.

bushrat011899:
~~CI Failure caused by this PR increasing MSRV to 1.87.0 (up from 1.79.0 within `dioxus-signals`).~~

I've adjusted the technique used here slightly to avoid changing MSRV. By extracting the `deref_impl` function out of `ReadableExt` and making it standalone, we can avoid implicitly capturing any types other than `Self`. I've also switched to a `PhantomData` based approach to naming anonymous types, which [generates even more favorable IR in debug builds with `impl_trait_in_assoc_type`](https://godbolt.org/z/571f4M3M3).

bushrat011899:
~~I don't believe the `lychee` failure is related to this PR, but if it is I'd appreciate some help working out what I can do to resolve it? I'm not familiar with the tool.~~
Lychee failure resolved upstream.

PR #4806, opened 2025-10-17
resolve the index before rendering

jkelleyrtp:


xsoheilalizadeh:
I was debugging this issue for 2 days, then I saw this PR :`)

yaymalaga:
Careful with the dbg! introduced in packages/fullstack-server/src/index_html.rs

PR #4811, opened 2025-10-18
👩‍🏫 Add support for `default-members`

Jasper-Bekkers:
This add pretty basic support for the `default-members` field in `Cargo.toml`, it's not perfect but it make things quite a bit better for larger projects using a bigger virtual workspace. 

For us it means that now we can just use `dx serve` instead of `dx serve -p the-project` which is a nice little productivity and onboarding win.


jkelleyrtp:
Thank you!

PR #4827, opened 2025-10-22
Fullstack: Fix missing query string on `#[get]` server functions

bwskin:
Fixes: #4828

### DIFF
--- a/packages/fullstack/src/client.rs
+++ b/packages/fullstack/src/client.rs
@@ -143,8 +143,19 @@ impl ClientRequest {
 
     #[cfg(feature = "web")]
     pub fn new_gloo_request(&self) -> gloo_net::http::RequestBuilder {
-        let mut builder =
-            gloo_net::http::RequestBuilder::new(self.url.path()).method(self.method.clone());
+        let mut builder = gloo_net::http::RequestBuilder::new(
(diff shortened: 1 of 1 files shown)


matchish:
I've faced similar issue in sse example
https://github.com/DioxusLabs/dioxus/blob/c55a29e930119d71ed2653b255dd1c75ab146935/examples/07-fullstack/server_sent_events.rs#L65

if I change signature this way, I also do not receive it
```
async fn listen_for_changes(config: String) -> Result<ServerEvents<MyServerEvent>> {
```

bwskin:
> I've faced similar issue in sse example
> 
> https://github.com/DioxusLabs/dioxus/blob/c55a29e930119d71ed2653b255dd1c75ab146935/examples/07-fullstack/server_sent_events.rs#L65
> 
> if I change signature this way, I also do not receive it
> 
> ```
> async fn listen_for_changes(config: String) -> Result<ServerEvents<MyServerEvent>> {
> ```

The same issue exists in the fullstack_hello_world example — it’s a global problem. Currently, Dioxus only passes the URL path to the gloo request builder, even though the initial URL (which is fully passed to the reqwest client when the web feature isn’t used) includes the complete address.

PR #4829, opened 2025-10-22
Fullstack: Introduce `serde_qs` extractor for query string

bwskin:
Replaces axum::extract::Query with a custom extractor based on serde_qs.
axum::extract::Query uses the serde_urlencoded crate, while on the client side, Dioxus currently relies on serde_qs.
This mismatch causes unexpected behavior and prevents proper deserialization of structs passed as arguments to #[get] server functions.

With this change, it becomes possible to use any type that serde_qs can serialize in a query string.

This PR is a followup to #4827 
It’s a bit unstructured at the moment, as the module is temporarily placed in the fullstack-server crate.
Suggestions on a better location for it are highly welcome.

### DIFF
--- a/packages/fullstack-server/src/extract.rs
+++ b/packages/fullstack-server/src/extract.rs
@@ -1,0 +1,36 @@
+use axum::extract::FromRequestParts;
+use axum_core::__composite_rejection as composite_rejection;
+use axum_core::__define_rejection as define_rejection;
+use http::request::Parts;
+use serde_core::de::DeserializeOwned;
+
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Query<T>(pub T);
+
+impl<T, S> FromRequestParts<S> for Query<T>
+where
+    T: DeserializeOwned,
+    S: Send + Sync,
+{
+    type Rejection = QueryRejection;
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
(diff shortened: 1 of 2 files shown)


bwskin:
Moved module to `fullstack-server/extract.rs`

jkelleyrtp:
Thanks! I just added your changes to #4852 to test other changes in the query system. Should hope to have it merged later today. Since I moved the changes there, I'm going to close this PR. 

Appreciate the contribution!

bwskin:
> Thanks! I just added your changes to #4852 to test other changes in the query system. Should hope to have it merged later today. Since I moved the changes there, I'm going to close this PR.
> 
> Appreciate the contribution!

Hi, thanks for including this extractor! However, the issue I was trying to solve still exists in version 0.7.0 — the main query extractor is still the one from axum. I still can’t pass a struct as a query parameter unless I use a catch-all, which I believe will also fail with nested structs and blocks me from using additional parameters.

I managed to get it working correctly with the following patch:

```diff
diff --git a/packages/fullstack-macro/src/lib.rs b/packages/fullstack-macro/src/lib.rs
index 325ef7748..5ddd0530e 100644
--- a/packages/fullstack-macro/src/lib.rs
+++ b/packages/fullstack-macro/src/lib.rs
@@ -763,7 +763,7 @@ impl CompiledRoute {
         });
 
         out.push(parse_quote!(
-            dioxus_server::axum::extract::Query(#query_tokens)
+            dioxus_fullstack::payloads::Query(#query_tokens)
         ));
 
         out
```

Would you consider fixing this?

PR #4833, opened 2025-10-23
fix(hotpatch): Gracefully handle missing '-flavor' flag during WASM link

luckybelcik:
Problem: The unwrap() panic that occurs for WASM targets when the "-flavor" flag is absent from the linker arguments. (It happened to me, error shown at the bottom 😢)

Solution: Graceful missing '-flavor' flag handling

The error:
[dev] Thread tokio-runtime-worker panicked at packages\cli\src\build\request.rs:2258:76:
called Option::unwrap() on a None value
[dev] Build failed: Build panicked! JoinError::Panic(Id(33), "called Option::unwrap() on a None value", ...)


jkelleyrtp:
Thanks for the PR! I removed the logs you added, but otherwise lgtm

PR #4839, opened 2025-10-24
Fix macro path sanitization to use absolute paths

### LINKED ISSUE #4781: Proc macros are improperly sanitized
**Problem**

When using the `component` and `rsx` macros (and possibly others) without importing the prelude it will complain about

- use of unresolved module or unlinked crate `dioxus_core`
- use of unresolved module or unlinked crate `dioxus_elements`
- cannot find type `GlobalSignal`

**Steps To Reproduce**

Cargo.toml
```toml
...
dioxus = "0.6.3"
```

```rust
use dixous::prelude::{Element, rsx, component};

#[component(App)]
fn app() -> Element {
    rsx! {
        div { "This should work" }
    }
}
```

**Expected behavior**

This should work entirely as normal.

**Environment:**

- Dioxus version: 0.6.3
- Rust version:  1.90
- OS info: 6.15.8-arch1-2
- App platform:  Irrelevant

**Questionnaire**

Himmelschmidt:
Fixes #4781 by using absolute paths (::dioxus_core::, ::dioxus_html::, etc.) in all macro-generated code to prevent compilation errors when the prelude isn't imported.

### DIFF
--- a/packages/core-macro/src/props/mod.rs
+++ b/packages/core-macro/src/props/mod.rs
@@ -204,7 +204,7 @@ mod field_info {
                 // children field is automatically defaulted to None
                 if name == "children" {
                     builder_attr.default =
-                        Some(syn::parse(quote!(dioxus_core::VNode::empty()).into()).unwrap());
+                        Some(syn::parse(quote!(::dioxus_core::VNode::empty()).into()).unwrap());
                 }
 
                 // String fields automatically use impl Display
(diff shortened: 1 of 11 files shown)


nicoburns:
This may make sense for dioxus_core references, but the dioxus_html references are deliberately unhygienic to allow people to define their own elements, either by extending dioxus_html or by defining a completely different element set.

Himmelschmidt:
> This may make sense for dioxus_core references, but the dioxus_html references are deliberately unhygienic to allow people to define their own elements, either by extending dioxus_html or by defining a completely different element set.

I did not realize that, just reverting the `dioxus_html` changes breaks everything so I'm going to withdraw this for now

PR #4842, opened 2025-10-26
permissions, `manganis::ffi`, full Info.plist/AndroidManifest.xml customization

wheregmis:
This PR introduces a comprehensive **linker-based permission management for automatic update of manifest files** and **android native plugin building foundation** for Dioxus. The implementation is inspired by Manganis's linker-based asset collection approach.

Build on top of #4932 so diff is huge for now

Two main motivation for this PR 
1. To automatically handle the manifest file and info.plist file instead of invoking, ejecting or providing custom file. 
2. Able to pass in the java file path through linker, copy and bundle them (Opening doors for android native apis development)

### Currently packages/geolocation and geolocation example resides within this PR, its just for the ease of testing, and will not be upstreamed

Here is a breakdown of how things are working for now.
---

### Source layout
- Kotlin sources + Gradle wrapper live under `packages/geolocation/android/`
- The module builds a release `.aar`: `android/build/outputs/aar/geolocation-plugin-release.aar`

### build.rs responsibilities
`packages/geolocation/build.rs` builds the Gradle module whenever the target triple contains
`android`:

1. Respects the toolchain env exported by the CLI (`DX_ANDROID_JAVA_HOME`, `DX_ANDROID_SDK_ROOT`,
   `DX_ANDROID_NDK_HOME`, `DX_ANDROID_ARTIFACT_DIR`). If those are absent (non-`dx` builds) it falls
   back to `ANDROID_*` or `JAVA_HOME`.
2. Runs `./gradlew assembleRelease` inside `packages/geolocation/android`.
4. Copies the resulting `.aar` into the CLI-provided artifact staging dir
   (`$DX_ANDROID_ARTIFACT_DIR/geolocation-plugin-release.aar` or
   `$OUT_DIR/android-artifacts/...` as a fallback).
5. Emits `cargo:rustc-env=DIOXUS_ANDROID_ARTIFACT=<absolute path>` so the Rust macro can reference
   the built artifact.

### Metadata embedded in the Rust binary
In `packages/geolocation/src/lib.rs` we declare the Android artifact like this:

```rust
#[cfg(all(feature = "metadata", target_os = "android"))]
dioxus_platform_bridge::android_plugin!(
    plugin = "geolocation",
    aar = { env = "DIOXUS_ANDROID_ARTIFACT" },
    deps = ["implementation(\"com.google.android.gms:play-services-location:21.3.0\")"]
);
```

The `aar` block only needs a stable way to refer to the file that `build.rs` produced. Emitting a
`cargo:rustc-env=...` from the build script is the idiomatic Rust approach for passing data from
`build.rs` to the crate being compiled, which is why each plugin publishes its own environment
variable. Nothing prevents you from using unique names—`DIOXUS_GEO_ANDROID_ARTIFACT`,
`DIOXUS_CAMERA_ANDROID_ARTIFACT`, etc.—as long as your macro invocation references the same key that
your build script sets. Automatically inferring the path from the plugin name is brittle because the
artifact location lives in `target/` (or another `$OUT_DIR`) that only `build.rs` knows about after
running Gradle, so the build script remains the single source of truth.

The macro serializes:
- `plugin` name
- Absolute `aar` path (resolved via the env var emitted by `build.rs`)
- Any Gradle dependency lines listed in `deps`

The metadata is wrapped in `SymbolData::AndroidArtifact` and stored under the shared
`__ASSETS__*` symbol prefix (the same one used for assets and permissions). Nothing runs at runtime—
the symbols are just scanned later by the CLI.

### What the CLI does (`dx serve --android`)
1. After building the Rust binary, `dx` walks every `__ASSETS__*` symbol once. When it encounters a
   `SymbolData::AndroidArtifact`, it adds it to the manifest alongside assets and permissions.
2. `install_android_artifacts()` (in `packages/cli/src/build/request.rs`) copies each `.aar` into
   `target/dx/.../app/libs/` in the generated Gradle project.
3. For every artifact it ensures the Gradle dependency file contains both:
   - `implementation(files("libs/<artifact>.aar"))`
   - Any additional strings from `deps` (e.g. the Play Services location runtime)
4. When Gradle runs for the host app, those libraries are already on disk and referenced in
   `app/build.gradle.kts`, so they get packaged automatically.

> **Note**: because the `.aar` is built inside the Rust crate, plugin authors do not touch the app’s
> Gradle project. Everything happens by copying artifacts + editing `app/build.gradle.kts` inside the
> generated dx workspace.

---

### Source layout
- Swift sources + `Package.swift` live under `packages/geolocation/ios/`
- The Swift Package exports a single product: `GeolocationPlugin`

### build.rs responsibilities
For Apple targets, `build.rs`:

1. Detects the desired triple (`aarch64-apple-ios`, `aarch64-apple-ios-sim`, `x86_64-apple-ios`) and
   finds the matching SDK via `xcrun --show-sdk-path`.
2. Runs `xcrun swift build --package-path ios --configuration <debug|release> --triple <...> --sdk <...> --product GeolocationPlugin`.
3. Locates the resulting static library (`libGeolocationPlugin.a`) and emits:
   - `cargo:rustc-link-search` for both the Swift build output and the Xcode toolchain Swift libs
   - `cargo:rustc-link-lib=static=GeolocationPlugin`
   - `cargo:rustc-link-lib=swiftCompatibility*` + `-force_load` flags so the ObjC runtime sees the
     plugin class
   - `cargo:rustc-link-lib=framework=CoreLocation` / `Foundation`

At this point the Swift code is already linked into the Mach-O produced by Rust; no additional build
steps are required later.

### Metadata embedded in the Rust binary
In `packages/geolocation/src/lib.rs` we declare the Swift package via:

```rust
#[cfg(all(feature = "metadata", any(target_os = "ios", target_os = "macos")))]
dioxus_platform_bridge::ios_plugin!(
    plugin = "geolocation",
    spm = { path = "ios", product = "GeolocationPlugin" }
);
```

This writes an entry into the shared `__ASSETS__*` stream as `SymbolData::SwiftPackage`. Again,
nothing is executed at runtime; it is just discoverable metadata alongside assets and permissions.

### What the CLI does (`dx serve --ios` / `--macos`)
1. The CLI sees the `SymbolData::SwiftPackage` entries while it’s already processing the asset
   stream.
2. When at least one Swift package is present, `embed_swift_stdlibs()` runs (see
   `packages/cli/src/build/request.rs`). This invokes `xcrun swift-stdlib-tool --scan-executable` on
   the final app binary and copies the required Swift runtime libraries into the bundle’s
   `Frameworks/` folder (iOS) or `Contents/Frameworks/` (macOS).

Because the Swift package was linked during the Rust build, the CLI’s only job is to make sure the
Swift runtime ships alongside the executable.

---

## Summary table

| Platform | Native sources live | Compiled by | Metadata stored | CLI responsibilities |
|----------|--------------------|-------------|-----------------|----------------------|
| Android  | `packages/geolocation/android` (Gradle lib) | `build.rs` running `gradlew assembleRelease` | `SymbolData::AndroidArtifact` (plugin, path, deps) | Copy `.aar` into `app/libs`, append Gradle dependencies |
| iOS/macOS | `packages/geolocation/ios` (SwiftPM) | `build.rs` running `xcrun swift build` | `SymbolData::SwiftPackage` (plugin, package path, product) | Embed Swift stdlibs when Swift metadata is present |

This setup keeps each plugin self-contained:
- All native code + build tooling stays inside the crate.
- `build.rs` produces platform artifacts and emits env vars for linker metadata.
- The Dioxus CLI only needs the linker symbols to know what to copy/embed when bundling user apps.

New crates: (Please suggest the names as i am bad at naming stuff)
- **permissions**: Public API that re-exports permissions-core and permissions-macro for easy integration.
- **permissions-core**: Types for permission declaration and platform identifiers (Android, IOS, MacOS) (We can extend the identifiers as needed or requested)
- **permissions-macro**: Procedural Macros (static_permission!) for declaring permissions via linker symbol (So cli can extract the symbols and update the manifest files automatically)
- **dx-macro-helpers**: Common helpers for linker based stuff, shared by manganis and permission
- **platform-bridge**: Bareminimal Cross-platform FFI utilities and plugin metadata for Android (JNI) and Darwin (objc2)
- **platform-bridge-macro**: android_plugin! macro for embedding Java source file paths in binaries via linker symbols

### DIFF
--- a/packages/cli/schema.json
+++ b/packages/cli/schema.json
@@ -11,7 +11,7 @@
       "$ref": "#/definitions/ApplicationConfig"
     },
     "background": {
-      "description": "Unified background mode configuration. Background capabilities declared here are mapped to platform-specific configurations. Use [ios], [android] sections for overrides.",
+      "description": "Unified background mode configuration. Background capabilities declared here are mapped to platform-specific configurations. Use `[ios]`, `[android]` sections for overrides.",
       "$ref": "#/definitions/BackgroundConfig"
     },
     "bundle": {
@@ -21,7 +21,7 @@
       "$ref": "#/definitions/ComponentConfig"
     },
     "deep_links": {
-      "description": "Unified deep linking configuration. URL schemes and universal links declared here are mapped to platform-specific configurations. Use [ios], [android], [macos] sections for overrides.",
+      "description": "Unified deep linking configuration. URL schemes and universal links declared here are mapped to platform-specific configurations. Use `[ios]`, `[android]`, `[macos]` sections for overrides.",
       "$ref": "#/definitions/DeepLinkConfig"
     },
     "ios": {
@@ -159,7 +159,7 @@
           ]
         },
         "intent_filters": {
-          "description": "Intent filters for deep linking. These extend the unified [deep_links] configuration with Android-specific options.",
+          "description": "Intent filters for deep linking. These extend the unified `[deep_links]` configuration with Android-specific options.",
           "type": "array",
           "items": {
             "$ref": "#/definitions/AndroidIntentFilter"
@@ -255,7 +255,7 @@
           "minimum": 0.0
         },
         "url_schemes": {
-          "description": "Additional URL schemes beyond unified [deep_links].schemes. These are merged with the unified schemes.",
+          "description": "Additional URL schemes beyond unified `[deep_links]`.schemes. These are merged with the unified schemes.",
           "type": "array",
           "items": {
             "type": "string"
@@ -400,14 +400,14 @@
       "type": "object",
       "properties": {
         "application": {
-          "description": "Raw XML inside <application> element.",
+          "description": "Raw XML inside `<application>` element.",
           "type": [
             "string",
             "null"
           ]
         },
         "application_attrs": {
-          "description": "Raw attributes for <application> element.",
+          "description": "Raw attributes for `<application>` element.",
           "type": [
             "string",
             "null"
@@ -920,7 +920,7 @@
       "type": "object",
       "properties": {
         "background_modes": {
-          "description": "Additional background modes beyond unified [background]. Valid values: \"audio\", \"location\", \"voip\", \"fetch\", \"remote-notification\", \"newsstand-content\", \"external-accessory\", \"bluetooth-central\", \"bluetooth-peripheral\", \"processing\"",
+          "description": "Additional background modes beyond unified `[background]`. Valid values: \"audio\", \"location\", \"voip\", \"fetch\", \"remote-notification\", \"newsstand-content\", \"external-accessory\", \"bluetooth-central\", \"bluetooth-peripheral\", \"processing\"",
           "type": "array",
           "items": {
             "type": "string"
@@ -1037,7 +1037,7 @@
           ]
         },
         "url_schemes": {
-          "description": "Additional URL schemes beyond unified [deep_links].schemes. These are merged with the unified schemes.",
+          "description": "Additional URL schemes beyond unified `[deep_links]`.schemes. These are merged with the unified schemes.",
           "type": "array",
           "items": {
             "type": "string"
@@ -1731,7 +1731,7 @@
           ]
         },
         "url_schemes": {
-          "description": "Additional URL schemes beyond unified [deep_links].schemes. These are merged with the unified schemes.",
+          "description": "Additional URL schemes beyond unified `[deep_links]`.schemes. These are merged with the unified schemes.",
           "type": "array",
           "items": {
             "type": "string"
--- a/packages/cli/src/config/dioxus_config.rs
+++ b/packages/cli/src/config/dioxus_config.rs
@@ -26,13 +26,13 @@ pub(crate) struct DioxusConfig {
 
     /// Unified deep linking configuration.
     /// URL schemes and universal links declared here are mapped to platform-specific
-    /// configurations. Use \[ios\], \[android\], \[macos\] sections for overrides.
+    /// configurations. Use `[ios]`, `[android]`, `[macos]` sections for overrides.
     #[serde(default)]
     pub(crate) deep_links: DeepLinkConfig,
 
     /// Unified background mode configuration.
     /// Background capabilities declared here are mapped to platform-specific
-    /// configurations. Use [ios], [android] sections for overrides.
+    /// configurations. Use `[ios]`, `[android]` sections for overrides.
     #[serde(default)]
     pub(crate) background: BackgroundConfig,
 
--- a/packages/cli/src/config/manifest.rs
+++ b/packages/cli/src/config/manifest.rs
@@ -347,12 +347,12 @@ pub struct IosConfig {
     pub raw: IosRawConfig,
 
     // === Platform-specific overrides (extend unified config) ===
-    /// Additional URL schemes beyond unified \[deep_links\].schemes.
+    /// Additional URL schemes beyond unified `[deep_links]`.schemes.
     /// These are merged with the unified schemes.
     #[serde(default)]
     pub url_schemes: Vec<String>,
 
-    /// Additional background modes beyond unified [background].
+    /// Additional background modes beyond unified `[background]`.
     /// Valid values: "audio", "location", "voip", "fetch", "remote-notification",
     /// "newsstand-content", "external-accessory", "bluetooth-central",
     /// "bluetooth-peripheral", "processing"
@@ -635,13 +635,13 @@ pub struct AndroidConfig {
     pub application: AndroidApplicationConfig,
 
     // === Platform-specific overrides (extend unified config) ===
-    /// Additional URL schemes beyond unified \[deep_links\].schemes.
+    /// Additional URL schemes beyond unified `[deep_links]`.schemes.
     /// These are merged with the unified schemes.
     #[serde(default)]
     pub url_schemes: Vec<String>,
 
     /// Intent filters for deep linking.
-    /// These extend the unified \[deep_links\] configuration with Android-specific options.
+    /// These extend the unified `[deep_links]` configuration with Android-specific options.
     #[serde(default)]
     pub intent_filters: Vec<AndroidIntentFilter>,
 
@@ -908,7 +908,7 @@ pub struct MacosConfig {
     pub raw: MacosRawConfig,
 
     // === Platform-specific overrides (extend unified config) ===
-    /// Additional URL schemes beyond unified \[deep_links\].schemes.
+    /// Additional URL schemes beyond unified `[deep_links]`.schemes.
     /// These are merged with the unified schemes.
     #[serde(default)]
     pub url_schemes: Vec<String>,


packages/mobile-geolocation/src/sys/android/LocationCallback.java
```diff
@@ -0,0 +1,76 @@
...
+    public void onLocationChanged(Location location) {
+        this.executing = true;
+        if (!this.doNotExecute) {
+            rustCallback(this.handlerPtrHigh, this.handlerPtrLow, location);
+        }
+        this.executing = false;
+    }
+
+    /**
+     * NOTE: Technically implementing this function shouldn't be necessary as it has
```
mohe2015: I think this has something to do with Java 8 desugaring, see https://developer.android.com/tools/d8#j8

wheregmis: I tried with --lib as the docs but still shows the some error.  I am getting some error on String concat i think as per adb logs. Thank you for the documentation i will try to see if i can fix it.
```rust
 let dex_success = android_build::JavaRun::new()
        .class_path(d8_jar_path)
        .main_class("com.android.tools.r8.D8")
        .args([
            "--classpath",
            &android_jar_str.clone(),
            "--classpath",
            &out_dir_str.clone(),
            "--lib",
            &android_jar_str.clone(),
            "--output",
            &out_dir_str.clone(),
        ])
        .arg(&class_file)
        .run()
        .map_err(|_| BuildError::DexCompilationFailed)?
        .success();

```

wheregmis: If the issue is on mobile-geolocation, i think its okay for now, as i dont want to upstream it to dioxus but instead keep it somewhere out, maybe sdk or seperate crate. Its kinda like experimental crate to build the foundation of permissions and mobile build process. As i also am kinda noob, any feedbacks or pointing the mistakes is super highly appreciated 😬

wheregmis:
Yey, Android Permission Dialog is working.
Screenshot_1761495440

packages/mobile-geolocation/src/sys/android/LocationCallback.java
mohe2015: There is also API desugaring see https://developer.android.com/studio/write/java11-default-support-table and https://developer.android.com/studio/write/java8-support-table but it would depend which error you get exactly whether this is another problem. I also don't know how to integrate this manually.

mohe2015: For me with `d8` the `--lib` worked with the android.jar (one that is 27 megabytes large) from ~/Android/Sdk/platforms/android-36/android.jar

wheregmis: Hi @mohe2015 yes i fixed it with --lib + few java bindings, Now basic geolocation example crate works with ios and android automatically (no manual AndroidManifest.xml or info.plist editing or anything)

wheregmis:
Web:
image

wheregmis:
MacOs:
image

packages/permissions/permissions/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+//! > to preserve backward compatibility with existing code.
+
+pub use permissions_core::{
+    LocationPrecision, Permission, PermissionKind, PermissionManifest, Platform, PlatformFlags,
+    PlatformIdentifiers,
+};
+pub use permissions_macro::{permission, static_permission};
+
+#[doc(hidden)]
+pub mod macro_helpers {
```
ealmloff: This looks very similar to the code in manganis for the macro helpers. Could we generalize that code maybe with some generics instead of duplicating it here?

packages/permissions/permissions-macro/src/linker.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Generate a linker section for embedding permission data in the binary
+///
+/// This function creates a static array containing the serialized permission data
+/// and exports it with a unique symbol name that can be found by build tools.
+/// The pattern follows the same approach as Manganis for asset embedding.
+pub fn generate_link_section(permission: impl ToTokens, permission_hash: &str) -> TokenStream2 {
+    let position = proc_macro2::Span::call_site();
+    let export_name = syn::LitStr::new(&format!("__PERMISSION__{}", permission_hash), position);
+
+    quote::quote! {
```
ealmloff: This looks very similar to the manganis macro. Might be worth pulling out a linker export macro both crates use

packages/permissions/permissions-macro/README.md
```diff
@@ -0,0 +1,45 @@
...
+alias is kept for backward compatibility.
+
+## Usage
+
+```rust
+use permissions_core::Permission;
+use permissions_macro::static_permission;
+
+// Basic permission
+const CAMERA: Permission = static_permission!(Camera, description = "Take photos");
```
ealmloff: Instead of a special syntax for permissions, can we use a const builder like the assets macro options?

We started with a custom parser similar to this implementation, but moved away from it over time because:
1) It is harder to autocomplete and provide good inline documentation for
2) It is unfamiliar and harder to document
3) You need to update the parser every time you adjust the settings struct

packages/permissions/permissions/src/lib.rs
wheregmis: Thank you for the tip. Common helpers are extracted into a seperate crate for now, naming can be changed afterwards but for now i have named it dx-macro-helper

packages/permissions/permissions-macro/README.md
wheregmis: This is definately something we should do. I will try work on this asap.

wheregmis:
@ealmloff Thank you for the feedback. Other than these nits, What do you think about the overall process? Do you think this is something we should proceed with to upstream it? Or maybe try and figure some other build processes, toolings?  I want to discuss some top level ideas on the whole process and how should we proceed.

packages/permissions/permissions-macro/README.md
wheregmis: Now, i am using builder method like for assets macro. Thank you for sharing your experience. This advice is definately something that i can apply to my other community crate.

ealmloff:
> @ealmloff Thank you for the feedback. Other than these nits, What do you think about the overall process? Do you think this is something we should proceed with to upstream it? Or maybe try and figure some other build processes, toolings? I want to discuss some top level ideas on the whole process and how should we proceed.

@jkelleyrtp may have more thoughts here as he set up most of the android/ios build support in the CLI

From my perspective, I think this is close to the right approach, but I'm not sure this is the right interface yet. The two approaches for collecting bindings seem to be either build scripts (like robius and tauri) or linker-based collection like this and wasm-bindgen. Between the two linker-based collection is a lot easier to consume which will be important as the number of native bindings grows.

Both tauri or robius require you to manually register permissions which adds extra setup for any system libraries. That approach makes permissions more visible, but automatically registering permissions is much easier to use. We may want to think about adding a prompt to accept or reject the new permissions collected by the linker, but overall I think the linker permissions approach is much nicer.

One concern with the linker approach for bindings is symbols that are optimized away. Looking at the example linked in your description, it looks like you/AI had some issues with symbols being dropped that required adding `__ensure_permissions_linked`. We will need to either make sure the symbol isn't dropped within the macro by adding `#[used]` or tie the symbol to the usage of the script that is linked so it is only dropped if the bindings are not used

wheregmis:
> > @ealmloff Thank you for the feedback. Other than these nits, What do you think about the overall process? Do you think this is something we should proceed with to upstream it? Or maybe try and figure some other build processes, toolings? I want to discuss some top level ideas on the whole process and how should we proceed.
> 
> @jkelleyrtp may have more thoughts here as he set up most of the android/ios build support in the CLI
> 
> From my perspective, I think this is close to the right approach, but I'm not sure this is the right interface yet. The two approaches for collecting bindings seem to be either build scripts (like robius and tauri) or linker-based collection like this and wasm-bindgen. Between the two linker-based collection is a lot easier to consume which will be important as the number of native bindings grows.
> 
> Both tauri or robius require you to manually register permissions which adds extra setup for any system libraries. That approach makes permissions more visible, but automatically registering permissions is much easier to use. We may want to think about adding a prompt to accept or reject the new permissions collected by the linker, but overall I think the linker permissions approach is much nicer.
> 
> One concern with the linker approach for bindings is symbols that are optimized away. Looking at the example linked in your description, it looks like you/AI had some issues with symbols being dropped that required adding `__ensure_permissions_linked`. We will need to either make sure the symbol isn't dropped within the macro by adding `#[used]` or tie the symbol to the usage of the script that is linked so it is only dropped if the bindings are not used

Yes, __ensure_permissions_linked does need some nits, i agree on that part. I started with build script like how robius was working but its a hastle to build dex files copy it and alot of issues when the native plugins will grow. Supporting tauri native plugins would be awesome but i tried and failed as it needs some extra setups that we need to deal with to provide seamless experience. With the growing number of contributors and community i think we can definately provide better abstraction and maybe just sharing android/kotlin files with those plugins should be good. Also, another thing is instead of .java file only,  i want to support ,kt file for registering kotlin files and placing it on copying it on correct path for kotlin so we dont have to use java only. (Some plugins are freely available for kotlin and java side of things). Another discussion is on the IOS side of things, building and supporting each native swift package is tricky and challenging like how we are providing seamless for android but i am also not sure if we can do almost everything with objc crates. But the way its growing i think we should be able to do almost majority of things with it so not handling each ios plugin as a native swift package + swift files seems more appealing to me atm.

jkelleyrtp:
Going through the linker seems to be the right approach for permissions. Both build scripts and linker will require an end tool to collect the permissions, and I much prefer infrastructure-as-code approaches than the implicit build script. It's nice that not *using* a permission means not inserting it the Info.plist. Instead of `#[used]` attributes I would prefer that we make the permission a required argument to the function that requests the permission, and then we perform a volatile read, forcing the symbol to remain. This is similar to how `Asset` remains because its `Display` impl forces a volatile read.

We might need to think about Koitlin/Swift source files a bit more since cargo's cache busting system will be more reliable than the linker-based system, at least early on.

However, I'd rather not create a new symbol type for collecting permissions and instead try to reuse the one we use for assets. Though, the asset symbol format is not very flexible (see https://github.com/DioxusLabs/dioxus/issues/4863), so it might require some work on our end to allow different variants as long as they're the same size. Adding a new symbol seems to require duplicating logic in the CLI which we can avoid if we stick with one symbol with different variants.


---


Also, FWIW, this would be at the top of the list for 0.8 features. If we could get it in 0.7, then it would be at the top of priority behind fixing bugs, that way we can ship native apis while still on 0.7

wheregmis:
@jkelleyrtp Thank you for the inputs. All the recommendation and inputs seems super valid and reasonable. After looking into #4863 I do think if we can somehow provide a appropriate way of creating different variants, and we dont need to create multiple symbols, alot of code on the CLI for extracting symbols can be alot more refined and cleaned. For kotlin/swift sources files, I also dont have super huge usecase to get into alot more depth to learn and define the fundamentals so i would definately leave these big chunks to dioxus core members. For the time being i will try to update my codes to more like an enum of LinkerSymbol which can hold either asset or permission and try to cleanup the CLI code abit so we dont go with multiple symbol type and can slip the permissions under __MANGANIS__ symbol. And when we have a concrete plan for #4863 I can update my code for it. Also maybe we can try to slip in the permissions related stuff if possible into the 0.7 that way in with permissions, followed by build system for kotlin/swift and native apis can be added and shipped while still on 0.7.

wheregmis:
I think i am able to package swift files as swift package and package it and run. (Like how tauri does for swift files)

wheregmis:
I am also able to build each android plugin as a gradle submodule and can be tested like this inside each plugin
```rust
./gradlew assembleRelease --no-daemon --warning-mode all
```

-Oops i broke it lol- my bad it worked

wheregmis:
So the current plugin structure is something like this.
image

wheregmis:
Running gradle assembly takes time now, And i expect with the number of apis you add, it will grow exponentially

wheregmis:
Here is a breakdown of how things are working for now.
---

## Android (Kotlin / Gradle)

### Source layout
- Kotlin sources + Gradle wrapper live under `packages/geolocation/android/`
- The module builds a release `.aar`: `android/build/outputs/aar/geolocation-plugin-release.aar`

### build.rs responsibilities
`packages/geolocation/build.rs` builds the Gradle module whenever the target triple contains
`android`:

1. Respects the toolchain env exported by the CLI (`DX_ANDROID_JAVA_HOME`, `DX_ANDROID_SDK_ROOT`,
   `DX_ANDROID_NDK_HOME`, `DX_ANDROID_ARTIFACT_DIR`). If those are absent (non-`dx` builds) it falls
   back to `ANDROID_*` or `JAVA_HOME`.
2. Runs `./gradlew assembleRelease` inside `packages/geolocation/android`.
3. Copies the resulting `.aar` into the CLI-provided artifact staging dir
   (`$DX_ANDROID_ARTIFACT_DIR/geolocation-plugin-release.aar` or
   `$OUT_DIR/android-artifacts/...` as a fallback).
4. Emits `cargo:rustc-env=DIOXUS_ANDROID_ARTIFACT=<absolute path>` so the Rust macro can reference
   the built artifact.

### Metadata embedded in the Rust binary
In `packages/geolocation/src/lib.rs` we declare the Android artifact like this:

```rust
#[cfg(all(feature = "metadata", target_os = "android"))]
dioxus_platform_bridge::android_plugin!(
    plugin = "geolocation",
    aar = { env = "DIOXUS_ANDROID_ARTIFACT" },
    deps = ["implementation(\"com.google.android.gms:play-services-location:21.3.0\")"]
);
```

The `aar` block only needs a stable way to refer to the file that `build.rs` produced. Emitting a
`cargo:rustc-env=...` from the build script is the idiomatic Rust approach for passing data from
`build.rs` to the crate being compiled, which is why each plugin publishes its own environment
variable. Nothing prevents you from using unique names—`DIOXUS_GEO_ANDROID_ARTIFACT`,
`DIOXUS_CAMERA_ANDROID_ARTIFACT`, etc.—as long as your macro invocation references the same key that
your build script sets. Automatically inferring the path from the plugin name is brittle because the
artifact location lives in `target/` (or another `$OUT_DIR`) that only `build.rs` knows about after
running Gradle, so the build script remains the single source of truth.

The macro serializes:
- `plugin` name
- Absolute `aar` path (resolved via the env var emitted by `build.rs`)
- Any Gradle dependency lines listed in `deps`

The metadata is wrapped in `SymbolData::AndroidArtifact` and stored under the shared
`__ASSETS__*` symbol prefix (the same one used for assets and permissions). Nothing runs at runtime—
the symbols are just scanned later by the CLI.

### What the CLI does (`dx serve --android`)
1. After building the Rust binary, `dx` walks every `__ASSETS__*` symbol once. When it encounters a
   `SymbolData::AndroidArtifact`, it adds it to the manifest alongside assets and permissions.
2. `install_android_artifacts()` (in `packages/cli/src/build/request.rs`) copies each `.aar` into
   `target/dx/.../app/libs/` in the generated Gradle project.
3. For every artifact it ensures the Gradle dependency file contains both:
   - `implementation(files("libs/<artifact>.aar"))`
   - Any additional strings from `deps` (e.g. the Play Services location runtime)
4. When Gradle runs for the host app, those libraries are already on disk and referenced in
   `app/build.gradle.kts`, so they get packaged automatically.

> **Note**: because the `.aar` is built inside the Rust crate, plugin authors do not touch the app’s
> Gradle project. Everything happens by copying artifacts + editing `app/build.gradle.kts` inside the
> generated dx workspace.

---

## iOS / macOS (Swift Package)

### Source layout
- Swift sources + `Package.swift` live under `packages/geolocation/ios/`
- The Swift Package exports a single product: `GeolocationPlugin`

### build.rs responsibilities
For Apple targets, `build.rs`:

1. Detects the desired triple (`aarch64-apple-ios`, `aarch64-apple-ios-sim`, `x86_64-apple-ios`) and
   finds the matching SDK via `xcrun --show-sdk-path`.
2. Runs `xcrun swift build --package-path ios --configuration <debug|release> --triple <...> --sdk <...> --product GeolocationPlugin`.
3. Locates the resulting static library (`libGeolocationPlugin.a`) and emits:
   - `cargo:rustc-link-search` for both the Swift build output and the Xcode toolchain Swift libs
   - `cargo:rustc-link-lib=static=GeolocationPlugin`
   - `cargo:rustc-link-lib=swiftCompatibility*` + `-force_load` flags so the ObjC runtime sees the
     plugin class
   - `cargo:rustc-link-lib=framework=CoreLocation` / `Foundation`

At this point the Swift code is already linked into the Mach-O produced by Rust; no additional build
steps are required later.

### Metadata embedded in the Rust binary
In `packages/geolocation/src/lib.rs` we declare the Swift package via:

```rust
#[cfg(all(feature = "metadata", any(target_os = "ios", target_os = "macos")))]
dioxus_platform_bridge::ios_plugin!(
    plugin = "geolocation",
    spm = { path = "ios", product = "GeolocationPlugin" }
);
```

This writes an entry into the shared `__ASSETS__*` stream as `SymbolData::SwiftPackage`. Again,
nothing is executed at runtime; it is just discoverable metadata alongside assets and permissions.

### What the CLI does (`dx serve --ios` / `--macos`)
1. The CLI sees the `SymbolData::SwiftPackage` entries while it’s already processing the asset
   stream.
2. When at least one Swift package is present, `embed_swift_stdlibs()` runs (see
   `packages/cli/src/build/request.rs`). This invokes `xcrun swift-stdlib-tool --scan-executable` on
   the final app binary and copies the required Swift runtime libraries into the bundle’s
   `Frameworks/` folder (iOS) or `Contents/Frameworks/` (macOS).

Because the Swift package was linked during the Rust build, the CLI’s only job is to make sure the
Swift runtime ships alongside the executable.

---

## Summary table

| Platform | Native sources live | Compiled by | Metadata stored | CLI responsibilities |
|----------|--------------------|-------------|-----------------|----------------------|
| Android  | `packages/geolocation/android` (Gradle lib) | `build.rs` running `gradlew assembleRelease` | `SymbolData::AndroidArtifact` (plugin, path, deps) | Copy `.aar` into `app/libs`, append Gradle dependencies |
| iOS/macOS | `packages/geolocation/ios` (SwiftPM) | `build.rs` running `xcrun swift build` | `SymbolData::SwiftPackage` (plugin, package path, product) | Embed Swift stdlibs when Swift metadata is present |

This setup keeps each plugin self-contained:
- All native code + build tooling stays inside the crate.
- `build.rs` produces platform artifacts and emits env vars for linker metadata.
- The Dioxus CLI only needs the linker symbols to know what to copy/embed when bundling user apps.

wheregmis:
I have added the geolocation package + example for geolocation in for easy testing so the diff is huge as of now. I need to delete them before we upstream it.

wheregmis:
image
image
image

wheregmis:
@jkelleyrtp I am sorry for a large diff atm, it has changes from #4932 as well as package/geolocation and example inside. So its easier for you to check. I am going out for few weeks after 2-3 days so i wanted to make sure this PR is in good shape before that. A small review on the build process would be really appreciated. I think the way it turned out is really interesting + approach is something that can be shared across other rust UI projects (I wonder someone can fork tauri plugins and make it compatiable with dioxus as this opens the door for that as well)

jkelleyrtp:
> @jkelleyrtp I am sorry for a large diff atm, it has changes from #4932 as well as package/geolocation and example inside. So its easier for you to check. I am going out for few weeks after 2-3 days so i wanted to make sure this PR is in good shape before that. A small review on the build process would be really appreciated. I think the way it turned out is really interesting + approach is something that can be shared across other rust UI projects (I wonder someone can fork tauri plugins and make it compatiable with dioxus as this opens the door for that as well)

Awesome! Will try to get it in next week and then hopefully in a patch release soon after that

hogyzen12:
as someone who will need native permissions for usbc and bluetooth access on android and iOs i love seeing this being built. thank you team and everyone involved

wheregmis:
The PR is ready for review. I have tried to clean things up. Before merge, we can remove the package/geolocation and gelocation example from this branch. Its only included in the repo for ease of testing.

jkelleyrtp:
I'm in the midst of tweaking this to allow manganis to submit the source files of swift/java to the linker to avoid the need to create `build.rs` files.

This means binding to a java/kotlin/swift library looks like this:

```rust
#[manganis::ffi("/src/ios")]
extern "Swift" {
    /// The native GeolocationPlugin class
    pub type GeolocationPlugin;

    /// Get current position as JSON string
    /// Swift signature: func getCurrentPositionJson(_ optionsJson: String) -> String
    pub fn get_current_position_json(this: &GeolocationPlugin, options_json: String) -> String;

    /// Check permissions and return status as JSON
    /// Swift signature: func checkPermissionsJson() -> String
    pub fn check_permissions_json(this: &GeolocationPlugin) -> String;

    /// Request permissions with optional types list as JSON, return status as JSON
    /// Swift signature: func requestPermissionsJson(_ permissionsJson: String) -> String
    pub fn request_permissions_json(this: &GeolocationPlugin, permissions_json: String) -> String;
}
```

You can call methods and functions directly from the swift/kotlin files and then wrap them in a Rust interface. This requires deeply modifying `request.rs`- quite complex of a change - but will be extraordinarily useful in building bindings at scale.



----

As far as feedback - I squashed the many numerous crates directly into manganis. Our long-term goal with manganis is to be a general framework for submitting metadata through the linker into a final build tool. The new ffi type just exists as a folder-like asset that the CLI knows what to do with.

wheregmis:
> I'm in the midst of tweaking this to allow manganis to submit the source files of swift/java to the linker to avoid the need to create `build.rs` files.
> 
> This means binding to a java/kotlin/swift library looks like this:
> 
> ```rust
> #[manganis::ffi("/src/ios")]
> extern "Swift" {
>     /// The native GeolocationPlugin class
>     pub type GeolocationPlugin;
> 
>     /// Get current position as JSON string
>     /// Swift signature: func getCurrentPositionJson(_ optionsJson: String) -> String
>     pub fn get_current_position_json(this: &GeolocationPlugin, options_json: String) -> String;
> 
>     /// Check permissions and return status as JSON
>     /// Swift signature: func checkPermissionsJson() -> String
>     pub fn check_permissions_json(this: &GeolocationPlugin) -> String;
> 
>     /// Request permissions with optional types list as JSON, return status as JSON
>     /// Swift signature: func requestPermissionsJson(_ permissionsJson: String) -> String
>     pub fn request_permissions_json(this: &GeolocationPlugin, permissions_json: String) -> String;
> }
> ```
> 
> You can call methods and functions directly from the swift/kotlin files and then wrap them in a Rust interface. This requires deeply modifying `request.rs`- quite complex of a change - but will be extraordinarily useful in building bindings at scale.
> 
> As far as feedback - I squashed the many numerous crates directly into manganis. Our long-term goal with manganis is to be a general framework for submitting metadata through the linker into a final build tool. The new ffi type just exists as a folder-like asset that the CLI knows what to do with.

This looks really good. Thank you for picking and making it this good. I was not sure about manganis's scope on long term and created other crates to not break the current existing stuff (As i dont know all ins and outs). But overall the way it turned out after your changes is really impressive. Thank you again and kuddos to our dioxus community. We have open skies with this PR and cant wait to see how far we take it.

nicoburns:
I'm not seeing the purpose of the typed API for permissions (the enum of all permissions). Or the permissions mapping in `Dioxus.toml`. Wouldn't it better to treat everything as "custom" permissions (specified as strings)?

That way there would be much less code to maintain, it should work for all permissions not just ones we define, and the code to write permissions in Dioxus would be 1:1 with what is documented in the platform docs, so there's no additional DSL/mapping for users to learn.

jkelleyrtp:
This PR has been expanded to include
- full customization of ios, android manifests
- deep linking, url schemes, intent filters
- unified permissions interface and per-platform permissions 
- dynamic entitlements generation
- live activities / widget extensions for ios
- schemars generation
- swift + kotlin ffi using the `ffi` macro
- inline dioxus.toml support for examples/bins using the leading doc comment

PR #4849, opened 2025-10-27
restore the extract function for server functions

### LINKED ISSUE #4810: Fullstack::extract() broken [0.7.0-rc.2]
This function cannot be used with causing compile errors.

I believe this is because it is an async function, but uses non-thread-safe values. In particular `Rc<RefCell<...>>`.

This means when the function calls await, it may cross a thread boundary, and this fails.

So these values really need to be Arc. And the RefCell needs to be replaced with something thread safe also.

I have a test case to show the compiler error which I will link to shortly.

_Originally posted by @brianmay in https://github.com/DioxusLabs/dioxus/discussions/4807#discussioncomment-14715411_

ealmloff:
Fixes #4810

### DIFF
--- a/packages/fullstack/src/payloads/cbor.rs
+++ b/packages/fullstack/src/payloads/cbor.rs
@@ -21,7 +21,7 @@ use serde::{de::DeserializeOwned, Serialize};
 /// *last* if there are multiple extractors in a handler.
 /// See ["the order of extractors"][order-of-extractors]
 ///
-/// [order-of-extractors]: crate::extract#the-order-of-extractors
+/// [order-of-extractors]: mod@crate::extract#the-order-of-extractors
 #[must_use]
 pub struct Cbor<T>(pub T);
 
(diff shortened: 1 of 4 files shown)


packages/fullstack-core/src/streaming.rs
```diff
@@ -56,17 +91,24 @@ impl FullstackContext {
...
+    /// Run a future within the scope of this FullstackContext.
+    pub async fn scope<F, R>(self, fut: F) -> R
+    where
+        F: std::future::Future<Output = R>,
+    {
+        FULLSTACK_CONTEXT.scope(self, fut).await
+    }
+
+    /// Extract an axum extractor from the current request.
+    pub async fn extract<T: FromRequestParts<()>>() -> Result<T, ServerFnError> {
```
ealmloff: This changed from `FromRequest` to `FromRequestParts` which only allows types that don't consume the body (which is empty here)

The only types `FromRequest` is implemented for that  `FromRequestParts` is not are types like `Body` itself which will not behave correctly with an empty body in server functions

ealmloff: I restored this in this PR as it is largely unrelated, but we may want to either tighten the bound or make the body extractor work in the future

PR #4882, opened 2025-11-01
Note "d" for debug

Craig-Macomber:
The new debugging feature is hard to discover. This adds it to the command list in dioxus-cli.

Related: https://dioxuslabs.com/learn/0.7/guides/testing/debugging seems empty and should probably have some more information.

### DIFF
--- a/packages/cli/src/serve/output.rs
+++ b/packages/cli/src/serve/output.rs
@@ -33,7 +33,7 @@ use super::AppServer;
 const TICK_RATE_MS: u64 = 100;
 const VIEWPORT_MAX_WIDTH: u16 = 90;
 const VIEWPORT_HEIGHT_SMALL: u16 = 5;
-const VIEWPORT_HEIGHT_BIG: u16 = 13;
+const VIEWPORT_HEIGHT_BIG: u16 = 14;
 
 /// The TUI that drives the console output.
 ///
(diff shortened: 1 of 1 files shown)


packages/cli/src/serve/output.rs
```diff
@@ -851,6 +851,7 @@ impl Output {
             "v: toggle verbose logs",
             "t: toggle tracing logs ",
             "c: clear the screen",
+            "d: attach debugger",
             "/: toggle more commands",
         ];
         let layout: [_; 8] = Layout::vertical(cmds.iter().map(|_| Constraint::Length(1)))
```
Craig-Macomber: If I don't set this to 9, then I get a panic:

>    3.492s ERROR Thread main panicked at /home/craig/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ratatui-0.29.0/src/layout/layout.rs:530:35:
>
>               invalid number of rects: TryFromSliceError(()) 

If I do set it to 9 then the `v: toggle verbose logs` line does not show.

I'm not sure how to fix it: the layout stuff here isn't something I have worked with.

packages/cli/src/serve/output.rs
```diff
@@ -851,6 +851,7 @@ impl Output {
             "v: toggle verbose logs",
```
Craig-Macomber, line 851: It seems odd that `packages/cli/src/serve/mod.rs` displays:

                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts

If I then press `/` we have unrelated code here with displays a similar list, but its not actually a superset since `Press `ctrl+c` to exit the server` is missing and the phrasing is a little different.

Craig-Macomber:
My first attempt at this broke the layout logic, causing a panic if you hit `/` to expand, but passed all the tests. Thus I suspect there is missing test coverage for `/`. I figured I'd mention it in case that is useful information.

jkelleyrtp:
Good catch, thanks!

I didn't include ctrl-c in the list since it *seems* obvious after the initial log message, but I might be wrong.

PR #4899, opened 2025-11-03
Fix dioxus-native to run without the net feature

### LINKED ISSUE #4898: dioxus-native cannot be used without the `net` feature flag.
**Problem**

dioxus-native only adds blitz-net as a dependency when the net feature flag is present, but blitz-net is referenced in the code regardless of the feature flag.

**Steps To Reproduce**

- Use [todomvc](https://github.com/DioxusLabs/blitz/tree/main/examples/todomvc) example (any project will do) without `dioxus-native/net` flag
  ```
  [dependencies]
  dioxus = "0.7.0"
  dioxus-native = { version = "0.7.0", features = ["accessibility", "clipboard", "file_dialog", "hot-reload", "html", "svg", "system-fonts"], default-features = false }
  ...
  ```
- Run `cargo build`

Result:

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `blitz_net`
  --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-native-0.7.0/src/assets.rs:20:28
   |
20 |         let net_provider = blitz_net::Provider::shared(net_callback.clone());
   |                            ^^^^^^^^^ use of unresolved module or unlinked crate `blitz_net`
   |
help: there is a crate or module with a similar name
   |
20 -         let net_provider = blitz_net::Provider::shared(net_callback.clone());
20 +         let net_provider = blitz_dom::Provider::shared(net_callback.clone());
   |

For more information about this error, try `rustc --explain E0433`.
error: could not compile `dioxus-native` (lib) due to 1 previous error
```

**Expected behavior**

No error

**Environment:**

- Dioxus version:  0.7
- Rust version:    nightly
- OS info:         wsl2
- App platform:    native

**Questionnaire**

omasakun:
Avoid referencing blitz_net when the net feature flag is not specified.

Fixes #4898

### DIFF
--- a/packages/native/src/lib.rs
+++ b/packages/native/src/lib.rs
@@ -27,7 +27,6 @@ use blitz_shell::{create_default_event_loop, BlitzShellEvent, Config, WindowConf
 use dioxus_core::{ComponentFunction, Element, VirtualDom};
 use link_handler::DioxusNativeNavigationProvider;
 use std::any::Any;
-#[cfg(feature = "html")]
 use std::sync::Arc;
 use winit::window::WindowAttributes;
 


nicoburns (changes requested):
This PR looks like it will also disable asset support when the `net` feature is disabled. Could you change it so that the `DioxusNativeNetProvider` is always defined and used, but only depends on `blitz_net` when the `net` feature is enabled?

omasakun:
Whoops, my bad. I didn't realize that pushing to github immediately reflects in the pull request, so I ended up pushing at an incomplete stage. Does current version seem ok?

nicoburns:
@omasakun This is basically fine, although you could compile out the `inner_net_provider` field entirely when the `net` feature is not enabled.

nicoburns (approved):

PR #4942, opened 2025-11-09
Skip default members when canonicalize fails

### LINKED ISSUE #4873: dx-cli 0.7.0 fails on default-members
**Problem**

Moving from 0.7.0-rc.0 to 0.7.0 causes the dioxus project to fail with error:
```
dx serve 
   0.482s ERROR Thread main panicked at packages/cli/src/workspace.rs:309:74:
called `Result::unwrap()` on an  `Err`  value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

**Steps To Reproduce**

-  Clone https://github.com/newfla/rEav
- uncomment default_members in workspace Cargo.toml
- cd examples/dioxus
- dioxus --serve

**Expected behavior**

Normal Dioxus serving

**Environment:**

- Dioxus version:  0.7.0
- Rust version:   1.9.1
- OS info:         macOS 26.0.1
- App platform:    N.A.

newfla:
**Summary:**
Avoid panicking when searching default members in a current directory different from the main workspace folder.
Unfortunately #4895 was not targeting the actual issue of #4873.

**Details:**
Basically it avoid panicking on line 309. 
Unwrapping on 322 should be fine but for code style consistency I've also edited it.

Fix #4873

### DIFF
--- a/packages/cli/src/workspace.rs
+++ b/packages/cli/src/workspace.rs
@@ -318,7 +318,7 @@ impl Workspace {
                                 return None;
                             }
                             if std::fs::canonicalize(krate.manifest_path.parent().unwrap())
-                                .is_ok_and(|memeber_path| memeber_path == default_member_path)
+                                .is_ok_and(|member_path| member_path == default_member_path)
                             {
                                 return Some(id);
                             }


ealmloff:
> Avoid panicking when searching default members in a current directory different from the main workspace folder.

We shouldn't be unwrapping in case the Cargo.toml is wrong, but I don't think that is the main issue here. Looking at your reproduction again, the original PR fixed `lib`s in the default members but you also use glob paths which the CLI doesn't handle. You have `"examples/*"` which fails to canonicalize because it isn't a path. We should expand the glob path before looking for the crates

newfla:
> You have `"examples/*"` which fails to canonicalize because it isn't a path. We should expand the glob path before looking for the crates

What actually is panicking is canonicalize on the first memeber in default-members.
Calling dx inside examples/dioxus causes `realpath examples/dioxus` to fail: with this pr dx will stop trying default members and defaulting to the current dir.

PR #4948, opened 2025-11-10
fix: check HTTP status in FileStream::from_response

### LINKED ISSUE #4947: `FileStream` error handling
**Problem**

FileStream server functions return `Ok(_)` on the client side even when the server returns an error, making it impossible for clients to properly handle errors using `use_action`.

The `FileStream::from_response()` implementation does not check the HTTP status code before attempting to stream the response as a file. This causes all responses, including errors, to be wrapped in `Ok(FileStream)` instead of being propagated as `Err(ServerFnError)`.

**Steps To Reproduce**

1. Create a server function that returns `Result<FileStream>`
2. Have the server function return an error
3. Call this function from the client using `use_action`
4. Observe that `use_action` shows the result as `Some(Ok(_))` instead of `Some(Err(_))`

**Expected behavior**

When a FileStream server function returns an error (resulting in HTTP 500 or other error status codes), the client should receive `Err(ServerFnError)` so that:
- `use_action` shows `Some(Err(_))` instead of `Some(Ok(_))`
- Error handling logic in the client can properly catch and handle these errors
- The behavior is consistent with other server function return types

**Environment:**

- Dioxus version: v0.7.1 / main
- Rust version: 1.91.0
- OS info: Linux
- App platform: fullstack

**Questionnaire**

Himmelschmidt:
Closes #4947

Adds HTTP status code checking to `FileStream::from_response()` so server errors are properly propagated to the client as `Err(ServerFnError)` instead of `Ok(FileStream)`.

### DIFF
--- a/packages/fullstack/src/payloads/files.rs
+++ b/packages/fullstack/src/payloads/files.rs
@@ -249,6 +249,24 @@ impl<S> FromRequest<S> for FileStream {
 impl FromResponse for FileStream {
     fn from_response(res: ClientResponse) -> impl Future<Output = Result<Self, ServerFnError>> {
         async move {
+            // Check status code first - don't try to stream error responses as files
+            if !res.status().is_success() {
+                let status_code = res.status().as_u16();
+                let canonical_reason = res
+                    .status()
+                    .canonical_reason()
+                    .unwrap_or("Unknown error")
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Thanks!

PR #4950, opened 2025-11-10
Fix stores deadlock

### LINKED ISSUE #4935: Store with SyncStorage deadlocks
**Problem**

When I create a sync store and write to it from another thread while writing/reading it from dioxus renders, I get a deadlock between the two threads.

```rust
fn app() -> Element {
    let mut store: Store<HashMap<usize, String>, WriteSignal<HashMap<usize, String>, SyncStorage>>
        = use_hook(move || Store::new_maybe_sync(HashMap::<usize, String>::default())).into();
    use_hook({
        let mut s = store.clone();
        move || {
            thread::spawn(move || {
                let (mut n, mut rng) = (0usize, rand::thread_rng());
                loop {
                    n = n.wrapping_add(1);
                    for i in 1..120 {
                        s.write().insert(i, n.to_string());
                    }
                    thread::sleep(Duration::from_millis(rng.gen_range(1..120)));
                }
            });
        }
    });

    let items: Vec<_> = store.read().clone().into_iter().collect();

    rsx! {
        for (k, v) in items {
            input {
                key: "{k}",
                value: "{v}",
                oninput: move |e| { store.write().insert(k, e.value().into()); }
            }
        }
    }
}
```
Results in this log when I use the `deadlock_detection` feature from `parking_lot`:
## [deadlock from detect_deadlocks feature](https://github.com/user-attachments/files/23426636/deadlock.txt)
Sometimes I have to resize the window or restart the app to make it occur.

**Expected behavior**
No deadlocking when writing/reading to SyncStorage stores from another thread and the Dioxus thread.

**Environment:**

- Dioxus version:  0.17.1
- Rust version:    1.92.0-nightly
- OS info:  Arch Linux x86_64
- App platform: Desktop/Webview

**Questionnaire**

I am trying to figure out how to make SelectorScope<Lens> on write return `WriteMetadata` that on drop will do the mark_dirty, but I'm a bit over my head with all the abstractions here diving in from scratch.

I believe that the issue is an AB BA locking problem where SelectorScope<Lens> immediately does `self.mark_dirty()` on writes: https://github.com/DioxusLabs/dioxus/blob/main/packages/stores/src/scope.rs#L212 which I believe tries to lock subscribers to notify them, then tries to grab the write lock on the value, but when rendering the scope with a read on the signal - I believe that the dioxus thread is trying to grab a lock on the value, then trying to grab a lock on the subscribers.

This is in contrast to signals which return a guard which notifies subscribers when dropped: https://github.com/DioxusLabs/dioxus/blob/main/packages/signals/src/signal.rs#L528

Also, a side note, I'm not sure if the way I made the hook is correct - but it would be nice to have this in `prelude`:
```rust
pub fn use_sync_store<T, F>(make_default: F) -> Store<T, WriteSignal<T, SyncStorage>>
where
    T: 'static + Send + Sync,
    F: FnOnce() -> T + 'static,
{
    use_hook(move || Store::new_maybe_sync(make_default())).into()
}
```

ealmloff:
We can't hold the stores subscription lock while calling `mark_dirty` because it runs arbitrary code which may try to add subscriptions. This PR fixes the issue by collecting a list of paths in the store that need to be marked as dirty first before marking each one as dirty without holding the lock over the `mark_dirty` call

Fixes #4935

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6351,6 +6351,7 @@ dependencies = [
  "dioxus-core",
  "dioxus-signals",
  "dioxus-stores-macro",
+ "generational-box",
 ]
 
 [[package]]
--- a/packages/stores/Cargo.toml
+++ b/packages/stores/Cargo.toml
@@ -13,6 +13,7 @@ description = "Server function macros for Dioxus"
 dioxus-core = { workspace = true }
 dioxus-signals = { workspace = true }
 dioxus-stores-macro = { workspace = true, optional = true }
+generational-box.workspace = true
 
 [dev-dependencies]
 dioxus = { workspace = true }
--- a/packages/stores/src/impls/btreemap.rs
+++ b/packages/stores/src/impls/btreemap.rs
@@ -1,12 +1,15 @@
 //! Additional utilities for `BTreeMap` stores.
 
-use std::{borrow::Borrow, collections::BTreeMap, hash::Hash, iter::FusedIterator};
+use std::{
+    borrow::Borrow, collections::BTreeMap, hash::Hash, iter::FusedIterator, panic::Location,
+};
 
 use crate::{store::Store, ReadStore};
 use dioxus_signals::{
     AnyStorage, BorrowError, BorrowMutError, ReadSignal, Readable, ReadableExt, UnsyncStorage,
     Writable, WriteLock, WriteSignal,
 };
+use generational_box::ValueDroppedError;
 
 impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
     Store<BTreeMap<K, V>, Lens>
@@ -248,15 +251,7 @@ impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
         Q: Hash + Ord + 'static,
         K: Borrow<Q> + Ord,
     {
-        self.contains_key(&key).then(|| {
-            self.into_selector()
-                .hash_child_unmapped(key.borrow())
-                .map_writer(move |writer| GetWrite {
-                    index: key,
-                    write: writer,
-                })
-                .into()
-        })
+        self.contains_key(&key).then(|| self.get_unchecked(key))
     }
 
     /// Get a store for the value associated with the given key without checking if the key exists.
@@ -273,16 +268,19 @@ impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
     /// store.insert(0, "value".to_string());
     /// assert_eq!(store.get_unchecked(0).cloned(), "value".to_string());
     /// ```
+    #[track_caller]
     pub fn get_unchecked<Q>(self, key: Q) -> Store<V, GetWrite<Q, Lens>>
     where
         Q: Hash + Ord + 'static,
         K: Borrow<Q> + Ord,
     {
+        let created = std::panic::Location::caller();
         self.into_selector()
             .hash_child_unmapped(key.borrow())
             .map_writer(move |writer| GetWrite {
                 index: key,
                 write: writer,
+                created,
             })
             .into()
     }
@@ -293,6 +291,7 @@ impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
 pub struct GetWrite<Index, Write> {
     index: Index,
     write: Write,
+    created: &'static Location<'static>,
 }
 
 impl<Index, Write, K, V> Readable for GetWrite<Index, Write>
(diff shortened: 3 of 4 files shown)


alextechcc:
So fast! Thank you for looking into this.

I still seem to be able to deadlock - working on a minimal reproduction at the moment and a trace. Happens in my bevy/dioxus sync library I am working on.

EDIT: This may have been a false positive. Going to let my app run for a while.

alextechcc:
I've found a probably separate but somewhat related issue where this unwrap fails when iterating over the list:
https://github.com/ealmloff/dioxus/blob/fix-store-deadlock/packages/stores/src/impls/hashmap.rs#L86

Crash Log:
[crash.txt](https://github.com/user-attachments/files/23464691/crash.txt)

I can make another issue if necessary.

This one is hard to replicate. So, I've put my test here:
https://github.com/alextechcc/bevy_store_deadlock

It crashes faster if I type in the input fields, but crashes eventually naturally.

The gist is
1. A thread is randomly adding, mutating, and removing hashmap store entries.
2. A dioxus thread is rendering the hashmap by iterating over it.

This results in a crash where I believe what is happening is that there isn't a lock held preventing removal of elements when this block is running:

```rust
        keys.into_iter().map(move |key| {
            let value = self.clone().get(key.clone()).unwrap();
            (key, value)
        })
```

So I think while the map is running, `key` no longer points to a valid entry, and unwrap fails.

ealmloff:
That specific issue should be fixed, but your reproduction still panics. You can't easily use fail-able mappings like `Store<Hashmap>::get` in multithreaded settings since the other thread could take the value out from under you after the item is removed

alextechcc:
I guess kinda the point of the design of stores is being able to grab handles to substructures specifically without moving ownership or holding a lock - so it's a natural consequence that the value may disappear under you while iterating. And if you could grab a lock on the structure and pass it around (to prevent the value disappearing) - then any callbacks defined would just prevent mutation of whatever they're grabbing a handle to until they go away - so a lock wouldn't work.

Maybe I'm missing something here. I can't do this:
```rust
        for (_entity , mut name) in store.iter() {
            input {
                value: format!(
                    "{}",
                    name.try_read().map_or(String::new(), |name| name.as_str().to_string()),
                ),
            }
        }
```
I think it'd be more ergonomic to allow iteration in sync, so long as you handle the error when you read the value. Otherwise it feels a bit difficult for someone new to Rust to figure out that you basically have to iterate over keys and the values are not safe to touch.

Does it make sense to do something like this change for GetWrite in HashMap/btreemap?
```diff
@@ -316,7 +316,7 @@ where
             Self::Storage::map(value, |value: &Write::Target| {
                 value
                     .get(&self.index)
-                    .expect("Tried to access a key that does not exist")
+                    .unwrap_or(BorrowError::Dropped(self.index))
             })
         })
     }

```

Also, not to expand on the scope much, but I believe the unwrap is also here:
https://github.com/ealmloff/dioxus/blob/fix-store-deadlock/packages/stores/src/impls/btreemap.rs#L80

ealmloff:
> I think it'd be more ergonomic to allow iteration in sync, so long as you handle the error when you read the value. Otherwise it feels a bit difficult for someone new to Rust to figure out that you basically have to iterate over keys and the values are not safe to touch.

I think this is reasonable behavior to expect when reading an index that no longer exists and both provides a better error message and makes some iteration more ergonomic.

However, in general iteration with sync stores is still fraught as items could also be added after the iterator is created.

alextechcc:
> > I think it'd be more ergonomic to allow iteration in sync, so long as you handle the error when you read the value. Otherwise it feels a bit difficult for someone new to Rust to figure out that you basically have to iterate over keys and the values are not safe to touch.
> 
> I think this is reasonable behavior to expect when reading an index that no longer exists and both provides a better error message and makes some iteration more ergonomic.
> 
> However, in general iteration with sync stores is still fraught as items could also be added after the iterator is created.

So I guess in this scenario - you may get in a state where the component is not rerendered despite a new hashmap entry being added? I guess Dioxus isn't really set up for this - where dirtyness can change while being rendered. I feel like this is a trap for Dioxus users and maybe sync stores should not be available until it's addressed.

_The error message is good - but what I was suggesting above was to make it so you're iterating over a Result type where you have to explicitly handle the fact that the key could've disappeared - instead of it panicking without you being able to handle it. If the error happens rarely - it could lead to a rare crash in a production application - rather than a user just handling the issue explicitly.

As it stands - sync hashmaps are a bit odd of an API surface - if you want to iterate, you have to use .iter()- throw out the value (because it can panic if you use it), then use `.get()` with the key - and there is no `.keys()` iterator, just a `.values()._
EDIT: I see the new changes now

PR #4959, opened 2025-11-11
Fix marking nodes as dirty when inserting into an existing hashmap entry

### LINKED ISSUE #4956: `WritableHashMapExt::insert` doesn't trigger rerender when overwriting an item.
Title says it all — I tried using a `Store<HashMap<K, V>>` and calling `store.insert(k, v)` repeatedly for the same `k` and different `v` but none of my components re-rendered.

I ended up having to do two codepaths: `.get_mut()` and replace v in there, that worked, or `.insert()` if it didn't contain it at all.

ealmloff:
Inserting into a hashmap or btreemap currently only marks the hashmap itself as dirty which will cause any `get` or `iter` operations to rerun, but it doesn't mark the existing value at the entry as dirty. This PR fixes that issue.

Fixes #4956

### DIFF
--- a/packages/stores/src/impls/btreemap.rs
+++ b/packages/stores/src/impls/btreemap.rs
@@ -136,7 +136,6 @@ impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
         // in the map and mark the whole map as dirty if it does.
         // In the 0.8 release, we should change this method to only mark the existing value as dirty.
         if self.peek().contains_key(&key) {
-            println!("marking map as dirty");
             self.selector().mark_dirty();
         } else {
             self.selector().mark_dirty_shallow();


packages/stores/src/impls/btreemap.rs
```diff
@@ -131,7 +131,16 @@ impl<Lens: Readable<Target = BTreeMap<K, V>> + 'static, K: 'static, V: 'static>
         K: Ord,
         Lens: Writable,
     {
-        self.selector().mark_dirty_shallow();
+        // TODO: This method was released in 0.7 without the hash bound so we don't have a way
+        // to mark only the existing value as dirty. Instead we need to check if the value already exists
+        // in the map and mark the whole map as dirty if it does.
+        // In the 0.8 release, we should change this method to only mark the existing value as dirty.
+        if self.peek().contains_key(&key) {
+            println!("marking map as dirty");
```
jkelleyrtp: println

jkelleyrtp:
Re: the comment on the changed bit, we do have the `Ord` bound which implies `Eq` + `PartialEq` - is there no way to use those to determine if it can be marked dirty?

ealmloff:
> Re: the comment on the changed bit, we do have the `Ord` bound which implies `Eq` + `PartialEq` - is there no way to use those to determine if it can be marked dirty?

Not that I can think of. Store selectors need to be stable over time and copy. Currently they are `[u32; LEN]` where each entry is either a list index or a hash of the type. The other index I could think of for BTrees is the order in the tree, but that is not stable over time if you insert an element before the element you are holding. For example, that would break this case:

```rust
let tree = store_hashmap();
// Creates path [0]
tree.insert(1, String::new());
// Holds path [0] in the selector subscription tree
let first = tree.get(1);
// Creates path [0] and shift the original value to the second item
tree.insert(0, "wrong value".to_string());
// This now references (the subscriptions for) "wrong value"
first.read();
```

PR #4961, opened 2025-11-13
fix(fullstack): Streaming::into_inner() returning empty stream

ryo33:
On 0.6.3, [JsonStream::into_inner](https://docs.rs/server_fn/0.6.15/src/server_fn/codec/json.rs.html#126-128) always returns the valid stream but [Streaming::into_inner on 0.7](https://github.com/DioxusLabs/dioxus/blob/03ab7fa4821ff18c175d8576e966aec994030ee5/packages/fullstack/src/payloads/stream.rs#L155-L157) returns an empty stream on both client and server sides where data was placed in output_stream.

This PR consolidates `input_stream`/`output_stream` fields to a single `stream` field that always returns the actual stream data when `into_inner()` is called.

Also adds examples demonstrating bidirectional streaming:
- echo_stream: passes through the received `Streaming`
- transform_stream: transforms client stream on server using filter_map (uses `.into_inner()` on both server and client)

### DIFF
--- a/examples/07-fullstack/streaming.rs
+++ b/examples/07-fullstack/streaming.rs
@@ -60,8 +60,9 @@ fn app() -> Element {
     let mut start_echo_stream = use_action(move || async move {
         continue_echo_stream.set(true);
         echo_responses.clear();
-        let stream = echo_stream(Streaming::new(
-            futures::stream::unfold(0, move |index| async move {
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Thank you!

Sorry this sat - the holidays were a bit busy for us.

PR #4965, opened 2025-11-14
install.sh: Change shell interpreter from sh to bash

raptorswing:
The `[[ ]]` syntax on line 16 and elsewhere, i.e. `if [[ -t 1 ]]; then` doesn't work in sh - it's a [bash feature](https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html#index-_005b_005b).

Tested on Ubuntu 25.10. The current install script errors with:

```
./install.sh: 16: [[: not found
./install.sh: 50: [[: not found
./install.sh: 69: [[: not found
#=O#-    #       #                                                                                                                                                                                                                    curl: (22) The requested URL returned error: 404
```

On this system, /bin/sh is symlinked to dash.

Changing to `#!/bin/bash` solves the problem, and the script works as expected.

It was probably working in many places because some systems symlink /bin/sh to /bin/bash.

### DIFF
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -eo pipefail
 
 # Reset


imbolc:
I don't think that's enough, as in docs curl piped to sh. Why not convert the script to posix instead? LLMs are good at that.

raptorswing:
> I don't think that's enough, as in docs curl piped to sh

D'oh, good point. Opened a separate PR to update the docs: https://github.com/DioxusLabs/docsite/pull/594

> Why not convert the script to posix instead? LLMs are good at that.

Rewriting it to be POSIX shell compatible is a good idea if it truly needs to run under /bin/sh, and maybe we'd get lucky asking an LLM to do it, but I'm not familiar enough with POSIX shell scripting to confidently review the LLM's output. 

If updating to bash is undesirable, that's OK, I can withdraw these PRs or you can close them.

jkelleyrtp:
Closing in favor of https://github.com/DioxusLabs/dioxus/pull/5058 which properly implements sh idioms

PR #4966, opened 2025-11-14
Disable strip setting in profile and manually strip with llvm the build

### LINKED ISSUE #4314: wasm-opt fails to optimize and produces empty output
**Problem**

This is probably not fully actionable but maybe it's good that you are aware of the issue. For my own project (see below) wasm-opt fails to optimize and produces and empty output. The build shows the following:
`ERROR wasm-opt failed with status code exit status: 1:`

**Steps To Reproduce**

- Compile https://github.com/tucant/tucant/tree/03c63e32e7b4a44662ff8e3457d366f8cf4517fc using `dx build --release` with current `main` of dioxus. Note that I depend on a fork of `dioxus` in the project so this may theoretically have some effect though I don't believe so.

**Expected behavior**

Compilation should succeed and produce working output.

**Environment:**

- Dioxus version: main
- Rust version: 1.87.0
- OS info: Fedora
- App platform: web

**Questionnaire**

### LINKED ISSUE #4874: dx-cli 0.7.0 fails to bundle assets when using cargo flag strip=true
**Problem**
Tested target in linux and android and both not worked
`target/dx/dioxus-test/release/linux/app` the folder assets appears empty when using the following flags:
`.cargo/config.toml`
```toml
[profile.release]
opt-level = "z"
debug = false
lto = true
codegen-units = 1
panic = "abort"
strip = true # <-- this is the problematic flag
incremental = false
```
**Steps To Reproduce**

- `dx new program`
- Add config file as mentioned in the problem
- run: `dx build --linux --release`, the final result will not contain the assets so my styles not appears

**Expected behavior**

Assets being bundled properly

**Screenshots**

Image

**Environment:**

- Dioxus version:  0.7.0
- Rust version:    1.92.0-nightly
- OS info:         Arch Linux, kernel: 6.17.5
- App platform:    Android and desktop (Linux)

ealmloff:
Instead of just warning when strip is enabled, this PR disables strip in our profile override and re-applies it after we are done with the symbols with rustc-objcopy.

This is working correctly with and without strip enabled in the profile on macos and WASM, but I still need to test this on linux and windows setups

Testing:
- [x] Macos
- [x] WASM
- [x] Windows
- [x] Linux

Fixes #4874
Fixes #4314

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -4045,7 +4045,7 @@ impl BuildRequest {
             || self.wasm_split
             || !self.release
             || ctx.mode == BuildMode::Fat;
-        let keep_names = ctx.mode == BuildMode::Fat;
+        let keep_names = self.wasm_split || ctx.mode == BuildMode::Fat;
         let demangle = false;
         let wasm_opt_options = WasmOptConfig {
             memory_packing: self.wasm_split,
@@ -4179,7 +4179,6 @@ impl BuildRequest {
         // Make sure to optimize the main wasm file if requested or if bundle splitting
         if should_bundle_split || self.release {
             ctx.status_optimizing_wasm();
-            tracing::info!("Optimizing wasm file");
             wasm_opt::optimize(&post_bindgen_wasm, &post_bindgen_wasm, &wasm_opt_options).await?;
         }
 


nicoburns:
@ealmloff I'm still not getting assets through with this.

To reproduce:
- Compile CLI from Dioxus main
- In Blitz (main) run `dx bundle --profile production -p browser`

The resultant bundle does not contain any assets.

ealmloff:
> @ealmloff I'm still not getting assets through with this.
> 
> To reproduce:
> 
> * Compile CLI from Dioxus main
> * In Blitz (main) run `dx bundle --profile production -p browser`
> 
> The resultant bundle does not contain any assets.

#5052 should fix that issue. I was testing with `strip = true` in the release profile which this PR fixes, but it skips modifying the profile if you explicitly pass a custom one

PR #4980, opened 2025-11-17
Fixing regression due to route preservation

### LINKED ISSUE #4934: Hot reload / hot patch problem
**Problem**

I use `dx serve --target i686-pc-windows-msvc --watch false --hot-reload false`.

And often when I press the R button (rebuild app), I wait for a while for the program to build, and then the program opens, but it doesn't start from the beginning, it continues from where it left off but resets the state that does not belong to dioxus.

Before that, I used dioxus-cli 0.6.3 and didn't have this problem.

**Screenshots**
After first start
Image

Before rebuild
Image

After rebuild

Image

**Environment:**

- Dioxus version:  0.7.0, 0.7.1
- Rust version: 1.91
- OS info:  Win 11
- App platform: desktop, webview

**Questionnaire**

RichardWGNR:
When closing the application, the URL was saved for subsequent restoration the next time it was opened.
**This is completely incorrect behavior (which breaks my development).**

This change was made in the very first release of 0.7.x by @jkelleyrtp 
Perhaps you would like to explain yourself?

Fixes #4934

### DIFF
--- a/packages/desktop/src/app.rs
+++ b/packages/desktop/src/app.rs
@@ -7,8 +7,7 @@ use crate::{
     shortcut::ShortcutRegistry,
     webview::{PendingWebview, WebviewInstance},
 };
-use dioxus_core::{consume_context, ScopeId, VirtualDom};
-use dioxus_history::History;
+use dioxus_core::VirtualDom;
 use std::{
     cell::{Cell, RefCell},
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


jkelleyrtp (approved):

jkelleyrtp:
The behavior was originally added to help improve development workflows where you would need to re-navigate to a particular route between rebuilds. Resuming from a route helps improve that roadblock.

With hot-patching enabled, resuming from a route is less important since URL state is preserved between reloads.

MRWGNRRR:
> The behavior was originally added to help improve development workflows where you would need to re-navigate to a particular route between rebuilds. Resuming from a route helps improve that roadblock.
> 
> With hot-patching enabled, resuming from a route is less important since URL state is preserved between reloads.

Thanks

PR #4991, opened 2025-11-18
Enable musl target builds with vendored libgit2

### LINKED ISSUE #4373: CLI: Statically linked musl binaries for x86_64 and aarch64
Right now, the release binaries are dynamically linked and require glibc. This makes it hard to use them in minimal Docker containers like Alpine or scratch-based images.

It would be great to have prebuilt statically linked binaries using musl for both x86_64 and aarch64 so we can run `dioxus-cli` in any environment without needing glibc.

Please consider adding musl builds to the release.

xsoheilalizadeh:
Fixes #4373

  Enable statically-linked musl binaries for x86_64 and aarch64 by using
  vendored libgit2. This resolves the linking issues that previously
  caused musl builds to fail in CI.

  Changes:
  - Add target-specific git2 dependency with vendored-libgit2 feature for
    musl targets in packages/cli/Cargo.toml
  - Re-enable x86_64-unknown-linux-musl and aarch64-unknown-linux-musl
    build targets in publish workflow

  The vendored-libgit2 feature builds libgit2 from source and statically
  links it, avoiding musl-specific system library linking issues. This
  only affects musl builds - other platforms continue using system
  libraries for faster builds.

### DIFF
--- a/packages/fullstack/src/payloads/websocket.rs
+++ b/packages/fullstack/src/payloads/websocket.rs
@@ -826,7 +826,6 @@ pub enum WebsocketError {
     Uninitialized,
 
     #[cfg(not(target_arch = "wasm32"))]
-    #[cfg_attr(docsrs, doc(cfg(not(target_arch = "wasm32"))))]
     #[error("websocket upgrade failed")]
     Handshake(#[from] native::HandshakeError),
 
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
LGTM

PR #5004, opened 2025-11-21
force the user to provide "." as path

thyseus:
when creating a project in the current directory is requested.

The default behavior simply creates the project at "." no matter where i are. Maybe i am not even in a rust project or in my home directory. This most probably is not the wanted behavior the user requested.

When making the parameter required the user is forced to consciously confirm that he wants that behavior by providing a .

```
error: the following required arguments were not provided:
  <PATH>

Usage: dx init <PATH>
```

```
Init a new project for Dioxus in a given directory. Will attempt to keep your project in a good state

Usage: dx init [OPTIONS] <PATH>

Arguments:
  <PATH>  Please provide "." if you are sure you want to create the project inside the current directory
```

### DIFF
--- a/packages/cli/src/cli/init.rs
+++ b/packages/cli/src/cli/init.rs
@@ -4,11 +4,13 @@ use cargo_generate::{GenerateArgs, TemplatePath, Vcs};
 #[derive(Clone, Debug, Default, Deserialize, Parser)]
 #[clap(name = "init")]
 pub struct Init {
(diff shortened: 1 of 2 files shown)


LeWimbes:
I think the current behavior is good as it is.
It matches the way [`cargo init`](https://doc.rust-lang.org/cargo/commands/cargo-init.html) works.

thyseus:
You are right. Closing the issue.

PR #5044, opened 2025-12-01
cli: Add a wait screen (HTML) instead of a plain text error when reverse proxy hits a non-started server

tekacs:
When waiting for the backend to load in fullstack builds, we currently see an error message that's a plaintext response from the reverse proxy, which... stalls out reload workflows, hard. If doing a full rebuild, you can end up hitting this screen, and then you have to go refresh all the windows that got stuck here.

This PR adds a simple HTML page that shows the error templated in, but tries to refresh periodically, trying to get to a working version of the page.

To save from it looking ugly as sin, I had an LLM generate inline styling here and it's... very aesthetic, but happy to adjust to match Dioxus's styling as appropriate!

### DIFF
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -132,6 +132,7 @@ itertools = { workspace = true }
 throbber-widgets-tui = "0.8.0"
 unicode-segmentation = "1.12.0"
 handlebars = "6.3.2"
+html-escape = "0.2.13"
 strum = { version = "0.27.1", features = ["derive"] }
 memmap = "0.7.0"
 walrus = { workspace = true, features = ["parallel"] }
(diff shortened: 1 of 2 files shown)


ealmloff:
Instead of making this error screen nicer, could we just avoid it entirely by waiting until the proxy responds before sending the reload message to `/_dioxus/build_status`?

tekacs:
So... that'd work, but it would presumably not work on startup – only on reload, right? As in... if you started `dx serve --fullstack` and then opened the browser... with that approach you'd still just see a plain text error message?

The only way I could see 'waiting' work is if the proxy just held the connection open and stalled (not responding) until the upstream responded?

Am I... understanding that right?

ealmloff:
> So... that'd work, but it would presumably not work on startup – only on reload, right? As in... if you started `dx serve --fullstack` and then opened the browser... with that approach you'd still just see a plain text error message?
> 
> The only way I could see 'waiting' work is if the proxy just held the connection open and stalled (not responding) until the upstream responded?
> 
> Am I... understanding that right?

Currently while building you get a loading screen that shows the progress of the build that redirects to the finished page once it is ready. Ideally we could re-use that same screen here by just not sending the reload request until the server boots. The loading screen uses an incomplete http response to keep the socket open so it doesn't register as loaded in the browser for playwright tests

PR #5047, opened 2025-12-02
Fix exit_on_last_window_close for UserWindowEvent::CloseWindow

Klemen2:


Klemen2:
Actually, is there a reason why UserWindowEvent::CloseWindow uses handle_close_msg instead of handle_close_requested?

jkelleyrtp:
> Actually, is there a reason why UserWindowEvent::CloseWindow uses handle_close_msg instead of handle_close_requested?

That's a very good point. Not sure why. I updated your PR to use handle_close_requested instead. Thanks!

PR #5049, opened 2025-12-02
Resolves #5045:  Trim slashes to prevent base path resolution issues in web

### LINKED ISSUE #5045: Setting base path to "/" will break asset macro path rendering
**Problem**
When the base path is set to `/` the asset macro will in turn render the full path as `//assets/asset.css` prepending an extra slash to the full path. This will cause the browser to treat the source as a protocol agnostic path and thus treat `assets` as a domain. This will not resolve and thus breaks asset loading.

**Steps To Reproduce**

- Add the line `base_path = "/"` to your `Dioxus.toml`
- Load an asset such as `document::Stylesheet { href: asset!("/assets/tailwind.css")}`
- Using `dx serve` the application will compile and load
- Inspect the HTML to check the source for your asset and you'll notice the path similar to `<link rel="stylesheet" href="//assets/tailwind-dxh1dfdfee79cb7291d.css" type="text/css">`
- The expected result is `<link rel="stylesheet" href="/assets/tailwind-dxh1dfdfee79cb7291d.css" type="text/css">`

**Expected behavior**

We expect the base path does not double slash when set to `/` and renders the asset path starting from the root `/assets/` and not `//assets/`.

**Screenshots**

We discussed this issue in the #help-chat discord.
https://discord.com/channels/899851952891002890/1444957504747602064

**Environment:**

- Dioxus version: v0.7.1
- Rust version: 1.19.1
- OS info: debian-slim
- App platform: web

**Questionnaire**

- I'm interested in fixing this myself but don't know where to start but can likely figure it out. :)

binarypie:
Normalize the base path when set to `/` to prevent routing and asset resolution errors in web. This resolves #5045.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -2801,11 +2801,8 @@ impl BuildRequest {
         // If this is a release build, bake the base path and title into the binary with env vars.
         // todo: should we even be doing this? might be better being a build.rs or something else.
         if self.release {
-            if let Some(base_path) = self.base_path() {
-                let trimmed = base_path.trim_matches('/');
-                if !trimmed.is_empty() {
-                    env_vars.push((ASSET_ROOT_ENV.into(), trimmed.to_string().into()));
-                }
+            if let Some(base_path) = self.trimmed_base_path() {
+                env_vars.push((ASSET_ROOT_ENV.into(), base_path.to_string().into()));
             }
             env_vars.push((
                 APP_TITLE_ENV.into(),
@@ -4901,11 +4898,8 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
 
         // Add the base path to the head if this is a debug build
         if self.is_dev_build() {
-            if let Some(base_path) = &self.base_path() {
-                let trimmed = base_path.trim_matches('/');
-                if !trimmed.is_empty() {
-                    head_resources.push_str(&format_base_path_meta_element(trimmed));
-                }
+            if let Some(base_path) = &self.trimmed_base_path() {
+                head_resources.push_str(&format_base_path_meta_element(base_path));
             }
         }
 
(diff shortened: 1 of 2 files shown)


binarypie:
> Thanks! Could you move this normalization to when `DIOXUS_ASSET_ROOT` is set instead of when it is read?
> 
> Then we don't need to rerun it every time we read the base path. It could also be in the head instead of the env var depending on the platform.

Sure can!

binarypie:
This approach covers all three ways (cli, toml, env var) the base_path can be set.

In fact it is likely true that this code in request.rs could be removed. However, I was unsure of the ramifications of doing so.
```
/// Get the normalized base path for the application with `/` trimmed from both ends. If the base path is not set, this will return `.`.
    pub(crate) fn base_path_or_default(&self) -> &str {
        let trimmed_path = self.base_path().unwrap_or_default().trim_matches('/');
        if trimmed_path.is_empty() {
            "."
        } else {
            trimmed_path
        }
    }
```

Another point to note. Simply modifying where we internally set `DIOXUS_ASSET_ROOT` with the value from base_path is not enough because it will break when/if someone overrides the environment variable but does not include a cli arg nor a declaration in the toml config.

ealmloff:
> Another point to note. Simply modifying where we internally set `DIOXUS_ASSET_ROOT` with the value from base_path is not enough because it will break when/if someone overrides the environment variable but does not include a cli arg nor a declaration in the toml config.

The CLI always either writes `DIOXUS_ASSET_ROOT` or the meta element. The way this path is communicated to the front end is an implementation detail. We don't support overwriting that environment variable

binarypie:
> > Another point to note. Simply modifying where we internally set `DIOXUS_ASSET_ROOT` with the value from base_path is not enough because it will break when/if someone overrides the environment variable but does not include a cli arg nor a declaration in the toml config.
> 
> The CLI always either writes `DIOXUS_ASSET_ROOT` or the meta element. The way this path is communicated to the front end is an implementation detail. We don't support overwriting that environment variable

I apologize for the confusion here. I assumed that the environment variable was accessible from the comments in the code. I'll revert these and make that change.

packages/cli/src/build/request.rs
```diff
@@ -4899,7 +4902,10 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
         // Add the base path to the head if this is a debug build
         if self.is_dev_build() {
             if let Some(base_path) = &self.base_path() {
-                head_resources.push_str(&format_base_path_meta_element(base_path));
+                let trimmed = base_path.trim_matches('/');
```
binarypie: This is specifically for the dev server which calls base_path before the environment variable is set.

ealmloff (approved):
Much better, thanks!

Issue #5065, opened 2025-12-06
Build failing with `strip = "debuginfo"` in cargo config


paul-hansen:
**Problem**

When I run dx serve when `strip = "debuginfo"` is set, I get a build error:
```
dx serve
12:06:36 [dev] -----------------------------------------------------------------
                Serving your app: test-dx! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
                Learn more at https://dioxuslabs.com/learn/0.7/getting_started
               ----------------------------------------------------------------
12:06:36 [dev] Build failed: Failed to strip binary
12:06:36 [dev] Build failed: Failed to strip binary
```

Running `dx serve --release` runs and I can open it in the browser, but it gives a different error in the terminal:
```
dx serve --release
12:12:32 [dev] -----------------------------------------------------------------
                Serving your app: test-dx! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
                Learn more at https://dioxuslabs.com/learn/0.7/getting_started
               ----------------------------------------------------------------
12:12:36 [dev] wasm-opt failed with status code signal: 6 (SIGABRT) (core dumped)
stderr:
stdout: compile unit size was incorrect (this may be an unsupported version of DWARF)

12:12:36 [dev] Build completed successfully in 446ms, launching app! 💫
12:12:36 [server]  INFO Registering server function: POST /api/echo
12:12:43 [server] [200] /
```
**Steps To Reproduce**

- Create a new project using `dx new` with fullstack, router, and tailwind enabled and default to web.
- Add this to your `~/.cargo/config.toml` file:
   ```toml
   [profile.dev]
   strip = "debuginfo"
   ```
- Run `dx serve`
- Observe error

**Expected behavior**

Build without error.

**Environment:**

- Dioxus version:  0.7.2
- Rust version:    v1.91.1 stable
- OS info: EndeavorOS (Arch based rolling distro)
- App platform: Web

paul-hansen:
Seems to be something in my `~/.cargo/config.toml`, commenting everything out in that makes it build fine. Will try to narrow it down and share which setting it was.

paul-hansen:
[...]

PR #5066, opened 2025-12-06
Fix call to rustc-objcopy to strip binary on Linux

nicoburns:
- Fixes https://github.com/DioxusLabs/dioxus/issues/5065

Error was:

> /home/nico/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/lib/rustlib/aarch64-unknown-linux-gnu/bin/rust-objcopy: error while loading shared libraries: libLLVM.so.20.1-rust-1.90.0-stable: cannot open shared object file: No such file or directory

Fix is to set the `LD_LIBRARY_PATH` env variable to the directory that contains that file when executing the command.

---

It would be possible to append to LD_LIBRARY_PATH (like with PATH). However, I think there's no need in this case as we're calling into a specific binary, not setting a general env var.

---

It would additionally be nice to print stdout/stderr when the command fails so that we get a debuggable error message. But I'm not sure what the idiomatic way to log in the CLI is.

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -3985,6 +3985,12 @@ impl BuildRequest {
             command.arg(strip_arg).arg(exe).arg(exe);
             let output = command.output().await?;
             if !output.status.success() {
+                if let Ok(stdout) = std::str::from_utf8(&output.stdout) {
+                    tracing::error!("{}", stdout);
+                }
+                if let Ok(stderr) = std::str::from_utf8(&output.stderr) {
+                    tracing::error!("{}", stderr);
+                }
                 return Err(anyhow::anyhow!("Failed to strip binary"));
             }
         }


paul-hansen:
I confirmed that this fixes the build error in https://github.com/DioxusLabs/dioxus/issues/5065 by installing the dioxus-cli from this MR like so:
```
cargo install dioxus-cli --git https://github.com/nicoburns/dioxus.git --rev=bd8ad33
```

Wasm-opt still fails in release builds but I'm now thinking that's a separate issue.

ealmloff:
> It would additionally be nice to print stdout/stderr when the command fails so that we get a debuggable error message. But I'm not sure what the idiomatic way to log in the CLI is.

Tracing is hooked up to the CLI so you should be able to do `tracing::error!` if we can't run the command

PR #5068, opened 2025-12-07
Update README links for version 0.7

jullanggit:
This seems to be out of sync right now, as the link itself points to 0.7, but the name still says 0.6


cramhead:
Thanks @nicoburns

PR #5076, opened 2025-12-09
SEO Optimization: Crawler and Indexes Enhancements

wheregmis:
Introduces a handler for /robots.txt requests to ensure the file is always served as plain text. If robots.txt exists in the public directory, it is served; otherwise, a default allowing all crawlers is returned. The handler is registered before static assets to guarantee correct content type and priority.

image


wheregmis:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/DioxusLabs/dioxus/pull/5076#pullrequestreview-3555027727)

wheregmis:
I think i. can handle this on client end

PR #5088, opened 2025-12-12
Add `read_async` to Resource

mcmah309:
There currently is no ergonomic way to `await` `Resource`s in an async closure. This PR adds the `read_async` method to `Resource`


examples/05-using-async/chaining_resource.rs
```diff
@@ -0,0 +1,144 @@
...
+        let count_write = request_count.write();
+
+        // ⚠️ WARNING: Normally, holding guards across await points is dangerous!
+        // `read_async()` solves this by:
+        // 1. Forcing you to acknowledge you have some or no guards and accepting them as parameters
+        // 2. Dropping them if the resource is currently unavailable
+        // 3. Returning them back with the resource if the resource is available
+
+        // Wait for the first resource and get guards back
+        let (result_ref, mut image_url_write, mut breed_info_write, mut count_write) = dog_image
```
ealmloff: One use case I see for this API is chaining a resource read with other async work. This example shows chaning a resource read with sync work which could have just been inlined into the original resource.

If there were async work based on the result of the resource read here (`result_ref`), it seems natural to use the reference directly but that would result in holding the lock over an await point which causes errors

packages/hooks/src/use_resource.rs
```diff
@@ -442,6 +442,165 @@ impl<T> Resource<T> {
...
+    where
+        M: MaybeDrop<generational_box::GenerationalRef<std::cell::Ref<'a, T>>>,
+    {
+        let read: generational_box::GenerationalRef<std::cell::Ref<'a, Option<T>>> = self.read();
+        if read.is_none() {
+            drop(read);
+            drop(maybe_drop);
+            let _: () = (*self).await;
+            // `.read()` should have panicked if not in the correct scope as well
+            unreachable!("Future should cancel when ready");
```
ealmloff: This API could be called inside of an event handler or spawn which won't reset the async block after the resource changes. How does this API behave in that use case:
```rust
button {
    onclick: |_| async move { println!("resolved resource value is {:?}", resource.read_async(()) }
}
```

examples/05-using-async/chaining_resource.rs
mcmah309: > One use case I see for this API is chaining a resource read with other async work. This example shows chaning a resource read with sync work which could have just been inlined into the original resource.

Admittedly, this example use case is a little weak, it shows how to use the api, but the scenario does not really need it. As you said, could have easily avoided using two resources here. I’ll improve the example

> If there were async work based on the result of the resource read here (`result_ref`), it seems natural to use the reference directly but that would result in holding the lock over an await point which causes errors

Yes, but that is also true with any signal - a sync signal with async work. As with all signals, the ref should be dropped before and retrieved again after the async work, or cloned then dropped before the async work.

packages/hooks/src/use_resource.rs
mcmah309: Just tested it quickly, If the resource is not ready, it hangs. I’ll look into a solution soon, but I can’t work on it today. Do you have any ideas how to solve this in the meantime?

examples/05-using-async/chaining_resource.rs
mcmah309: I improved the example to use 3 resources connected to each other and some more async work.

May be related to your last comment, but if we required `T: Clone` to use the api, a big problem I see is that most resources return a `Result<T, E>` and most `E` is not `Clone`. Like `anyhow::Error`, so this would make chaining `Resource` for these types more difficult. Especially when a user may not need to clone at all, but we force that decision on them.

packages/hooks/src/use_resource.rs
mcmah309: Looks like there was an issue with the original `Future` implementation for resource. No futures ever resolved if not already complete when first polled. I fixed this and added a way for `Resource` futures to work here. I removed the accepting other guards to drop in the api too because given these futures won't be canceled, the behavior won't work.

ealmloff: I think we are just missing a call to `use_waker().wake()` when the value is resolved? I would prefer to keep using the waker hook so we can maintain one version of that pattern.

examples/05-using-async/chaining_resource.rs
ealmloff: Just adding a `cloned_async` function in addition to `read_async` would be fine. I can see some use cases for `read_async` and clippy should warn against misuse if it is configured correctly, but I expect most of the time you will need to clone the result

packages/hooks/src/use_resource.rs
mcmah309: We do call `.wake()` https://github.com/DioxusLabs/dioxus/pull/5088/files#diff-cffff841d5cfd91aa084e148058c59051af7956b3269eb35ac0db3811ac63c75L67 . I took a look and re-verified the issue exists. 

The issue is `self.waker.clone().poll_unpin(cx)` will always return `Pending` (If the resource is already resolved we don't `await`/`poll` so this is never hit and that is why this case does not hang). 

When `.wake()` is called it wakes up all the futures. The futures then re-poll and since `.wake()` replaces the internal waker channel, the new channel returns `Pending` again. Thus, I believe it was wrong for `UseWaker` to implement `Future` in its current form in the first place. I removed the implementation. It was only used in the original `Future` implementation for `Resource` so removing now has no effect.

examples/05-using-async/chaining_resource.rs
mcmah309: Okay :+1:  I added `cloned_async` and went ahead and added `peek_async` as well.

packages/hooks/src/use_resource.rs
mcmah309: I had some time today so I optimized the `HashMap` away with a versioned slab. It should surely be more efficient than the original `UseWaker`/`SharedFuture` `Future` implementation now too, since that used a lot of clones, channels, Mutexes, Arcs, and even Slab internally.

packages/hooks/src/use_waker.rs
```diff
@@ -50,20 +50,6 @@ impl<T: Clone + 'static> UseWaker<T> {
     }
 }
 
-// Can await the waker to be woken.
-// We use `.peek()` here to avoid reacting to changes in the underlying task_rx which could lead
-// to an effect/future loop.
-impl<T: Clone + 'static> std::future::Future for UseWaker<T> {
```
ealmloff, line 56: This makes the change breaking. Could you restore this and mark it as deprecated so we can remove it in the next major release? We may need to remove IntoFuture

ealmloff (changes requested):

packages/hooks/src/use_resource.rs
ealmloff: It looks like the issue was cloning the shared waker every time we polled. https://github.com/DioxusLabs/dioxus/pull/5088/commits/6a5eeea54e8269e924b7e991c29b234452559a48 seems to work with `UseWaker`. This isn't performance critical code and I would rather we keep the implementation centralized and avoid unsafe if we can

packages/hooks/src/use_waker.rs
mcmah309: Unless I am missing something, if we restore this, then we have to remove `IntoFuture`. If we remove `IntoFuture` then the bug in the future implementation for `UseWaker` remains. If this is considered a breaking change, then removing `Future` directly on `Resource` should also be considered a breaking change?

It may come down to this being an acceptable "breaking change". The only code this would break is if someone did something like `let x: UseWaker = use_waker.into_future();` (now needs to be `let x: UseWakerFuture = use_waker.into_future();`), which I imagine never happens in practice. Code like `use_waker.await` or `resource.await` is unaffected. I know libraries like tokio have done these type of breaking changes in the past, I think this may be one of those cases we don't have to be religious about semver.

mcmah309:
@ealmloff Can we merge this now that the next major release is being worked on?

PR #5095, opened 2025-12-14
Add auxclick and scrollend to dioxus events

rhaskia:


nicoburns (approved):
This looks good to me, but one of @ealmloff and @jkelleyrtp should probably review as they are more familiar with how this code affects the web and desktop renderers.

jkelleyrtp (approved):

PR #5107, opened 2025-12-16
Add support for HTTP redirect from extractor

dprkh:
This is a breaking change, because it adds a new variant to the `ServerFnError` enum.

I used AI to generate this to unblock myself, so feel free to reject, but if it is within your vision to have this feature, I am open on working on this further.

Thanks.


dprkh:
Here is what my code looks like with this feature, so that you know the motivation behind this.

```rust
use crate::{extensions, types::user_session};

use dioxus::{
    fullstack::{Cookie, HeaderMap, HeaderValue, extract::FromRequestParts, headers::HeaderMapExt},
    prelude::StatusCode,
    server::http::{header::LOCATION, request::Parts},
};

const SESSION_COOKIE_NAME: &'static str = "session";

impl<S> FromRequestParts<S> for user_session::Id
where
    S: Send + Sync,
{
    type Rejection = (StatusCode, HeaderMap);

    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
        let Some(auth) = parts.extensions.get::<extensions::Auth>() else {
            tracing::error!("extension `auth` is not initialized");

            return Err((StatusCode::INTERNAL_SERVER_ERROR, HeaderMap::default()));
        };

        let redirect = |to: &'static str| {
            return (StatusCode::TEMPORARY_REDIRECT, HeaderMap::from_iter([(LOCATION, HeaderValue::from_static(to))].into_iter()));
        };

        let Some(cookie) = parts.headers.typed_get::<Cookie>() else {
            return Err(redirect("/sign-up"));
        };

        let Some(session_str) = cookie.get(SESSION_COOKIE_NAME) else {
            return Err(redirect("/sign-up"));
        };

        let Some((data_str, tag_str)) = session_str.split_once('.') else {
            return Err(redirect("/sign-in"));
        };

        if !auth.verify(data_str.as_bytes(), tag_str.as_bytes()) {
            return Err(redirect("/sign-in"));
        }

        let Ok(session_id) = user_session::Id::try_from(data_str.to_string()) else {
            return Err(redirect("/sign-in"));
        };

        Ok(session_id)
    }
}
```

Without this change, dioxus would treat this as an error, and redirect would not work.

jkelleyrtp:
We're not currently merging breaking PRs right now; however, if you wanted to get it into current stable, we could just reuse the ServerError type using the redirect status code, and then attach the required details. The two variants seem similar in structure that this might be able to work.

dprkh:
Hi @jkelleyrtp,

Thank you for taking a look at my PR.

I made changes accordingly.

Let me know if there's anything else.

PR #5113, opened 2025-12-18
feat: add onunmounted lifecycle event to close #5108

### LINKED ISSUE #5108: Element Lifecycle Event: `onunmounted ` ━ **ESSENTIAL**
Element Lifecycle Event `onunmounted`

## The Problem

An animation engine _(and other types of visualization controllers)_ need to track element lifecycles to manage state. Dioxus provides `onmounted` to notify when an element enters the DOM, but there's no corresponding notification for when an element leaves, nor any viable workarounds.

Without exit notifications the engine cannot clean up state associated with removed elements. This is a fundamental gap in element lifecycle management in Dioxus as of `0.7.2`.

## Proposed Solution

Add `onunmounted` to complement the existing `onmounted`:

| Stage         | Event         | Status      | When it Fires                     |
| ------------- | ------------- | ----------- | --------------------------------- |
| After mount   | `onmounted`   | Exists      | Element is in DOM                 |
| After unmount | `onunmounted` | **Missing** | Element has been removed from DOM |

```
[INSERT INTO DOM] ──► onmounted
                          │
                          ▼
                    (element live)
                          │
                          ▼
                   [REMOVE FROM DOM] ──► onunmounted
```

### `onmounted` (after mount) - Already Exists

Element is in the DOM and can be measured/manipulated. Animation engines use this to register the element and begin tracking its state.

_This event exists and works. The problem is the missing unmount event._

### `onunmounted` (after unmount)

**When it fires**: Element has been removed from the DOM.

**Use case**: Clean up tracking state.

This is when the animation engine deregisters the element and frees associated resources. Without this notification:

| Problem                    | Consequence                                                                                  |
| -------------------------- | -------------------------------------------------------------------------------------------- |
| **Unbounded state growth** | HashMap of tracked elements grows forever. 10,000 list operations = 10,000 orphaned entries. |
| **Wasted computation**     | Animation loop iterates over stale entries every frame.                                      |
| **Stale references**       | Calling methods on removed elements.                                                         |

```rust
// Proper lifecycle management
div {
    key: "{item.id}",
    onmounted: move |e| {
        animation_engine.register(item.id, e.data());
    },
    onunmounted: move |_| {
        // Element is gone - clean up tracking state
        animation_engine.deregister(item.id);
    },
}
```

## Why `use_drop` is Insufficient

`use_drop` operates at the **component scope level**, not the **element level**. If you have multiple elements in a single component, `use_drop` only fires when the entire component unmounts, not when individual elements are removed.

```rust
fn MyComponent(show_card: bool) -> Element {
    use_drop(|| {
        // This fires when MyComponent unmounts,
        // NOT when the card is conditionally removed
    });

    rsx! {
        div {
            if show_card {
                div {
                    class: "card",
                    onmounted: |e| animation_engine.track(e.data()),
                    // No way to cleanup when this specific div is removed!
                    // Animation engine now has orphaned state
                    "Animated Card"
                }
            }
        }
    }
}
```

**The problem**: When `show_card` becomes `false`, the card div is removed but `use_drop` doesn't fire (the component is still mounted). The animation engine still thinks the element exists.

## Current Workaround is Impractical

Wrap every element that needs lifecycle detection in its own component:

```rust
#[component]
fn TrackedDiv(
    on_mount: EventHandler<Rc<MountedData>>,
    on_unmount: EventHandler<()>,
    children: Element,
) -> Element {
    use_drop(move || {
        on_unmount.call(());
    });

    rsx! {
        div {
            onmounted: move |e| on_mount.call(e.data()),
            {children}
        }
    }
}
```

### Problems with this workaround

1. **Poor DX**: Need wrapper components for every element type (`TrackedDiv`, `TrackedSpan`, etc.)
2. **Breaks composition**: Users can't use standard `div`/`span` with animation systems
3. **No DOM access on unmount**: `use_drop` fires during scope teardown - element may already be gone
4. **Not scalable**: Animation libraries can't provide wrappers for all 100+ HTML elements

## Other Frameworks

All other frameworks provide a an event / solution for unmounted elements.

rapporian:
## Summary

Dioxus provides `onmounted` to notify when elements enter the DOM, but no corresponding notification for when an element leaves.

The existing `use_drop` hook doesn't solve this as it operates at component scope, not element level. When a single element conditionally disappears while its parent component remains mounted, `use_drop` never fires. This leaves orphaned entries in animation engines, visualization controllers, and any system tracking element state. Please see Issue #5108 for more detail.

This PR adds `onunmounted` to fire when an element is removed from the DOM, enabling proper cleanup and preventing unbounded state growth.
 
Closes #5108

## Usage

```rust
div {
    onmounted: move |e| engine.register(id, e.data()),
    onunmounted: move |_| engine.deregister(id),
}
```

## Changes

|File|Change|
|---|---|
|`core-types/src/bubbles.rs`|Register `unmounted` as non-bubbling event|
|`html/src/events/mounted.rs`|Add `UnmountedData` struct and `onunmounted` event|
|`html/src/events/mod.rs`|Add `From<&PlatformEventData>` for `UnmountedData`|
|`web/src/dom.rs`|Track elements with unmount listeners|
|`web/src/mutations.rs`|Fire event on `remove_node` / `replace_node_with`; fix hydration bug|
|`playwright-tests/fullstack-unmounted/`|New test verifying the event fires correctly|

## Platform Support

- **Web**: ✅ Implemented

## Test Plan

- [x]  `cargo check -p dioxus-web --features hydrate,mounted`
- [x]  Playwright tests pass (element removal + toggle cycles after hydration)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6215,6 +6215,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "dioxus-playwright-fullstack-unmounted-test"
+version = "0.1.0"
+dependencies = [
+ "dioxus",
+ "serde",
+ "tokio",
+ "web-sys",
+]
+
+[[package]]
 name = "dioxus-playwright-liveview-test"
 version = "0.0.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,6 +130,7 @@ members = [
     "packages/playwright-tests/fullstack",
     "packages/playwright-tests/fullstack-errors",
     "packages/playwright-tests/fullstack-mounted",
+    "packages/playwright-tests/fullstack-unmounted",
     "packages/playwright-tests/fullstack-spread",
     "packages/playwright-tests/fullstack-routing",
     "packages/playwright-tests/fullstack-hydration-order",
--- a/packages/core-types/src/bubbles.rs
+++ b/packages/core-types/src/bubbles.rs
@@ -99,6 +99,7 @@ pub fn event_bubbles(evt: &str) -> bool {
         "toggle" => false,
         "beforetoggle" => false,
         "mounted" => false,
+        "unmounted" => false,
         "visible" => false,
         _ => true,
     }
--- a/packages/html/src/events/mod.rs
+++ b/packages/html/src/events/mod.rs
@@ -223,6 +223,13 @@ impl From<&PlatformEventData> for MountedData {
     }
 }
 
+impl From<&PlatformEventData> for UnmountedData {
+    fn from(_val: &PlatformEventData) -> Self {
+        // UnmountedData contains no element information since the element is already gone
+        UnmountedData::new()
+    }
+}
+
 impl From<&PlatformEventData> for MouseData {
     fn from(val: &PlatformEventData) -> Self {
         with_event_converter(|c| c.convert_mouse_data(val))
--- a/packages/html/src/events/mounted.rs
+++ b/packages/html/src/events/mounted.rs
@@ -280,6 +280,72 @@ impl_event! [
 
 pub use onmounted as onmount;
 
+/// Data associated with an unmounted event.
+///
+/// This event fires after an element has been removed from the DOM.
+/// Since the element no longer exists, no element data is available.
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnmountedData {}
+
+impl UnmountedData {
+    /// Create a new UnmountedData
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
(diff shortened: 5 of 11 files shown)


rapporian:
### Documentation has been pre-written:

https://github.com/DioxusLabs/docsite/pull/612

rapporian:
Closing to revised strategy per conversation w/ @ealmloff:
https://github.com/DioxusLabs/dioxus/issues/5108#issuecomment-3672558945

Return cleanup in `unmounted`:

```rust
div {
    onmounted: move |e| {
        start_animation(e.data());
        move || stop_animation(e.data())
    },
}
```

PR #5132, opened 2025-12-23
fix dx fmt span attribute issue

jkelleyrtp:
Fixes autoformatting for this test case:

```
//! A comment about this file

use dioxus::prelude::*;

#[component]
pub fn Sample() -> Element {
    let is_active = use_signal(|| false);

    rsx! {
        div { class: if is_active() { "active" } else { "inactive" },
            div { class: if is_active2() { "a" } else { "b" } }
        }
    }
}
```

We were accidentally dropping span locations for if-chain attributes, causing the spans location to default to 0. Now, we we properly quote-span these attributes. I double checked that the other attributes are properly quote spanned.

Also, as a guard rail for future whitespace writing, if the span of the token is at byte 0 in the file, we don't write its whitespace. This means if we have issues in the future, whitespace might be deleted, but the file won't be corrupted.

### DIFF
--- a/packages/autofmt/src/writer.rs
+++ b/packages/autofmt/src/writer.rs
@@ -487,7 +487,7 @@ impl<'a> Writer<'a> {
                     .comma
                     .as_ref()
                     .map(|c| c.span())
-                    .unwrap_or_else(|| self.totoal_span_of_attr(attr)),
+                    .unwrap_or_else(|| self.total_span_of_attr(attr)),
                 AttrType::Spread(attr) => attr.span(),
             };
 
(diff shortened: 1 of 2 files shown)


packages/autofmt/src/writer.rs
```diff
@@ -1069,17 +1080,13 @@ impl<'a> Writer<'a> {
         }
     }
 
-    fn final_span_of_attr(&self, attr: &Attribute) -> Span {
+    fn totoal_span_of_attr(&self, attr: &Attribute) -> Span {
```
nicoburns: Typo here

PR #5147, opened 2025-12-28
CLI: don't enable default features in `--no-default-features` builds

nicoburns:
- Depends on https://github.com/DioxusLabs/dioxus/pull/5143. Seperate PRs because I'm less sure about this fix.

The CLI currently enables "renderless" default features (default features which don't enable a renderer) even when `--no-default-features` is specified as a command line argument. It also unconditionally set the variable tracking the `--no-default-features` flag to true unless the renderer is `Unknown`.

This PR removes the code modifying the `--no-default-features` flag, and PR inverts the condition so that default features are enabled in the `--no-default-features` flag is *not* set.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9018,6 +9018,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "harness-web-with-default-features"
+version = "0.0.1"
+dependencies = [
+ "dioxus",
+]
+
+[[package]]
+name = "harness-web-with-no-default-features"
+version = "0.0.1"
+dependencies = [
+ "dioxus",
+]
+
+[[package]]
 name = "hash32"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/packages/cli-harnesses/harness-web-with-default-features/Cargo.toml
+++ b/packages/cli-harnesses/harness-web-with-default-features/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "harness-web-with-default-features"
+version = "0.0.1"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+publish = false
+
+[dependencies]
+dioxus = { workspace = true }
+
+
+[features]
+default=["other"]
+other=[]
+
+    
--- a/packages/cli-harnesses/harness-web-with-default-features/src/main.rs
+++ b/packages/cli-harnesses/harness-web-with-default-features/src/main.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("Hello, world!");
+}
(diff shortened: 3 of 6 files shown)


nicoburns:
@jkelleyrtp This is failing CI on the following test:

```rust
        TestHarnessBuilder::new("harness-default-to-non-default")
            .deps(r#"dioxus = { workspace = true, features = [] }"#)
            .fetr(r#"default=["web"]"#)
            .fetr(r#"web=["dioxus/web"]"#)
            .asrt(
                r#"dx build --ios"#,
                |targets| async move {
                    let t = targets.unwrap();
                    assert!(t.server.is_none());
                    assert_eq!(t.client.bundle, BundleFormat::Ios);
                    assert_eq!(t.client.triple, TestHarnessBuilder::host_ios_triple_sim());
                    assert!(t.client.no_default_features);
                },
            ),
```

In my mind, the test is wrong because `dx build --ios` shouldn't imply `no_default_features`. That should only be configured if you run something like `dx build --ios --no-default-features`. However, I'm wondering if the `no_default_features` setting has some other meaning within the Dioxus CLI that I'm missing?

jkelleyrtp:
> @jkelleyrtp This is failing CI on the following test:
> 
> ```rust
>         TestHarnessBuilder::new("harness-default-to-non-default")
>             .deps(r#"dioxus = { workspace = true, features = [] }"#)
>             .fetr(r#"default=["web"]"#)
>             .fetr(r#"web=["dioxus/web"]"#)
>             .asrt(
>                 r#"dx build --ios"#,
>                 |targets| async move {
>                     let t = targets.unwrap();
>                     assert!(t.server.is_none());
>                     assert_eq!(t.client.bundle, BundleFormat::Ios);
>                     assert_eq!(t.client.triple, TestHarnessBuilder::host_ios_triple_sim());
>                     assert!(t.client.no_default_features);
>                 },
>             ),
> ```
> 
> In my mind, the test is wrong because `dx build --ios` shouldn't imply `no_default_features`. That should only be configured if you run something like `dx build --ios --no-default-features`. However, I'm wondering if the `no_default_features` setting has some other meaning within the Dioxus CLI that I'm missing?

The reason we specify no_default_features here is because `--ios` is overriding the default `web` platform, so we need to strip it down and then add back everything except for the web feature.

Passing `--ios` implies `--features mobile` if it exists, but we need to disable `web` to get the feature set we want.

nicoburns:
> The reason we specify no_default_features here is because --ios is overriding the default web platform, so we need to strip it down and then add back everything except for the web feature.

I believe in the current version of the code this happens even without the `no_default_features` flag being set. Scroll down to where the `no_default_features` variable is read, and you'll see that the "feature set" it controls is the "renderless" set of features (which I believe excludes any feature which enables a renderer)..

nicoburns:
@jkelleyrtp This is now passing CI, and the following commands in the Dioxus repo all work:

- `dx serve --example todomvc` (uses default `desktop` renderer)
- `dx serve --example todomvc --web` (uses web renderer, doesn't enable `desktop` feature)
- `dx serve --example todomvc --renderer native` (uses native renderer, doesn't enable `desktop` feature)

PR #5162, opened 2025-12-29
Enable GameActivity

nicoburns:
We probably can't merge this as-is. But putting this up to track version of CLI that can compile Blitz with GameActivity.

### DIFF
--- a/packages/cli/assets/android/gen/app/build.gradle.kts.hbs
+++ b/packages/cli/assets/android/gen/app/build.gradle.kts.hbs
@@ -59,4 +59,5 @@ dependencies {
     implementation("androidx.webkit:webkit:1.6.1")
     implementation("androidx.appcompat:appcompat:1.6.1")
     implementation("com.google.android.material:material:1.8.0")
+    implementation("androidx.games:games-activity:2.0.2")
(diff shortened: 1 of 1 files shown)


chtpl:
can you explain the context a bit more please? what is this GameActivity necessary for?

nicoburns:
Along with `NativeActivity`, it's one of two [Android Activities](https://developer.android.com/guide/components/activities/intro-activities) that Google provides for native code (i.e. C/C++/Rust) Android apps to use: https://developer.android.com/games/agdk/game-activity. Compared to `NativeActivity` it provides more advanced functionality.

For Dioxus Native on Android we need to use either `NativeActivity`, `GameActivity`, or build our own similar to https://github.com/rust-mobile/android-view. We probably want to do the latter eventually, but getting that right is tricky. In the short-term, `GameActivity` provides more functionality than `NativeActivity`.

chtpl:
I'm not familiar with android development - what is the purpose of these `Activity` things?

nicoburns:
They are the entry point for Android Apps. The OS calls methods on them at various points in the apps lfiecycle.

Issue #5079, opened 2025-12-10
Some linker args (atomics & bulk-memory) breaks assets


wheregmis:
For any app, just following the instructions for the following crate and modifying the .cargo/config.toml as below will break the assets
https://github.com/RReverser/wasm-bindgen-rayon

```rust
[target.wasm32-unknown-unknown]
rustflags = [
    "-C",
    "target-feature=+atomics,+bulk-memory",
    "-C",
    "link-arg=--shared-memory",
    "-C",
    "link-arg=--max-memory=1073741824",
    "-C",
    "link-arg=--import-memory",
    "-C",
    "link-arg=--export=__wasm_init_tls",
    "-C",
    "link-arg=--export=__tls_size",
    "-C",
    "link-arg=--export=__tls_align",
    "-C",
    "link-arg=--export=__tls_base",
]

[unstable]
build-std = ["panic_abort", "std"]

```

repro: https://github.com/wheregmis/rustflags_dioxus

wheregmis:
I will try to give an attempt for the fix

wheregmis:
couldnt fix it sadly waiting for some pros

jkelleyrtp:
I don't think that setup will work with hotpatching enabled, but from your issue this seems to just break asset collection in regular mode? Not sure if we can get fixed memory setups working with hotpatching.

wheregmis:
> I don't think that setup will work with hotpatching enabled, but from your issue this seems to just break asset collection in regular mode? Not sure if we can get fixed memory setups working with hotpatching.

Yes, in my current scenerio, its breaking assets collection in regular mode.  (PS: I wasnt running it with hotpatching, tried with dx bundle to see maybe for release build it works, but that wasnt a success either)

Robert-MacWha:
Seconding this issue.  I also explored using Trunk instead of dioxus-cli or downgrading dioxus to 0.6 in case it would help, but nada.  The memory alignment issues seem persistent.

In case this helps anyone else - I ultimately ended up creating a separate "coordinator" web worker which can have shared memory with the various "compute" workers, and then used worker messages (`postMessage` / `onmessage handler`) to communicate between the main dioxus worker and the "compute" worker.  It's hardly perfect but it works.

wheregmis:
> Seconding this issue. I also explored using Trunk instead of dioxus-cli or downgrading dioxus to 0.6 in case it would help, but nada. The memory alignment issues seem persistent.
> 
[...]

PR #5163, opened 2025-12-30
allow passive wasm segments for manganis

jkelleyrtp:
Fixes https://github.com/DioxusLabs/dioxus/issues/5079


wheregmis:
Just tested and It worked. Thank you. 🎉

PR #5169, opened 2026-01-01
Dioxus plugin use case

### LINKED ISSUE #5170: dioxus-desktop: add embedded-safe "returning" launcher (avoid terminating host process on Windows)
Fixed by https://github.com/DioxusLabs/dioxus/pull/5169

  - We build a Rust add-in/plugin that runs inside an existing Windows application's process.
  - We use dioxus-desktop to render UI in a child/embedded window.
  - The plugin can be enabled/disabled (loaded/unloaded) at runtime.

  Problem
  - When the plugin is disabled/unloaded, we shut down the dioxus-desktop event loop (e.g. close the window / post WM_QUIT).
  - On Windows, tao::EventLoop::run ultimately calls std::process::exit(...) when the event loop exits.
  - This terminates the entire host process, even though the UI/event loop is running on a worker thread.
  - Result: disabling/unloading the plugin unexpectedly closes the host application.

  Expected
  - Shutting down an embedded dioxus-desktop UI should stop only the embedded UI/event-loop thread.
  - It must not call std::process::exit or terminate the host process.

dxvid-pts:
Fixes: https://github.com/DioxusLabs/dioxus/issues/5170

  Proposed change (backwards compatible)
  1) dioxus-desktop: add new opt-in returning launch APIs that use tao::platform::run_return::EventLoopExtRunReturn::run_return where
  supported:
     - launch_virtual_dom_blocking_return(VirtualDom, Config) -> ()
     - launch_virtual_dom_return(VirtualDom, Config) -> ()
     - launch_return(root, contexts, platform_config) -> ()
     Keep existing launch/launch_virtual_dom* -> ! unchanged.

  2) dioxus: add LaunchBuilder::embedded() (desktop feature) as an ergonomic wrapper that switches the launcher to
  dioxus_desktop::launch::launch_return.
     This is optional sugar; embedders can also use LaunchBuilder::custom(dioxus_desktop::launch::launch_return).

### DIFF
--- a/packages/desktop/src/launch.rs
+++ b/packages/desktop/src/launch.rs
@@ -8,6 +8,169 @@ use dioxus_document::eval;
 use std::any::Any;
 use tao::event::{Event, StartCause, WindowEvent};
 
+/// Launch the WebView and run the event loop, returning to the caller when the event loop exits.
+///
+/// This is useful when embedding Dioxus Desktop inside another host process where calling
+/// `std::process::exit` would terminate the host.
+///
+/// On Windows, `tao::event_loop::EventLoop::run` will call `std::process::exit` when the event loop
+/// ends. This function uses `run_return` where supported so the event loop can shut down without
+/// exiting the process.
+pub fn launch_virtual_dom_blocking_return(virtual_dom: VirtualDom, mut desktop_config: Config) {
+    #[cfg(target_os = "ios")]
+    {
+        // `run_return` is not available on iOS in tao; fall back to the normal launcher.
+        launch_virtual_dom_blocking(virtual_dom, desktop_config);
+    }
+
+    #[cfg(not(target_os = "ios"))]
+    {
+    let mut custom_event_handler = desktop_config.custom_event_handler.take();
+    let (mut event_loop, mut app) = App::new(desktop_config, virtual_dom);
+
+        use tao::platform::run_return::EventLoopExtRunReturn;
+        event_loop.run_return(move |window_event, event_loop, control_flow| {
+            // Set the control flow and check if any events need to be handled in the app itself
+            app.tick(&window_event);
+
+            if let Some(ref mut f) = custom_event_handler {
+                f(&window_event, event_loop)
+            }
+
+            match window_event {
+                Event::NewEvents(StartCause::Init) => app.handle_start_cause_init(),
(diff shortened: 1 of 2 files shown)


dxvid-pts:
Example usage with this change

```rust
dioxus::LaunchBuilder::new()
   .embedded()
   .with_cfg(desktop! { cfg })
   .launch(app_fn);
```

jkelleyrtp:
Have been thinking about this and ideally we can make the default `launch` allow returning by default - with something like `app::exit()` that will allow returning back to the execution under main. Will think on this a bit more

dxvid-pts:
> Have been thinking about this and ideally we can make the default `launch` allow returning by default - with something like `app::exit()` that will allow returning back to the execution under main. Will think on this a bit more

I thought about this too. My original `embedded` naming was misleading because it sounds like embedded devices, while this is really about running Dioxus Desktop inside another host process.

My use case is a Windows plugin DLL: Dioxus is running inside a host application that owns the process lifetime. In that setup, Dioxus should be able to shut down its event loop without calling `std::process::exit`.

I think there are two pieces here:

1. Make desktop `dioxus::launch(app)` return by default when the Dioxus app exits:

```rust
fn main() {
    dioxus::launch(app);

    cleanup(); // runs after the Dioxus app exits
}
```

Second, add an app-scoped shutdown API for host/plugin scenarios. This is similar to your `app::exit()` idea, but uses a handle so the host can shut down the specific Dioxus app it started:

```rust
static SHUTDOWN_HANDLE: Mutex<Option<dioxus::desktop::ShutdownHandle>> = Mutex::new(None);

pub extern "system" fn plugin_start() {
    let app = dioxus::desktop::prepare(app);
    let shutdown = app.shutdown_handle();

    *SHUTDOWN_HANDLE.lock().unwrap() = Some(shutdown);

    std::thread::spawn(move || {
        app.launch();
    });
}

pub extern "system" fn plugin_shutdown() {
    if let Some(handle) = SHUTDOWN_HANDLE.lock().unwrap().take() {
        handle.exit();
    }
}
```

This avoids a process-global `app::exit()` ambiguity if multiple Dioxus apps exist in the same host process, while still allowing `launch` to return back to the caller once the event loop exits.

What is your opinion on this? If you like it I will adjust this PR.

PR #5179, opened 2026-01-04
Fix more examples to work with `dioxus-native`

nicoburns:
## Changes to non-example code 

- Re-export `dioxus-native` as `dioxus::native`
- Add `native!` macro to `dioxus-config-macro`
- Add `Config` struct/builder to `dioxus-native`

### DIFF
--- a/examples/01-app-demos/calculator_mutable.rs
+++ b/examples/01-app-demos/calculator_mutable.rs
@@ -6,22 +6,29 @@
 //! Generally, you'll want to split your state into several signals if you have a large application, but for small
 //! applications, or focused components, this is a great way to manage state.
 
-use dioxus::desktop::tao::dpi::LogicalSize;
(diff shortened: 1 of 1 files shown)


nicoburns:
> Getting the examples running on native is nice. I'm less sure about the config struct for native. It is nice as a centralized place to look for config items that native supports, but the version in this PR only supports window attributes, not the wgpu config that native can accept. Is there a reason those options are not included?

The reasoning for the `Config` struct was to mirror the API in Desktop with an eye on merging them in future. But we could definitely do that differently if we think there's a better way. The reason it only supports one thing in this PR is that that was what was needed to get the example running, and I figured the build pattern would mean that further things could be added backwards compatibly in future.

ealmloff (approved):
Fair enough, might be worth revisiting when we switch to winit on desktop which will break most of the config apis regardless

Issue #4142, opened 2025-05-19
Access to the event target property


filips123:
## Feature Request

There is currently no way to access the [`event.target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) property (like the clicked element for `onclick` events) from the Rust event handlers. This makes it hard or impossible to do dynamic events based on the clicked elements or events that affect the clicked elements).

This was already requested in #2590, but that issue was closed without a proper solution, as `web_event` does not work for native and mobile targets.

## Implement Suggestion

There should be a way to access (and potentially modify) the event target element of the event.

PR #5194, opened 2026-01-09
Get wasm bindgen working on desktop

ealmloff:
This PR adds support for a fork of wasm bindgen, web sys and js sys. It moves the virtual dom to a separate thread from the event loop. This lets us share a lot more code between desktop and web including the logic for:
- events
- mounted data (web_sys::Element)
- eval

## Breaking changes

- Many of the web-sys types we export are now `web-sys-x` types which use `wasm-bindgen-x` for desktop compatibility
- Some hooks like the wry event handler hooks now require a `Send` bound. This breaks a couple of examples

## Status of wasm-bindgen-x

Where we still aren't compatible with wasm-bindgen:
- Low level wasm access like `memory()`
- Unstable apis in wasm bindgen like `wasm_bindgen::convert`

Publishing:
An initial version is published [here](https://docs.rs/wasm-bindgen-x/latest/wasm_bindgen_x/) which works for dioxus' use cases without a patch. For patching it may be easier to rely on the git tag instead as cargo doesn't let you patch to the same source without a registry alias.

The current published version of web-sys-x has only a limited number of features because of the crates.io limit of 300 features per crate. I have reached out to the crates.io team to see if we can remove that limit to publish the full version

## TODO
- [x] Move the virtual doms to a new thread
- [x] Get wasm bindgen working on that new thread
- [x] Fix event conversion for desktop-specific events
- [x] Get it working with multiple webviews
- [x] Create shims for all of the main thread-only functionality we previously exposed
- [x] Integrate into the CLI to either auto patch wasm-bindgen or recommend patching wasm-bindgen when we detect it
- [x] Fix android
- [x] Documentation

Partially addresses https://github.com/DioxusLabs/dioxus/issues/4142 since `as_web_event` now works on desktop and mobile


packages/web-sys-events/src/events/form.rs
```diff
@@ -68,15 +72,29 @@ impl HasFormData for WebFormData {
     fn values(&self) -> Vec<(String, FormValue)> {
         let mut values = Vec::new();
 
+        // Try to find the form element - either the element itself is a form,
+        // or we need to find the closest ancestor form
+        let form_element = self
+            .element
+            .dyn_ref::<web_sys_x::HtmlFormElement>()
+            .cloned()
```
ealmloff, line 47: This was the existing behavior we tested for in dioxus desktop

ealmloff:
> No space left on device : '/home/runner/actions-runner/cached/2.331.0/_diag/pages/17dfb33f-0037-46e5-ae39-d9ea7ee4fdbf_d3e09c59-d36d-43fe-9d07-2d68e31abafe_1.log'

CI is running out of storage with the new crates this PR adds. Running locally on linux gnu, the same tests pass

jkelleyrtp:
can we move https://github.com/DioxusLabs/wasm-bindgen-wry 's fork of wasm-bindgen to dioxuslabs as owner? and make sure all these crates are publishable by dioxuslabs/publish?

Also, I presume we don't actually need a fork of wasm-bindgen, right? Just an alternate interface and then 1:1 api match depending on the target? We just need to track v0.2.x of wasm-bindgen and then we should be shielded against breaking changes?

If we are able to just paper over the differences in API, I would prefer all the code to sit here in the dioxus repo instead of sprawling over several different repos.

ealmloff:
> can we move https://github.com/DioxusLabs/wasm-bindgen-wry 's fork of wasm-bindgen to dioxuslabs as owner? and make sure all these crates are publishable by dioxuslabs/publish?

Moved and dioxuslabs/publish is added on all the crates

> Also, I presume we don't actually need a fork of wasm-bindgen, right? Just an alternate interface and then 1:1 api match depending on the target? We just need to track v0.2.x of wasm-bindgen and then we should be shielded against breaking changes?

wasm-bindgen itself isn't changed, but we need a few changes to js-sys since their assumptions about the wasm environment it runs in are sightly different than a desktop environment. Crates like web-sys that just call the wasm bindgen macro are unchanged except for the patch pointing to our shim macro.

js_sys::global() uses a thread local lazy to cache the document. With multi-window apps we run all windows in the same thread. wasm-bindgen-x patches this to cache based on the current execution context instead of assuming it is one per thread. There are a few other minor changes like we rely on std support that could probably be worked around with some time.

> If we are able to just paper over the differences in API, I would prefer all the code to sit here in the dioxus repo instead of sprawling over several different repos.

Similar to blitz and the dioxus-cli, wasm-bindgen-x will have a different more frequent release cadence to keep up with wasm-bindgen versions. Ideally we will also have crates outside of dioxus like the gloo family of crates and yew add support for wasm-bindgen-x

The only public api dependency we expose is http 1.4.0 for the request handler. We could remove that dependency with our own request type or move more of the handler logic into this repo if that is a concern. It is fully decoupled from wry/tao types so we shouldn't need to update when dependencies in desktop change. It mostly pulls in utility crates like futures-util and base64.

examples/08-apis/multiwindow.rs
```diff
@@ -12,7 +12,7 @@ fn main() {
 
 fn app() -> Element {
     let onclick = move |_| {
-        dioxus::desktop::window().new_window(VirtualDom::new(popup), Default::default());
```
ealmloff, line 15: new window now requires a callback to create the window config instead of accepting the window config directly because some settings do not implement send

examples/08-apis/wgpu_child_window.rs
```diff
@@ -16,22 +16,6 @@ use wgpu::CurrentSurfaceTexture;
 fn main() {
     let config = Config::new()
         .with_window(WindowBuilder::new().with_transparent(true))
-        .with_on_window(|window, dom| {
```
ealmloff, line 19: We cannot support this api as is anymore because the virtual dom is no longer on the main thread. This example works without it in the current form on macos, but flagging this in case this was required for specific platforms

This api was added in https://github.com/DioxusLabs/dioxus/pull/4919

packages/desktop/src/ipc.rs
```diff
@@ -1,9 +1,116 @@
...
+    fn from(variant: UserWindowEventVariant) -> Self {
+        Self { variant }
+    }
+}
 
 #[non_exhaustive]
-#[derive(Debug, Clone)]
-pub enum UserWindowEvent {
+#[derive(Debug)]
+pub(crate) enum UserWindowEventVariant {
```
ealmloff, line 104: This is now private which hides it from the public api. We can't make the event type private because tao forces us to expose the type with the window target api, but the variants can be private if we make the enum private and wrap it in a public struct with no api. Adds a bit of boilerplate to this pr, but should make things easier to change in minor releases

packages/desktop/src/event_handlers.rs
```diff
@@ -1,70 +1,160 @@
...
     window_id: WindowId,
 
     #[allow(clippy::type_complexity)]
-    handler:
-        Box<dyn FnMut(&Event<UserWindowEvent>, &EventLoopWindowTarget<UserWindowEvent>) + 'static>,
+    handler: Box<
+        dyn for<'a> FnMut(
+                Event<'a, UserWindowEvent>,
+                &EventLoopWindowTarget<UserWindowEvent>,
+            ) -> Event<'a, UserWindowEvent>
```
ealmloff, line 33: This type shuffling is required by tao since they only provide methods to turn `Event<'a, T> -> Option<Element<'a, U>>`, not the reference version and the enum is non exhaustive

packages/fullstack/Cargo.toml
```diff
@@ -39,6 +39,8 @@ serde_qs = { workspace = true, default-features = true}
 mime = "0.3.17"
 reqwest = { workspace = true, features = ["json", "rustls-tls", "cookies", "stream", "multipart"] }
 url = { workspace =  true }
+# Resolver guard for reqwest's cookie stack; remove once cookie supports time 0.3.48+.
+time = { workspace = true }
```
ealmloff: https://github.com/time-rs/time/issues/783

ealmloff:
High level overview for the review

## Thread model

There are now two threads for every desktop app:
1) The event loop thread which handles webview and window communication
2) The virtual dom thread which runs the app

There are synchronous calls between the two for wasm-bindgen calls and callbacks.

image

## API changes

- web-sys-events crate was split out so we could share event definitions between the web based platforms (web/desktop/mobile). We still need the existing serialization based events for liveview
- Desktop includes some breaking changes around tao/wry event loop hooks that now require sync. We offer async versions that don't require send where possible. **`with_on_window` was removed entirely**. We cannot support a method that runs on the main thread with the virtual dom since the vdom thread owns the vdom. see the open review comment for more details
- Tokio is now required on desktop unless you provide a custom async driver to the desktop config

PR #5200, opened 2026-01-10
Drag drop enhancements

kevinwright:
Added `FileData.url` and `DataTransfer.items`


packages/html/src/data_transfer.rs
```diff
@@ -57,12 +64,17 @@ impl DataTransfer {
...
+        self.inner.items()
+    }
+
     pub fn files(&self) -> Vec<crate::file_data::FileData> {
         self.inner.files()
     }
 }
 
 pub trait NativeDataTransfer: Send + Sync {
+    fn items(&self) -> Vec<DataTransferItem>;
```
nicoburns: Can we give this method a default implementation (that just returns an empty `Vec` + a TODO comment to remove it again). That way I think we can release this as a non-breaking change.

packages/html/src/data_transfer.rs
```diff
@@ -2,6 +2,13 @@ pub struct DataTransfer {
     inner: Box<dyn NativeDataTransfer>,
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct DataTransferItem {
```
nicoburns, line 13: Can we make `SerializedDataTransferItem` an alias of this type (we can derive `Serialize`/`Deserialize` on this type conditionally on the `serialize` feature being enabled.

kevinwright: I wanted to do this, but my Rust-fu isn't up to conditional derivation, so I have no idea how...

kevinwright: Okay, think I got it

kevinwright:
Having investigated thorougly, a bit more explanation...

Files dropped on a WebView are supplied to the (javascript) event handler as a `Blob` - of which `File` is a subclass.

A Blob can be converted to a `blob:` url via the `URL.createObjectURL()` method. But the generated url is only valid within the WebView itself until the next page load. It's a weak opaque reference.

The **only** way to get the underlying content in a manner usable outside the WebView is via a `FileReader`, which can generate an `ArrayBuffer` or base64-encode it into a `data:` url. In both cases it returns via an event listener (much like XMLHttpRequest), which made the use of `async` unavoidable.

Given that the _whole point_ was to yield something that could be serialised across the ipc boundary, the data url was the obvious and natural choice.

packages/interpreter/src/ts/native.ts
```diff
@@ -350,10 +350,10 @@ export class NativeInterpreter extends JSChannel_ {
     }
   }
 
-  handleEvent(event: Event, name: string, bubbles: boolean) {
+  async handleEvent(event: Event, name: string, bubbles: boolean) {
```
nicoburns: Unfortunately `handleEvent` needs to remain synchronous. This is necesary for the `preventDefault` functionality below:

```typescript
    const response = this.sendSerializedEvent(body);
    // capture/prevent default of the event if the virtualdom wants to
    if (response) {
      if (response.preventDefault) {
        event.preventDefault();
```

If handleEvent is async then this `event.preventDefault();` won't do anything which means that https://docs.rs/dioxus-core/latest/dioxus_core/struct.Event.html#method.prevent_default will stop working.

---

I think what you need to somehow pass a reference to the browser's event object through the serializing boundary. You can then make the `items` function on `DataTransfer` async and call back into the webview to obtain to decode the `Blob`s and obtain the data at that point.

@ealmloff ^ Does that sounds right to you? And also, how can we actually do that using Dioxus's data serialization setup?

kevinwright: The problem here is:
- drag-drop ONLY supplies a Blob
- Blobs are ONLY readable within the javascript environment of the page (WebView stupidly has no methods to access them)
- The ONLY way to get the content of a Blob is via a `FileReader`
- FileReader doesn't have synchronous methods,it can ONLY be used via callbacks

nicoburns: @kevinwright Yes, ultimately accessing the file data is going to need to be asynchronous. And it's going to need to happen from inside the context of the webpage. The idea here is to *defer* the asychrony to the `items()` method rather than the `handleEvent`() method so that `preventDefault()` can still be called in `handleEvent()`. And to implement that by calling back into the WebView via the FFI/IPC mechanism in the `items()` method.

(exactly how that is done I'm not sure, but I'm pretty sure we already do a similar thing for other events like `MountedData`, so it must be possible).

kevinwright: One other possibility did occur to me. Is there any way we could:
- Return a `blob:...` uri, which can be completely synchronous
- In parallel, launch the FileReader
- When the FileReader completes, it calls a separate event handler to say "blob uri <x> corresponds to data uri <y>"

Alternately...
- Have the FileReader get a byte array instead of a data uri
- on completion, it prepends the `blob:...` url to be an `announce-blob:...` uri
- then makes a `PUT` request to this
- which we pick up via a custom protocol handler

This also has the advantage of not potentially running into any string length limits with data uris. Plus Dioxus gets to announce that they've solved an issue which I know for a fact has been plaguing Tauri for years and has never been cleanly solved.

kevinwright: I never thought XMLHttpRequest would ever be the right answer to *anything*.

But... here we are. No more async.

nicoburns: Yeah, Dioxus Desktop is heavily reliant of synchronous http requests so that we can get sync communication to the outside world. Bit of a hack, but it works quite well in practice.

nicoburns (approved):
Looks good to me, athough it may be that there's a better approach by using native drag-drop events to bypass the webview.

kevinwright:
Native window events CANNOT do this. Both winit and wry can only handle local files in native drop events, and require that you sacrifice DOM-based drop events to do so.

- https://github.com/rust-windowing/winit/issues/720
- https://v2.tauri.app/release/wry/v0.38.0/
- https://github.com/tauri-apps/wry/blob/51d06d0cecdf69652a0ae636b1b875e427534bc3/src/lib.rs#L2181

This is a solution that allows us to continue using WebView drop events for images dropped from a browser (which will be `data:` urls, url lists, or HTML fragments). But to also be able to handle dropped local files, for which WebView hides the full path.

jkelleyrtp:
> Native window events CANNOT do this. Both winit and wry can only handle local files in native drop events, and require that you sacrifice DOM-based drop events to do so.
> 
> * [Drag & Drop meta-issue rust-windowing/winit#720](https://github.com/rust-windowing/winit/issues/720)
> * https://v2.tauri.app/release/wry/v0.38.0/
> * https://github.com/tauri-apps/wry/blob/51d06d0cecdf69652a0ae636b1b875e427534bc3/src/lib.rs#L2181
> 
> This is a solution that allows us to continue using WebView drop events for images dropped from a browser (which will be `data:` urls, url lists, or HTML fragments). But to also be able to handle dropped local files, for which WebView hides the full path.

key point: you sacrifice DOM-based drop events to do so.

Yes, but because we control the webview, we can synthesize the actual event and tie the IDs together as the event traverses from the webview to native. I'd rather we capture everything on the native end and then replay whatever the webview needs than let the webview be the source of truth.

Let me take a closer look at this. We can definitely support file drags/drops without needing to catch it in the webview.

If someone drops a 1gb file in the webview, we definitely don't want to send it via the ipc boundary.

kevinwright:
So the underlying issue here is in *what* can be captured natively.

Drag & drop events are usually considered in two forms:
- Files dragged from your computer, which is handled via an os-specific mechanism
- Other stuff internal to the app, such as rearranging columns in a table

But it's also possible for the os-native drag and drop to handle content besides files. This is typically done using a mechanism very similar to the clipboard and can handle arbitrary data, but details vary by OS. Neither wry nor winit can handle such drops. Winit has said they have no intent to and wry renamed some stuff in anticipation two years ago then did nothing more.

If you want to handle such events in dioxus-desktop without webview you'd have to implement that logic, separately for each OS, in wry. That's not a small undertaking. Or... you can just use WebView which has already done it. And **that** is what I don't want to give up.

Because my own use case is "accept drag/dropped images, from anywhere". This means I 100% cannot live without the WebView support because I need to be able to accept images dragged from websites. Sometimes these are data urls, or then can be url lists, or even HTML fragments, but **not files** and so not usable by native drop handlers.

But I also want to be able to drop local files. In which case webview hides the file path (I searched hard for a way to disable that) and only exposes a blob which can only be read in the WebView's own js engine.

This is a hack, I freely admit. But it's a hack around the design of webview, which is something that isn't going to change any time soon. Though we can do a bit better on the IPC boundary. Instead of base64-encoding the blob data, the typescript drop handler could (asynchronously!) PUT the raw bytes to a custom protocol handler where we could cache them keyed by the corresponding `blob:` uri (which is all the event would return). To avoid memory leaks this cache should then be cleared on a page navigation event. That would then be a general-purpose mechanism which we could use in any other case where we needed to break blobs out of the webview sandbox. I didn't do that because it's beyond my current comfort level with the dioxus codebase.


Electron can do a bit better, but only by embedding the entirety of chromium which they then only have to patch once to add the necessary behaviour.

nicoburns:
> Winit has said they have no intent to and wry renamed some stuff in anticipation two years ago then did nothing more.

I think winit will get there eventually. They seems to have been a lot more accepting of contributions recently. I'm also quite willing to soft-fork Winit at this point if need be (Wry unfortunately seems to be a dying project that has fallen quite far behind Winit and even Tauri are looking to migrate off of).

---

That being said, I agree that it makes sense in the short term to pull the data out of the webview so that we can support this use case ASAP.

kevinwright:
Yeah. And this is a use-case that's had people tearing their hair out for **years** with no definitive/consistent answer.

If it was easily done without something a bit hacky, I rather think it would have been.

For me, this approach works, and it works now. I can get file bytes from the drop event data without having to care about where they came from, and a subsequent pivot to the native handler approach (or anything else) wouldn't be a breaking change.

kevinwright:
This is a terrible PR, and I'm going to self-reject it. Because I finally have the context to understand what @jkelleyrtp  was saying.

We are, indeed, getting native file drop events, and they get saved to `file_hover` here: https://github.com/DioxusLabs/dioxus/blob/15bf75b7b0df82384c98872eae80c12c299abf50/packages/desktop/src/webview.rs#L306

That's then used to swap full paths into the HTML drop event here: https://github.com/DioxusLabs/dioxus/blob/15bf75b7b0df82384c98872eae80c12c299abf50/packages/desktop/src/webview.rs#L160

At first I thought this logic was part of some windows-specific workaround, but I was wrong there.
Thanks to some judicious eprintln! usage, I can now see that it can't work, because the events are handled in the wrong order. `<element>.ondrop` is called *BEFORE* `file_drop_handler` fires. So `file_hover` doesn't contain the necessary full path at the point in time when we need it to (and drag-over events don't contain the path either).

I'm going to play a bit with the idea of capturing paths from the drag-over event then re-using them to enrich the drop event.

PR #5204, opened 2026-01-11
Bump rfd to 0.17.2

nicoburns:
This bumps MSRV to 1.90 which is non-ideal for a patch release. But currently the Dioxus Desktop build is completely broken on linux.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: blacksmith-4vcpu-ubuntu-2404
     steps:
       - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@1.90.0
+      - uses: dtolnay/rust-toolchain@1.88.0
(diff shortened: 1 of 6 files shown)


LeWimbes:
The MSRV has been lowered back to 1.88. Therefore, it might make sense to wait for rfd 0.17.2.

See https://github.com/PolyMeilex/rfd/pull/303.

nicoburns:
Yes, although it looks like the immediate issue has been resolved upstream. So we can probably defer this to 0.8.0 (at which point the MSRV bump is probably fine).

ealmloff:
It looks like the apt cache action we are using is flakey. Last run failed to load the cache and silently continued to the next step, but it looks like it is doing better after resetting the cache

ealmloff (approved):
Looks like this works with the current msrv after the new minor version of rfd, so this should be good for 0.7.3? It doesn't seem like the ashpd patch fixed things

PR #5211, opened 2026-01-12
Full file paths in drop handlers

kevinwright:
Use native drop handler to push full file paths into element drop handler

### DIFF
--- a/packages/desktop/src/webview.rs
+++ b/packages/desktop/src/webview.rs
@@ -158,14 +158,20 @@ impl WebviewEdits {
                 let full_file_paths = hovered_file.current_paths();
 
                 let xfer_data = drag.data_transfer.clone();
-                let new_file_data = xfer_data.files.iter().map(|f| {
-                    let new_path =
-                        full_file_paths.iter().find(|p| p.ends_with(&f.path)).unwrap_or(&f.path);
-                    SerializedFileData {
-                        path: new_path.clone(),
-                        ..f.clone()
-                    }
-                }).collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


packages/desktop/src/file_upload.rs
```diff
@@ -305,15 +305,25 @@ impl HasFormData for DesktopFormData {
...
     event: Rc<RefCell<Option<DragDropEvent>>>,
+    paths: Rc<RefCell<Vec<PathBuf>>>,
 }
 impl NativeFileHover {
     pub fn set(&self, event: DragDropEvent) {
+        match event {
+            DragDropEvent::Enter { ref paths, .. } => self.paths.borrow_mut().clone_from(paths),
+            DragDropEvent::Drop { ref paths, .. } => self.paths.borrow_mut().clone_from(paths),
+            _ => {}
+        }
```
nicoburns, line 316: Why is this being stored here rather than extracted from the `event` like before? Is it actually necessary for the new functionality, or just preference?

kevinwright: Because the event bubbles up from WebView to Wry, so the native handler is called **after** the element handler. So we can't inject paths from the native handler via that route.

However... paths are supplied in two cases, for both Enter and Drop. So the sequence is:

- element OnDragEnter
- native OnDragEnter <- paths are saved here
- ...
- a bunch of OnDragOver events without paths that aren't relevant for this
- ...
- element OnDrop <- paths are injected here
- native OnDrop <- paths are saved here too, but not used

nicoburns (approved):
This also needs to be formatted

PR #5212, opened 2026-01-12
Implement `DataTransfer::items()`

kevinwright:
Adds an `items()` method that can be used to iterate over dropped data.

### DIFF
--- a/packages/html/src/data_transfer.rs
+++ b/packages/html/src/data_transfer.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "serialize")]
 pub use ser::*;
 #[cfg(feature = "serialize")]
-use serde::{Serialize, Deserialize};
+use serde::{Deserialize, Serialize};
 
 pub struct DataTransfer {
     inner: Box<dyn NativeDataTransfer>,


jkelleyrtp:
Can you combine this with the other PR so we have one PR that adjusts this plumbing code instead of multiple?

kevinwright:
Are you sure? This is completely independent of the paths PR.

I actively try to avoid unrelated things in a PR if I can help it, especially if the commits are squashed, as it makes it so much harder for the next person who tries to understand the evolution of the codebase by looking at commits.

It also makes it harder to rollback just the one part if a bug is found.

packages/html/src/data_transfer.rs
```diff
@@ -1,7 +1,21 @@
...
+#[cfg(feature = "serialize")]
+use serde::{Serialize, Deserialize};
+
 pub struct DataTransfer {
     inner: Box<dyn NativeDataTransfer>,
 }
 
+#[derive(Debug, PartialEq, Clone)]
+#[cfg(feature = "serialize")]
+#[derive(Serialize, Deserialize)]
```
nicoburns: This needs to be `#[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]`. What's you've written is conditionally compiling the struct itself.

nicoburns:
@kevinwright this needs to be formatted with `rustfmt` / `cargo fmt`

PR #5242, opened 2026-01-19
fix: remove "rust-lld" as custom linker on windows

### LINKED ISSUE #4694: `dx serve --hot-patch` fails on Windows with `linker = "rust-lld.exe"` set
**Problem**

Explicitly opting into lld on Windows (a practice recommended, for example, by [Bevy's getting started guide](https://bevy.org/learn/quick-start/getting-started/setup/#alternative-linkers)) causes `dx serve` to fail consistently with the following errors (verbose logging):
```
01:16:36 [dev] Linker warnings: Linker error:
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
 
01:16:36 [dev] Failed to generate fat binary: lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
01:16:36 [dev] Fat linking completed in 146771us
01:16:36 [dev] Setting builder to failed state
01:16:36 [dev] Build failed: The system cannot find the file specified. (os error 2)
```
The first error seems to suggest that `rust-lld.exe` isn't being called with the right semantics (to inform it what linker flavor we want).

**Steps To Reproduce**

- ensure you have `rust-lld.exe`:
```
cargo install -f cargo-binutils
rustup component add llvm-tools-preview
```
- create a skeleton project:
```
cargo new repro
cd repro
```
- add a `.cargo/config.toml` using `rust-lld.exe` for windows:
```toml
[target.x86_64-pc-windows-msvc]
linker = "rust-lld.exe"
```
- run with `dx serve --hot-patch`

**Expected behavior**

No error should occur.

**Environment:**

- Dioxus version: dioxus 0.7.0-rc.0 (a519fba)
- Rust version:  rustc 1.90.0 (1159e78c4 2025-09-14)
- OS info:  Windows 10
- App platform: desktop

alex-ds13:
While porting some of the windows fixes into `cargo-hot` ([this PR](https://github.com/hecrj/cargo-hot/pull/5)), I discovered and fixed another issue on windows that also exists on dioxus cli.

The issue is that the `lld` packed by rust by default as `rust-lld.exe` on windows needs a `-flavor` flag to tell it which flavor to use. When executing `cargo` normally it seems to be passing a `Z linker-flavor=link` to rustc which ends up calling "rust-lld.exe" with the proper flavor, that is it calls `rust-lld.exe -flavor link`.

We could do the same here and add the `-flavor link` to the "rust-lld.exe" command on Windows when we detect that the user is using the "rust-lld" linker. However I believe that is just some unnecessary complexity, since we already use the lld version packed by rust on windows by default, we can simply set `custom_linker` to `None` when the user is using "rust-lld" as linker. And then it will behave properly.

Also, I've used `cfg!(windows)` instead of checking for `BundleFormat::Windows`, since from what I understood the `BundleFormat` is related to the target you are building and not the system you are currently using, but if you think there is a better way to do this, let me know!

Also let me know if you think this should be fixed in another manner, like passing the flavor flag to the custom linker.

Fixes #4694

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -861,6 +861,20 @@ impl BuildRequest {
         let mut custom_linker = cargo_config.linker(triple.to_string()).ok().flatten();
         let mut rustflags = cargo_config2::Flags::default();
 
+        // Remove "rust-lld" as a custom linker on Windows, since that is already the linker we
+        // default to.
+        if let Some(linker) = custom_linker.as_ref() {
+            if (linker == "rust-lld" || linker == "rust-lld.exe") && cfg!(windows) {
+                // When using "rust-lld.exe" as linker on windows, it still needs to have a flavor
+                // given to it. rustc appears to be passing `-flavor "link"` when none is set by the
+                // user. If no flavor is given, it fails with 'lld is a generic driver'.
+                // We already use the existing lld-link by default on windows, so we can simply set the
+                // `custom_linker` to `None` in these cases, since we end up using "lld-link" anyway
+                // which is the same as "rust-lld.exe -flavor link".
+                custom_linker = None;
+            }
+        }
+
         // Make sure to take into account the RUSTFLAGS env var and the CARGO_TARGET_<triple>_RUSTFLAGS
         for env in [
             "RUSTFLAGS".to_string(),


jkelleyrtp (approved):

jkelleyrtp:
Thanks!

PR #4916, opened 2025-11-05
Replace unwraps with proper error handling in fullstack client

Himmelschmidt:
Closes #4915

PR #5248, opened 2026-01-21
fix: replace unwraps with errors in fullstack WebSockets

### LINKED ISSUE #5247: Panics on network errors due to .unwrap() calls in dioxus-fullstack WebSockets
**Problem**

Basically #4915, but with WebSockets. For instance, I got

```
thread 'main' (24899) panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-fullstack-0.7.3/src/payloads/websocket.rs:624:22:
called `Result::unwrap()` on an `Err` value: Reqwest(reqwest::Error { kind: Request, url: "http://100.64.0.11:8000/api/subscribe_to_updates", source: hyper_util::client::legacy::Error(Connect, ConnectError("tcp connect error", 100.64.0.11:8000, Os { code: 110, kind: TimedOut, message: "Connection timed out" })) })
```

**Environment:**

- Dioxus version:  v0.7.3
- Rust version:    rustc 1.92.0 (ded5c06cf 2025-12-08)
- OS info:         Arch Linux
- App platform:    web, mobile

**Questionnaire**

matous-volf:
Similarly to #4916, closes #5247.

### DIFF
--- a/packages/fullstack/src/payloads/websocket.rs
+++ b/packages/fullstack/src/payloads/websocket.rs
@@ -840,6 +840,14 @@ pub enum WebsocketError {
     /// Error during serialization/deserialization.
     #[error("error during serialization/deserialization")]
     Serialization(Box<dyn std::error::Error + Send + Sync>),
+
+    /// Error during serialization/deserialization.
+    #[error("serde_json error")]
+    Json(#[from] serde_json::Error),
+
+    /// Error during serialization/deserialization.
+    #[error("ciborium error")]
+    Cbor(#[from] ciborium::de::Error<std::io::Error>),
 }
 
 #[cfg(feature = "web")]
@@ -1551,6 +1559,8 @@ mod native {
                 WebsocketError::Tungstenite(error) => error.into_request_error(),
                 WebsocketError::Serialization(error) => Self::Serialization(error.to_string()),
                 WebsocketError::Deserialization(error) => Self::Decode(error.to_string()),
+                WebsocketError::Json(error) => Self::Decode(error.to_string()),
+                WebsocketError::Cbor(error) => Self::Decode(error.to_string()),
             }
         }
     }


packages/fullstack/src/payloads/websocket.rs
```diff
@@ -843,14 +840,6 @@ pub enum WebsocketError {
...
     #[error("error during serialization/deserialization")]
     Serialization(Box<dyn std::error::Error + Send + Sync>),
-
-    /// Error during serialization/deserialization.
-    #[error("serde_json error")]
-    Json(#[from] serde_json::Error),
-
-    /// Error during serialization/deserialization.
-    #[error("ciborium error")]
-    Cbor(#[from] ciborium::de::Error<std::io::Error>),
```
matous-volf, line 853: These are used nowhere.

packages/fullstack-core/src/error.rs
```diff
@@ -312,3 +312,60 @@ impl RequestError {
...
+        } else if value.is_redirect() {
+            Self::Redirect(string)
+        } else if value.is_status() {
+            Self::Status(
+                string,
+                value
+                    .status()
+                    .as_ref()
+                    .map(StatusCode::as_u16)
+                    .unwrap_or(DEFAULT_STATUS_CODE),
```
matous-volf: I am not sure how you guys handle these kinds of cases. This code returns a zero status code, which is not very valid (even the `StatusCode` type accepts only `NonZeroU16`).

I guess after checking that `value.is_status()` (which this code does), we could just do something like `.expect("status code not present when is_status is true")`, but that could theoretically panic, which is exactly what this PR is trying to avoid.

packages/fullstack/src/payloads/websocket.rs
```diff
@@ -619,14 +619,11 @@ impl IntoRequest<UpgradingWebsocket> for WebSocketOptions {
...
-                let response = native::send_request(request, &self.protocols)
-                    .await
-                    .unwrap();
+                let response = native::send_request(request, &self.protocols).await?;
 
                 let (inner, protocol) = response
                     .into_stream_and_protocol(self.protocols, None)
-                    .await
-                    .unwrap();
+                    .await?;
```
matous-volf, line 626: This is the whole point of this PR. Everything else are required error type conversions.

jkelleyrtp:
I would prefer we not pull in large deps in the fullstack-core crate. Can we just place the conversions inline in the higher level crates?

Placing them too low bloats the compile graph and slows down our compile times.

matous-volf:
Sure, is it okay like this?

PR #5267, opened 2026-01-26
fix: allow let chains in rsx! by updating rust editions

RichardEpure:
I noticed that I couldn't use let chains in my dioxus project despite it using rust edition 2024. This is due to a couple of packages in dioxus still using rust edition 2021 which I have updated.


jkelleyrtp:
I ended up wanting to port more of the workspace over and the fallout was bigger than I expected.

Going to close this and continue in my new branch

https://github.com/DioxusLabs/dioxus/pull/5502

PR #5300, opened 2026-02-08
fix(fullstack): prevent AlreadyBorrowed panic in WebSocket recv()

kyle-rader:
## Problem

When using `use_websocket` in a component where the WebSocket connection needs to be recreated (e.g., when navigating between different endpoints), the `recv()` method could panic with an `AlreadyBorrowed` error.

### Root Cause

`UseWebsocket::recv()` held a read borrow of the Resource's internal Signal across an `.await` point. When a reactive dependency (like a `group_id` signal) changed, `use_resource` would try to update the connection, requiring a write borrow. This panicked because `recv()` was still holding its read borrow.

**Panic message:**
```
panicked at dioxus-fullstack-0.7.3/src/payloads/websocket.rs:246:25:
called Result::unwrap() on an Err value: AlreadyBorrowed(AlreadyBorrowedError)
```

### Scenario

1. Component loads with WebSocket connection for endpoint A
2. `recv()` is awaiting a message (holding read borrow)
3. User navigates to endpoint B, reactive signal changes
4. `use_resource` detects change and tries to update connection
5. `value.set()` requires write borrow → **PANIC**

## Solution

Changed `recv()` to use `peek()` to get a short-lived borrow, extract a raw pointer from it, then drop the borrow guard before awaiting. This is safe because:

- The Resource is `'static` and won't be dropped
- The `UseWebsocket` handle is held by the component
- Even when `use_resource` updates, the old Websocket isn't immediately dropped - it only drops after outstanding futures complete

## Additional Changes

Added `tokio-util` `rt` feature which was missing and causing compilation errors in some scenarios.

## Testing

Tested in a real-world application (https://github.com/kyle-rader/automata.games) where navigating between different group WebSocket connections previously panicked consistently. With this fix, navigation works smoothly without any panics.

Fixes panic when navigating between WebSocket endpoints in reactive contexts.

### DIFF
--- a/examples/07-fullstack/websocket.rs
+++ b/examples/07-fullstack/websocket.rs
@@ -27,7 +27,7 @@ fn app() -> Element {
     let mut messages = use_signal(std::vec::Vec::new);
 
     // This signal is read inside the use_websocket closure, making it a reactive dependency.
-    // Whenver it changes, the websocket will automatically re-connect.
+    // Whenever it changes, the websocket will automatically re-connect.
     let mut name = use_signal(|| "John Doe".to_string());
 
     let mut socket =


jkelleyrtp:
We don't want unsafe to fix this, so I added an Rc under the hood. I also implemented another fix to properly re-connect when the socket changes.

kyle-rader:
Ty so much for looking at this @jkelleyrtp !

PR #5302, opened 2026-02-09
fix(fullstack): use target_arch instead of feature flags for server/web platform detection

raxosiris:
## Problem

In fullstack workspaces with multiple crates, Cargo's feature unification enables both `web` and `server` features simultaneously on `dioxus-fullstack-core`. This causes SSR hydration to break because code guarded by `#[cfg(feature = "server")]` and `#[cfg(feature = "web")]` both compiles and executes.

For example, in `history.rs`, the server branch (`entry.insert()`) runs in WASM instead of the client branch (`entry.get()`), causing hydration data to be overwritten instead of read.

## Reproduction

1. Create a fullstack workspace with a shared UI crate:
```
workspace/
  packages/app/     # depends on dioxus with features ["server"] or ["web"]
  packages/ui/      # depends on dioxus with features ["router", "fullstack"]
```

2. The `ui` crate unconditionally enables `fullstack`, which pulls in both `dioxus-fullstack-core/web` and `dioxus-fullstack-core/server` through transitive feature resolution.

3. Run `dx serve` — SSR renders correctly on the server, but WASM hydration fails because server-side code paths execute in the browser (e.g., `entry.insert()` instead of `entry.get()` in history.rs).

## Fix

Replace `cfg(feature = "server")` / `cfg(feature = "web")` with `cfg(not(target_arch = "wasm32"))` / `cfg(target_arch = "wasm32")` in platform-specific code paths.

**Why this works:** `target_arch` is determined by the compilation target, not Cargo features, so it can never be both values simultaneously. `dx serve` compiles the server for native (x86_64/aarch64) and the client for wasm32 — these are separate compilation units immune to feature unification.

**Why this is safe for other platforms:**
- Desktop, mobile, native, and liveview have **no dependency** on `fullstack-core`
- `fullstack-core` is only compiled for server (via `dioxus-server`) and web client (via `dioxus-web` with `hydrate`)
- The `fullstack` crate itself already uses `target_arch = "wasm32"` for websocket transport in `client.rs`, establishing this as the correct pattern

## Changes

- **`fullstack-core/src/history.rs`** — Route finalization: server inserts route, client reads it
- **`fullstack-core/src/loader.rs`** — Data loader serialization direction
- **`fullstack-core/src/server_cached.rs`** — Cache write (server) vs read (client)
- **`fullstack-core/src/server_future.rs`** — Server future serialization direction
- **`fullstack-core/src/transport.rs`** — `is_hydrating()` and `serialize_context()` platform detection
- **`fullstack-server/src/ssr.rs`** — Added explicit type annotation to `has_context::<HydrationContext>()`
- **`web/src/lib.rs`** — `should_hydrate` now respects runtime config only, removing `cfg!(feature = "hydrate")` override that could be incorrectly triggered by feature unification

### DIFF
--- a/packages/fullstack-core/src/history.rs
+++ b/packages/fullstack-core/src/history.rs
@@ -38,12 +38,14 @@ pub(crate) fn finalize_route() {
         return;
     };
 
-    if cfg!(feature = "server") {
+    // Use target_arch instead of cfg(feature) because cargo feature unification
+    // can enable both "web" and "server" features simultaneously.
+    if cfg!(not(target_arch = "wasm32")) {
         let history = history();
         let route = history.current_route();
         entry.insert(&route, std::panic::Location::caller());
         provide_context(ResolvedRouteContext { route });
-    } else if cfg!(feature = "web") {
+    } else if cfg!(target_arch = "wasm32") {
         let route = entry
             .get()
             .expect("Failed to get initial route from hydration context");
--- a/packages/fullstack-core/src/loader.rs
+++ b/packages/fullstack-core/src/loader.rs
@@ -42,11 +42,13 @@ where
     let storage_entry: crate::transport::SerializeContextEntry<Result<T, CapturedError>> =
         use_hook(|| serialize_context.create_entry());
 
-    #[cfg(feature = "server")]
+    // Use target_arch instead of cfg(feature) because cargo feature unification
+    // can enable both "web" and "server" features simultaneously.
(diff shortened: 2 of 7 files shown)


ealmloff:
We don't rely on the target to determine web vs server builds because the server can also be compiled to wasm on serverless targets like spin/fly.io. Your setup instructions sound very similar to the dioxus template workspace fullstack + router setup which does work correctly with hydration. Can you compare your feature setup to the template to narrow down the issue?

PR #5313, opened 2026-02-14
Avoid eval on web

mohe2015:
Probably the simplest way to fix this.

I use dioxus for developing web extensions and there is a really strict Content-Security-Policy


ealmloff:
Thanks!

PR #5315, opened 2026-02-14
Avoid inline css and js in dev

mohe2015:


packages/cli/src/build/request.rs
```diff
@@ -6217,7 +6229,7 @@ __wbg_init({{module_or_path: "/{}/{wasm_path}"}}).then((wasm) => {{
     /// Replace any special placeholders in the HTML with resolved values
     fn replace_template_placeholders(&self, html: &mut String, wasm_path: &str, js_path: &str) {
         let base_path = self.base_path_or_default();
-        *html = html.replace("{base_path}", base_path);
+        *html = html.replace("{base_path}", &format!("/{base_path}"));
```
mohe2015: I'm aware that this is technically a breaking change but I think we should do it as it avoids bugs where these imports only work depending on from where you import. This way they either work all the time or not at all.

jkelleyrtp:
Is this solving an issue in the tracker? I can't find anything. Why do we need to do this?

mohe2015:
Oh sorry for the missing description etc, there is no existing issue. When developing a browser extension using dioxus you are not allowed to use inline JavaScript because of a strict enforced content security policy. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy for Manifest V3.
~~The change I commented on could probably be dropped if you don't like it's side effects. (Removed)~~

packages/cli/src/build/request.rs
mohe2015: Removed

mohe2015:
Would it be possible to merge this please? I removed all controversal changes. This is needed to develop browser extensions using dioxus.

PR #5316, opened 2026-02-14
Support hotpatching in web extensions

mohe2015:
This is the KISS (Keep it simple) approach. I think it is not worth it to make this configurable for now as web extension allow localhost / 127.0.0.1 only anyways and this does not break anything. If somebody wants to e.g. support other ports at some point I think they can simply create a followup PR.


mohe2015:
Would it be possible to merge this please?

PR #5323, opened 2026-02-20
Introduce a crate to write unit tests of Dioxus apps

### LINKED ISSUE #5324: Dioxus testing library
## Feature Request

Right now it's hard to write tests against Dioxus apps following the [testing guide](https://dioxuslabs.com/learn/0.7/guides/testing/web). The guide offers two options, both of which have severe limitations:

* **Use dioxus-ssr to render and assert on the resulting DOM.** This does not provide any way to interact with the components in the test. As soon as the test needs to click a button, this is no longer an option.
* **Use Playwright to write end-to-end tests.** Such tests are very slow and notoriously hard to write and maintain. The resulting tests are in another language such as JavaScript, making it really difficult to instrument the system under test.

I would like writing tests to be really easy, without requiring too much thought into how to set up the test harness. Other frameworks like [React](https://testing-library.com/docs/react-testing-library/intro/) and [Flutter](https://api.flutter.dev/flutter/flutter_test/) have dedicated testing libraries, so I propose the same for Dioxus.

## Implement Suggestion

I have prepared a [PR](https://github.com/DioxusLabs/dioxus/pull/5323) with a minimal version of such a library. This uses the dioxus-native crate and Blitz to simulate layout and rendering. The test can then interact with the `VirtualDom`, dispatching events and driving the event loop. It can use the `DioxusDocument` to which the DOM was rendered to query elements.

See also [Discord thread](https://discord.com/channels/899851952891002890/1455297524960198758).

hovinen:
This is a preliminary version of a library in the spirit of [React](https://testing-library.com/docs/react-testing-library/intro/) or [Flutter](https://api.flutter.dev/flutter/flutter_test/) to allow easy unit testing of Dioxus apps.

Right now it only supports the `click` event, but I believe that is sufficient to show the design. Once the overall design is finalised, it can be fully fleshed out.

I placed this directly in the Dioxus repository. I'm also fine with keeping it as a separate crate if desired. However, this requires a few small changes to the dioxus-native crate to allow the creation of synthetic events. Otherwise the existing visibility restrictions make it impossible to dispatch DOM events from the test.

I decided to add fairly fleshed out documentation including doctests so as to illustrate the use of the library. I have tested it against a personal project and it is working quite well.

## Alternatives considered

- **Writing a new renderer**. I had considered this but found that it would be pretty much a duplicate of the existing dioxus-native.
- **Dispatching events via `Document::handle_ui_event`**. This would mean that the events would act more like real user-generated events, where, say, a click happens at particular coordinates and the framework identifies the element which is hit. This would be more "realistic" in the sense that a click on a button which is covered by another element would not hit the button but rather the element covering it. However, the `UiEvent` enum from Blitz does not support a "click" event and I did not want to create multiple PRs in different repositories for now.

Fixes #5324


packages/test/src/tester.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// # }
+    /// ```
+    ///
+    /// If this method is invoked with no pending asynchronous operations, then it times out after
+    /// one second and returns `Err(Elapsed)`.
+    pub async fn pump(&mut self) -> Result<(), Elapsed> {
+        timeout(PUMP_TIMEOUT, self.document.vdom.wait_for_work()).await?;
+        while self.document.poll(None) {}
+        Ok(())
+    }
```
ealmloff, line 137: Very excited about this, thanks for working on it!

Instead of requiring explicit event pumping, we could mimic the Playwright test API and make queries asynchronously run the event loop:
```rust
let mut tester = dioxus_test::render(AComponent).build();
// Wait for make-request-button to load and click it to make a network request.
tester.find_by_test_id("make-request-button").click().await?;
// Assert on the state of the UI while the network request is in flight.
expect(tester.find_by_test_id("loading-indicator")).to_be_visible().await?;
// Receive the server response and assert on the state of the UI after the response
// is received and the UI has been rerendered.
expect(tester.find_by_test_id("resolved-content")).to_be_visible().await?;
```
In some cases, waiting for work and applying a single tick of DOM operations could provide more control over testing, but I think the familiarity of the Playwright-style API and intuitive behavior in the presence of feedback loops with DOM events are typically more important.

For example, `onmounted` and `use_effect` will only trigger after the document has applied edits. With the current model, I think that would require rendering, then waiting for a 0ms timeout, then asserting the state has changed based on the `onmounted` event. If asserting on that state also pumped the DOM, you could just assert the state is correct directly.

hovinen: Thanks -- let me try this and see what I can come up with.

hovinen: Hi Evan, thanks for your patience on this. It took me a while to find a workable structure for this, and I admittedly didn't have as much to work on it as I would have liked.

I settled on a second layer on top of the initial API. So one can use the `pump` method to drive the event loop, but there's a higher level API which lets the caller await conditions on the DOM and assert on those.

Since the tester doesn't know when all async operations have fully settled, it has to rely on heuristics to some degree. It tries up to a maximum number of attempts to obtain an element or to make an assertion and gives up eventually. The `pump` method is similarly equipped with a timeout since I don't have an easy way to detect whether it would just wait forever.

I had to introduce some new concepts to support assertions in this model. A given element on which the test is to assert might already be present but have the wrong state at the time the assertion is invoked. So there needs to be a concept of "assert that this is _eventually_ true, even if it's not true right now." To support this, I introduced matchers similar to the [GoogleTest crate](https://crates.io/crates/googletest). These can be passed into the tester which can just repeatedly try the assertion until it's true, giving up after a maximum number of tries.

I've tried this out on a private project and the API seems reasonable.

There are some more TODOs on this: I need to update the overall docs and I'd like to make some improvements to the API. But this should at least convey the idea.

Please let me know what you think. Thanks!

ealmloff: Much nicer! We could make this more composable by extracting a poll once or await helper so you can do something like:
```rust
tester.get("selector").immediately()?;
tester.get("selector").await?;
tester.get("selector").expect(inner_html(to_contain("hello"))).await?;
tester.get("selector").expect(inner_html(to_contain("hello"))).immediately()?;
```

It isn't fully built out yet, but I have a prototype in this commit: https://github.com/DioxusLabs/dioxus/commit/358b9ede63bf44af41eff9a597cd2f36cc690716

hovinen: Interesting! I had played around trying to build a fluent interface like that but found that whatever the query method on `DocumentTester` returned had to contain an exclusive reference to `DocumentTester`, which caused no end of problems. I admittedly didn't consider implementing `IntoFuture` directly.

I'll play around with your prototype and see how well it works in the project I am doing.

hovinen: Hi Evan, I've built on your commit and managed to get it to "feature-parity" with the previous solution. I've pushed what I have, but it's going to need a lot more cleanup before it's ready for merge. Still, I felt this is a good checkpoint to show you what I have.

One can construct a query, then await on it or invoke `immediately()` to assert that it already resolves to an element. Similarly, one can call `expect()` and supply a `Matcher` on that query. The expectation then follows the same pattern: one can `.await` or invoke `.immediately()`.

Here's how its use looks:

```rust
let mut tester = create_tester(...);

tester.query(by_testid("some-button")).click().await?;

tester
    .query(by_testid("some-state"))
    .expect(inner_html(not(contains_string("Button not yet clicked"))))
    .await?;
Ok(tester
    .query(by_testid("some-state"))
    .expect(inner_html(contains_string("Button now clicked")))
    .immediately()?)
```

(I'll update the docs again once the API is reasonably settled.)

This was quite tricky to get right due to all the lifetimes in play here. I settled on having the asynchronous loop part just return node IDs so that they don't hold any references, then resolving them afterwards.

ealmloff: Yes, that looks much nicer!

Cargo.toml
```diff
@@ -220,7 +222,7 @@ tracing = "0.1.41"
 tracing-futures = "0.2.5"
 tracing-subscriber = { version = "0.3.19", default-features = false }
 toml = "0.8"
-tokio = "1.48"
+tokio = { version = "1.48", features = ["time"] }
```
nicoburns, line 225: This feature specification probably ought to go in the `dioxus-test` crate's `Cargo.toml` so that build's that don't use `dioxus-test` don't have to have this feature enabled.

packages/test/src/document.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// # }
+    /// ```
+    ///
+    /// If this method is invoked with no pending asynchronous operations, then it times out after
+    /// one second and returns `Err(Elapsed)`.
+    pub async fn pump(&mut self) -> Result<(), Elapsed> {
+        timeout(PUMP_TIMEOUT, self.document.vdom.wait_for_work()).await?;
+        while self.document.poll(None) {}
+        Ok(())
+    }
```
nicoburns, line 117: Doesn't necessarily block merging this PR, but we probably ought to integrate resource fetching (linked stylesheets, fonts, images, etc), and have the tester be aware of whether there are in-flight resource requests.

nicoburns: Especially if we want to add rendering and visual snapshot testing to this at some point.

packages/native-dom/src/events.rs
```diff
@@ -229,3 +230,7 @@ impl HasFocusData for NativeFocusData {
         self as &dyn std::any::Any
     }
 }
+
+pub fn synthetic_click_event(node: &Node, modifiers: Modifiers) -> Box<dyn Any> {
+    Box::new(NativeClickData(node.synthetic_click_event_data(modifiers)))
+}
```
nicoburns, line 236: This seems totally fine to me :) We'll probably want to add something more generic (synthetic `DomEvent`s in general) at some point.

packages/test/src/document.rs
```diff
@@ -0,0 +1,464 @@
...
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + 'vdom>>;
+
+    fn into_future(mut self) -> Self::IntoFuture {
+        Box::pin(async move {
+            let node_ids = self.into_waitable_future().await?;
+            Ok(node_ids
+                .into_iter()
+                .map(|node_id| self.data.node_id_to_element(node_id))
+                .collect())
+        })
```
nicoburns: Might it be possible to add a `map_result` method to `Waitable`, then make the `IntoFuture` impl generic for all `Waitable`? I think `.immediately()` could then also be a provided method of the `Waitable` trait.

packages/test/src/document.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+pub trait ImmediateCondition<'output> {
+    type Output: 'output;
+    fn immediately(&'output self) -> Result<Self::Output, TesterError>;
+}
+
+pub struct ElementCondition<'vdom> {
+    data: &'vdom mut DocumentTester,
```
nicoburns: Would it make things simpler if the "conditions" didn't own the `DocumentTester` but just had it passed into their methods? Probably then `pump` wouldn't be needed on the "conditions" and could be handled by the thing polling them?

packages/test/src/document.rs
```diff
@@ -0,0 +1,464 @@
...
+                Ok(node)
+            }
+        }
+    }
+}
+
+pub struct AllElementsCondition<'vdom> {
+    data: &'vdom mut DocumentTester,
+    query: SelectorList,
+}
```
nicoburns: Can we a short doc comment on each struct? i.e. "A test matcher that waits for all provided selectors to be present in the document and then returns their node IDs" or something

packages/test/src/element.rs
```diff
@@ -0,0 +1,174 @@
+use blitz_dom::Node;
```
nicoburns, line 1: Quite a lot of this file probably ought to move into `blitz_dom` (e.g. method on `Node`) at some point.

packages/test/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+//!              "Hello, world!"
+//!         }
+//!     }
+//! }
+//!
+//! #[test]
+//! fn my_component_renders_correctly() {
+//!     let tester = render(MyComponent).build();
+//!     assert_eq!(
+//!         tester.find_by_css_selector(".test-component").unwrap().inner_html(),
```
nicoburns: Should `find_by_css_selector` just panic if it fails? `.unwrap()` here seems like noise.
(we could also always have a non-panicking `try_find_by_css_selector` in case someone does need that).

packages/test/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+//! The method [DocumentTester::pump] returns control to the async runtime and thus drives any
+//! asynchronous operations such as requests to the backend. If any rendering depends on the result
+//! of a request, then the test must invoke [DocumentTester::pump] to resolve that.
+//!
+//! The test can also assert on the state of the DOM while backend requests are in flight.
+//!
+//! ## Limitations
+//!
+//! Interactions with the DOM operate directly on elements, not on the screen. So if, say, the test
+//! dispatches a click on an element which is covered by a frost, the element will respond as though
```
nicoburns, line 131: What's a frost?

packages/test/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+//! it were reachable even though it would not be in reality.
+//!
+//! The layout system is limited by what the Blitz layout system can support. Since Blitz is not
+//! complete as of the time of writing, computed layouts will often not be as in reality.
```
nicoburns, line 135: > computed layouts will often not be as in reality.

Perhaps we ought to specify "may not match the web or desktop renderers"? As they will match if people's apps are also using the native renderer.

packages/test/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+//!     }
+//! }
+//!
+//! #[tokio::test]
+//! async fn my_component_changes_button_text_on_click() {
+//!     let mut tester = render(MyComponent).build();
+//!     tester.find_first_by_css_selector(".test-button").unwrap().click();
+//!     tester.pump().await;
+//!     assert_eq!(
+//!         tester.find_first_by_css_selector(".test-button").unwrap().inner_html(),
```
nicoburns: Would it make sense to have `find_*` method implicitly pump? (do they not already?). If, so we could remove the explicit pump above.

nicoburns (commented):
This overall looks really good to me. I've made some suggestions, some of which are a lot more speculative than others.

packages/test/src/lib.rs
hovinen: These docs are outdated -- there is now just such an API. I just haven't updated the docs yet.

packages/test/src/lib.rs
hovinen: I'm referring to a div which covers the page and prevents interaction with elements covered by it, similar to https://css-tricks.com/frosting-glass-css-filters/.

packages/test/src/document.rs
hovinen: I plan to do so, but want the API to stabilise a little first.

packages/test/src/document.rs
hovinen: Not sure how you mean that.

Suppose I have something like:

```rust
tester.query(by_testid("something")).expect(inner_html(contains("Some text"))).await
```

`ElementCondition` is the output of `query()`. It needs the `DocumentTester` any time it wants to drive the DOM, e.g., in the `IntoFuture` implementation behind `await`. If it doesn't hold a reference to `DocumentTester`, then there's no other opportunity in the line above to deliver that reference to the `IntoFuture` implementation. Or am I missing something?

packages/test/src/document.rs
hovinen: Thanks for the suggestion! I've tried something like this but ran into lots of obscure lifetime issues. But I'll see what I can do.

packages/native-dom/src/events.rs
hovinen: Good to know, thanks!

packages/test/src/document.rs
hovinen: Noted, thanks!

Cargo.toml
hovinen: Understood, thanks.

packages/test/src/lib.rs
hovinen: I'm following the philosophy in [GoogleTest](https://crates.io/crates/googletest): don't panic, but rather return a `Result` and let the caller decide how to handle that.

I'm fine with panicking if that's what you prefer, though.

hovinen:
Thanks for your review! I'll address the code changes as soon as I get a chance.

packages/test/src/tester.rs
hovinen: So, after what I'm sure was the most challenging lifetime jujitsu I have ever experienced, I'm happy to say that I got everything working with the fluent API. I've cleaned things up a bit and updated this PR with my current state. I still have quite a few things to do before this is ready, especially updating docs and seeing to decent error messages when assertions fail.

One caveat: I'm not sure the compiler error messages will be especially helpful in the case that one makes a mistake, such as by using a matcher in `expect` which doesn't make sense in that context. There might be need to iterate on that design a little.

I'm also thinking that implementing a full set of matchers would essentially duplicate a lot of the [GoogleTest crate](https://crates.io/crates/googletest). So I'm musing over whether it makes sense just to introduce a dependency on that crate. I'd be happy to hear your thoughts on that.

I'll continue updating this branch in the coming days, as I find the time.

packages/test/src/lib.rs
hovinen: I've updated the docs now to reflect the current API. Please take a look.

packages/test/src/tester.rs
hovinen: All right, I think I've settled on a reasonable solution which is ready for another round of review. Thanks!

Cargo.toml
hovinen: Done

packages/test/src/document.rs
hovinen: After a lot of revisions, I landed somewhere slightly different:
- `AllElementsCondition` doesn't implement `Waitable` any more. Unlike `ElementCondition`, there is no clear unambiguous condition for which to wait. The query could, after all, legitimately yield no elements.
- The 'immedately()` methods are now just ordinary methods on each of the structs. There is no trait for them. This is because there is nothing which needs to be generic over the type on which to call `immediately()`.
- The implementation of `Matchable` for `AllElementsCondition` now just references `immediately()`, eliminating the code duplication.
- `ElementCondition` has a method which basically does what `map_result` would do. But since it's the only thing which needs it, it's not on any trait.
- There are now only two `IntoFuture` implementations rather than three. It's still very slight duplication, but I think this makes the design a bit easier.

packages/test/src/lib.rs
hovinen: Done

packages/test/src/element.rs
hovinen: Acknowledged

ealmloff:
I started porting some of the component library tests to this crate in this branch to try out the api https://github.com/DioxusLabs/dioxus-components/tree/testing-lib (using this branch with a driver trait, more events and selectors https://github.com/ealmloff/dioxus/tree/testing-tweaks)

Overall, its nice to have everything in rust and much faster, but runs into a few issues:
- Playwright lets you create selectors separately from assertions which makes tests less verbose. Currently, assertions in this pr own the selector which means you have to duplicate it. It would be nice if you could write:
```rust
let mut tester = dioxus_test::render(AComponent).build();
let click_count = tester.query("#click-count");
click_count.expect(inner_html(contains_string("Click count: 0"))).await?;
tester.query("button").click().await?;
click_count.expect(inner_html(contains_string("Click count: 1"))).await?;
```
instead of:
```rust
let mut tester = dioxus_test::render(AComponent).build();
tester.query("#click-count").expect(inner_html(contains_string("Click count: 0"))).await?;
tester.query("button").click().await?;
tester.query("#click-count").expect(inner_html(contains_string("Click count: 1"))).await?;
```
- Blitz crashes if you run too many tests in parallel https://github.com/DioxusLabs/blitz/issues/430
- Missing some common selectors to assert things about attribute values, focus status, etc
- Missing inputs like keydown, and hover
- Only works in the blitz renderer. While blitz can test the majority of functionality for the component library, some tests with js for things like focus traps only work in full browser. With #5194, hopefully we can pull out a driver trait and implement one backing for renderers that implement js

hovinen:
Thanks for the feedback!

I respond to your points inline below.

> I started porting some of the component library tests to this crate in this branch to try out the api https://github.com/DioxusLabs/dioxus-components/tree/testing-lib (using this branch with a driver trait, more events and selectors https://github.com/ealmloff/dioxus/tree/testing-tweaks)
> 
> Overall, its nice to have everything in rust and much faster, but runs into a few issues:
> 
> * Playwright lets you create selectors separately from assertions which makes tests less verbose. Currently, assertions in this pr own the selector which means you have to duplicate it. It would be nice if you could write:
> 
> 
> ```rust
> let mut tester = dioxus_test::render(AComponent).build();
> let click_count = tester.query("#click-count");
> click_count.expect(inner_html(contains_string("Click count: 0"))).await?;
> tester.query("button").click().await?;
> click_count.expect(inner_html(contains_string("Click count: 1"))).await?;
> ```
> 
> instead of:
> 
> ```rust
> let mut tester = dioxus_test::render(AComponent).build();
> tester.query("#click-count").expect(inner_html(contains_string("Click count: 0"))).await?;
> tester.query("button").click().await?;
> tester.query("#click-count").expect(inner_html(contains_string("Click count: 1"))).await?;
> ```

This would be nice but will require some more thought. Right now the types returned by `query` and `query_all` hold an exclusive reference to the `DocumentTester`. This is needed so that they can drive the event loop when one `await`'s on them. I'll see what I can do.

> * Blitz crashes if you run too many tests in parallel [Resolving HtmlDocument in many different threads panics blitz#430](https://github.com/DioxusLabs/blitz/issues/430)
> 
> * Missing some common selectors to assert things about attribute values, focus status, etc
> 
> * Missing inputs like keydown, and hover

I intend to add these, but wanted to get the basic design right first. Would you consider these to be blockers for merging this PR, or can they be added after it is merged?

> * Only works in the blitz renderer. While blitz can test the majority of functionality for the component library, some tests with js for things like focus traps only work in full browser. With [Get wasm bindgen working on desktop #5194](https://github.com/DioxusLabs/dioxus/pull/5194), hopefully we can pull out a driver trait and implement one backing for renderers that implement js

This would also be nice but I couldn't find an easy way to make it work without the Blitz renderer when I built it. I'm interested to see any changes which make such a generalisation possible.

ealmloff:
> I intend to add these, but wanted to get the basic design right first. Would you consider these to be blockers for merging this PR, or can they be added after it is merged?

Events and more attribute values, no. Crashing when there are more than a few tests, yes

> This would also be nice but I couldn't find an easy way to make it work without the Blitz renderer when I built it. I'm interested to see any changes which make such a generalisation possible.

I'm working on a couple of bugs with the wasm-bindgen bindings today. Might be able to get a prototype of this later this week

hovinen:
Sorry for the slow reply. I've had a couple of things going on recently.

> > I intend to add these, but wanted to get the basic design right first. Would you consider these to be blockers for merging this PR, or can they be added after it is merged?
> 
> Events and more attribute values, no. Crashing when there are more than a few tests, yes

I've applied the workaround Nico mentioned in https://github.com/DioxusLabs/blitz/issues/430#issuecomment-4434025490. Please let me know whether that helps with the crashes you have seen.

> > This would also be nice but I couldn't find an easy way to make it work without the Blitz renderer when I built it. I'm interested to see any changes which make such a generalisation possible.
> 
> I'm working on a couple of bugs with the wasm-bindgen bindings today. Might be able to get a prototype of this later this week

Looking forward!

hovinen:
>     * Playwright lets you create selectors separately from assertions which makes tests less verbose. Currently, assertions in this pr own the selector which means you have to duplicate it. It would be nice if you could write:
> 
> 
> ```rust
> let mut tester = dioxus_test::render(AComponent).build();
> let click_count = tester.query("#click-count");
> click_count.expect(inner_html(contains_string("Click count: 0"))).await?;
> tester.query("button").click().await?;
> click_count.expect(inner_html(contains_string("Click count: 1"))).await?;
> ```
> 
> instead of:
> 
> ```rust
> let mut tester = dioxus_test::render(AComponent).build();
> tester.query("#click-count").expect(inner_html(contains_string("Click count: 0"))).await?;
> tester.query("button").click().await?;
> tester.query("#click-count").expect(inner_html(contains_string("Click count: 1"))).await?;
> ```

I've made some adjustments to enable this: Relevant methods on `ElementCondition` and `AllElementsCondition` now take a shared reference rather then consuming the struct.

The downside is that this means having to use interior mutability on the `DioxusDocument`. So there is a risk of panics if the tester does odd things like trying to concurrently pump the event loop and interact with the DOM. But they just shouldn't do that :slightly_smiling_face:.

hovinen:
Since we have agreed to move the testing library first to a separate repository in https://github.com/DioxusLabs/dioxus-test, I'm closing this PR now.

PR #5338, opened 2026-02-28
feat(signals): add WritableResultExt trait

### LINKED ISSUE #5293: [Feature Request] `WriteableResultExt` in addition to `ReadableResultExt`?
## Feature Request

Currently there is a `ReadableResultExt`, but no `WritableResultExt`. You can make a signal holding a `Result` readable with `.as_ref()`, but not `Writable` with an `.as_mut()`, There is a `WritableOptionExt`, but a `ResultOptionExt` would be useful for not needing a binding in order to create writable result value.

costajohnt:
## Summary

Adds `WritableResultExt` trait for `Writable<Target = Result<T, E>>`, mirroring the existing `ReadableResultExt` on the read side. This provides mutable access to signal values holding `Result<T, E>` without needing an intermediate binding.

### Methods

- **`as_mut()`** — returns `Result<WritableRef<T>, WritableRef<E>>`, giving mutable access to whichever variant the Result holds
- **`unwrap_mut()`** — returns `WritableRef<T>`, panicking if the Result is `Err`

### Pattern

Follows the same extension trait pattern used by `WritableOptionExt` and `ReadableResultExt`:
- Blanket impl over all `W: Writable<Target = Result<T, E>>`
- Uses `WriteLock::map` / `WriteLock::filter_map` for mapped mutable references
- `#[track_caller]` on all methods for better panic diagnostics

Closes #5293

### DIFF
--- a/packages/signals/src/write.rs
+++ b/packages/signals/src/write.rs
@@ -467,19 +467,18 @@ pub trait WritableResultExt<T, E>: Writable<Target = Result<T, E>> {
         T: 'static,
         E: 'static,
     {
-        let is_ok = self.read().is_ok();
-        if is_ok {
-            Ok(WriteLock::map(self.write(), |v| {
+        let write = self.write();
+        match write.as_ref() {
+            Ok(_) => Ok(WriteLock::map(write, |v| {
                 v.as_mut()
                     .ok()
                     .expect("Result variant changed between read and write")
-            }))
-        } else {
-            Err(WriteLock::map(self.write(), |v| {
+            })),
+            Err(_) => Err(WriteLock::map(write, |v| {
                 v.as_mut()
                     .err()
                     .expect("Result variant changed between read and write")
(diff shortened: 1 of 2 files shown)


packages/signals/src/write.rs
```diff
@@ -447,6 +447,45 @@ pub trait WritableOptionExt<T>: Writable<Target = Option<T>> {
...
+
+    /// Attempts to mutably access the inner value of the Result.
+    #[track_caller]
+    fn as_mut(&mut self) -> Result<WritableRef<'_, Self, T>, WritableRef<'_, Self, E>>
+    where
+        T: 'static,
+        E: 'static,
+    {
+        let is_ok = self.read().is_ok();
+        if is_ok {
```
ealmloff, line 471: Concerned this could cause race conditions if the value is changed after the read before the write

PR #5340, opened 2026-03-01
fix: Content-hash-based filename when bundling with wasm_split

### LINKED ISSUE #5339: Content hash based filename is disabled when `--wasm-split` is present
**Problem**

When invoking `dx bundle --wasm-split`, content-hash-based filename is not used. 

**Steps To Reproduce**

- `dx bundle --wasm-split`

**Expected behavior**

All assets files (html, css, js, wasm) has a content hash part.

**Screenshots**

Not applicable

**Environment:**

- Dioxus version:  0.7.2
- Rust version:    1.92.0
- OS info:         Linux
- App platform:    web

**Questionnaire**

I don't have time to fix this now, but I roughly know the way to fix (because claude told me):

In `/Users/ken/Projects/dioxus/packages/cli/src/build/request.rs`, Line 3998-4000:

```rust
fn should_bundle_to_asset(&self) -> bool {
  self.release && !self.wasm_split && self.bundle == BundleFormat::Web
}
```

We should remove `&& !self.wasm_split` here. According to claude's analysis, this won't cause any other problems.

SuperKenVery:
Add content hash to asset filenames. Without this, client may continue to use cached old versions even if the server has been updated.

---

I tried directly removing the `self.wasm_split` check. Seems working well.

Closes #5339


chtpl:
could you please add an explanation to the body of the commit message why this is done?

My assumption is that it helps with caching?

SuperKenVery:
Yes exactly, this fixes caching. Without this, clients will use old html or wasm files even if server has updated.

---

I added description to PR body. Adding to commit body would require force push, tell me if you want that.

Issue #5363, opened 2026-03-11
Store children writes don't always cause store parent subscribers to be notified


dunkyl:
**Problem**

When a store is written to through its children, reads don't see all the changes. Interestingly, some subscribers seem to be notified while others are not. In the example below, the effect that calls `store.read()` is rerun, but the higher scoped read in app is not. The effect running is dependent on the read of store.val in the child component.

**Steps To Reproduce**

- Read a store
- (Optionally) read through a lens
- Write through a lens

```rust
use dioxus::prelude::*;

#[derive(Store, Debug)]
struct MyStruct {
    val: i32,
}

fn use_my_store() -> Store<MyStruct> {
    let store = use_store(|| MyStruct { val: 0 });
    use_effect(move || {
        println!("use_my_store effect {:?}", store.read());
    });
    store
}

#[component]
fn DisplayVal(val_lens: ReadSignal<i32>) -> Element {
    rsx! {
        p {"store.val read = {val_lens}" }
    }
}

fn main() {
    dioxus::launch(|| {
        let store = use_my_store();
        let val_lens = store.val();
        rsx! {
            p { "store read = {store.read():?}" }
            DisplayVal { val_lens }
            button {
                onclick: move |_| *store.val().write() += 1,
                "increment val"
            }
        }
    });
}
```

- Implementation of PartialEq seems to not impact this case

**Expected behavior**

All writes through lenses to children notify subscribers to the parent.

**Environment:**

- Dioxus version:  0.7.3
- Rust version:   1.93.1 stable
- OS info:  Windows 11 26200.7840
- App platform:   desktop, web, did not test mobile

PR #5365, opened 2026-03-12
Fix: notify parent store when child is marked dirty

dunkyl:
To address #5363

Too easy to be real?

In practice, I was able to point my more complex app at this PR and its behavior was fixed. I tried to also include some tests to cover basic behavior (do writes spread down, up, and not to siblings?). The additional dev dependencies are for testing convenience.

What I am most unsure about is if `StoreSubscriptions` is the correct place for this fix, if it might need to instead be moved up into `SelectorScope`, or if it should be moved down into `SelectorNode`.

### DIFF
--- a/packages/stores/src/subscriptions.rs
+++ b/packages/stores/src/subscriptions.rs
@@ -286,7 +286,7 @@ impl StoreSubscriptions {
             paths
         };
         for path in descendant_paths {
-            self.mark_node_dirty(&path);
+            self.mark_node_subscribers_dirty(&path);
         }
     }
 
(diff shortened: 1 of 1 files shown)


ealmloff:
Thanks, the test here helps to narrow things down! This is the right issue, but the solution is too narrowly scoped. Parent's are currently tracking children by getting the list of child nodes which were already materialized in the subscription tree. That breaks down if you write to a child that didn't exist when the parent's list was created.

Instead of manually adding parents to the childs subscriber lists, the children should look up into the parents to see if they are deeply subscribed to all children and notify them if they are. This needs to be through all ancestors and only if the subscription is deep (eg. `read` can access the contents of all children, `len` cannot).

Can you confirm your app is working with the new approach?

dunkyl:
Thank you for the comprehensive fix! This works in my app exactly as expected.

Issue #5375, opened 2026-03-15
BorrowMutError with 'onkeydown' listener


Thibaud379:
I had a weird panic and managed to recreate it in a minimal example bellow

When clicking or selecting a suggestion using the keyboard on an `<input>` linked to a `<data-list>`, I get the following error
```
19:39:37 [web] wasm-bindgen: imported JS function that was not marked as `catch` threw an error:
19:39:38 [web] panicked at [...]/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-core-0.7.3/src/events.rs:665:28:
already borrowed: BorrowMutError
Stack:
Error
    at http://127.0.0.1:8081/wasm/front.js:954:25
    at logError (http://127.0.0.1:8081/wasm/front.js:1815:18)
    at __wbg_new_194e7e70e82f7747 (http://127.0.0.1:8081/wasm/front.js:953:57)
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[2946]:0xdd6a6
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[720]:0x7e13c
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[1393]:0xb9d37
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[1637]:0xc5742
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[3619]:0xe00fb
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[2566]:0xda8ce
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[2567]:0xda8fa

19:39:38 [web] panicked at  [...]/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dioxus-core-0.7.3/src/events.rs:665:28:
already borrowed: BorrowMutError
Stack:
Error
    at http://127.0.0.1:8081/wasm/front.js:954:25
    at logError (http://127.0.0.1:8081/wasm/front.js:1815:18)
    at __wbg_new_194e7e70e82f7747 (http://127.0.0.1:8081/wasm/front.js:953:57)
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[2946]:0xdd6a6
    at http://127.0.0.1:8081/wasm/front_bg.wasm:wasm-function[720]:0x7e13c
[...]

Issue #4079, opened 2025-05-06
Cannot read properties of undefined (reading 'length') error on Chrome mobile after autocomplete


brianmay:
**Problem**

When selecting an autocomplete suggestion in Chrome on Android, website stops responding.

Have not been able to reproduce with other browsers or on desktop because I don't see autocomplete there.

Code in question is available at https://github.com/brianmay/penguin_nurse/

**Steps To Reproduce**

- Using Chrome on Android, bring up form in Dioxus application.
- Form works as expected.
- Chrome helpfully gives suggestions for text fields.
- Select any suggestion, and error occurs.

Error is only seen when setting up Chrome for remote debugging. User sees website stops responding to any input as a result.

**Expected behavior**

Error should not happen. Website should continue to be responsive.

**Screenshots**

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at passStringToWasm0 (penguin_nurse-b10dc3314abaed41.js:1:20102)
    at imports.wbg.__wbg_key_7b5c6cb539be8e13 (penguin_nurse-b10dc3314abaed41.js:2:20109)
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x2cd12e
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x2c5020
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x25dcdb
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x21d362
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x185cb
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x113ab6
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x2e4f28
    at penguin_nurse_bg-7128d0be0aa2f7e8.wasm:0x2e9014
```

The code at that point:

```javascript
function passStringToWasm0(arg, malloc, realloc) {
    if (realloc === undefined) {
        const buf = cachedTextEncoder.encode(arg);
        const ptr = malloc(buf.length, 1) >>> 0;
        getUint8ArrayMemory0().subarray(ptr, ptr + buf.length).set(buf);
        WASM_VECTOR_LEN = buf.length;
        return ptr;
    }
    let len = arg.length;
    let ptr = malloc(len, 1) >>> 0;
    const mem = getUint8ArrayMemory0();
    let offset = 0;
[...]

PR #5379, opened 2026-03-16
Fix autofill event downcast - ignore events of the wrong type

ealmloff:
autofill events don't show up as the right type of event. this pr adds checks for event types and ignores events we don't recognize instead of panicing

It also introduces a unified source for events and event types which deduplicates some code

Fixes https://github.com/DioxusLabs/dioxus/issues/4079
Fixes https://github.com/DioxusLabs/dioxus/issues/5375

### DIFF
--- a/packages/html/src/events/mod.rs
+++ b/packages/html/src/events/mod.rs
@@ -21,9 +21,6 @@ macro_rules! impl_event {
             /// </details>
             ///
             #[doc = include_str!("../../docs/common_event_handler_errors.md")]
-            $(
-                #[doc(alias = $js_name)]
-            )?
             #[inline]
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I will say, glad this fixes the bug, but I don't think the code for the html module really got any cleaner. I think we should look into using proper Rust codegen (build.rs -> html.rs) in a way that's easy to expand for other element systems like TUI / uikit.

acheul:
Hi! I just found out that ResizeEvent and VisibleEvent are not working on web with this patch.
It seems that they fail "$event_name.is_instance_of::<$web_ty>()" condition (in "web/src/events/mod.rs"). I'd like to contribute a fix, but tbh not very sure where to start.

ealmloff:
> Hi! I just found out that ResizeEvent and VisibleEvent are not working on web with this patch. It seems that they fail "$event_name.is_instance_of::<$web_ty>()" condition (in "web/src/events/mod.rs"). I'd like to contribute a fix, but tbh not very sure where to start.

#5483 should fix this

Issue #5371, opened 2026-03-14
dx serve not watching sub-modules


drdozer:
**Problem**

I have a dioxus project with various modules in src, and then a sub-module for charting. Changes to source in the top-level files triggers `dx serve` to rebuild and launch the app. Changes to source in the nested folder does not.

**Expected behavior**

Changes to code in nested modules should trigger a rebuild and relaunch. It does not. Pressing `r` does a manual rebuild and relaunch, pulling in any changes in the nested module.

**Environment:**

- Dioxus version:  0.7.3 (22b06ba)
- Rust version:    1.96.0-nightly (80381278a 2026-03-01)
- OS info:         Devuan excalibur 
- App platform:   linux

**Questionnaire**

I'm interested in fixing this myself but don't know where to start.

PR #5381, opened 2026-03-17
fix(cli): watch all workspace members in dx serve

BillionClaw:
`dx serve` only watched crates that were direct dependencies of the main crate. Workspace members not in the dependency graph were ignored, so changes to sub-modules didn't trigger rebuilds.\n\nNow `all_watched_crates()` also iterates over all workspace members and adds their directories to the watch list.\n\nFixes #5371

### DIFF
--- a/packages/cli/src/serve/runner.rs
+++ b/packages/cli/src/serve/runner.rs
@@ -1173,6 +1173,30 @@ impl AppServer {
             krates.extend(server_krates);
         }
 
+        // Also add all workspace members that are local (not from registry/git)
+        // This ensures that sub-modules (workspace members not directly depended on) are watched
+        for member in self.workspace.krates.workspace_members() {
+            if let krates::Node::Krate { krate, .. } = member {
+                // Skip if the crate is from the registry or git - we only want local path dependencies
+                // and workspace members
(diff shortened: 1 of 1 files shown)


packages/cli/src/serve/runner.rs
```diff
@@ -1173,6 +1173,30 @@ impl AppServer {
             krates.extend(server_krates);
         }
 
+        // Also add all workspace members that are local (not from registry/git)
+        // This ensures that sub-modules (workspace members not directly depended on) are watched
+        for member in self.workspace.krates.workspace_members() {
+            if let krates::Node::Krate { krate, .. } = member {
+                // Skip if the crate is from the registry or git - we only want local path dependencies
+                // and workspace members
+                if krate
```
ealmloff, line 1182: when would workspace members be inside of the .cargo/git/registry directory?

ealmloff:
appears to be a bot. newlines escaped in the description, comments on https://github.com/DioxusLabs/dioxus/issues/5403, ~500 pull requests in the last month to 100 different repos

For local dependencies we can read cargo's generated dependency files and watch them instead of just workspace dependencies

PR #5382, opened 2026-03-17
fix(cli): prevent dx build stall on macOS arm64 with Tailwind

### LINKED ISSUE #5369: dx build / dx serve can stall on macOS arm64 when automatic Tailwind is enabled
Hi team — thank you for maintaining Dioxus.

**Problem**

On macOS arm64, dx build / dx serve can stall when automatic Tailwind integration is enabled.

In the repro, the last visible log line is where dx spawns the managed Tailwind binary for the default root-level tailwind.css. After that, the build does not continue into the normal Cargo/WebAssembly build flow.

I reproduced this with a fresh app outside the Dioxus monorepo as well, so it does not appear to depend on the Dioxus workspace itself.

I have only reproduced this on macOS arm64 so far.

**Steps To Reproduce**

Create a fresh Dioxus app with Tailwind enabled.
From the project root, run dx build --platform web --verbose or dx serve --platform web --verbose.
Observe that the last visible log line is the Tailwind spawn, and the build does not continue.
Example last log line:
```text
0.290s DEBUG dx::tailwind: Spawning tailwindcss@v4.1.5 with args: ["~/.dx/tailwind/tailwindcss-v4.1.5", "--input", "<project>/tailwind.css", "--output", "<project>/assets/tailwind.css", "--watch"]
```
**Expected behavior**

dx should run the automatic Tailwind step and then continue into the normal web build process.

**Screenshots**

Not included. The most relevant artifact is the verbose log output above.

**Environment:**
```text
Dioxus version: 0.7.3
Rust version: rustc 1.93.1 (01f6ddf75 2026-02-11)
OS info: macOS arm64
App platform: web
```
**Questionnaire**

I would like to help fix the macOS case.

BillionClaw:
## Summary

Fixes #5369 - `dx build` and `dx serve` would stall on macOS arm64 when automatic Tailwind integration is enabled.

## Root Cause

In `TailwindCli::serve()`, the Tailwind v4 process was spawned with piped stdin. On macOS arm64, the Tailwind process in watch mode blocks waiting for stdin. The original code:

```rust
let stdin = proc.stdin.take();
proc.wait().await?;
drop(stdin);
```

This caused a deadlock because `wait()` waits for the process to exit, but the Tailwind process waits for stdin to close. Since stdin was only dropped after `wait()` returned, neither could proceed.

## Fix

Drop stdin BEFORE calling `wait()`:

```rust
drop(proc.stdin.take());
proc.wait().await?;
```

## Testing

Added unit tests to verify the autodetect logic and document the fix.

Fixes #5369

### DIFF
--- a/packages/cli/src/tailwind.rs
+++ b/packages/cli/src/tailwind.rs
@@ -66,9 +66,10 @@ impl TailwindCli {
             // the tw watcher blocks on stdin, and `.wait()` will drop stdin
             // unfortunately the tw watcher just deadlocks in this case, so we take the stdin manually
             let mut proc = tailwind.run(&manifest_dir, input_path, output_path, true)?;
-            let stdin = proc.stdin.take();
+            // Drop stdin immediately to prevent the process from blocking on macOS arm64.
+            // Tailwind v4 in watch mode waits for stdin; if we don't close it, wait() never returns.
+            drop(proc.stdin.take());
             proc.wait().await?;
-            drop(stdin);
 
             Ok(())
         })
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
I cannot reproduce the issue this PR is addressing. I also run arm64 macos and tailwind works fine. I suspect if there is a hang somewhere it is not in the tailwind watcher.

PR #5383, opened 2026-03-17
fix(file-explorer): prevent entering files as directories

### LINKED ISSUE #5368: "file_explorer" example does not work
**Problem**

I am trying to get _something_ working, but `dioxus` seems not to work at all for me:

**Steps To Reproduce**

- `dx serve --desktop` inside the `file_explorer` example

**Screenshots**

Image

**Environment:**

- Dioxus version:  main
- Rust version:    1.93.1
- OS info:         NixOS 25.11
- App platform:    desktop

---

Desktop Environment is KDE.

BillionClaw:
Fixes #5368

The file_explorer example had two issues:
1. Used heuristic (contains '.') to detect files vs directories, which is unreliable (directories can contain dots)
2. Tried to enter files as directories when clicked, causing errors

Changes:
- Use `path.is_file()` for accurate file detection in icon rendering
- Use `path.is_dir()` check in `enter_dir()` before attempting to navigate
- Rename shadowed `path` variable to `path_display` for clarity

This ensures clicking on files doesn't cause navigation errors, and files/directories are correctly identified.

### DIFF
--- a/examples/01-app-demos/file-explorer/src/main.rs
+++ b/examples/01-app-demos/file-explorer/src/main.rs
@@ -31,12 +31,13 @@ fn app() -> Element {
                 for (dir_id, path) in files.read().path_names.iter().enumerate() {
                     {
                         let path_end = path.components().next_back().map(|p|p.as_os_str()).unwrap_or(path.as_os_str()).to_string_lossy();
-                        let path = path.display();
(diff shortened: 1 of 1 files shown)


BillionClaw:
Thank you for the review! This is ready to merge when you get a chance.

PR #5388, opened 2026-03-19
fix(router): decode '+' as space in URL query strings

### LINKED ISSUE #5280: Dioxus Router fails to parse `+` in URL query string
**Problem**

The router does not parse `+` as space ` ` in query string. This is because the router will never put the `+` there, only `%20`. But other websites and services will change that `%20` into `+` in the URL when they process URLs, for example facebook products.

**Steps To Reproduce**

1. make app with router with query string route like `?:q`

```rs
#[derive(Debug, Clone, Routable, PartialEq)]
#[rustfmt::skip]
pub enum Route {
    #[layout(Navbar)]
    #[route("/?:query")]
    Home {query: String},
}
```

2. navigate to page with space in q, for example query=`Shis Kebap` = https://dj-vaslui.alt-f4.ro/?query=Shis%20Kebap

3. copy/paste above URL into facebook messenger
4. click the processed URL from facebook messenger
5. Zuck will do MITM redirects on the URL and change it into https://dj-vaslui.alt-f4.ro/?query=Shis+Kebap and thus save on 2 bytes
6. dioxus router fails to parse the `+` as ` ` as per the URL spec, and my query string is now query=`Shis+Kebap`

**Expected behavior**

Router should parse `+` as ` ` as per URL spec, even if it never put the `+` there itself.

https://en.wikipedia.org/wiki/Query_string#URL_encoding

**Screenshots**

---

before zuck

Image

---

after zuck

Image

---

**Environment:**

- Dioxus version:  0.7.3
- Rust version: 1.93
- OS info:        debian 11 on docker
- App platform:   web

**Questionnaire**

mango766:
## Summary

Fixes #5280

The Dioxus router currently uses `percent_decode_str` to decode query strings, but this only handles `%XX`-style percent encoding. Per the [application/x-www-form-urlencoded](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) spec, `+` is a valid encoding for space in query strings. Some services (e.g. Facebook Messenger) rewrite `%20` to `+` when processing URLs, causing the router to pass through the literal `+` character.

This PR replaces `+` with space in the query string before percent-decoding, matching the expected URL behavior.

### Changes

- **`packages/router-macro/src/lib.rs`**: In the generated `FromStr` implementation, replace `+` with space in the query string before calling `percent_decode_str`
- **`packages/router/tests/parsing.rs`**: Added `query_plus_decoded_as_space` test verifying `+` is decoded as space, `%20` still works, and multiple `+` signs are handled

### Example

```
Before: /?query=Shis+Kebap → query = "Shis+Kebap" ❌
After:  /?query=Shis+Kebap → query = "Shis Kebap" ✅
```

### DIFF
--- a/packages/router-macro/src/lib.rs
+++ b/packages/router-macro/src/lib.rs
@@ -532,9 +532,14 @@ impl RouteEnum {
                     // Remove any trailing slashes. We parse /route/ and /route in the same way
                     // Note: we don't use trim because it includes more code
                     let route = route.strip_suffix('/').unwrap_or(route);
-                    let query = dioxus_router::exports::percent_encoding::percent_decode_str(query)
+                    // Replace '+' with space in query strings before percent-decoding.
+                    // Per the application/x-www-form-urlencoded spec, '+' is an
+                    // alternative encoding for space in query strings. Some services
+                    // (e.g. Facebook) rewrite '%20' to '+' when processing URLs.
+                    let query_replaced = query.replace('+', " ");
+                    let query = dioxus_router::exports::percent_encoding::percent_decode_str(&query_replaced)
                         .decode_utf8()
-                        .unwrap_or(query.into());
+                        .unwrap_or(query_replaced.into());
                     let hash = dioxus_router::exports::percent_encoding::percent_decode_str(hash)
                         .decode_utf8()
                         .unwrap_or(hash.into());
--- a/packages/router/tests/parsing.rs
+++ b/packages/router/tests/parsing.rs
@@ -162,3 +162,47 @@ fn optional_query_segments_parse() {
         route_without_query_and_other
     );
 }
+
+// Regression test for https://github.com/DioxusLabs/dioxus/issues/5280
+// '+' in query strings should be decoded as space per the
+// application/x-www-form-urlencoded spec.
+#[test]
+fn query_plus_decoded_as_space() {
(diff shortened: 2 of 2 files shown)


packages/router-macro/src/lib.rs
```diff
@@ -532,9 +532,14 @@ impl RouteEnum {
                     // Remove any trailing slashes. We parse /route/ and /route in the same way
                     // Note: we don't use trim because it includes more code
                     let route = route.strip_suffix('/').unwrap_or(route);
-                    let query = dioxus_router::exports::percent_encoding::percent_decode_str(query)
+                    // Replace '+' with space in query strings before percent-decoding.
+                    // Per the application/x-www-form-urlencoded spec, '+' is an
+                    // alternative encoding for space in query strings. Some services
+                    // (e.g. Facebook) rewrite '%20' to '+' when processing URLs.
+                    let query_replaced = query.replace('+', " ");
```
nerjs, line 539: I'm not sure this actually solves the problem and it might even introduce a new one.

According to [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-2.2), `+` is a valid literal character in the query component and has no special meaning by default. Forcing `application/x-www-form-urlencoded` semantics here changes that behavior globally.

I don't think it's a good idea to strictly enforce this interpretation based on behavior of a few services.

PR #5390, opened 2026-03-19
feat(desktop): expose navigation_handler in Config

FreeSynergy:
## Summary

The navigation handler in `dioxus-desktop` is currently hardcoded in `webview.rs`:

```rust
.with_navigation_handler(|var| {
    if var.starts_with("dioxus://") || var.starts_with("http://dioxus.") {
        true
    } else {
        if var.starts_with("http://") || var.starts_with("https://") {
            _ = webbrowser::open(&var);
        }
        false
    }
})
```

This unconditionally opens every external `http`/`https` URL in the system browser and blocks it in the WebView. Apps that intentionally embed web content (e.g. via `<iframe>`) have no way to opt out of this behavior or to allow specific URLs.

## Changes

- **`config.rs`**: adds `navigation_handler: Option<Box<dyn Fn(String) -> bool + 'static>>` field to `Config`, initialized to `None`.
- **`config.rs`**: adds `pub fn with_navigation_handler(mut self, handler: impl Fn(String) -> bool + 'static) -> Self` builder method with full doc-comment and example.
- **`webview.rs`**: if `cfg.navigation_handler` is `Some`, the closure delegates entirely to the user-supplied handler; otherwise the existing hardcoded behavior is preserved exactly.

## Non-breaking

When `with_navigation_handler` is not called (the default), behavior is **identical** to before — no existing apps are affected.

## Example

```rust
let cfg = Config::new()
    .with_navigation_handler(|url| {
        // Allow dioxus internal URLs and a trusted external origin
        url.starts_with("dioxus://") || url.starts_with("https://trusted.example.com")
    });
```

### DIFF
--- a/packages/desktop/src/config.rs
+++ b/packages/desktop/src/config.rs
@@ -63,6 +63,7 @@ pub struct Config {
     pub(crate) background_color: Option<(u8, u8, u8, u8)>,
     pub(crate) last_window_close_behavior: WindowCloseBehaviour,
     pub(crate) custom_event_handler: Option<CustomEventHandler>,
+    pub(crate) navigation_handler: Option<Box<dyn Fn(String) -> bool + 'static>>,
 }
 
 impl LaunchConfig for Config {}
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
We added this to 0.7.8 release, however it doesn't currently allow you to intercept non-https links due to a separate issue. I'm gonna close this PR since that fix is a bit more complex. Thanks for the PR anyhow!

PR #5406, opened 2026-03-23
docs: clarify difference between use_context and consume_context

### LINKED ISSUE #3862: Document the difference between `use_context` and `consume_context`
I'm currently learning Dioxus, but I often feel that the documentation regarding the differences in similar features is insufficient.

I've read the following pages, but they don't explain the difference between `use_context` and `consume_context`.

- https://dioxuslabs.com/learn/0.6/guide/state#global-state-with-context
- https://docs.rs/dioxus/latest/dioxus/prelude/fn.use_context.html
- https://docs.rs/dioxus/latest/dioxus/prelude/fn.consume_context.html

Could you document the difference between them and when to use which?

costajohnt:
## Summary

- Document the key distinction between `use_context` and `consume_context`:
  - `use_context` is a **hook** that captures the context value once on first render and caches it, following the rules of hooks
  - `consume_context` is **not a hook** — it reads directly from the runtime each time and can be called from anywhere (event handlers, async tasks, spawned futures)
- Add cross-references between the two functions so users can discover the alternative

## Test plan

- [x] `cargo check -p dioxus-core -p dioxus-hooks` passes
- [x] Verify rustdoc renders correctly

Fixes #3862

### DIFF
--- a/packages/core/src/global_context.rs
+++ b/packages/core/src/global_context.rs
@@ -32,20 +32,20 @@ pub fn throw_error(error: impl Into<CapturedError> + 'static) {
 
 /// Try to consume context from the current scope, returning `None` if not found.
 ///
-/// Unlike [`use_context`](dioxus_hooks::use_context), this is **not** a hook. It reads the context
-/// value directly from the runtime each time it is called, rather than caching it on first render.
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thanks, this is more clear. Looks good once ci passes

PR #5417, opened 2026-03-26
fix(core): drain scheduler events after RerunScope in render_immediate

### LINKED ISSUE #5415: render_immediate doesn't pick up signal writes from scope re-runs
## Problem

When a child component writes to a parent's signal during render, the parent doesn't re-render in the same `render_immediate` pass. The dirty notification sits in the scheduler channel because `queue_events()` is never called after `RerunScope`.

This is invisible in desktop/web renderers because they run continuous event loops — the missed dirty scope gets picked up next frame. But in event-driven renderers (TUI, SSR, testing), calling `render_immediate` once per event leaves stale content on screen.

## Reproduction

```rust
fn parent() -> Element {
    let val: Signal<String> = consume_context();
    rsx! { "{val}" ChildWriter {} }
}

#[component]
fn ChildWriter() -> Element {
    let mut val: Signal<String> = consume_context();
    let trigger: Signal<bool> = consume_context();
    if *trigger.read() {
        val.set("updated".to_string());  // writes to parent's signal
    }
    rsx! {}
}

// After triggering:
// render_immediate #1 → child runs, writes signal, parent does NOT re-run
// render_immediate #2 → parent runs, picks up "updated"
```

## Root cause

In `virtual_dom.rs`, `render_immediate` calls `queue_events()` after `PollTask` but not after `RerunScope`:

```rust
while let Some(work) = self.pop_work() {
    match work {
        Work::PollTask(task) => {
            _ = self.runtime.handle_task_wakeup(task);
            self.queue_events();  // ← drains channel here
        }
        Work::RerunScope(scope) => {
            // ...runs scope, signal writes go to channel...
            // ← but never drained here
        }
    }
}
```

`render_suspense_immediate` doesn't have this gap — it calls `queue_events()` after every work item.

## Attempted fix

Adding `queue_events()` (or a selective `Immediate`-only drain) after the match block fixes the signal propagation but changes mutation ordering within a pass, breaking `toggle_suspense` and `nested_suspense_resolves_client` tests. Those tests assert on exact mutation sequences that depend on the current ordering.

## Current workaround

Calling `render_immediate` twice:

```rust
vdom.render_immediate(&mut tree);
vdom.render_immediate(&mut tree); // picks up cascading signal writes
```

This works but is wasteful. Happy to help with a proper fix if someone can advise on the right approach that preserves suspense ordering.

rlch:
Fixes #5415

Moves `self.queue_events()` from inside the `PollTask` arm to after the match block in `render_immediate`, matching the pattern already used in `render_suspense_immediate`. Without this, signal writes during scope re-runs aren't picked up in the same render pass.


rlch:
Sorry, accidentally including changes from my other branch. Will resolve!

rlch:
Closing this — the simple queue_events move breaks more than just the two suspense tests. My TUI app also breaks with it, likely due to render ordering changes when signals are drained mid-pass.

The issue still stands though — signal writes during scope re-runs aren't picked up in the same render_immediate pass.

PR #5430, opened 2026-03-31
feat(core): add wakeup callback for external event loop integration

rlch:
## Summary

Adds `VirtualDom::set_wakeup_callback()` — a hook that fires when async tasks complete, allowing external event loops to schedule a repaint without polling.

**The problem:** Frameworks that drive the VirtualDom synchronously (calling `render_immediate()` in response to input events) have no way to know when async work needs processing. A `use_resource` fetch that completes while the user isn't interacting sits in the VirtualDom's queue until the next input event. This causes visible delays — e.g., images only appearing when the user hovers over the UI.

**The solution:** `set_wakeup_callback` registers a `Send + Sync` callback on the runtime that gets invoked from `ArcWake::wake_by_ref` alongside the existing channel send. External event loops use this to schedule a `render_immediate` call:

```rust
// winit example
let proxy = event_loop.create_proxy();
dom.set_wakeup_callback(move || {
    let _ = proxy.send_event(UserEvent::NewWork);
});
```

This follows the universal pattern used by every major GUI framework (Flutter's `scheduleFrame`, egui's `request_repaint`, Iced's `EventLoopProxy`, Qt's `postEvent`, GTK's `g_idle_add`).

**Changes:**
- `Runtime`: new `wakeup_callback: RefCell<Option<Arc<dyn Fn() + Send + Sync>>>` field
- `LocalTaskHandle`: carries the callback, invokes it in `wake_by_ref`
- `VirtualDom::set_wakeup_callback()`: public API to set the callback
- 2 new tests in `packages/core/tests/wakeup_callback.rs`

## Test plan

- [x] All existing dioxus-core tests pass (71 passed, 7 ignored)
- [x] New test: callback fires when spawned async task completes
- [x] New test: callback does NOT fire when no async work is spawned
- [ ] No breaking changes — fully backwards compatible (callback is optional, defaults to None)

### DIFF
--- a/packages/core/src/runtime.rs
+++ b/packages/core/src/runtime.rs
@@ -18,6 +18,7 @@ use slab::Slab;
 use slotmap::DefaultKey;
 use std::any::Any;
 use std::collections::BTreeSet;
+use std::sync::Arc;
 use std::{
     cell::{Cell, Ref, RefCell},
     rc::Rc,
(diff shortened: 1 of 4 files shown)


ealmloff:
The `wait_for_work` method on the virtual dom will poll any futures that have woken up and return when a scope needs to be rerun. We use this with a custom waker to integrate with the window event loop here in desktop: https://github.com/DioxusLabs/dioxus/blob/45442af41d50b74d51304e7736b7b15fd35364dd/packages/desktop/src/webview.rs#L539

jkelleyrtp:
Closed this because Evan's point is right - `poll_vdom` shows how to integrate the virtualdom wait_for_work with a custom waker setup. This PR's approach tries to add a callback to the runtime, but the custom waker system is our blessed approach

PR #5442, opened 2026-04-02
Add support for aarch64 Windows platform in wasm_opt

hadeutscher:
Heyo

`dx bundle --release` fails on my Surface Laptop; please accept this fix :)

Also, is it possible to manually supply wasm-opt? I put it in path but `dx` still tries to download it.


jkelleyrtp:
> Also, is it possible to manually supply wasm-opt? I put it in path but dx still tries to download it.

We have a `no-downloads` feature flag for the CLI which will defer to your system tools, but AFAIK I don't think there is a way to manually configure tools like wasm-opt without sym-linking within the `.dx` folder.

PR #5446, opened 2026-04-03
Autodetect android location

ealmloff:
Try to auto detect android install location when the environment variable isn't set. Tested with the android studio install on:
- [x] Windows
- [x] Macos
- [ ] Linux

### DIFF
--- a/packages/cli/src/build/tools.rs
+++ b/packages/cli/src/build/tools.rs
@@ -1,7 +1,10 @@
 use crate::Result;
 use anyhow::Context;
 use itertools::Itertools;
-use std::{path::PathBuf, sync::Arc};
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 use target_lexicon::{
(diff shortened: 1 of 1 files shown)


packages/cli/src/build/tools.rs
```diff
@@ -19,31 +22,13 @@ pub(crate) struct AndroidTools {
...
-        .or_else(|| var_or_debug("ANDROID_SDK"))
-        .or_else(|| var_or_debug("ANDROID_HOME"));
+    let sdk_from_env = android_sdk_from_env();
+    let ndk_from_env = android_ndk_from_env();
 
-    // Check the ndk. We look for users's overrides first and then look into the SDK.
-    // Sometimes users set only the NDK (especially if they're somewhat advanced) so we need to look for it manually
-    //
-    // Might look like this, typically under "sdk":
-    // "/Users/jonkelley/Library/Android/sdk/ndk/25.2.9519653/toolchains/llvm/prebuilt/darwin-x86_64/bin/aarch64-linux-android24-clang"
```
nicoburns, line 32: Woudl be nice to keep the example path

PR #5466, opened 2026-04-08
feat(cli): add ios.frameworks for embedding dylibs/frameworks

voxelbee:
## Summary

Adds `[ios] frameworks` paralleling `[macos] frameworks`, for embedding `.framework` directories or bare `.dylib` files into the iOS `.app` bundle. This closes the gap that forced users wrapping a C/C++ runtime (llama.cpp, whisper.cpp, ONNX Runtime, LiteRT-LM, etc.) through a `-sys` crate to write post-build shell scripts to get their vendored dylibs into `App.app/Frameworks/`.

The existing iOS build already injects `-Wl,-rpath,@executable_path/Frameworks` at link time, so the rpath resolution story is already solved — the only missing piece was the copy + codesign step.

## Motivation

I ran into this shipping a Dioxus iOS app that uses Google's [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM) runtime for on-device LLM inference. LiteRT-LM ships two prebuilt dylibs alongside its static archive:

- `libLiteRtMetalAccelerator.dylib` — the Metal GPU backend
- `libGemmaModelConstraintProvider.dylib` — the Gemma model constraint provider

Both are referenced by the main static archive via `@rpath/` and must live inside `App.app/Frameworks/` at launch or dyld fails with "image not found" before any Rust code runs. `[macos] frameworks` handles exactly this case on desktop; there was no counterpart for iOS.

## Changes

- **`IosConfig::frameworks: Vec<String>`** — new manifest field mirroring `MacosConfig::frameworks`. Schema regenerated via `dx config schema`.
- **`copy_framework` promoted** from a private helper in `bundler/macos.rs` to `pub(crate)` in `bundler/mod.rs`, next to `copy_dir_recursive`. Handles both `.framework` directories and bare `.dylib` files.
- **`write_frameworks` in `build/request.rs`** — after the existing link-args loop (which only catches dylibs that happen to appear as full paths on the rustc link line), walks the config's platform-appropriate frameworks list and copies each entry. Resolves relative paths against `crate_dir()`. Missing paths are tolerated, matching existing macOS semantics for system framework references.
- **`codesign_apple` in `build/request.rs`** — signs embedded frameworks individually (deepest-first) before signing the main bundle. This is the order Apple's `codesign` requires: nested Mach-O files must carry their own signature before the container bundle's signature is valid. The per-framework codesign invocation doesn't pass `--entitlements` since those only apply to the main executable. This is also a latent bug fix for any macOS or iOS app that was already placing libraries under `Frameworks/` by other means — the current code would fail `codesign --verify --deep --strict` on such an app.

## Backward compatibility

`ios.frameworks` defaults to an empty `Vec` via `#[serde(default)]`. Projects that don't set the field see identical behavior. No breaking changes to `MacosConfig` — `copy_framework` was only moved, not modified.

## Test plan

- [x] `cargo fmt --all -- --check` (pinned rustc 1.94.0 matching CI)
- [x] `cargo check -p dioxus-cli --tests --examples`
- [x] `cargo clippy -p dioxus-cli --tests -- -D warnings`
- [x] `cargo test -p dioxus-cli --tests`
- [x] End-to-end: real Dioxus app built for iOS simulator, embeds two vendored LiteRT-LM dylibs, launches successfully, FFI calls into LiteRT-LM return from Rust code at runtime

## Follow-ups (out of scope)

- `write_frameworks` could also auto-discover dylibs referenced via `cargo:rustc-link-lib=dylib=NAME` by walking `-L` search paths for a matching `libNAME.dylib`. That would let most `-sys` crates get free embedding without any config. Leaving that as a future PR — this one matches the existing explicit `[macos] frameworks` convention to keep the scope focused.

### DIFF
--- a/packages/cli/schema.json
+++ b/packages/cli/schema.json
@@ -972,6 +972,13 @@
             "$ref": "#/definitions/IosTypeIdentifier"
           }
         },
+        "frameworks": {
+          "description": "Frameworks to embed.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "icon": {
           "description": "Icons for the app. Overrides `bundle.icon` for iOS builds.",
           "type": [
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -971,6 +971,17 @@ impl BuildRequest {
             );
         }
 
+        // Apple bundles route through dx-as-linker so we can capture resolved -l/-L
+        // flags and bundle -sys crate dylibs into Frameworks/ (forwards to `cc`).
+        if custom_linker.is_none()
+            && matches!(
+                triple.operating_system,
+                OperatingSystem::IOS(_) | OperatingSystem::Darwin(_) | OperatingSystem::MacOSX(_)
+            )
+        {
+            custom_linker = Some(PathBuf::from("cc"));
+        }
+
         let target_dir = std::env::var("CARGO_TARGET_DIR")
             .ok()
             .map(PathBuf::from)
@@ -1057,6 +1068,7 @@ impl BuildRequest {
         let cache_dir = self.session_cache_dir();
         _ = std::fs::create_dir_all(&cache_dir);
         _ = std::fs::create_dir_all(self.rustc_wrapper_args_dir());
+        _ = std::fs::create_dir_all(self.link_args_file().parent().unwrap());
         _ = std::fs::File::create_new(self.link_err_file());
         _ = std::fs::File::create_new(self.link_args_file());
         _ = std::fs::File::create_new(self.windows_command_file());
@@ -1376,16 +1388,17 @@ impl BuildRequest {
             }
         }
 
-        // Collect the linker args and attach them to the tip crate's bin entry
+        // Always attach the persisted linker args, even on cached cargo builds where
+        // the rustc wrapper didn't run and the bin entry doesn't exist yet.
         let tip_crate_name = self.tip_crate_name();
         let tip_bin_key = format!("{tip_crate_name}.bin");
-        if let Some(tip_args) = workspace_rustc_args.get_mut(&tip_bin_key) {
-            tip_args.link_args = std::fs::read_to_string(self.link_args_file())
-                .context("Failed to read link args from file")?
-                .lines()
-                .map(|s| s.to_string())
-                .collect();
-        }
+        let link_args: Vec<String> = std::fs::read_to_string(self.link_args_file())
+            .map(|s| s.lines().map(String::from).collect())
+            .unwrap_or_default();
+        workspace_rustc_args
+            .entry(tip_bin_key.clone())
+            .or_default()
+            .link_args = link_args;
 
         let exe = output_location.context("Cargo build failed - no output location. Toggle tracing mode (press `t`) for more information.")?;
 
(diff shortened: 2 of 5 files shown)


jkelleyrtp:
Thanks for the PR! Something I'm curious about - does the built-in build.rs + our automatic copying not adequately solve this?

---

In your `build.rs`, use `cargo:rustc-link-lib` and `cargo:rustc-link-search` directives printed to stdout:

**Linking a dynamic library (dylib):**

```rust
fn main() {
    // Add a search path for the linker
    println!("cargo:rustc-link-search=native=/path/to/lib");

    // Link against libfoo.so (or libfoo.dylib on macOS)
    println!("cargo:rustc-link-lib=dylib=foo");
}
```

**Linking a macOS framework:**

```rust
fn main() {
    println!("cargo:rustc-link-lib=framework=CoreFoundation");
}
```

**The link kinds you can specify:**

- `dylib=name` — dynamic library (default if you omit the kind)
- `static=name` — static library
- `framework=name` — macOS framework
- `raw-dylib=name` — Windows raw dylib (niche)

**A few practical notes:**

- The `name` should omit the `lib` prefix and file extension, so for `libssl.so` you'd write `dylib=ssl`.
- `cargo:rustc-link-search=native=/some/path` tells the linker where to find the library. You can also use `framework=/some/path` as the search kind for framework search paths.
- If you just write `println!("cargo:rustc-link-lib=foo");` without a kind, Cargo defaults to `dylib`.
- You can use the `pkg-config` or `cc` crates to automate finding libraries and emitting these flags for you, which is often cleaner for real-world projects.

voxelbee:
I am using the `build.rs` in my crate as you explained above (it is a library crate though not the root binary package). The build works fine it's just when you package and try to run on the simulator for iOS it can't find the dylib's that we have specified as they don't exist (the don't seem to have been copied to the Frameworks folder). I think this is because it's coming from a build file in a library crate not the root binary crate.

There also seems to be another issue that the signing of the libraries is skipped so this will not work when bundled. The macOS frameworks configuration does this correctly we just need the same code for iOS. Which currently that is what this pull request does and right now if you specify the libraries in the frameworks setting for iOS it works and bundles / signs for a device correctly.

But maybe there is a better way to set this up so it does automatically copy by reading the cargo build output like you suggested. That would probably be better as you need to link the right dylib's built for `aarch64-apple-ios` or `aarch64-apple-ios-sim` and that's best decided in the `build.rs`. Whereas right now with the iOS frameworks config that is a bit ambiguous as you can only specify for any iOS not sim or device.

It seems right now that the automatic copying doesn't work properly for iOS. I can look into how we can get that working properly for iOS by looking into fixing the code that exists right now?

jkelleyrtp:
> It seems right now that the automatic copying doesn't work properly for iOS. I can look into how we can get that working properly for iOS by looking into fixing the code that exists right now?

If you can, that'd be great! We have the `copy_frameworks` step - it sounds like we're just missing the inspection for the specific flag? I believe because we see the link args for the final crate, we should be able to see everything, but there is a chance the linker is running before the final link we don't see it.

Also, yes, you're right about code signing for framework folders / plugins, but I *believe* code signing is recursive, meaning if you sign the outer library the inner ones get signed too. Again, could also be wrong.

voxelbee:
> If you can, that'd be great! We have the `copy_frameworks` step - it sounds like we're just missing the inspection for the specific flag?

Took a look and pushed a follow-up. With it, the `[ios] frameworks` config from the original commit becomes redundant for the common `-sys` crate case — projects can go back to plain `cargo:rustc-link-lib=dylib=Foo` + `cargo:rustc-link-search=native=...` and dx picks up the dylibs on its own. We could drop the framework settings for iOS if we wanted then and just have the fixes for the auto copying?

**What was actually broken**

`write_frameworks` does iterate `direct_rustc.link_args`, but two things conspired against it:

1. The dx linker wrapper wasn't installed for Apple Base bundles, so `link_args.json` was always empty for `dx bundle`. Works for `dx serve` Fat mode (which installs its own linker), but iOS/macOS Base had no equivalent.
2. The existing scan only matched entries literally ending in `.dylib` — i.e. absolute paths on the link line. The wrapper output instead has `-lFoo` + `-L /vendor`, because the resolution from `cargo:rustc-link-lib` happens via rlib metadata at link time and only ever appears on the actual linker invocation.

**What the patch does** (one file, `build/request.rs`)

- Auto-set `custom_linker = Some("cc")` for iOS/macOS bundles, mirroring the Android branch a few lines above. Routes the link through dx's existing transparent wrapper at `cli/link.rs`, which captures argv to `link_args.json` and forwards to `cc`.
- Walk `-l`/`-L` in `write_frameworks` and copy `lib<name>.dylib` into Frameworks/. Anything ld linked from a project-local `-L` is by definition a `-sys` dylib (system libs come from SDK paths that aren't in `-L`), so no path filtering needed.
- Persist `link_args.json` under `target/dx/` so cached `dx bundle` runs (where cargo skips the link and the wrapper doesn't re-run) still see the previous link's args. Otherwise a no-op rebundle would silently produce a broken IPA.

**Code signing**

You were right — `codesign` recursion handles this correctly. 

**Backward compatibility**

The `[ios]/[macos] frameworks` config block in `write_frameworks` is untouched and still works for users who want to declare frameworks explicitly (e.g. `.framework` directories that don't come through `-l`). The `IosConfig::frameworks` field stays in the manifest as the explicit escape hatch.

**Testing**
```
cargo +1.94.0 fmt --all -- --check        ✓
cargo +1.94.0 check -p dioxus-cli --tests --examples   ✓
cargo +1.94.0 clippy -p dioxus-cli --tests -- -D warnings   ✓
cargo +1.94.0 test -p dioxus-cli --tests   ✓ (69 passed)
```

End-to-end on a real Dioxus iOS app linking `libLiteRtMetalAccelerator.dylib` + `libGemmaModelConstraintProvider.dylib` from a `-sys` crate's `vendor/aarch64-apple-ios/`: both dylibs land in `Payload/<App>.app/Frameworks/`, the IPA installs and launches on device, FFI calls into LiteRT-LM return from Rust at runtime.

voxelbee:
I've discovered another issue that happens when running with iOS `dx serve --ios` the dylibs need to be copped in and not using symlinks because `installd` rejects any symlinks inside an iOS app's `Frameworks/`
directory. This didn't happen with the `dx bundle --ios` because that does the proper copy.

Issue #5467, opened 2026-04-08
Allow subsecond to be used in release mode


voidentente:
## Feature Request

Subsecond is currently written in a way that it becomes basically a no-op when debug assertions aren't enabled without a way to disable this behavior. This is unnecessarily restrictive.

The motivation behind this is that debug builds often become excessively huge, making hotpatching take much longer than a second (thus ruining the whole subsecond name thing). Some applications need to be run in release mode to function at an acceptable performance. Having to configure debug mode to resemble release just so subsecond works isn't the point.

## Implement Suggestion

Add a feature `enable_always` which is disabled by default and which disables this behavior. Replace occurrences of

```rs
if !cfg!(debug_assertions) {}
```

with

```rs
if cfg!(all(not(feature = "enable_always"), not(debug_assertions))) {}
```

This is backwards compatible. This feature is directed at end-users so libraries shouldn't set this feature. Then, end users can disable the check for `debug_assertions` by enabling the `enable_always` feature of subsecond.

PR #5469, opened 2026-04-08
Allow subsecond to be used in release mode

voidentente:
Implement #5467

### DIFF
--- a/packages/subsecond/subsecond/Cargo.toml
+++ b/packages/subsecond/subsecond/Cargo.toml
@@ -9,6 +9,10 @@ repository = "https://github.com/DioxusLabs/dioxus/tree/main/packages/subsecond"
 keywords = ["hotpatch", "engine", "subsecond", "dioxus", "hot-reload"]
 readme = "../README.md"
 
+[features]
+default = []
+enable_always = []
+
 [dependencies]
 serde = { workspace = true, features = ["derive"] }
 subsecond-types = { workspace = true }
--- a/packages/subsecond/subsecond/src/lib.rs
+++ b/packages/subsecond/subsecond/src/lib.rs
@@ -247,9 +247,10 @@ use std::{
 ///
 /// However, if you wrap your calling code in a future, you *can* simply drop the future which will
 /// cause `drop` to execute and get something similar to unwinding. Not great if refcells are open.
-pub fn call<O>(mut f: impl FnMut() -> O) -> O {
+pub fn call<O>(f: impl FnMut() -> O) -> O {
     // Only run in debug mode - the rest of this function will dissolve away
-    if !cfg!(debug_assertions) {
+    if cfg!(all(not(feature = "enable_always"), not(debug_assertions))) {
+        let mut f = f;
         return f();
     }
 
@@ -409,7 +410,7 @@ impl<A, M, F: HotFunction<A, M>> HotFn<A, M, F> {
     /// then this function will emit an [`HotFnPanic`] which can be unwrapped and handled by next [`call`]
     /// instance.
     pub fn try_call(&mut self, args: A) -> Result<F::Return, HotFnPanic> {
-        if !cfg!(debug_assertions) {
+        if cfg!(all(not(feature = "enable_always"), not(debug_assertions))) {
             return Ok(self.inner.call_it(args));
         }
 
@@ -456,7 +457,7 @@ impl<A, M, F: HotFunction<A, M>> HotFn<A, M, F> {
         ptr: HotFnPtr,
         args: A,
     ) -> Result<F::Return, HotFnPanic> {
-        if !cfg!(debug_assertions) {
+        if cfg!(all(not(feature = "enable_always"), not(debug_assertions))) {
             return Ok(self.inner.call_it(args));
         }
 


nicoburns:
Have you tested this? Does `subsecond` actually work without debuginfo?

voidentente:
> Have you tested this? Does `subsecond` actually work without debuginfo?

Work? Yes. I'm using it right now, actually. I don't know if subsecond subtly relies on debug info somewhere, though.

voidentente:
Example usage: I have this custom egui runtime on top of wgpu and winit.

```rs
fn main() -> Result<(), EventLoopError> {
    #[cfg(feature = "subsecond")]
    dioxus_devtools::connect_subsecond();

    let initial_window = WindowBuilder {
        attributes: WindowAttributes::default(),
        handler: Box::new(MaybeSubsecond(WindowState::new())),
    };

    start(initial_window)
}

struct MaybeSubsecond<T>(T);

#[cfg(feature = "subsecond")]
impl<T: WindowHandler> WindowHandler for MaybeSubsecond<T> {
    #[inline]
    fn close_requested(&mut self, proxy: &EventLoopProxy<Message>) -> ControlFlow<()> {
        dioxus_devtools::subsecond::call(|| self.0.close_requested(proxy))
    }

    #[inline]
    fn update(&mut self, ui: &mut ide_rt::egui::Ui, proxy: &EventLoopProxy<Message>) {
        dioxus_devtools::subsecond::call(|| self.0.update(ui, proxy))
    }
}

#[cfg(not(feature = "subsecond"))]
impl<T: WindowHandler> WindowHandler for MaybeSubsecond<T> {
    #[inline]
    fn close_requested(&mut self, proxy: &EventLoopProxy<Message>) -> ControlFlow<()> {
        self.0.close_requested(proxy)
    }

    #[inline]
    fn update(&mut self, ui: &mut ide_rt::egui::Ui, proxy: &EventLoopProxy<Message>) {
        self.0.update(ui, proxy);
    }
}

#[derive(Clone)]
struct WindowState {}

impl WindowState {
    fn new() -> Self {
        Self {}
    }
}

impl WindowHandler for WindowState {
    fn close_requested(&mut self, _proxy: &EventLoopProxy<Message>) -> ControlFlow<()> {
        ControlFlow::Break(())
    }

    fn update(&mut self, ui: &mut Ui, proxy: &EventLoopProxy<Message>) {
        CentralPanel::default().show_inside(ui, |ui| {
            if ui.button("Exit").clicked() {
                let _ = proxy.send_event(Message::Exit);
            }

            if ui.button("Clone window").clicked() {
                let _ = proxy.send_event(Message::SpawnWindow(WindowBuilder {
                    attributes: WindowAttributes::default(),
                    handler: Box::new(self.clone()),
                }));
            }
        });
    }
}
```

Hotpatching in debug mode takes 6252ms, most of which is spent on extracting assets. Hotpatching in release mode only takes 1644ms. That's almost 4 times faster.

jkelleyrtp:
Sorry, why not just adjust the opt-level of your app's web-dev / desktop-dev profile?

Also, I'm digging into hotpatch perf right now, what platform/target are you seeing the big timing swings?

voidentente:
> Sorry, why not just adjust the opt-level of your app's web-dev / desktop-dev profile?

Because the hotpatching performance isn't degraded by opt-level but by artifact size. If I have to strip out all debug info from my debug artifact to achieve release mode hotpatching speed that's counterproductive.

And actually, I tried running debug with `opt-level = 3` for dev profile right now, and hotpatching now takes 9154ms. So that's worse somehow.

EDIT: That's an outlier. Hotpatching time effectively doesn't change.

> Also, I'm digging into hotpatch perf right now, what platform/target are you seeing the big timing swings?

I'm on Arch Linux, using mold linker. I only target desktop. Most of the time to hotpatch (about 80%) is spent on "extracting assets".

jkelleyrtp:
> > Sorry, why not just adjust the opt-level of your app's web-dev / desktop-dev profile?
> 
> Because the hotpatching performance isn't degraded by opt-level but by artifact size. If I have to strip out all debug info from my debug artifact to achieve release mode hotpatching speed that's counterproductive.
> 
> And actually, I tried running debug with `opt-level = 3` for dev profile right now, and hotpatching now takes 9154ms. So that's worse somehow.
> 
> EDIT: That's an outlier. Hotpatching time effectively doesn't change.
> 
> > Also, I'm digging into hotpatch perf right now, what platform/target are you seeing the big timing swings?
> 
> I'm on Arch Linux, using mold linker. I only target desktop. Most of the time to hotpatch (about 80%) is spent on "extracting assets".

Opt-level does tend to affect artifact size. Faster builds (opt=3) tend to produce smaller code.

There are other levers involved that make dev different than release. Notably, debug

- debug = "line-tables-only"
- debug = 0

can be used to eliminate debug info while retaining debug_assertions. Do these improve performance at all?

voidentente:
> Opt-level does tend to affect artifact size.

I would actually say the opposite is true: many optimizations like unrolling and inlining actually increases the number of instructions in order to reduce the amount of branches.

> There are other levers involved that make dev different than release. Notably, debug

I'm aware, but that's my "if I have to strip out all debug info from my debug artifact to achieve release mode hotpatching speed that's counterproductive" argument. Even if I set `debug = 0` it's slightly slower:

| Release [57.8 MB] | Debug (opt-level = 3) [408.8 MB] | Debug (opt-level = 3, debug = 0) [62.9 MB] |
|-------------------|----------------------------------|--------------------------------------------|
| 1501ms            | 6170ms                           | 1165ms                                     |
| 1482ms            | 6622ms                           | 1796ms                                     |
| 1713ms            | 5438ms                           | 1745ms                                     |
| 1609ms            | 5959ms                           | 1602ms                                     |
| 1583ms            | 5000ms                           | 2652ms                                     |
| 1617ms            | 2501ms                           | 1277ms                                     |
| 1546ms            | 3277ms                           | 1817ms                                     |
| 1621ms            | 5889ms                           | 1586ms                                     |
| 1538ms            | 4671ms                           | 2202ms                                     |
| 1648ms            | 5651ms                           | 1332ms                                     |
| Avg: 1585.8ms     | Avg: 5117.8ms                    | Avg: 1717.4ms                              |

nicoburns:
@jkelleyrtp I'd also like this ability. I could modify the dev profile, but it's nice to be able to easily switch between profiles, and sometimes I do want the regular (unoptimised, debug info) debug builds. My thinking (assuming that there are no technical blockers) is that it ought to be `dx serve --hotpatch` that enables hotpatching regardless of profile (we could have the CLI enable the feature).

PR #5473, opened 2026-04-09
feat: provide api to customize server address

### LINKED ISSUE #5263: Assigning a port number / bind address when using LaunchBuilder
**Problem**

I could not find a way to set the port number via the launch builder.
There is [https://github.com/DioxusLabs/dioxus/issues/2380#issuecomment-2089103296](https://github.com/DioxusLabs/dioxus/issues/2380#issuecomment-2089103296) but the API has changed since this comment and this no longer works. 
I found it can be done by setting env variables: IP and PORT when looking at:
dioxus-cli-config-0.7.3/src/lib.rs fullstack_address_or_localhost()
as called by
dioxus-cli-config-0.7.3/src/launch.rs serve_server()
As far as I can set tell it is not longer possible to set these programmatically. 
I require to fetch IP and PORT over as a CLI argument and then pass to the dioxus::LaunchBuilder

**Environment:**

- Dioxus version:  0.7.3
- Rust version:    1.92
- OS info:         Debian 13
- App platform:    web fullstack

**Questionnaire**
Is it possible to do this, or is there a workaround for this?

seflue:
Fixes #5263

Before 0.7, `fullstack::Config::new().addr(...)` allowed setting the server address programmatically. This was removed in the 0.7 refactoring, leaving `IP`/`PORT` env vars as the only option. On Rust edition 2024, that means calling `unsafe { std::env::set_var(...) }` before `serve()`.

`serve_at` is a new entry point that takes a `SocketAddr` and a callback, which is just wrapped by `serve()`, so existing behavior is unchanged. It is re-exported as `dioxus::serve_at` and included in the prelude.

### DIFF
--- a/packages/fullstack-server/src/launch.rs
+++ b/packages/fullstack-server/src/launch.rs
@@ -69,14 +69,14 @@ async fn serve_server(
 
     let cb = move || {
         let cfg = cfg.clone();
-        Box::pin(async move {
+        async move {
             Ok(apply_base_path(
                 Router::new().serve_dioxus_application(cfg.clone(), original_root),
                 original_root,
                 cfg.clone(),
                 base_path().map(|s| s.to_string()),
             ))
-        }) as _
+        }
     };
 
     serve_router(cb, dioxus_cli_config::fullstack_address_or_localhost()).await;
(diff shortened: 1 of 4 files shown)


packages/fullstack-server/src/launch.rs
```diff
@@ -111,16 +111,44 @@ pub fn router(app: fn() -> Element) -> Router {
...
+/// let addr = SocketAddr::from(([0, 0, 0, 0], 3000));
+/// dioxus::serve_at(addr, || async {
+///     Ok(dioxus::server::router(app)
+///         .route("/health", axum::routing::get(|| async { "ok" })))
+/// });
+/// ```
+///
 /// This function uses axum to block on serving the application, and will not return.
-pub fn serve<F>(mut serve_it: impl FnMut() -> F) -> !
+pub fn serve_at<F>(addr: SocketAddr, mut serve_it: impl FnMut() -> F) -> !
```
ealmloff: It looks like this has the same api as serve_router except it doesn't require you to box the resulting future? Can we just generalize serve_router to box for you instead of introducing a new public api?

ealmloff (changes requested):

packages/fullstack-server/src/launch.rs
seflue: If I understood you correctly, you'd rather expose serve_router (which was private) than introduce a new public API.

That's what I'm doing now: box the future inside serve_router and expose it (prelude + re-exports) so users can call it directly. I removed serve_at, but anyone who needs a custom bind address can still do it from their own tokio::main using serve_router, so they don't lose hot-reload.

PR #5481, opened 2026-04-13
fix(cli): display user-specified address in serve output

### LINKED ISSUE #5477: dx serve --addr 0.0.0.0 should report http://0.0.0.0:8080 as the address
## Feature Request

I'm accustomed to tools where you can set 0.0.0.0 showing THAT address when you start the dev server. Mine is still showing

> Serving at: http://127.0.0.1:8080

It also seems to redirect to 127 if I manually visit the all 0s version.

I think it would be more intuitive to switch the address to all 0s both in the `dx serve` printout and to NOT redirect when you're making that the address in the browser. Obviously, you can run something like `sudo ss -tunlp` to verify but it's an extra step and I am fundamentally lazy.

costajohnt:
## Summary

Fixes #5477

When running `dx serve --addr 0.0.0.0`, the CLI overrode the displayed address to `127.0.0.1`. This was confusing for users who explicitly chose `0.0.0.0` and expected the output to reflect their choice.

Removed the override in `displayed_address()` and the now-unused `devserver_bind_ip` struct field. The default behavior (showing `127.0.0.1` when no `--addr` flag is given) is unchanged since the default bind IP is already `127.0.0.1`.

## Test plan

- [x] Added unit tests for `displayed_address()` covering both `0.0.0.0` and `127.0.0.1`
- [x] All existing tests pass (`cargo test -p dioxus-cli`)


jkelleyrtp:
The problem with 0.0.0.0 is that it's not a real address. It doesn't make a whole lot of sense to show it.

The best alternative is to display the *bound* address of the socket, ie a 192.x.y.z address. However, it's not exactly clear which network device the socket is bound to and which address is the network address in this situation.

The goal with the displayed address is to provide a clickable link for the end user to open the page in their browser. Unfortunately 0.0.0.0 is not a valid address which makes it less useful to present.

costajohnt:
Good point, the displayed address is basically a clickable link so 0.0.0.0 doesn't help there. What if we resolve the actual bound address when listening on 0.0.0.0? The UDP socket trick (bind 0.0.0.0:0, connect to a public IP, read local_addr) would give us the real 192.x.y.z address without sending traffic. Fall back to 127.0.0.1 if that fails.

Also fine closing this if you'd rather keep things as-is.

jkelleyrtp:
> The UDP socket trick (bind 0.0.0.0:0, connect to a public IP, read local_addr) would give us the real 192.x.y.z address without sending traffic. Fall back to 127.0.0.1 if that fails

huh, didn't realize that was a thing but makes sense! feel free to take a stab at it. we'll want to use an array of ips for fallback esp across regions, but I think this would work!

jkelleyrtp (approved):

jkelleyrtp:
Thanks!

Issue #4426, opened 2025-07-16
using asset! to import build.rs outputs in a crate in a workspace


mwkohout:
**Problem**

app I'm writing that lives in a multi-crate workspace, I've got a build script that generates an asset.  Generally, with build scripts you're encouraged to write the results out to a location pointed to by the env var "OUT_DIR".

OUT_DIR is set by cargo and is a subdirectory in $WORKSPACE_ROOT/target

It seems that asset! doesn't allow users to point at things in $OUT_DIR.  

**Steps To Reproduce**

Create a new workspace based project, and write a build.rs that writes some asset to $OUT_DIR/junk.css.
import it as an asset:
`const JUNK_CSS: Asset = asset!(concat!(env!("OUT_DIR"), "/junk.css"));`

1) asset requires a static str as a parameter.
2) even if I manually expand it to a static str(in my case "../target\debug\build\ui-82050a64fa3cc57c\out\junk.css"), since OUT_DIR is outside the crate(but still in the workspace) it'll fail:
`const JUNK_CSS: Asset = asset!("../target\debug\build\ui-82050a64fa3cc57c\out\junk.css");`
with
```
Asset path C:\dioxus_app_ui\target\debug\build\ui-82050a64fa3cc57c\out\junk.css is invalid. Make sure the asset exists within this crate.
```

here's my build.rs that lives in a "ui" crate:
```
use std::env;
use std::fs;
use std::path::PathBuf;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let out_dir = env::var("OUT_DIR").unwrap();

    let junk_css_path = PathBuf::from(out_dir).join("junk.css");

    println!("cargo:info=junk.css path= {junk_css_path:?}");

    let css_contents = "body { background: pink; }";

    fs::write(&junk_css_path, css_contents)?;

    println!("cargo:rerun-if-changed=build.rs");

    Ok(())
}
```
- 
- 
- 

**Expected behavior**

I'd hope for a way where I could use $OUT_DIR, and also point at a file outside the current crate

**Screenshots**

**Environment:**

- Dioxus version:  0.6.3 
- Rust version:    1.88.0  
- OS info:   Windows 11      
- App platform:  web,desktop, and mobile but I'm using asset here in a shared "ui" crate   

**Questionnaire**

I'm interested in fixing this myself but don't know where to start.  Perhaps another macro in the "asset" family?

cprussin:
[...]

PR #5490, opened 2026-04-15
Allow using generated assets

skreborn:
Currently, all assets are resolved either relative to the current file (if the path begins with a `.`) or the crate root (in all other cases).

## Changes

This PR introduces a new option for asset path resolution that enables reading generated assets.

```rust
const _: Asset = asset!(concat!(env!("OUT_DIR"), "/generated-asset.txt"));
```

The path resolution heuristics are as follows:
- paths beginning with `.` or `..` both resolve relative to the current file;
- paths beginning with the output directory are used as-is;
- all other paths are resolved relative to the crate root (with the leading `/` ignored).

I also updated the documentation of the macros, added some tests for new and existing behavior, and refactored the internal error handling to improve clarity and consistency.

## Alternatives considered

Following are a number of alternatives considered before settling on the current solution.

### Explicit path resolution strategy

Explicitly declaring the resolution base would make the code more explicit.

```rust
asset!(crate "assets/asset.txt"); // Resolved relative to the crate root
asset!(self "../assets/asset.txt"); // Resolved relative to the current file
asset!(out "generated-asset.txt"); // Resolved relative to the output folder
```

Skipping the explicit strategy would fall back to the original behavior by automatically detecting whether `crate` or `self` should be used.

This was ultimately dropped in response to https://github.com/DioxusLabs/dioxus/pull/5490#issuecomment-4263057892 to avoid introducing new syntax.

### Prefix character

A special character (like `"!"`) at the beginning of the path could denote a generated path.

```rust
asset!("!/generated.css");
```

While the likelihood of anyone having a local folder named `"!"` is practically non-existent, it's still _technically_ allowed by the operating system and thus this would constitute a breaking change. On Linux particularly, the only character that's not allowed in a file name is `/`.

### Separate macros

A new `generated_asset` could be introduced so no changes would need to be made to the existing ones. However, this would lead to needless duplication without any practical gains.

```rust
generated_asset!("generated.css");
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8117,6 +8117,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
+ "tempfile",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -291,6 +291,7 @@ rustversion = "1.0.21"
 rand = "0.9"
 longest-increasing-subsequence = "0.1.0"
 trybuild = "1.0.116"
+tempfile = "3.27.0"
 dirs = "6.0.0"
 cargo-config2 = "0.1.34"
 criterion = { version = "0.6" }
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -124,7 +124,7 @@ shell-words = { workspace = true }
 log = { version = "0.4.29", features = ["max_level_off", "release_max_level_off"] }
 
 # link intercept
-tempfile = "3.27.0"
+tempfile = { workspace = true }
 manganis = { workspace = true }
 manganis-core = { workspace = true }
 target-lexicon = { version = "0.13.2", features = ["serde", "serde_support"] }
--- a/packages/manganis/manganis-macro/.cargo/config.toml
+++ b/packages/manganis/manganis-macro/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustdocflags = ['--cfg', 'skip_out_tests']
--- a/packages/manganis/manganis-macro/Cargo.toml
+++ b/packages/manganis/manganis-macro/Cargo.toml
@@ -27,3 +27,4 @@ default = []
 
 [dev-dependencies]
 manganis = { workspace = true }
+tempfile = "3.19.1"
--- a/packages/manganis/manganis-macro/Cargo.toml
+++ b/packages/manganis/manganis-macro/Cargo.toml
@@ -27,4 +27,4 @@ default = []
 
 [dev-dependencies]
 manganis = { workspace = true }
-tempfile = "3.19.1"
+tempfile = { workspace = true }
--- a/packages/manganis/manganis-macro/Cargo.toml
+++ b/packages/manganis/manganis-macro/Cargo.toml
@@ -2,6 +2,7 @@
 name = "manganis-macro"
 version.workspace = true
 edition = "2024"
+rust-version = "1.85.0"
 authors = ["Evan Almloff"]
 description = "Ergonomic, automatic, cross crate asset collection and optimization"
 license = "MIT OR Apache-2.0"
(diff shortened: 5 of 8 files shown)


ealmloff:
We do need to expand the path check to allow the out dir path for assets, but we can do this without extra syntax. We support the same macros inside the path as `include` or `include_str` macros. You can use those to link to a path from the out dir:

```rust
const GENERATED_ASSET: Asset = asset!(concat!(env!("OUT_DIR"), "/file.css"));
```

Linking this pr with #4426

skreborn:
That's what I originally imagined would be possible, but like the linked issue, I was met with resistance.

Making no changes to the macro syntax obviously has an upside—even if the changes are optional.

Checking if an absolute path resolves to an actual file before trying it as a crate-relative one is easy enough, but in case no file is found, it's difficult to provide the user with correct and exact feedback about where the asset is supposed to be, because we can't reasonably tell whether they _meant_ to use an absolute or a crate-relative path.

It's also _technically_ a breaking change because if I happen to have an `/assets/asset.txt` file in my actual file system root, the resolution fails—unless of course the check for whether the absolute path is inside the allowed folders is repeated between those two.

The explicit resolution strategy also allows using `asset!(self "asset.txt")` instead of `asset!("./asset.txt")`, which doesn't seem like a big deal, but it definitely makes the _intent_ clearer and more explicit. Whether that's important enough to the average developer, I don't know; it's mostly a personal preference.

In any case, if you'd like me to head in a different direction, please let me know, and I'll adjust the PR accordingly. I'd be more than happy to implement an alternative solution if that's what you think is best.

ealmloff:
If we resolve assets like this, it seems unlikely there would be much overlap:
```rust
asset!("/assets/asset.txt"); // Resolved relative to the crate root because it starts with slash and isn't point to the dir `OUT_DIR` resolves to in target
asset!("./assets/asset.txt"); // Resolved relative to the current file because it starts with `./`
asset!(concat!(env!("OUT_DIR"), "/generated-asset.txt")); // Resolved relative to the output folder because the file starts with the output folder
```

You would need to have something like `target/{debug|release}/build/<pkg>-<hash>/out` both in your root and the current crate. This could be the case if you have a global target dir for caching, but I assume you would want to link to the version you are building, (`env!("OUT_DIR")`) not the one in the workspace.

```
/
├─ Users/
│  └─ desktop/
│     └─ code/
│        └─ my-lib/
│            ├─ src/
│            │  └─ lib.rs
│            ├─ assets/
│            │  └─ asset.txt
│            ├─ Cargo.toml
│            └─ target/
│                └─ debug/
│                   └─ build/
│                      └─ my-lib-abc123/
│                         └─ out/
│                            └─ generated-asset.txt # This overlaps if you build with `TARGET_DIR=/target`, but intending to point to the lib local target dir.
└─ target/
   └─ debug/
      └─ build/
         └─ my-lib-abc123/
            └─ out/
               └─ generated-asset.txt
```


The restriction about which assets you can link to is intended to prevent builds passing locally, but failing elsewhere like when you publish to crates.io. For libraries, only inside the workspace or build directory should be allowed.

jkelleyrtp:
I think I would prefer something like

```
asset!(concat!(env!("OUT_DIR"), "/generated-asset.txt"));
```

or

```
asset!("{OUT_DIR}/generated-assets.txt")
```

where OUT_DIR and other env vars or hard-coded replacements can be passed through the macro system in a way that the proc macros can see. I lean towards the second approach a bit since it's cleaner and easy to parse. One issue that OUT_DIR for the proc macro might not be the same OUT_DIR for the target crate. I think a dtolnay crate has a solution for that somewhere

skreborn:
@ealmloff @jkelleyrtp Thank you both for the feedback! I've rebased and updated to PR to address your comments. The updated version doesn't introduce any new syntax and still avoids breaking changes (except in the most unrealistic conditions).

ealmloff:
Looks like we may need to canonicalize both paths before checking if it starts with the out dir from the failing tests? Api looks good after that fix

skreborn:
I added canonicalization to the output directory path in the test context. Hopefully that'll do it!

skreborn:
I rebased everything—hoping to push it over the finish line. Is there anything else you'd like me to change?

ealmloff (approved):
Thanks!

PR #5513, opened 2026-04-28
fix(cli): align SSG server address

### LINKED ISSUE #5503: Bug: `--ssg` fails with `--web --fullstack true` — server binary not started before static routes call
## **Problem**

When running `dx build --verbose --trace --web --fullstack true --features fullstack --release --ssg`, the build fails because `dx` attempts to call `http://127.0.0.1:9999/api/static_routes` but no server is running at that address.
```rs
ERROR dx build: Failed to get static routes from server. Make sure you have a server function at the `/api/static_routes` endpoint that returns Vec<String> of static routes.

Caused by:
    0: error sending request for url (http://127.0.0.1:9999/api/static_routes)
    1: client error (Connect)
    2: tcp connect error
    3: Connection refused (os error 111)

Stack backtrace:
   0: <E as anyhow::context::ext::StdError>::ext_context
   1: <tokio::future::maybe_done::MaybeDone<Fut> as core::future::future::Future>::poll
   2: <core::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
   3: dx::main::{{closure}}::{{closure}}::{{closure}}
   4: <core::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
   5: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::future::future::Future>::poll
   6: dx::logging::TraceController::main::{{closure}}
   7: tokio::runtime::park::CachedParkThread::block_on
   8: tokio::runtime::context::runtime::enter_runtime
   9: dx::main
  10: std::sys::backtrace::__rust_begin_short_backtrace
  11: std::rt::lang_start::{{closure}}
  12: std::rt::lang_start_internal
  13: main
  14: __libc_start_call_main
  15: __libc_start_main_alias_2
  16: _start
```

## **Steps To Reproduce**

- create a new dioxus workspace project (crates: app, pages)
- put your router inside pages crate and make it public
- add this to app/src/main.rs
```rs
use dioxus::prelude::*;

#[cfg(not(target_arch = "wasm32"))]
#[cfg(feature = "server")]
#[tokio::main]
async fn main() {
    rustls::crypto::ring::default_provider()
        .install_default()
        .expect("Failed to install ring crypto provider");

    let certs_dir = std::env::var("CERTS_DIR").expect("CERTS_DIR must be set");
    let rustls_config = axum_server::tls_rustls::RustlsConfig::from_pem_file(
        format!("{}/server.crt", certs_dir),
        format!("{}/server.key", certs_dir),
    )
    .await
    .expect("Failed to load TLS config");

    let address = std::net::SocketAddr::from(([0, 0, 0, 0], 8080));

    let router = axum::Router::new().serve_dioxus_application(get_serve_config(), App);

    axum_server::bind_rustls(address, rustls_config)
        .serve(router.into_make_service())
        .await
        .unwrap();
}

#[cfg(target_arch = "wasm32")]
fn main() {
    dioxus::logger::init(tracing::Level::DEBUG).expect("failed to init logger");

    dioxus::LaunchBuilder::new()
        .with_cfg(server_only! { get_serve_config() })
        .launch(App);
}

#[cfg(feature = "fullstack")]
#[server(endpoint = "static_routes", output = server_fn::codec::Json)]
async fn static_routes() -> Result<Vec<String>, ServerFnError> {
    Ok(pages::Route::static_routes()
        .iter()
        .map(ToString::to_string)
        .collect())
}

#[cfg(feature = "server")]
fn get_serve_config() -> ServeConfig {
    ServeConfig::builder()
        .incremental(
            dioxus::server::IncrementalRendererConfig::new()
                .static_dir(static_dir())
                .clear_cache(false),
        )
        .enable_out_of_order_streaming()
}

fn static_dir() -> std::path::PathBuf {
    std::env::current_exe()
        .unwrap()
        .parent()
        .unwrap()
        .join("public")
}

#[component]
fn App() -> Element {
    rsx! {
        Router::<pages::Route> {}
    }
}
```
- run `dx build --verbose --trace --web --fullstack true --features fullstack --release --ssg`

## **Expected behavior**

SSG should work on running `dx build --verbose --trace --web --fullstack true --features fullstack --release --ssg`.

## **Environment:**

- Dioxus version: 0.7.6
- Rust version: 1.95.0
- OS info: NixOS (nixpkgs unstable)
- App platform: web

imwyvern:
Fullstack SSG release builds can bake the server address at compile time, while the prerender step chose the address later at runtime. That left the server listening on its compiled default while `dx build --ssg` queried `127.0.0.1:9999`.

The SSG address is now resolved before the server build, passed into the server build environment, and reused by the prerenderer when it starts and queries the server. Added harness coverage that confirms `dx build --ssg` records an SSG address for fullstack builds.

Verification: `git diff --check`. Could not run cargo tests locally because this environment does not have the Rust toolchain installed.

Fixes #5503

### DIFF
--- a/packages/cli/src/build/request.rs
+++ b/packages/cli/src/build/request.rs
@@ -207,12 +207,12 @@ use cargo_metadata::diagnostic::Diagnostic;
 use cargo_toml::{Profile, Profiles, StripSetting};
 use depinfo::RustcDepInfo;
 use dioxus_cli_config::PRODUCT_NAME_ENV;
-use dioxus_cli_config::{APP_TITLE_ENV, ASSET_ROOT_ENV};
+use dioxus_cli_config::{APP_TITLE_ENV, ASSET_ROOT_ENV, SERVER_IP_ENV, SERVER_PORT_ENV};
 use krates::{NodeId, cm::TargetKind};
 use manganis::BundledAsset;
 use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use serde::Deserialize;
-use std::{borrow::Cow, collections::VecDeque, ffi::OsString};
+use std::{borrow::Cow, collections::VecDeque, ffi::OsString, net::SocketAddr};
 use std::{
     collections::HashSet,
     path::{Path, PathBuf},
@@ -272,6 +272,7 @@ pub(crate) struct BuildRequest {
     pub(crate) session_cache_dir: PathBuf,
     pub(crate) raw_json_diagnostics: bool,
     pub(crate) windows_subsystem: Option<String>,
+    pub(crate) ssg_address: Option<SocketAddr>,
 }
 
 /// dx can produce different "modes" of a build. A "regular" build is a "base" build. The Fat and Thin
@@ -911,6 +912,7 @@ impl BuildRequest {
             apple_team_id: args.apple_team_id.clone(),
             raw_json_diagnostics: args.raw_json_diagnostics,
             windows_subsystem: args.windows_subsystem.clone(),
+            ssg_address: None,
         })
     }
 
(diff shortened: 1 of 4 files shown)


eaves-dropper:
@ealmloff is this pr going to be merged?

packages/cli/src/build/request.rs
```diff
@@ -1900,6 +1902,14 @@ impl BuildRequest {
             ));
         }
 
+        // Release builds read fullstack address configuration at compile time. When SSG runs the
```
ealmloff, line 1905: We shouldn't need to bake the address in at compile time. I don't think we read these, if we do at compile time, we need to fix that issue, not supply them in the build arguments

ealmloff:
The new argument effectively looks like it duplicates the `IP`/`PORT` env vars but only for ssg builds.

> Verification: git diff --check. Could not run cargo tests locally because this environment does not have the Rust toolchain installed.

Assuming this is AI generated and untested. I can take another look if you verify this fixes the original reproduction

PR #5524, opened 2026-05-01
Fix bundling on aarch64 Alpine Linux 

hadeutscher:
Heyo

`dx bundle --release` fails on Alpine Linux on aarch64 CPUs, because we always download the GNU version of wasm-bindgen and it doesn't run on Alpine. Since we already use the musl version for x86_64, I guess we can just switch to musl on aarch64 too.

Also, is it maybe possible to add a musl build of `dioxus-cli` to your CI? I have to compile it from source right now on all my alpine docker builds and it's... not fun :X


jkelleyrtp:
Thanks! Would be nice to distribute a musl cli build as well. We'd need to update our `install.sh` to select musl builds.

PR #5531, opened 2026-05-06
feat:Add opt‑in navigation handler for external URLs

martinmr:
Pretty much the same as #5231, which was closed without comment.

Without this change, iframes will open in an external browser, breaking existing apps and disallowing many legitimate uses like embedding a YouTube video, etc.

One change was made after discussing it with @jkelleyrtp, which was to make all URLs navigable by default when the handler is not set. That way, only users which need finer control over this have to set the handler.

### DIFF
--- a/packages/desktop/src/config.rs
+++ b/packages/desktop/src/config.rs
@@ -20,6 +20,8 @@ type CustomEventHandler = Box<
         ),
 >;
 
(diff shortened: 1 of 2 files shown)


martinmr:
I have tested this in a local app and works as expected.

PR #5181, opened 2026-01-06
Rust Only Builder Pattern: Alternative to RSX Macro

wheregmis:
Super draft, Not all code reviewed yet (Opus Generated), but things just works seamlessly without breaking anything from RSX. Dumping as we have this on roadmap's stretched goals.

PR #5554, opened 2026-05-12
Markerless placeholders/Portals/Multiwindow/const builder/optimized templates

### LINKED ISSUE #628: Portals
## Specific Demand

I'm looking for something like React's Portals, but I couldn't find anything in the docs. Not sure if it's available but under a different name or simply not available yet.

## Implement Suggestion

Basically the same as what React offers: https://reactjs.org/docs/portals.html

### LINKED ISSUE #1627: Typed elements
## Specific Demand

We should allow other libraries to define custom types for attributes that accept AnyAttribute. This would be useful for Freya and viewbuilder.

## Implement Suggestion

We could make attribute definitions something like this:
```rust
mod dioxus_elements {
   struct div;
   impl div {
       const const_#attribute: AttributeDiscription = ...;
       fn dyn_#attribute(cx: &ScopeState, param: i32) {...}
   }
}
```

Where you only include the const version if a string literal is valid for your attribute. 

Then instead of expanding dynamic attributes like this:
```rust
__cx
    .attr(
        dioxus_elements::div::count.0,
        #value,
        dioxus_elements::div::count.1,
        dioxus_elements::div::count.2,
    ),
```

We should expand them like this:
```rust
dioxus_elements::div::dyn_count(__cx, #value)
```

### LINKED ISSUE #2040: Elements should be easier to extend
## Feature Request

To support a number of usecases like resize observer, we should consider changing how we do element attributes and element instantiation. In theory if we used traits, we could allow any third party to extend existing elements with traits.

This would also let us drop the need for the `use dioxus_elements::*` hack we need to do to get attributes in scope.

### LINKED ISSUE #2467: Support prop extensions for callbacks
## Feature Request

It would be nice to be able to create wrapper components that also can pass down callbacks. 
For example, I'm creating a small select wrapper with some tailwind styling and a label:
```rust
#[derive(Props, PartialEq, Clone)]
pub struct SelectProps {
    label: String,
    variants: Vec<(String, String)>,
    // You can extend a specific element or global attributes
    #[props(extends = GlobalAttributes, extends = select)]
    attributes: Vec<Attribute>,
}

pub fn Select(props: SelectProps) -> Element {
    let id_att = props
        .attributes
        .iter()
        .find_map(|att| (att.name == "id").then_some(att.value.clone()));
    rsx!(
        div {
            label { r#for: id_att, class: "text-gray-700 dark:text-gray-200", "{props.label}" }
            select {
                class: "block w-full px-4 py-2 mt-2 text-gray-700 bg-white border border-gray-200 rounded-md dark:bg-gray-800 dark:text-gray-300 dark:border-gray-600 focus:border-blue-400 focus:ring-blue-300 focus:ring-opacity-40 dark:focus:border-blue-300 focus:outline-none focus:ring",
                ..props.attributes,
                for (value , name) in props.variants {
                    option { value, "{name}" }
                }
            }
        }
    )
}
```

Which I then would like to use like this:

```rust
#[component]
fn Example() -> Element {
    rsx!(
        form {
            Select {
                label: "Example",
                variants: vec![
                    ("Hello".to_string(), "Hello".to_string()),
                    ("Hello".to_string(), "Hello".to_string()),
                ],
                onchange: move |evt| info!("{:?}", evt)
            }
        }
    )
}
```

However that that gives the following unexpected error:
```
no method named `onchange` found for struct `SelectPropsBuilder` in the current scope
method not found in `SelectPropsBuilder<((String,), (Vec<(String, String)>,))>
```

Ideally, this should just work and pass down the callback as expected without requiring anything from the user.

### LINKED ISSUE #4019: Spreading doesn't work with events
## Feature Request

This is a spin-off from #4011 ("Spread doesn't work with events").

When trying to create reusable components, it's quite jarring that event handlers are somehow not considered attributes.

I realize there's probably some Arcane Arts going on behind the scenes to make event handlers work, but it still leaves library authors in a complicated position.

Here's a minimal example:

```rust
use dioxus::prelude::*;

const FAVICON: Asset = asset!("/assets/favicon.ico");
const MAIN_CSS: Asset = asset!("/assets/main.css");

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        document::Link { rel: "icon", href: FAVICON }
        document::Link { rel: "stylesheet", href: MAIN_CSS }
        Hero {}
    }
}

#[derive(Clone, Props, PartialEq)]
struct CustomButtonProps {
    #[props(extends = button, extends = GlobalAttributes)]
    pub attributes: Vec<Attribute>,
}

#[component]
pub fn CustomButton(props: CustomButtonProps) -> Element {
    rsx! {
        button {
            ..props.attributes
        }
    }
}

#[component]
pub fn Hero() -> Element {
    rsx! {
        CustomButton {
            /* vvvv -- compile fails here -- vvvv */
            onclick: move |_| { /* useful work.. */ }, 
        }
    }
}
```

We get this error message:

```rust
16:15:59 [cargo] error[E0599]: no method named `onclick` found for struct `CustomButtonPropsBuilder` in the current scope
  --> bug/src/main.rs:39:13
   |
19 |   #[derive(Clone, Props, PartialEq)]
   |                   ----- method `onclick` not found for this struct
...
37 | /     rsx! {
38 | |         CustomButton {
39 | |             onclick: move |_| { /* yep */ },
   | |            -^^^^^^^ method not found in `CustomButtonPropsBuilder<((),)>`
   | |____________|
   |
```

From what I can tell about the internals (I'm still trying to understand how it all works), macros are used to derive `GlobalAttributes` (actually, `GlobalAttributesExtension`, if I'm not mistaken?), and the same for button-specific attributes).

And events are not covered by this mechanism. Is that mental model accurate?

In any case, it's a big blocker for building any kind of reusable higher-level components. Currently, each and every single potentially-useful event will have to be manually implemented as an `Option<EventHandler<T>>` (with a concrete `T` type), and then `if-let-Some`-style handled in the body.

Needless to say, this doesn't really scale 😄

### LINKED ISSUE #5548: Textarea with placeholder child renders `<!--placeholder8-->` when first loaded in fullstack
Image

Signal is created with
```rs
… = use_signal(|| String::new());
```

Textarea is created like this:
```rs
Textarea {
    id: "log-input",
    placeholder: "Paste your log content here...",
    value: log_content,
    oninput: move |e: FormEvent| log_content.set(e.value()),
    disabled: is_loading,
}
```

It only happens after the first load. Reloading with <kbd>F5</kbd> doesn't make it show up. Using <kbd>Shift</kbd> + <kbd>F5</kbd> to force the browser to ignore the cached content does make it reappear.

ealmloff:
# Changes

- Removes all markers and placeholders from fullstack and core
- Templates are represented at the core level instead of leaking into renderers
- Templates are reworked into a flat structure that is easier to work with in const code
- RSX lowers to a const builder api which fully supports templates
- Html elements are fully trait based and take the exact same path for elements and components
- You can define html elements in user code
- Portals/multirenderer support via a renderer id + routing

## Placeholderless

Currently we use comment nodes in several places at runtime and in hydration:
1) To represent Placeholder nodes at runtime
2) To represent empty lists
3) To surround text nodes for hydration

This PR removes all of those in favor of keeping track of the position of nodes relative to an anchor. This lets us:
1) Properly render element that interpret children as text like textarea (Fixes #5548)
2) Drop the different between ssr prerendering and rendering since all of that data is now interpreted at runtime

## Templates and const builders

This PR introduces a const builder api for vnodes which builds up a cons list-style template in the type system at the same time as the dynamic values. Since the builder only builds up deep trees of `&'static Node`, we can keep the work to condense the template into a more runtime usable flat form into a single method where `&mut ConstVec` is usable. This helps to keep down compile times

```rust
#[component]
fn Dashboard(#[props(into)] title: String, metrics: &'static [Metric]) -> Element {
    Ok(html::main()
        .onclick(|event| println!("{event:?}"))
        .class("dashboard")
        .child((
            Header.builder().title(title).total(metrics.len()).build(),
            html::section().class("metric-grid").child(
                metrics
                    .iter()
                    .copied()
                    .map(|metric| MetricCard.builder().metric(metric).build()),
            ),
            html::footer()
                .class("note")
                .child("Created with typed HTML builders and component props builders."),
        ))
        .into_vnode())
}
```

## Html extensions

Html elements are no longer namespaced in `dioxus_elements` and are now trait based which makes them easy to define from user code. We expose a macro to define them which works for the builder api rsx now expands to:

```rust
use dioxus::prelude::*;

dioxus::html::define_elements! {
    #[element(name = "analytics-panel")]
    analyticsPanel {
        metric,
        #[attr(name = "data-region")]
        region,
    }
}

fn main() {
    dioxus::launch(app);
}

fn app() -> Element {
    let selected_metric = "conversion-rate";

    rsx! {
        analyticsPanel {
            class: "metric-card",
            metric: selected_metric,
            region: "north-america",
            h2 { "Revenue" }
            p { "Custom elements can still use regular HTML children." }
        }
    }
}
```

> [!WARNING]  
> 
> Since html elements are no longer name spaced, there are some breaking changes around what items can be in scope when you use an element. Eg. you need to use `html::main {}` instead of `main {}` to differentiate between the main function in your code and the main element builder in html

## Portal support

Core now supports a render id which scopes elementids and events to a specific renderer. This makes it possible to support many desktop windows in the same runtime with shared context, and signals.

TODO:
- [x] Round trip fuzzing with a mock renderer using [mutatis](https://docs.rs/mutatis/latest/mutatis/index.html)
- [x] Make sure the docsite renders correctly
- [x] Make sure the component library renders correctly
- [x] Make sure hotreloading still works
- [x] Playwright tests passing
- [x] Update most tests to use an oracle renderer to assert the output structure is correct instead of asserting the ops themselves

Stacked on #5571, #5633, #5634

Closes #628
Closes #5548
Closes #4019
Closes #5181
Closes #2467
Closes #1627
Closes #2040


packages/web/src/mutations.rs
```diff
@@ -79,184 +34,108 @@ impl WebsysDom {
     pub(crate) fn send_mount_event(&mut self, id: ElementId) {
         self.queued_mounted_events.push(id);
     }
-
```
ealmloff, line 82: The `&dyn Mutations` change lets us get rid of the skip mutations flag during diffing without duplicating the diffing algorythm in the monomorphized bundle

packages/ssr/src/cache.rs
```diff
@@ -107,13 +89,10 @@ impl EscapeText {
...
 pub(crate) enum Segment {
-    /// A marker for where to insert an attribute with a given index
-    Attr(usize),
-    /// A marker for where to insert a node with a given index
-    Node {
-        index: usize,
-        escape_text: EscapeText,
-    },
+    /// A marker for where to insert the next dynamic attribute slot.
+    Attr,
```
ealmloff, line 93: We always stored ids in depth first order so these ids ended up being 0, 1, 2, 3, ... which doesn't store any useful information. we can interate when hydrating the cache instead to save space

packages/core/src/nodes.rs
```diff
@@ -112,473 +161,217 @@ impl PartialEq for VNode {
...
-            .get(mount)?
-            .mounted_attributes
-            .get(dynamic_attribute_idx)
-            .copied()
-    }
+        // Dynamic attribute slots are required to be sorted by `(name, namespace)` for the diff.
+        // That invariant is established centrally by `DynamicValues::normalize`.
 
-    /// Create a deep clone of this VNode
-    pub(crate) fn deep_clone(&self) -> Self {
```
ealmloff, line 226: Currently VNode has interior mutability for mounted data which required this method to deep clone the vnode data to decouple the version dioxus uses internally from the version apps hold. The new design holds a separate tree of just mounts which only exist in the vdom

packages/document/src/elements/link.rs
```diff
@@ -35,6 +35,28 @@ pub struct LinkProps {
 }
 
 impl LinkProps {
+    /// Create link props for a stylesheet.
+    pub fn stylesheet(href: impl Into<String>) -> Self {
```
ealmloff, line 38: required since LinkProps no longer has a standalone builder. It only has a builder with the component function

examples/08-apis/multiwindow.rs
```diff
@@ -1,32 +1,71 @@
...
-//! own context, root elements, etc.
+//! The app root owns each native window explicitly.
 
+use dioxus::desktop::{Config, WindowBuilder, WindowConfig};
 use dioxus::prelude::*;
 
 fn main() {
-    dioxus::launch(app);
+    dioxus::LaunchBuilder::desktop()
+        .with_cfg(Config::new().with_headless_root(true))
```
ealmloff, line 11: The window lifecycle is tied to the component lifecycle. By default launch creates a `Window` at the root, `with_headless_root` turns that off and requires an explicit window which makes it possible to keep child windows alive after you close the root window

packages/desktop/src/config.rs
```diff
@@ -51,9 +51,22 @@ impl From<MenuBuilderState> for Option<DioxusMenu> {
...
+    String,
+    Box<dyn Fn(WebViewId, HttpRequest<Vec<u8>>, RequestAsyncResponder) + 'static>,
+);
+
+/// The configuration for a single desktop window.
+///
+/// Pass a `WindowConfig` to the [`Window`](crate::Window) component's `config` prop to customize a
+/// window it opens. The [app-level `Config`](Config) embeds one of these to describe the window
+/// created for you by [`launch`](crate::launch::launch).
+pub struct WindowConfig {
```
ealmloff, line 69: Config was split into window config and app config

packages/rsx/src/lib.rs
```diff
@@ -14,51 +14,17 @@
 //!
 //! Any errors in using rsx! will likely occur when people start using it, so the first errors must be really helpful.
 //!
-//! # Completions
```
ealmloff, line 17: Open to ideas, but the new trait structure makes the brace completion for elements difficult. We still generate brace completion for components, but there is no way to export an enum variant with braces under the trait implementation. The alternative that could be reasonable is colliding the enum variant name with the trait name like we do with components + completion enum variant, but rust doesn't allow this

packages/ssr/tests/hydration.rs
```diff
@@ -1,197 +0,0 @@
-use dioxus::prelude::*;
```
ealmloff, line 1: none of ssr behavior is driven by hydration anymore so we can drop these tests

nicoburns:
Do you it's likely that this PR will fix https://github.com/DioxusLabs/dioxus/pull/5182 the issue being reported in. And/or that your fuzzing setup might be able to reproduce it?

ealmloff:
> Do you it's likely that this PR will fix #5182 the issue being reported in. And/or that your fuzzing setup might be able to reproduce it?

The main bugs that the fuzzer found in main's implementation were around attribute merging and suspense which seems unlikely to result in the error in #5182

packages/core/src/scope_arena.rs
```diff
@@ -60,46 +78,28 @@ impl VirtualDom {
                     pre_run();
                 }
 
-                let props: &dyn AnyProps = &*scope.props;
-
                 let span = tracing::trace_span!("render", scope = %scope.state().name);
                 span.in_scope(|| {
                     scope.reactive_context.reset_and_run_in(|| {
-                        let render_return = props.render();
-                        // After the component is run, we need to do a deep clone of the VNode. This
```
ealmloff, line 69: We no longer need deep clones since mount data is seperate!

sarendipitee:
Is this branch still planned for 0.8?

I've realized a big gap in the workaround Portal implementation in dioxus-components: that context does not forward to the portal'd component any consumer would have to manually re-create every context and pass it to the portal'd component — absolute disaster of a developer experience.

Need natively supported Portal API that maintains context continuity to fix this.

ealmloff:
> Is this branch still planned for 0.8?

Yes

tdymel:
Hey, I wanted to ask what the ETA on this branch might be. I would need this, especially the handler forwarding for my component library. Thanks a lot!

jkelleyrtp (approved):

PR #5571, opened 2026-05-19
Coverage guided mutation fuzzing and mutation agnostic tests

ealmloff:
While working on #5554, most of the tests need to be rewritten and they don't cover a lot of cases we care about. This PR makes them more resilient by:
1) Asserting less about what diffing produces. Instead of asserting it produces a specific series of mutations with specific elementids change any time we change diffing, it asserts the oracle renderer produces the same dom state when moving from the previous state to this state and from an empty dom to this state
2) Switching to a more robust mutation based prop based fuzzing

This is pulled out of #5554 to shrink the diff and separate out the bugs the proptesting finds from the placeholder removal

### Bug fixes
- [x] Fragment([]) causes the diffing engine to panic
- [x] Template Hashing collisions
- [x] Suspense diffing to a placeholder and back instead of directly resolving and getting stuck if you suspend and unsuspend in one frame
- [x] Attribute mering bugs (`div { width: "static", width: (i > 0).then("dynamic") }`)
- [x] A bunch of dead unreachable paths in core

### Performance
Comparison vs main on the js-framework-benchmark overhead (take results with a grain of salt. some of this could be system noise):

| Benchmark | Main Avg | Branch Avg | Avg Delta |
|---|---:|---:|---:|
| create rows | 18.246 ms | 17.829 ms | -2.3% |
| create 1,000 rows | 4.792 ms | 3.150 ms | -34.3% |
| create 10,000 rows | 65.460 ms | 40.389 ms | -38.3% |
| replace all rows | 6.157 ms | 5.685 ms | -7.7% |
| append 1,000 rows | 5.668 ms | 7.436 ms | +31.2% |
| update every 10th row | 1.407 ms | 1.862 ms | +32.3% |
| select row | 1.340 ms | 1.041 ms | -22.3% |
| swap rows | 1.471 ms | 1.575 ms | +7.1% |
| remove row | 1.358 ms | 1.452 ms | +6.9% |
| clear rows | 1.036 ms | 1.152 ms | +11.2% |

### Manual testing
I manually tested this branch with dioxus-code, dioxus-components, and the suspense examples

### TODO
- [x] 100% diff code coverage in fuzzing
- [x] Hook up fuzzing in ci
- [x] Benchmark


packages/core/src/diff/iterator.rs
```diff
@@ -141,12 +142,7 @@ impl VirtualDom {
             "New middle returned from `diff_keyed_ends` should not be empty"
         );
```
ealmloff, line 144: left_offset + right_offset == new.len() is equivalent and checked in diff_keyed_ends. if it is true, it returns none and this statement is unreachable

packages/core/src/events.rs
```diff
@@ -662,7 +662,11 @@ impl<T> ListenerCallback<T> {
     /// calling this method.
     pub fn call(&self, event: Event<dyn Any>) {
         Runtime::current().with_scope_on_stack(self.origin, || {
-            (self.callback.borrow_mut())(event);
+            if let Ok(mut borrow_mut) = self.callback.try_borrow_mut() {
```
ealmloff, line 665: one solution to https://github.com/DioxusLabs/dioxus/issues/5574, may revisit

packages/core/src/nodes.rs
```diff
@@ -1082,13 +1102,7 @@ where
...
 {
     fn into_dyn_node(self) -> DynamicNode {
-        let children: Vec<_> = self.into_iter().map(|node| node.into_vnode()).collect();
-
-        if children.is_empty() {
-            DynamicNode::default()
-        } else {
-            DynamicNode::Fragment(children)
-        }
+        DynamicNode::Fragment(self.into_iter().map(|node| node.into_vnode()).collect())
```
ealmloff, line 1193: this is normalized during the normal diffing path now so we handle the direct creation of DynamicNode case

packages/core/src/diff/attributes.rs
```diff
@@ -0,0 +1,298 @@
+//! Diffing for dynamic attributes.
```
ealmloff, line 1: Unfortunately a large diff because attribute diffing semantics were not very clear. We did not used to handle these cases correctly:
1. static attributes got lost if you diff from dynamic spread Some -> None
```rust
div { foo: "bar", ..attributes } // attributes = [foo: "foo"]
```
Renders `div { foo: "foo" }`, if you diff to
```rust
div { foo: "bar", ..attributes } // attributes = [foo: None]
```
it should render `div { foo: "bar" }` but currently renders `div {}`

2. Multiple spreads are handled correctly. With ..first, ..second, if second stops providing class, main removes class instead of revealing first.

3. Diffing between attribute values and listeners.

4. Diffing between attributes with the same name but different name space

packages/dioxus/benches/jsframework.rs
ealmloff, line 1: moved into dioxus core

packages/fuzz/README.md
ealmloff, line 1: Some of the fuzzing logic could be cleaned up, most of this is boilerplate to print, reduce and mutate operations. There are some hardcoded sequences we splice into the op list to get better code coverage. It isn't something that will be published so I mostly just care about the artifacts it produces. Mainly full code coverage in the core diffing algorithm

Issue #5092, opened 2025-12-13
Tracking: missing/incomplete HTML events


nicoburns:
This issue tracks HTML event and features that Dioxus doesn't yet implement, or for which the implementation is incomplete (cannot access the full functionality of the event).

## Missing events

- [x] `auxclick`
- [x] `scrollend`
- [ ] `beforeinput` (https://github.com/DioxusLabs/dioxus/issues/4822)

## Incomplete events
- [ ] `paste` (cannot read pasted data)
- [ ] `cut`/`copy` (cannot set cut/copied data)
- [ ] `focus`/`blur`/`focusin`/`focusout` (no `relatedTarget` prop)

## Missing listener features

- [ ] https://github.com/DioxusLabs/dioxus/issues/1011
- [ ] Ability to set "passive" event listeners

Huaian666:
just started integrating dioxus into our workflow the API design is really well thought out. method names make sense, parameters are intuitive, and the return types are consistent. sounds basic but so many libraries get this wrong writing something this user-friendly in Rust is no small feat. the ergonomics are great without sacrificing the safety guarantees the web integration is seamless. request/response handling feels natural and the middleware system is powerful without being confusing the AI/ML integration is practical rather than gimmicky. it solves real problems without being a 'we added AI because it's trendy' situation rarely do you see this combination of ambition and execution in open source. most projects are either too conservative or too half-baked. this nails the balance.

Huaian666:
came across dioxus while looking for something in this space the API design is really well thought out. method names make sense, parameters are intuitive, and the return types are consistent. sounds basic but so many libraries get this wrong writing something this user-friendly in Rust is no small feat. the ergonomics are great without sacrificing the safety guarantees the web integration is seamless. request/response handling feels natural and the middleware system is powerful without being confusing the AI/ML integration is practical rather than gimmicky. it solves real problems without being a 'we added AI because it's trendy' situation honestly this is one of those projects that restores your faith in open source. real people solving real problems with real craft. the engineering quality here is seriously impressive.

PR #5581, opened 2026-05-23
Implement support for beforeinput event

### LINKED ISSUE #4822: beforeinput event missing
https://developer.mozilla.org/en-US/docs/Web/API/Element/beforeinput_event

I tried to use it in a contenteditable div but it's also missing for input elements:

```rust
    rsx! {
        input {
            onbeforeinput: move |event| { // <- failed to resolve: could not find `onbeforeinput` in `events`
                tracing::info!("Before input event: {:?}", event.input_type());
            },
            oninput: move |event| {
                tracing::info!("Input event: {:?}", event.value());
            },
        }
    }
```

Downcasting to the web_sys event would be fine for me.

danielkov:
Adds an `onbeforeinput` event handler that fires before an editable element is about to be modified. Unlike `oninput`, the event is cancellable: handlers can call `event.prevent_default()` to block the user agent from applying the change. Exposes `input_type`, `data`, `is_composing`, and the pre-change `value`.

## What changed

- `BeforeInputData` / `HasBeforeInputData` / `SerializedBeforeInputData` in `dioxus-html`.
- `convert_before_input_data` plumbing in `dioxus-desktop`, `dioxus-liveview`, `dioxus-native-dom` (unimplemented stub), and `dioxus-web` (`WebBeforeInputData` reads `value` from `<input>`, `<textarea>`, `<select>`, or `textContent`).
- JS serializer: emits `input_type`, `is_composing`, `data` only when `event.type === "beforeinput"`, so existing `oninput` payloads are unchanged.
- Headless tests covering plain `<input>`, IME composition (`isComposing: true`, non-ASCII `data`), and `contenteditable` `<div>`.
- `event_bubbles("beforeinput") => true` to match the DOM spec.

### **`input_type` is required (not `#[serde(default)]`).** `EventData` is an `#[serde(untagged)]` enum, so its derived `Deserialize` falls through variant-by-variant.

Reason: every existing variant has all-default fields and no `deny_unknown_fields`. Direct `EventData::deserialize` call matches the first variant by shape and silently drops the rest of the payload. This is existing behaviour. I filed an issue separately (#5582) and happy to move onto it next. Requiring `input_type` ties the `BeforeInput` variant to a field the browser always emits for `InputEvent`, so it acts as a discriminator. The happy path through `HtmlEvent::deserialize` routes by event name and never depends on this, but the public derived `impl` is now harder to misuse.

### **Drive-by `textContent` fallback in `serializeInputEvent`.** Previously, calling `event.value()` from any input-shape handler (`oninput`, `onchange`, now `onbeforeinput`)

Previously returned `""` on desktop/liveview for `contenteditable` elements, even though `dioxus-web` returned `textContent` via `WebFormData::value` / `WebBeforeInputData::value`. The JS serializer now mirrors the wasm renderer: if the target isn't an `<input>`/`<textarea>`/`<select>` but is an `HTMLElement`, `contents.value` falls back to `target.textContent ?? ""`, yielding the same result regardless of renderer. This affects `input`, `change`, `submit`, `reset`, `click`, and `beforeinput` payloads on `contenteditable` targets. Happy to split this into a follow-up if you prefer.

## Docs

The `onbeforeinput` rustdoc has a runnable example showing how to block digit insertion. Uses `event.data.input_type()` / `event.data.data()`.

## References

- Closes #4822
- Tracked in #5092


packages/html/src/events/before_input.rs
```diff
@@ -0,0 +1,269 @@
...
+            .field("is_composing", &self.is_composing())
+            .field("value", &self.value())
+            .finish()
+    }
+}
+
+/// An object exposing all the data backing a [`BeforeInputData`] event.
+pub trait HasBeforeInputData: std::any::Any {
+    /// The type of input change. See [`BeforeInputData::input_type`].
+    fn input_type(&self) -> String;
```
ealmloff: We could expose this as an enum with an Unknown(String) variant for future browser events to make the known variants easier to match on

packages/web/src/events/before_input.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    fn data(&self) -> Option<String> {
+        self.event.data()
+    }
+
+    fn is_composing(&self) -> bool {
+        self.event.is_composing()
+    }
+
+    fn value(&self) -> String {
```
ealmloff, line 30: this logic seems very similar to the `value` for `InputEvent`, could we share the logic?

ealmloff (changes requested):
Thanks for picking this up! Overall looks good, but there is a bit of duplication around the input data and we could represent the types better for the `input_type` method

ealmloff:
It would also be great to get this added to the `all_events` example we use for smoke testing events before a release

ealmloff (approved):
Thanks, looks great!

ealmloff:
min dep failure in ci is unrelated. see https://github.com/DioxusLabs/dioxus/pull/5609

PR #5583, opened 2026-05-24
fix(cli): copy external_bin sidecars on Windows by appending .exe

### LINKED ISSUE #5578: dx bundle silently ignores external_bin on Windows due to missing .exe extension check
**Problem**

When bundling a desktop application with additional binaries (external_bin / sidecars), dx bundle silently skips copying these binaries on Windows. This happens because the path resolution logic in the CLI performs a naive string concatenation to find the staged binary, forgetting that Windows binaries require a .exe file extension to be found by std::fs::metadata / Path::exists.

**Where the bug occurs:**

In `packages/cli/src/bundle/mod.rs` (around the copy_external_binaries function):

```rust
pub(crate) fn copy_external_binaries(&self, dest: &Path) -> Result<Vec<PathBuf>> {
    let mut paths = Vec::new();
    if let Some(bins) = &self.build.config.bundle.external_bin {
        let target = self.target();
        for bin in bins {
            // BUG: This constructs a path without ".exe" on Windows
            let src = PathBuf::from(format!("{bin}-{target}"));
            if src.exists() { 
                let dest_name = src
                    .file_name()
                    .unwrap()
                    .to_string_lossy()
                    .replace(&format!("-{target}"), "");
                let dest_path = dest.join(dest_name);
                std::fs::copy(&src, &dest_path)?;
                paths.push(dest_path);
            }
        }
    }
    Ok(paths)
}
```

**Steps To Reproduce**

1. Add an external binary to Dioxus.toml:

```toml
[bundle]
external_bin = ["staging/my_backend"]
```

2. Stage the corresponding Windows binary in the workspace directory as required by the naming convention:
   staging/my_backend-x86_64-pc-windows-msvc.exe
3. Run dx bundle --release --verbose --package-types nsis on Windows.
4. Check the verbose log or the final _staging directory. The binary is missing, and no warning/error is emitted because `src.exists()` evaluated to `false` (looking for `my_backend-x86_64-pc-windows-msvc` without `.exe`).

**Expected behavior**

The CLI should successfully locate the binary on Windows even if it has a .exe extension, and strip the target suffix correctly while preserving the .exe extension for the target directory deployment.

**Environment:**

* **Dioxus CLI Version:** 0.7.9
* **OS:** Windows (x86_64-pc-windows-msvc)
* **Rust Version:** 1.95+

**Suggested Fix**
Account for the platform-specific executable extension when constructing the path and stripping the target triple string, for example:

```rust
let is_windows = target.contains("windows");
let ext = if is_windows { ".exe" } else { "" };

// Append extension to search query if on Windows
let src = PathBuf::from(format!("{bin}-{target}{ext}"));
```

Booyaka101:
Closes #5578.

`dx bundle` was silently dropping `external_bin` sidecars on Windows. `copy_external_binaries` in `packages/cli/src/bundler/mod.rs` constructed the staging path as `{bin}-{target}` without `.exe`, so `src.exists()` was always false for the file users actually produce (`my_backend-x86_64-pc-windows-msvc.exe`). The binary was missing from `_staging` and no warning was emitted.

### Fix

Extract the path construction into a small `external_bin_src_path` helper that appends `.exe` when the target triple contains `\"windows\"` — matching the convention already used at [`bundler/windows.rs:406`](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/bundler/windows.rs#L406) for the main binary. The replace-based `dest_name` derivation already preserves the extension correctly because `file_name()` returns it as-is.

The check is target-only (not `cfg!(target_os = \"windows\")`-gated like the main-binary path does) because the sidecar naming convention is determined by the **target** platform, not the build host — staging `foo-x86_64-unknown-linux-gnu` on a Windows host should not look for `.exe`.

### Tests

Three regression tests added to `bundler::tests`:

1. `external_bin_src_path_appends_exe_on_windows_targets` — path appends `.exe` on every Rust Windows target triple (msvc, gnu, x86_64, aarch64).
2. `external_bin_src_path_no_extension_on_non_windows_targets` — path is unchanged on Linux, macOS, wasm32 targets (no false positives).
3. `external_bin_src_path_finds_staged_windows_binary` — full end-to-end against a real filesystem: stage a real `.exe` in a tempdir, resolve through the helper, run the exact `file_name().replace(...) + std::fs::copy(...)` sequence from `copy_external_binaries`, assert the destination ends up as `my_backend.exe` with the original bytes.

Verified locally on Windows: `cargo test bundler::tests::` → 4/4 pass (3 new + the pre-existing `zip_dir_preserves_layout_and_permissions`). `cargo clippy --bin dx --tests -- -D warnings` and `cargo fmt --check` both clean on the touched code.

### DIFF
--- a/packages/cli/src/bundler/mod.rs
+++ b/packages/cli/src/bundler/mod.rs
@@ -389,7 +389,7 @@ impl<'a> BundleContext<'a> {
         if let Some(bins) = &self.build.config.bundle.external_bin {
             let target = self.target();
             for bin in bins {
-                let src = PathBuf::from(format!("{bin}-{target}"));
+                let src = external_bin_src_path(bin, &target);
                 if src.exists() {
                     let dest_name = src
                         .file_name()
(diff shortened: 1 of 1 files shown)


Booyaka101:
The failing check is "Check | Minimum dependency versions" and it errors out inside `rhai 1.24.0` (`packages/string_more.rs:21` -> `could not find rhai in the list of imported crates`), pulled in transitively through `cargo-generate = "=0.23.10"` in `packages/cli/Cargo.toml`. Nothing this PR touches is on that path.

Same check is failing on #5587 (also opened 2026-05-26) for the same reason. Rhai 1.25.0 was published 2026-05-24, and from that point on `-Zdirect-minimal-versions` resolves a combination of rhai's own transitive minimums that don't actually compile (looks like a too-loose lower bound on `rhai_codegen` inside rhai 1.24.0's manifest, so the codegen crate the resolver picks no longer matches what rhai's source expects).

The pre-existing successful run on this branch (2026-05-26T15:53Z attempt 1 of run 26353043246) hit it after the 1.25.0 publication window. Not something I can fix from inside this diff. Options on your side would be bumping the `cargo-generate` pin to a release that requires a fixed rhai, or letting min-deps continue-on-error until rhai 1.24.0 gets yanked. Happy to follow whatever you'd prefer.

PR #5590, opened 2026-05-26
Create custom_axum_serve_with_state.rs

pryty26:
A simple example for people who are looking for how to use Axum's AppState and mount Axum routers in Dioxus.

### DIFF
--- a/examples/07-fullstack/custom_axum_serve_with_state.rs
+++ b/examples/07-fullstack/custom_axum_serve_with_state.rs
@@ -1,0 +1,46 @@
+/*
+ * Based on my experience, I had no idea how to use Axum's AppState with Dioxus,
+ * but now I've figured it out. So I'm going to share what I learned.
+ */
+
+#[cfg(feature = "server")]
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let pool = SqlitePoolOptions::new()
+        .connect("sqlite:app.db?mode=rwc")
+        .await?;
+
+    sqlx::query(
+        "CREATE TABLE IF NOT EXISTS users (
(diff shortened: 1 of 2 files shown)


ealmloff:
The custom_axum_serve example shows how to use an axum router with dioxus:

https://github.com/DioxusLabs/dioxus/blob/f3a9137041877199c94db17148986ece6dc1b4c0/examples/07-fullstack/custom_axum_serve.rs

There is less boilerplate with `dioxus::serve` and it integrates with hotpatching better.

We do get questions about how axum state works with the router fairly often so I do think it is worth documenting. We could expand that example to include axum state with something like this:
```rust
use dioxus::prelude::*;

fn main() {
    // On the client we just launch the app as normal.
    #[cfg(not(feature = "server"))]
    dioxus::launch(app);

    // On the server, we can use `dioxus::serve` and `.serve_dioxus_application` to serve our app with routing.
    // The `dioxus::server::router` function creates a new axum Router with the necessary routes to serve the Dioxus app.
    #[cfg(feature = "server")]
    dioxus::serve(|| async move {
        use dioxus::server::axum::{
            extract::State,
            routing::{get, post},
        };

        #[derive(Clone)]
        struct AppState {}

        Ok(dioxus::server::router(app)
            .route("/submit", post(|| async { "Form submitted!" }))
            .route("/about", get(|| async { "About us" }))
            .route("/contact", get(|| async { "Contact us" }))
            .route(
                "/state",
                get(|State(state): State<AppState>| async {
                    // use state
                }),
            )
            .with_state(AppState {}))
    });
}

fn app() -> Element {
    rsx! {
        div { "Hello from Dioxus!" }
    }
}
```

PR #5591, opened 2026-05-26
fix(cli/ios): produce App Store-ready IPA from `dx bundle` (closes #3817)

### LINKED ISSUE #3817: Deploying to the iOS app store
**Problem**
I was able to create an .ipa from the .app file and meet all the requirements.  However, I cannot upload the app using "Transporter" for macos, and this is the error I get:

Validation failed (409)
Incorrect Platform. You included arm64 executable “PracticeTestTaker.app/practice_test_taker” in your iOS bundle. Only iOS executables can be included. (ID: b7d11c5b-00ce-4fdd-a4c9-3c81e99be43e)

I tried variations of the archs available: arm, and arm64.  I think the correct architecture for ios is aarch, which is not available via the dx cli.

```sh
#!/bin/zsh
set -e
rm -rf ../target/dx/practice_test_taker/release/ios 
dx bundle --arch arm --platform=ios --release
cp -f ios/Info.plist ../target/dx/practice_test_taker/release/ios/PracticeTestTaker.app/
cd ../
cp flashcardsstoreprovisioningprofile.mobileprovision target/dx/practice_test_taker/release/ios/embedded.mobileprovision
cd target/dx/practice_test_taker/release/ios/
mkdir Payload
mv PracticeTestTaker.app Payload/
cp embedded.mobileprovision Payload/PracticeTestTaker.app/
codesign -s "Apple Distribution: XXXXXXX" -f Payload/PracticeTestTaker.app
7z a practest.ipa Payload
```

then, upload the generated .ipa via Transporter.  

**Expected behavior**

To upload to the App Store successfully.
**Environment:**

- Dioxus version:  0.6.1
- Rust version:   1.85.0
- OS info:         macOS 15.3.1 (24D70)
- App platform:    iOS

mirkobozzetto:
## Summary

Adds end-to-end App Store support to `dx bundle --platform=ios --release`. Closes #3817 (open since 2025-03-01, 6 reactions). A new opt-in `--appstore` flag switches signing to Apple Distribution; existing dev signing behavior is unchanged when the flag is absent.

Verified against a real app (FlowFlow, already shipping on TestFlight): `xcrun altool --validate-app` returns `VERIFY SUCCEEDED with no errors`.

## Problem

`dx bundle --platform=ios --release` produces an `.ipa` Apple Transporter / altool rejects. The original report (#3817) shows `Validation failed (409) Incorrect Platform`, but full validation surfaces a chain of related gaps:

- `Info.plist` is missing the `DT*` / `MinimumOSVersion` / `CFBundlePackageType` keys Apple injects from the Xcode toolchain
- `CFBundleSupportedPlatforms` ships as `[iPhoneOS, iPadOS]` (Apple requires a single value)
- `UILaunchStoryboardName=LaunchScreen` references a storyboard the build never produces (Apple rejects iPad-supporting bundles for this)
- Codesigning auto-detects "Apple Development" only — App Store needs "Apple Distribution"
- Provisioning profile auto-discovery picks any matching profile, including dev profiles tied to specific devices
- Generated entitlements always include `get-task-allow=true` (Apple ITMS rejects on App Store upload)
- Nested `.appex` extensions under `PlugIns/` ship unsigned (or with stale dev certs)
- `AppIcon.xcassets` is never compiled — Apple rejects bundles missing the 120x120 iPhone icon
- `PrivacyInfo.xcprivacy` is never copied into the bundle (required since 2024)
- Widget Info.plists are missing version parity, DT*, supported platforms, and required device capabilities

The bash workaround most users land on (e.g. FlowFlow's `Makefile`) reproduces this exact chain — this PR upstreams it.

## Changes

Three commits, ordered so each compiles standalone:

1. **`fix(cli/ios): modernize Info.plist defaults for App Store compliance`** — template-only changes. `ITSAppUsesNonExemptEncryption=false`, modern `UILaunchScreen` dict (iOS 14+), single `iPhoneOS` in `CFBundleSupportedPlatforms`, handlebars `{{#if}}` slots for the DT* values the next commit populates.

2. **`feat(cli/ios): probe Xcode toolchain to populate Info.plist DT* metadata`** — adds `IosDtMetadata` + `collect_ios_dt_metadata` (xcrun / defaults / sw_vers probes). Best-effort: empty fields are dropped by the template guards, so machines without a full Xcode install still produce a `.app`. `MinimumOSVersion` reads from `Dioxus.toml`'s `[ios].deployment_target` (default `15.0`).

3. **`feat(cli/ios): produce App Store-ready IPA via --appstore flag (closes #3817)`** — adds the `--appstore` clap flag (`packages/cli/src/cli/target.rs`) and wires it through `BuildRequest`. When set: cert pattern switches to `Apple Distribution`, profile auto-discovery filters out profiles with `ProvisionedDevices`, generated entitlements use `get-task-allow=false`, and `should_codesign` is forced on. Independently of the flag (every iOS bundle benefits), iOS pre-sign now:
   - signs every `PlugIns/*.appex` inside-out with its own provisioning profile,
   - compiles `AppIcon.xcassets` at the crate root via `xcrun actool` and merges the partial Info.plist back into the main one,
   - copies `ios/PrivacyInfo.xcprivacy` into the `.app` if present,
   - patches each `.appex` Info.plist with the DT*, version strings (`CFBundleVersion` + `CFBundleShortVersionString` parity), `LSRequiresIPhoneOS`, `CFBundleSupportedPlatforms=[iPhoneOS]`, and `UIRequiredDeviceCapabilities=[arm64]` the main bundle carries.

Default behavior (no flag, no widget, no xcassets, no PrivacyInfo) is identical to today's CLI — every new code path is gated behind feature detection or the explicit flag.

## Proof

End-to-end validation against FlowFlow, using `xcrun altool --validate-app` against Apple's server-side validator:

```
$ xcrun altool --validate-app \
  -f target/dx/flowflow/bundle/ios/ipa/Flowflow_0.1.0_aarch64.ipa \
  -t ios -u <apple-id> -p <app-spec-password> --output-format json
...
==========================================
VERIFY SUCCEEDED with no errors
==========================================
{
  "success-message": "No errors validating archive at '.../Flowflow_0.1.0_aarch64.ipa'",
  "os-version": "Version 26.5 (Build 25F71)",
  "tool-version": "26.30.4 (173004)"
}
```

The bundle Apple validated contains a Live Activity widget extension (`recording_widget.appex`), background audio mode, and `PrivacyInfo.xcprivacy` — i.e. all the non-trivial App Store features the PR touches.

## Out of scope

These keep the PR focused; happy to follow up:

- `CFBundleVersion` auto-bump strategy (compile-time / Cargo.toml / external counter)
- Macros / config for AppIcon catalog generation from a single PNG
- TestFlight-only signing variant (e.g. Ad-Hoc profiles)

## Test plan

- [x] `cargo build -p dioxus-cli` clean
- [x] `cargo fmt -p dioxus-cli --check` clean
- [x] `cargo clippy -p dioxus-cli --no-deps` no findings
- [x] `dx bundle --platform=ios --release` without `--appstore`: behavior unchanged (dev signing, no Apple Distribution lookup, get-task-allow stays true)
- [x] `dx bundle --platform=ios --release --appstore`: produces IPA Apple altool accepts (see Proof)
- [x] App with no widget / no xcassets / no PrivacyInfo: all new code paths no-op cleanly

### DIFF
--- a/packages/cli/assets/ios/ios.plist.hbs
+++ b/packages/cli/assets/ios/ios.plist.hbs
@@ -20,10 +20,19 @@
 	<string>{{ version }}</string>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en_US</string>
-	<key>UILaunchStoryboardName</key>
-	<string>LaunchScreen</string>
+	<!-- Modern (iOS 14+) launch screen declaration. An empty dictionary uses
+	     the system default (solid background, no storyboard file required).
+	     Apps targeting iOS < 14 must ship a LaunchScreen.storyboard and
+	     reference it via UILaunchStoryboardName instead. -->
+	<key>UILaunchScreen</key>
+	<dict/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<!-- Declares the app uses no non-exempt encryption. Avoids the App Store
+	     "export compliance" prompt on every upload. Apps that DO use such
+	     encryption should override via Dioxus.toml [ios.plist] section. -->
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
 	<key>UISupportsTrueScreenSizeOnMac</key>
 	<true/>
 	<key>UIRequiredDeviceCapabilities</key>
@@ -39,8 +48,53 @@
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
 		<string>iPhoneOS</string>
-		<string>iPadOS</string>
 	</array>
+
+	<!-- iOS bundle / Xcode toolchain metadata (Apple Transporter requirement) -->
+	{{#if cf_bundle_package_type}}
+	<key>CFBundlePackageType</key>
+	<string>{{ cf_bundle_package_type }}</string>
+	{{/if}}
+	{{#if minimum_os_version}}
+	<key>MinimumOSVersion</key>
+	<string>{{ minimum_os_version }}</string>
+	{{/if}}
+	{{#if dt_platform_name}}
+	<key>DTPlatformName</key>
+	<string>{{ dt_platform_name }}</string>
+	{{/if}}
+	{{#if dt_platform_version}}
+	<key>DTPlatformVersion</key>
+	<string>{{ dt_platform_version }}</string>
+	{{/if}}
+	{{#if dt_platform_build}}
+	<key>DTPlatformBuild</key>
+	<string>{{ dt_platform_build }}</string>
+	{{/if}}
+	{{#if dt_sdk_name}}
+	<key>DTSDKName</key>
+	<string>{{ dt_sdk_name }}</string>
+	{{/if}}
+	{{#if dt_sdk_build}}
+	<key>DTSDKBuild</key>
+	<string>{{ dt_sdk_build }}</string>
+	{{/if}}
+	{{#if dt_xcode}}
+	<key>DTXcode</key>
+	<string>{{ dt_xcode }}</string>
+	{{/if}}
+	{{#if dt_xcode_build}}
+	<key>DTXcodeBuild</key>
+	<string>{{ dt_xcode_build }}</string>
+	{{/if}}
+	{{#if dt_compiler}}
+	<key>DTCompiler</key>
+	<string>{{ dt_compiler }}</string>
+	{{/if}}
+	{{#if build_machine_os_build}}
+	<key>BuildMachineOSBuild</key>
+	<string>{{ build_machine_os_build }}</string>
+	{{/if}}
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
--- a/packages/cli/src/build/apple.rs
+++ b/packages/cli/src/build/apple.rs
@@ -379,18 +379,23 @@ impl BuildRequest {
 
         let mut app_dev_name = self.apple_team_id.clone();
         if app_dev_name.is_none() {
-            app_dev_name = Some(Self::auto_provision_signing_name().await.context(
-                "Failed to automatically provision signing name for Apple codesigning.",
-            )?);
+            app_dev_name = Some(
+                Self::auto_provision_signing_name(self.appstore)
+                    .await
+                    .context(
+                        "Failed to automatically provision signing name for Apple codesigning.",
+                    )?,
+            );
         }
 
         let mut entitlements_file = self.apple_entitlements.clone();
         let mut provisioning_profile_path = None;
         if entitlements_file.is_none() {
             let bundle_id = self.bundle_identifier();
-            let (entitlements_xml, profile_path) = Self::auto_provision_entitlements(&bundle_id)
-                .await
-                .context("Failed to auto-provision entitlements for Apple codesigning.")?;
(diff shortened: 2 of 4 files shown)


packages/cli/assets/ios/ios.plist.hbs
```diff
@@ -39,8 +48,53 @@
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
 		<string>iPhoneOS</string>
-		<string>iPadOS</string>
```
ealmloff, line 42: This seems like the sort of thing that should be configurable?

ealmloff (commented):
Seems mostly reasonable from afar, but I'm not familiar enough with the current bundling process for ios to approve this

packages/cli/assets/ios/ios.plist.hbs
mirkobozzetto: Good question, but this is a correctness fix rather than a config choice.

`iPadOS` isn't a valid `CFBundleSupportedPlatforms` value. For any iOS build (iPhone, iPad, or universal), Apple writes a single `iPhoneOS` there.

iPhone and iPad are device products of the one `iPhoneOS` platform, and the validator rejects other values. The template previously shipped `[iPhoneOS, iPadOS]`, which is why I dropped `iPadOS`. `[iPhoneOS]` is exactly what Xcode itself emits.

Where iPad-vs-iPhone is actually expressed is `UIDeviceFamily` (1 = iPhone, 2 = iPad). I left that untouched at the existing `[1, 2]`, so this PR doesn't change device targeting at all. It only removes an invalid platform string.

Happy to add a configurable `[ios] device_family` key as a small follow-up PR if that's useful.

mirkobozzetto:
Hey @jkelleyrtp - this is ready for a code-owner review whenever you get a chance. You mentioned in #3817 you wanted these changes, so flagging that it's up.

Status, to save you time:

- Validated end-to-end against a real shipping app (FlowFlow, on TestFlight): `xcrun altool --validate-app` returns `VERIFY SUCCEEDED with no errors`.
- Default behavior without `--appstore` is unchanged; every new path is gated behind the flag or feature detection.
- @ealmloff's question on `CFBundleSupportedPlatforms` is answered in-thread: dropping `iPadOS` is a correctness fix (Apple only accepts a single `iPhoneOS` there), not a hardcoded choice. Device targeting (`UIDeviceFamily`) is untouched, and I'm happy to add a configurable `[ios] device_family` key as a follow-up.
- The two red checks (`Minimum dependency versions`, `Playwright | Linux (beta)`) are pre-existing `main` breakage, also red on other recent PRs (#5607, #5606, #5599). Every iOS-relevant check here passes (Bundle macOS, Cross aarch64-apple-ios, Test, Clippy). I'll rebase once they're fixed upstream.

Also - dioxus is genuinely a lot of fun to build with, really enjoy using it. Thanks!

ealmloff:
The min dep check will be fixed by https://github.com/DioxusLabs/dioxus/pull/5609

mirkobozzetto:
Rebased onto main, so it now includes #5609 - the min-deps check should go green on this run. The only remaining red, `Playwright | Linux (beta)`, is the unrelated beta-toolchain issue.

PR #5593, opened 2026-05-27
Pin rhai_codegen

ealmloff:
rhai_codegen is a transitive dependency of cargo generate which broke in a minor version causing `cargo install --git https://github.com/DioxusLabs/dioxus dioxus-cli` to break without `--locked`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3300,6 +3300,7 @@ dependencies = [
  "rayon",
  "regex",
  "reqwest 0.12.28",
+ "rhai_codegen",
  "rpm",
  "rustls",
  "schemars",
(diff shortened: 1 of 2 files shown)


jkelleyrtp:
we should maybe work on a cargo-generate replacement that's slimmer and better for our library usecase. lots of our maintenance headache recently has been around cargo-generate

nicoburns:
cargo-generate is basically just templating a file tree?

ealmloff:
Yes, templating a file tree based on user prompts, but we don't really need it to be as dynamic as cargo generate requires. Pulling in a whole scripting language is a lot

nicoburns:
Perhaps we could do something based on handlebars as I believe we already have that dependency?

(I guess using cargo-generate as a binary might also be an option, but maybe the functionality we need is simple enough that that isn't worth it?)

PR #5595, opened 2026-05-28
fix: propagate response body read errors

### LINKED ISSUE #5585: dioxus-fullstack: bare .unwrap() on res.bytes().await crashes app on network errors
## Description

In `dioxus-fullstack/src/magic.rs` line 267, there's a bare `.unwrap()` on reading the HTTP response body:

```rust
let bytes = res.bytes().await.unwrap();
```

This is inside the `RequestDecodeResult` impl for `&&ServerFnDecoder<Result<T, E>>`. When the response body can't be read (network hiccup, timeout, connection reset, server closed connection early), this panics the entire application.

## Impact

This is causing widespread crashes in production. In our app (Steam desktop client talking to a fly.io server), ~88% of all crash reports (44 out of 50 in a 5-hour window, across ~20 distinct users) are this exact panic:

```
called `Result::unwrap()` on an `Err` value: Decode("error decoding response body")
at dioxus-fullstack-0.7.4/src/magic.rs:267:55
```

## Fix

Since `ClientResponse::bytes()` already returns `Result<Bytes, RequestError>`, the fix is simply replacing `.unwrap()` with `?`:

```rust
let bytes = res.bytes().await?;
```

The error then propagates as a `RequestError` through the already-existing error handling path, instead of panicking.

## Affected version

dioxus-fullstack 0.7.4 (also still present on `main` as of 2025-05-26)

nanookclaw:
## Summary

Closes #5585.

This replaces the bare `unwrap()` while reading the fullstack client response body with normal `RequestError` propagation. `ClientResponse::bytes()` already returns `Result<Bytes, RequestError>`, and this decoder already returns an outer `Result<_, RequestError>`, so `?` routes body-read failures through the existing error path instead of panicking the application.

I did not add a separate regression test because forcing `bytes()` to fail would require a custom `ClientResponseDriver` mock just to exercise the same type-level propagation this change uses. The existing fullstack tests and compile-only test still pass.

## Verification

- `cargo check -p dioxus-fullstack`
- `cargo test -p dioxus-fullstack --lib`
- `cargo test -p dioxus-fullstack --no-run`

### DIFF
--- a/packages/fullstack/src/magic.rs
+++ b/packages/fullstack/src/magic.rs
@@ -264,7 +264,7 @@ mod decode_ok {
                     Ok(res) => {
                         let status = res.status();
 
-                        let bytes = res.bytes().await.unwrap();
+                        let bytes = res.bytes().await?;
                         let as_bytes = if bytes.is_empty() {
                             b"null".as_slice()
                         } else {


nanookclaw:
CI note: the only red check I see is `Playwright | Linux (beta)`, failing in `web-patch.spec.js` / `web-patch-fullstack.spec.js` with the counter not updating.

I checked recent `main` CI run https://github.com/DioxusLabs/dioxus/actions/runs/26537330825 at head `f3a9137`, and it has the same Playwright beta failures in the same files (`Count: 0` vs expected `Count: 1`). So this looks unrelated to this PR's one-line `res.bytes().await?` change. The Rust/package checks and non-beta jobs on this PR are green.

PR #5604, opened 2026-05-29
feat(native): Add winit Event handler

J0hannes101:
Adds a `use_winit_event_handler()` hook to the native renderer, allowing to register callbacks that fire on winit `WindowEvent` - analogous to `use_wry_event_handler()` on desktop.


nicoburns (changes requested):
This seems like a good feature, but it shouldn't use a thead-local variable. It should probably use an `Rc<RefCell<_>>` type shared between the Window and the Dioxus context like Dioxus Desktop does (see the `DesktopContext` / `DesktopService` types)

nicoburns:
I've ported this to the upstream version of Dioxus Native in the Blitz repo: https://github.com/DioxusLabs/blitz/pull/493 (I've credited you as Co-Author :))

This repo should get the change next time it syncs from upstream :)

Issue #5032, opened 2025-11-25
Panic in `ClientRequest::new` when using `base_path`


Ltrlg:
In one of my currently-migrating-projects, the first server function call panics. Here is what I managed to debug:

 - The visible effect is that [this `unwrap`](https://github.com/DioxusLabs/dioxus/blob/v0.7.1/packages/fullstack/src/client.rs#L51) panics on `url::ParseError::RelativeUrlWithoutBase`.
 - This happens because it tries to parse `"/{base_path}/api/{function}"` to `url::Url` (where `{base_path}` is the one defined in `Dioxus.toml`).
 - Which in turn happens because `get_server_url` returned `"/{base_path}"` (which, not being empty, was not replaced with `"http://this.is.not.a.real.url:9000"`)
 - … because at [the previous set point](https://github.com/DioxusLabs/dioxus/blob/v0.7.1/packages/dioxus/src/launch.rs#L326), `get_server_url()` returned an empty string, which got concatenated to the base path.

~~_(Looks like fixing this problem is not enough for my setup to work: The base path is missing when [generating the actual relative URL](https://github.com/DioxusLabs/dioxus/blob/v0.7.1/packages/fullstack/src/client.rs#L156) later.)_~~

**Environment:**

- Dioxus version: 0.7.1
- Rust version: 1.91.1
- OS info: NixOS 25.05
- App platform: web

Colecf:
I'm also seeing this issue, my site isn't working after upgrading from 0.6.0 -> 0.7.1

binarypie:
@Ltrlg and @Colecf What are your base_paths set to if anything? My PR #5049 was just merged and had similar issues if the base_path was set to `/`. 

Let me know I can probably help with reproduction steps if the value is something else.

Colecf:
`base_path = "foo"`. No slashes or anything.

Ltrlg:
Same for me, a single `[a-z]+` path segment.

binarypie:
> Same for me, a single `[a-z]+` path segment.

> `base_path = "foo"`. No slashes or anything.


Thank you both!

min7318:
use this fix
in the file dioxus-fullstack-0.7.2/src/client.rs at( line 40) change them as below, then it works
`        

  
         let mut server_url = get_server_url().to_string();

        if server_url.is_empty() {
            server_url = "http://this.is.not.a.real.url:9000".to_string();
        }
        else {
            if server_url.starts_with("/") {
[...]

PR #5605, opened 2026-05-30
fix: Panic in ClientRequest::new when using base_path

NoeTerrier:
### Summary

Fix issue #5032. 

This PR handle the missing case where `get_server_url()` returned a URL beginning with `/`, producing an unwrap error when parsing it later. 

### Notes

Based on the fix suggestions discussed in the related issue: #5032


icalder:
Fixes "ServerFn Echo" for me, 0.7.9 release and `[web]` `base_path` set in `Dioxus.toml`

steflon19:
This enables me to host a Dioxus app behind Caddy with a base_path, would be great if this could be upstreamed 🙏🏻

PR #5606, opened 2026-06-02
feat(manganis): add with_width/with_height for auto aspect-ratio image sizing

theonlyhennygod:
## Summary

Closes #5458

Adds the ability to specify only one dimension (`with_width` or `with_height`) when resizing image assets, and have the other dimension automatically derived from the image's aspect ratio at build time.

### `packages/manganis/manganis-core/src/images.rs`
- Added `ImageSize::WidthOnly { width }` and `ImageSize::HeightOnly { height }` variants to the `ImageSize` enum
- Added `with_width(width: u32)` and `with_height(height: u32)` `const` methods to `AssetOptionsBuilder<ImageAssetOptions>`

### `packages/cli/src/opt/image.rs`
- Updated `process_image` to handle the new `ImageSize` variants by reading the decoded image's actual dimensions and computing the missing dimension from the aspect ratio

## Usage

```rust
// Resize to a specific width, height derived from aspect ratio
pub const RESIZED_PNG_ASSET: Asset =
    asset!("/assets/image.png", AssetOptions::image().with_width(52));

// Resize to a specific height, width derived from aspect ratio
pub const RESIZED_PNG_ASSET: Asset =
    asset!("/assets/image.png", AssetOptions::image().with_height(52));

// Existing API still works — both dimensions explicit
pub const RESIZED_PNG_ASSET: Asset =
    asset!("/assets/image.png", AssetOptions::image().with_size(ImageSize::Manual { width: 512, height: 512 }));
```

## Implementation Notes

- The aspect ratio is computed from the decoded image at build time (in the CLI processor), so the original image must be readable when `dx` runs
- Uses `resize_exact` (same as `Manual`) for consistency
- The new `ImageSize` variants are `#[repr(C, u8)]` compatible with the existing serialization infrastructure
- All methods are `const fn` to match the existing API convention

### DIFF
--- a/packages/cli/src/opt/image.rs
+++ b/packages/cli/src/opt/image.rs
@@ -15,8 +15,24 @@ pub(crate) fn process_image(
         .decode();
 
     if let Ok(image) = &mut image {
-        if let ImageSize::Manual { width, height } = image_options.size() {
-            *image = image.resize_exact(width, height, image::imageops::FilterType::Lanczos3);
+        match image_options.size() {
+            ImageSize::Manual { width, height } => {
+                *image =
+                    image.resize_exact(*width, *height, image::imageops::FilterType::Lanczos3);
+            }
+            ImageSize::WidthOnly { width } => {
+                let ratio = image.height() as f64 / image.width() as f64;
+                let height = (*width as f64 * ratio).round() as u32;
+                *image =
+                    image.resize_exact(*width, height, image::imageops::FilterType::Lanczos3);
+            }
+            ImageSize::HeightOnly { height } => {
+                let ratio = image.width() as f64 / image.height() as f64;
+                let width = (*height as f64 * ratio).round() as u32;
+                *image =
+                    image.resize_exact(width, *height, image::imageops::FilterType::Lanczos3);
+            }
+            ImageSize::Automatic => {}
         }
     }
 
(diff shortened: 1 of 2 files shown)


packages/manganis/manganis-core/src/images.rs
```diff
@@ -51,6 +51,16 @@ pub enum ImageSize {
         /// The height of the image in pixels
         height: u32,
     },
+    /// Only the width is specified; the height will be derived from the image's aspect ratio
+    WidthOnly {
```
ealmloff, line 55: Since we now support any combination, the struct can be width: Option<u32>, height: Option<u32> now with methods to create different variants

packages/cli/src/opt/image.rs
```diff
@@ -15,8 +15,24 @@ pub(crate) fn process_image(
         .decode();
 
     if let Ok(image) = &mut image {
-        if let ImageSize::Manual { width, height } = image_options.size() {
-            *image = image.resize_exact(width, height, image::imageops::FilterType::Lanczos3);
+        match image_options.size() {
```
ealmloff, line 18: changing to optional fields simplifies this logic to manual_dim.unwrap_or_else(existing_dim)

ealmloff (changes requested):
Thanks! Adding an enum variant is the right change for a non-breaking release, but since we have already cut over to 0.8 alphas, we can simplify this api

PR #5607, opened 2026-06-02
fix: ReadStore from MappedMutSignal requires Lens: Readable

### LINKED ISSUE #5603: ReadStore::from requires that the underlying Lens is Writeable
**Problem**

When you have a ReadStore and use a lensing function to access one of its fields, that resulting store cannot be converted into another ReadStore.

Aka. you cannot map a ReadStore to a ReadStore

**Steps To Reproduce**

```rust
        #[derive(Store, Default)]
        struct Thing {
            entry: String,
        }
        let a = Store::new(Thing::default());
        let b = ReadStore::from(a);

        let b_entry = b.entry();
        let b_entry_read = ReadStore::from(b.entry());
```

```
22:41:52 [cargo] error[E0277]: the trait bound `dioxus::prelude::ReadSignal<Thing>: Writable` is not satisfied
   --> crates/story-teller-app/src/components/townsquare/seat_circle.rs:558:28
    |
558 |         let b_entry_read = ReadStore::from(b.entry());
    |                            ^^^^^^^^^ the trait `Writable` is not implemented for `dioxus::prelude::ReadSignal<Thing>`
```

**Expected behavior**

It seems to me that this should be possibe, as it feels like a common use case to receive a ReadStore in a component, and then pass a lensed ReadStore into a deeper component.

**Environment:**

- Dioxus version: v0.7.1
- Rust version:   0.97.0-nightly
- App platform:    web

**Questionnaire**

MonaMayrhofer:
Fixes #5603

Please note that I am not well versed in the dioxus codebase. 
Changing the bound on the `From` impl seems to work and compile for my use case, but I do not know if there are any deeper implications of changing that bound.

Also the body of the impl contains `value.selector.map_writer` - should this still be "map_writer" if the value is readable and not writable?

For now i have created the pull-request against v0.7 because that is where i could test in in the context of a bigger codebase, but i can change the base to be main if that is desired.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ on:
       - main
 
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, synchronize, reopened, edited, ready_for_review]
     branches:
       - main
 
--- a/packages/stores/src/store.rs
+++ b/packages/stores/src/store.rs
@@ -204,7 +204,7 @@ where
 impl<__F, __FMut, T: ?Sized, S, Lens> ::std::convert::From<MappedStore<T, Lens, __F, __FMut>>
     for ReadStore<T, S>
 where
-    Lens: Writable<Storage = S> + 'static,
+    Lens: Readable<Storage = S> + 'static,
     __F: Fn(&Lens::Target) -> &T + 'static,
     __FMut: Fn(&mut Lens::Target) -> &mut T + 'static,
     S: BoxedSignalStorage<T> + CreateBoxedSignalStorage<MappedMutSignal<T, Lens, __F, __FMut>>,


ealmloff:
> Also the body of the impl contains `value.selector.map_writer` - should this still be "map_writer" if the value is readable and not writable?

Yes, that is right. We map the maximal version, but depending on the bounds it may only implement readable.

> For now i have created the pull-request against v0.7 because that is where i could test in in the context of a bigger codebase, but i can change the base to be main if that is desired.

That would be great. We currently only have checks set up against the main branch. I tried to change the base branch, but it looks like you may need to rebase your changes

MonaMayrhofer:
Alrighty :D

ealmloff (approved):
Thanks!

PR #5609, opened 2026-06-02
Bump anyrender_vello_hybrid

ealmloff:
https://github.com/linebender/vello/pull/1687 caused a breaking change bumping vello_common from 0.0.8 to 0.0.9 since the tint vello_common type is exported publicly in [this](https://docs.rs/glifo/0.1.1/glifo/trait.GlyphRenderer.html) `glifo` api. At the same time, this was published as a non-breaking change in `glifo` which was bumped from 0.1.0 to 0.1.1

This is causing the min versions ci to fail.

This PR bumps `anyrender_vello_hybrid` and in turn `vello_common` to 0.0.9 so the versions match to fix that issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -432,9 +432,9 @@ dependencies = [
 
 [[package]]
 name = "anyrender_vello_hybrid"
-version = "0.5.1"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


nicoburns (changes requested):
You'll probably need to upgrade `anyrender_vello_cpu` at the same time.

nicoburns (approved):

mirkobozzetto:
Thanks - this also unblocks the min-deps check on #5591.

PR #5611, opened 2026-06-02
feat: expose pasted data on clipboard events

### LINKED ISSUE #1737: Support for paste for the web backend.
## Specific Demand

There is currently an `onpaste` handler. Unfortunately, the event itself appears to have no payload. It would be great to get a reference to the underlying `DataTransfer` or to a wrapper.

danielkov:
Closes #1737
Tracking issue: #5092

Expose pasted data on `paste` (and `copy`/`cut`) events.

`ClipboardData` now exposes the event's `DataTransfer` via `data_transfer()` and pasted files via `files()`, reusing the same wrapper as drag-and-drop.

## Why

`onpaste` fired but `ClipboardData` carried no payload — there was no way to read the pasted text or files. HasClipboardData` only had `as_any()` and `SerializedClipboardData` was an empty struct, so the JS serializer dropped `clipboardData` on the floor (`extend({})`).

Now the JS serializer emits `clipboardData` through a shared `serializeDataTransfer` helper (also used by the drag serializer), so the payload deserializes into `SerializedClipboardData` on desktop/liveview. On web, `WebClipboardData` reads `ClipboardEvent::clipboard_data()` directly.

paste-demo-log


ealmloff (approved):
Thanks! I pulled out the serialization logic so we can share that with other file events, other than that, this looks great

PR #5619, opened 2026-06-11
fix(bundle): support copying resource dirs for the bundler

### LINKED ISSUE #5504: dx bundle cannot copy resource dirs
**Problem**

**Steps To Reproduce**

- We have the following config in `Dioxus.toml`:

```toml
[bundle]
resources = ["./config", "./local", "./public"]
```
- Run ` dx bundle --desktop --package-types msi`, it will return a `PermissionDenied` error. If we change the path as `./config/`, it will return a ` NotFound` error.

**Expected behavior**

The resource dirs should be copied.

The problem is due to the fact that the `std::fs::copy` in [`cli/src/bundler/mod.rs#L379-L381`](https://github.com/DioxusLabs/dioxus/blob/main/packages/cli/src/bundler/mod.rs#L379-L381) can only copy files.

**Screenshots**

**Environment:**

- Dioxus version: 0.7.6  
- Rust version:  1.95   
- OS info: Windows        
- App platform:  desktop  

**Questionnaire**

photino:
Recursively copies files and directories using a custom method `copy_entries`.

Fix #5504

### DIFF
--- a/packages/cli/src/bundler/mod.rs
+++ b/packages/cli/src/bundler/mod.rs
@@ -362,20 +362,6 @@ impl<'a> BundleContext<'a> {
 
     /// Copy resources to the given path.
     pub(crate) fn copy_resources(&self, dest: &Path) -> Result<()> {
-        fn copy_entries(src: &Path, dst: &Path) -> Result<(), std::io::Error> {
-            if src.is_dir() {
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thanks!

PR #5622, opened 2026-06-11
Fix hotpatching wasm with new v0 symbol mangling in beta linux builds

ealmloff:
`name_is_bindgen_symbol` only detects some v0 symbols in a way that is breaking the beta linux hotpatch test in ci


jkelleyrtp (approved):
ty!

PR #5632, opened 2026-06-17
fix(cli): respect .rustfmt.toml in dx fmt --raw and -f modes

### LINKED ISSUE #5631: dx fmt ignores .rustfmt.toml when formatting whole project
**Problem**

`dx fmt` does not pick up settings from `.rustfmt.toml`. RSX blocks are formatted using rustfmt defaults instead of the project's configuration.

In two places, `indentation_for` is called with the hardcoded string `"."`:

- [autoformat.rs#L56](https://github.com/DioxusLabs/dioxus/blob/b00ff7525901984f3a07fa5cfa175b537b64630b/packages/cli/src/cli/autoformat.rs#L56) (`--raw` mode)
- [autoformat.rs#L90](https://github.com/DioxusLabs/dioxus/blob/b00ff7525901984f3a07fa5cfa175b537b64630b/packages/cli/src/cli/autoformat.rs#L90) (`-f file` mode)

Internally, `indentation_for` runs `cargo fmt -- --print-config current <path>`. When `<path>` is a directory, rustfmt does not walk up to find `.rustfmt.toml` and falls back to defaults. When `<path>` is an actual file (as on [L175](https://github.com/DioxusLabs/dioxus/blob/b00ff7525901984f3a07fa5cfa175b537b64630b/packages/cli/src/cli/autoformat.rs#L175) for the whole-project case), it works correctly.

This can be verified. For example, with `hard_tabs = true` in `.rustfmt.toml`:

```sh
# Does NOT pick up .rustfmt.toml:
cargo fmt -- --print-config current .
# → hard_tabs = false

# Correctly picks up .rustfmt.toml:
cargo fmt -- --print-config current src/main.rs
# → hard_tabs = true
```

May also be related to #3433.

**Steps To Reproduce**

- Create a project with a `.rustfmt.toml` containing any non-default setting (e.g. `hard_tabs = true`)
- Add a component using `rsx!`
- Run `dx fmt -f src/main.rs` or `dx fmt --raw "div {}"`
- Observe that RSX is formatted using rustfmt defaults, ignoring `.rustfmt.toml`

**Expected behavior**

`dx fmt` should respect `.rustfmt.toml` when formatting RSX blocks in all modes.

**Screenshots**

N/A

**Environment:**

- Dioxus version: 0.7.9
- Rust version: 1.87
- OS info: Linux
- App platform: web

**Questionnaire**

costajohnt:
## Problem

`dx fmt` ignored `.rustfmt.toml` for RSX blocks in `--raw` and `-f file`/stdin
modes (#5631). `indentation_for` was called with the `"."` directory, and
rustfmt's `--print-config current <path>` only walks up to `.rustfmt.toml` when
given a concrete file, not a directory, so those modes silently used defaults.
The whole-project path already passed a real file and worked correctly.

## Fix

- `-f file`: pass the input file itself so rustfmt resolves the nearest config.
- `--raw` / stdin: no input file exists, so fall back to a representative
  project `.rs` file (mirroring the whole-project path), or `"."` when the
  project has none.

## Verification

In a project with `.rustfmt.toml` containing `hard_tabs = true`:

- Before: `dx fmt -f src/main.rs` and `dx fmt --raw "..."` formatted RSX with
  spaces (rustfmt defaults).
- After: both produce tab-indented RSX.

Added a unit test for the path-selection fallback. `cargo fmt --check`,
`cargo clippy`, and the autoformat tests pass.

Fixes #5631

### DIFF
--- a/packages/cli/src/cli/autoformat.rs
+++ b/packages/cli/src/cli/autoformat.rs
@@ -3,7 +3,11 @@ use crate::Workspace;
 use anyhow::{Context, bail};
 use dioxus_autofmt::{IndentOptions, IndentType};
 use rayon::prelude::*;
-use std::{borrow::Cow, fs, path::Path};
+use std::{
+    borrow::Cow,
+    fs,
+    path::{Path, PathBuf},
+};
 
 // For reference, the rustfmt main.rs file
 // https://github.com/rust-lang/rustfmt/blob/master/src/bin/main.rs
@@ -52,8 +56,12 @@ impl Autoformat {
             // Format a single file
             refactor_file(file, split_line_attributes, format_rust_code)?;
         } else if let Some(raw) = raw {
-            // Format raw text.
-            let indent = indentation_for(".", self.split_line_attributes)?;
+            // Format raw text. There is no input file, so anchor rustfmt's
+            // `.rustfmt.toml` lookup on a representative project file (#5631).
+            let indent = indentation_for(
+                representative_config_file(&project_rs_files()),
+                self.split_line_attributes,
+            )?;
             let formatted =
                 dioxus_autofmt::fmt_block(&raw, 0, indent).context("error formatting codeblock")?;
             println!("{}", formatted);
@@ -87,7 +95,6 @@ fn refactor_file(
     split_line_attributes: bool,
     format_rust_code: bool,
 ) -> Result<(), Error> {
-    let indent = indentation_for(".", split_line_attributes)?;
     let file_content = if file == "-" {
         let mut contents = String::new();
         std::io::stdin().read_to_string(&mut contents)?;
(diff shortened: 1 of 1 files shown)


RuboGubo:
I would love to see this merged, can I help push this along somehow?

costajohnt:
Sorry for the slow reply, and thanks for the offer.

Nothing is blocked on my end. The branch is one file, +59/-4, still merges cleanly against main, and CI was green across the full matrix on the last run. It is just waiting on a maintainer review.

The most useful thing you could do is confirm it fixes your case. Point `dx` at the branch, run `dx fmt --raw` and `dx fmt -f` in a project with a `.rustfmt.toml` that differs from the defaults (`hard_tabs`, `tab_spaces`, or `max_width` show the difference clearly), and check that the output honors the config. A second person saying "this fixed the formatting mismatch I was hitting" is worth more on a stale PR than anything I can add to it myself, and if it does not fix your case I would rather find that out now than after it merges.

Happy to rebase or split it up if a reviewer wants it in a different shape.

PR #5636, opened 2026-06-18
fix(rsx): `quote_spanned!` -> `quote!`. Resolves clippy::useless_format lints.

### LINKED ISSUE #5635: Nightly clippy emits `clippy::useless_format` for bare interpolations in dev builds.
**Problem**

`rust-clippy` [PR#17060](https://github.com/rust-lang/rust-clippy/pull/17060) fixed a false-negative for `clippy::useless_format` by changing how it resolves a `format!` call's context inside macros. 

In dev builds, this hits `ifmt.rs`'s [simple expression block](https://github.com/DioxusLabs/dioxus/blob/b00ff7525901984f3a07fa5cfa175b537b64630b/packages/rsx/src/ifmt.rs#L231) where `quote_spanned! { raw.span() => ::std::format!(#raw) }` gives the generated `format!` the user's source span. As a result, the format call is analyzed as the user's code rather than belonging to the macro expansion.

**Steps To Reproduce**

- Create a bare interpolation element.
```rust
pub fn clippy_render(label: &str) -> Element {
    rsx! {
        span { "{label}" }
    }
}
```
- Run `cargo +nightly clippy` and notice the lint fires.
- Run `cargo +nightly clippy --release` and notice the lint doesn't fire.

**Expected behavior**

Clean `cargo clippy`

**Environment:**

- Dioxus version:  v0.7.9
- Rust version:    rustc 1.98.0-nightly (9e2abe0c6 2026-06-16), should hit on any nightly since 2026-05-30
- OS info:         Linux (7.0.11-arch1-1)
- App platform:   Web

**Questionnaire**

scottfones:
## Problem

`ifmt.rs`'s `is_simple_expr` path emits `format!` with `quote_spanned! { raw.span() => ... }`, giving the call the user's source span. After rust-lang/rust-clippy#17060 , `useless_format` resolves a `format!`'s macro context from its span, so clippy isn't aware of the call's macro-generated context.

## Fix

Using `quote!` gives it the macro's span instead, so clippy treats it as part of the `rsx!` expansion and ignores it.

The `#raw` literal keeps its source span, so rust-analyzer is unaffected.

## Verification

Confirmed: clippy no longer lints bare interpolations as before.

Confirmed rust-analyzer functionality in-editor: renaming a variable still rewrites its `"{var}"` inside `rsx!`.

Fixes #5635

### DIFF
--- a/packages/rsx/src/ifmt.rs
+++ b/packages/rsx/src/ifmt.rs
@@ -230,7 +230,7 @@ impl ToTokens for IfmtInput {
         // If the segments are not complex exprs, we can just use format! directly to take advantage of RA rename/expansion
         if self.is_simple_expr() {
             let raw = &self.source;
-            return quote_spanned! { raw.span() => ::std::format!(#raw) }.to_tokens(tokens);
+            return quote! { ::std::format!(#raw) }.to_tokens(tokens);
         }
 
         // build format_literal


ealmloff:
We forward the span to help rust analyzers highlighting/rename functionality. We should keep the span and `#[allow(clippy::useless_format)]` instead so we get both the rust analyzer features and no clippy warning

scottfones:
Thanks for the quick feedback @ealmloff !

I saw the comment about the RA functionality and wanted to maintain that with the fix. I did test that highlighting and renaming work. My understanding is that the interpolated `#raw` literal will keep its own source span with `quote!`.

Here's the code I'm testing on:

```rust
use dioxus::prelude::*;

pub fn render(renamed: &str) -> Element {
    rsx! {
        span { "{renamed}" }
    }
}
```

Here, `renamed` is an artifact of the renaming test.  Querying rust analyzer for references returns: `[[3, 14], [5, 17]]`, which matches the parameter definition and the reference inside the `rsx!` block.

I also tested a `format!` error with `{x}`. That pointed back to the user source (identical to using `quote_spanned!`).

I'm happy to go with the allow attribute if you still prefer that.

scottfones:
I've updated the commit, which should fix the CI. 

The first commit was a hybrid of the original code and my fix; I messed up when I was cleaning up my edits. I left the `raw.span() =>` from the `quote_spanned!`, which is no longer necessary. 

The corrected line is `return quote! { ::std::format!(#raw) }.to_tokens(tokens);`

Sorry for the mix up.

PR #5641, opened 2026-06-20
fix(manganis): support 32-bit Android targets in JNI callback

### LINKED ISSUE #5637: Support for the armv7-linux-androideabi target was dropped starting with Dioxus 0.7.4
**Problem**

Up to Dioxus 0.7.3, building and running on 32-bit Android targets worked correctly. Starting with 0.7.4, the following compilation error is emitted:

```
   2.418s  INFO  error: Only 64-bit Android targets are supported
   --> /Users/a13400/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/manganis-0.7.9/src/android/callback.rs:122:5
    |
122 |     compile_error!("Only 64-bit Android targets are supported");
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This check was introduced in #4842.

While 32-bit Android is largely obsolete on smartphones, it is still actively used on devices such as Sony Android TVs and various Android-based set-top boxes (STBs). I would appreciate continued support for this target.

In my case, support for 32-bit Android is a hard requirement, so I am currently unable to upgrade beyond Dioxus 0.7.3.

**Steps To Reproduce**

- `dx bundle --android --release --target armv7-linux-androideabi`

**Expected behavior**

The project should build successfully for 32-bit Android targets (armv7-linux-androideabi), as it did prior to Dioxus 0.7.4.

**Environment:**

- Dioxus version: 0.7.9
- Rust version: 1.93.1
- OS: macOS
- App platform: Android

**Questionnaire**

costajohnt:
Building dioxus for `armv7-linux-androideabi` (and any 32-bit Android target) fails since #4842 because `rust_callback` gates compilation behind `target_pointer_width = "64"` with a `compile_error!`.

The handler pointer crosses the JNI boundary as two `jlong` halves. The existing reconstruction already yields a pointer-width `usize`: on 64-bit both halves are significant; on 32-bit the pointer fits in the low half (high half zero), so the `as usize` cast drops the unused high bits and recovers the original pointer. Removing the gate lets the same code compile and run on both widths, per @ealmloff's suggestion in the issue.

Verified with `cargo check --target armv7-linux-androideabi -p manganis` (failed before, passes after) and `cargo check --target aarch64-linux-android -p manganis` (unchanged). The encode side that splits the pointer into the two halves lives in plugin-provided DEX bytecode, so this preserves the existing low-half wire contract rather than changing it.

Closes #5637

### DIFF
--- a/packages/manganis/manganis/src/android/callback.rs
+++ b/packages/manganis/manganis/src/android/callback.rs
@@ -117,10 +117,12 @@ unsafe extern "C" fn rust_callback<'a>(
     handler_ptr_low: jlong,
     location: JObject<'a>,
 ) {
-    // Reconstruct the pointer from two i64 values (for 64-bit pointers)
-    #[cfg(not(target_pointer_width = "64"))]
-    compile_error!("Only 64-bit Android targets are supported");
-
+    // Reconstruct the handler pointer from the two `jlong` halves it was split
+    // into for the JNI boundary. On 64-bit targets both halves are significant;
+    // on 32-bit targets (e.g. `armv7-linux-androideabi`) the pointer fits in the
+    // low half and the high half is zero, so casting the recombined value to a
+    // pointer-width `usize` drops the unused high bits and yields the original
+    // pointer either way.
     let handler_ptr_raw: usize =
         ((handler_ptr_high as u64) << 32 | handler_ptr_low as u64) as usize;
 


costajohnt:
Just checking in on this one, is there anything else you'd like me to adjust before it can go in? Happy to add tests or docs if that would help.

atty303:
I applied this PR to the latest `main` (ad05395479fa01f3eb7fe394661770c88334141c) branch and confirmed that it builds successfully for `armv7-linux-androideabi` and that the app launches.

The WebView is outdated, so the styling is broken in some places, but the app appears to be minimally functional.

```
rk3288:/ $ getprop ro.product.manufacturer
rockchip
rk3288:/ $ getprop ro.product.model
S7
rk3288:/ $ getprop ro.product.cpu.abi
armeabi-v7a
rk3288:/ $ getprop ro.product.cpu.abilist
armeabi-v7a,armeabi
rk3288:/ $ getprop ro.build.version.release
7.1.2
rk3288:/ $ getprop ro.build.version.sdk
25
```

jkelleyrtp (approved):

PR #5613, opened 2026-06-04
fix(router): preserve query and hash segments on child routes

josephgilby:
Disclosure: Fix and test case were written with assistance from Claude Code (Opus 4.8). 

Example application demonstrating the issue. Enter `/search?query=hello&word_count=1` into the address bar and see that the input box and counter box don't have the values from query segment. The address bar updates if you then edit the input box, but if you then increment the counter, the address loses the query. Can the fix for this be backported to 0.7 as well?

```rust

use dioxus::prelude::*;

fn main() {
    dioxus::launch(|| {
        rsx! {
            Router::<Route> {}
        }
    });
}

#[derive(Routable, Clone, Debug, PartialEq)]
#[rustfmt::skip]
enum Route {
    #[route("/")]
    Home {},

    #[layout(AppLayout)]
        #[child("")]
        App { child: AppRoutes },
}

#[derive(Routable, Clone, Debug, PartialEq)]
#[rustfmt::skip]
enum AppRoutes {
    #[route("/search?:query&:word_count")]
    Search {
        query: String,
        word_count: usize,
    },
}

#[component]
fn AppLayout() -> Element {
    rsx! {
        div {
            p { "[AppLayout shell — child routes render below]" }
            Outlet::<Route> {}
        }
    }
}

#[component]
fn Home() -> Element {
    rsx! {
        h1 { "Home" }
        ul {
            li {
                Link {
                    to: "/search?query=hello&word_count=1",
                    "Go to /search?query=hello&word_count=1"
                }
            }
        }
    }
}

#[component]
fn Search(query: ReadSignal<String>, word_count: ReadSignal<usize>) -> Element {
    use_effect(move || {
        dioxus::logger::tracing::info!(
            "Search received: query = {:?}, word_count = {}",
            query(),
            word_count()
        );
    });

    const ITEMS: &[&str] = &[
        "hello",
        "world",
        "hello world",
        "hello dioxus",
        "hello dioxus-router",
    ];

    let results = use_memo(move || {
        ITEMS
            .iter()
            .filter(|item| {
                item.contains(&*query.read()) && item.split_whitespace().count() >= word_count()
            })
            .collect::<Vec<_>>()
    });

    rsx! {
        h1 { "parsed query = '{query}'  |  parsed word_count = {word_count}" }
        p {
            "If this shows query = '' and word_count = 0 while the URL says "
            "?query=hello&word_count=1, the query was dropped at the #[child] boundary."
        }

        input {
            oninput: move |e| {
[...]

PR #5655, opened 2026-07-01
Fix/router child dynamic prefix pr

rjharmon:
### Intent

Make `#[child("/path/:dyn")]` and multi-level `#[child]` composition work end-to-end at parse, render, and `Link` href generation.
- End-to-end support for `#[child("/path/:dyn")]`: parent dynamic-segment values now round-trip through parse, render, and Link href correctly.
- Support for 2+ (arbitrary) levels of #[child] composition, so deeply nested child routers no longer drop the deepest dynamic at render or drop outer captured values from `Link` hrefs.
- `#[child("/:..rest")]` composes with children that carry only query or hash specs

### Approach

Preserves PR #5613's query/hash forwarding and PR #4994's child-router layout shape; layered on
top of both.  Backstory:

> PR #5613 built trailing from all remaining URL segments and delegated the whole
> thing to `child::from_str` FIRST, then parsed the parent's route-segment definitions inside the
> `Ok` arm.  So its ordering was ***parent-last*** 

Implementation changed so the parent route's (dynamic) segment definition are consumed via the
nested try_parse chain first, and child::from_str(&trailing) runs only after segments has been
advanced to the child's tail.

## test notes

- Depth-2 `Link` href exercises the `Arc<dyn Fn(R) -> String>` widening: the emitted closure move-captures the parent's dynamic-segment values.
- Depth-3 `Link` href exercises `format_route_via_chain<R>`, which walks any outer `ChildRouteMapping<R>` chain. The multi-level composition the depth-2 test cannot reach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4041,6 +4041,7 @@ dependencies = [
  "serde",
  "tokio",
  "tracing",
+ "trybuild",
  "url",
 ]
 
--- a/packages/router-macro/src/lib.rs
+++ b/packages/router-macro/src/lib.rs
@@ -289,7 +289,9 @@ impl RouteEnum {
         let mut endpoints = Vec::new();
 
         let mut layouts: Vec<Layout> = Vec::new();
-        let mut layout_stack = Vec::new();
+        // Each active layout is paired with the nest depth at which it was opened, so
+        // #[end_nest] can detect a layout that would outlive its enclosing nest.
+        let mut layout_stack: Vec<(LayoutId, usize)> = Vec::new();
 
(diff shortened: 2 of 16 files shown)


rjharmon:
Redundantly, I'll say:
- Added diagnostic error for subroutes missing args
- Added diagnostic error for missing #[end_layout] before #[end_nest] before later-route-variant

Rebased on latest main branch too.  Sorry for the extra force-push; I'd forgotten to run `cargo fmt`

rjharmon:
@ealmloff :eye: in case you missed it

PR #5658, opened 2026-07-01
native-dom: Add a way to generate synthetic click events

hovinen:
This is needed for the new dioxus testing library. Tests need a way to interact with the the system under test -- in this case, to tell them that the user has clicked a component. Previously there was no way to construct the required `NativePointerData` struct outside the native-dom crate itself.


nicoburns (approved):
A bit awkward, but seems fine. I feel like we could probably come up with something a bit more elegant. But I'm not quite sure what yet.

PR #5664, opened 2026-07-03
fix: avoid Google Fonts in dev templates

### LINKED ISSUE #5660: `dx serve`'s dev-mode HTML template hardcodes a Google Fonts request
**Problem**

The HTML shell that `dx serve` injects in dev mode (used to render the
"Your app is being rebuilt" hot-reload toast) hardcodes an `@import` of
Inter from Google Fonts, unconditionally, on every page load:

```html
<style>
    /* Inter Font */
    @import url('https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap') layer;

    #dx-toast-template {
        display: none;
        visibility: hidden;
    }

    .dx-toast-inner {
        ...
        font-family: "Inter", sans-serif;
        ...
    }
</style>
```

This is generated by the CLI/dev server itself — it does not come from
application code, there's no config flag to disable it or point it at a
self-hosted font.

### Why this is a problem

- It makes running tests locally add 300ms for each.
- It adds an external network dependency (`fonts.googleapis.com` /
  `fonts.gstatic.com`) to every dev page load, purely to style a toast most
  developers rarely see.
- On a cold cache (first request in a session, CI runners, offline/airgapped
  dev environments, flaky networks) this can add 150-300ms+ to page load,
  which is easy to mistake for an app performance regression until you
  inspect the rendered HTML and realize it's unrelated to the app's own
  assets.
- It's silent — nothing in the project's own source (CSS, HTML template,
  `Dioxus.toml`) references Google Fonts or "Inter", so this is easy to miss
  when auditing where a page's network requests come from.
- No offline/self-hosted alternative: teams that vendor all fonts locally
  for privacy/reliability/CSP reasons have no way to make the dev toast use
  a local font instead.

** Suggested fix **

- Self-host (inline or embed) whatever font the toast needs so `dx serve`
  never depends on an external network call, or
- Fall back to a system font stack (e.g. `system-ui, sans-serif`) for the
  toast instead of pulling in a webfont at all, or
- At minimum, make this opt-out/configurable (e.g. a `Dioxus.toml` flag) so
  projects that already avoid third-party font CDNs can disable it.

**Steps To Reproduce**

- Open network tab
- Run `dx serve`

**Expected behavior**

No external requests

**Screenshots**

Image

**Environment:**

- **Dioxus version**: v0.7.9
- **Rust version**: 1.96.0
- **OS info**: Ubuntu 24.04.4 LTS
- **App platform**: web (default feature), with server/mobile features also available

**Questionnaire**

cyphercodes:
Fixes #5660.

This removes the Google Fonts imports from the built-in development HTML templates and switches the dev UI to system font stacks instead. That keeps `dx serve` from making external font requests for the Dioxus dev overlay/loading page.

Tests:
- `cargo test -p dioxus-cli build::web::tests::dev_templates_do_not_request_google_fonts -- --exact`
- `cargo fmt --check`
- `git diff --check`


Greenheart (commented):
Thanks for the contribution! I think this looks good.

jkelleyrtp (approved):

PR #5675, opened 2026-07-11
fix(cli): preserve original image bytes when asset!() requests no transform (#5642)

### LINKED ISSUE #5642: asset!() re-encodes PNG by default and significantly increases file size
## Description

In a Dioxus 0.7 web app, PNG assets included via `asset!()` appear to be re-encoded by default during `dx bundle`.

This can significantly increase the size of already optimized production images.

For example, in my project:

```text
public/assets/forest-background.png        193K
dist/public/assets/forest-background.png   976K
dist/public/assets/forest-background-dxh...png 976K
```

The same problem happens with other optimized PNG files:

```text
public/assets/forest-background-winter.png        224K
dist/public/assets/forest-background-winter.png   806K

public/assets/satchel-background-winter.png       212K
dist/public/assets/satchel-background-winter.png  1.1M
```

The main issue is that `asset!()` appears to optimize/re-encode images by default, but for already optimized PNGs this makes the files much larger.

## Expected behavior

By default, `asset!()` should not change image bytes.

I would expect `asset!()` to:

```text
1. copy/bundle the asset
2. optionally hash the filename
3. preserve the original file bytes unless image processing is explicitly requested
```

In other words, this should preserve the original optimized PNG:

```rust
img {
    src: asset!("public/assets/forest-background.png")
}
```

Image transcoding/resizing/re-encoding should only happen when explicitly requested with something like:

```rust
AssetOptions::image()
    .with_format(...)
    .with_size(...)
```

or another explicit optimization option.

## Actual behavior

`asset!()` re-encodes PNG files by default, and the output PNG can become much larger than the source PNG.

Example:

```text
source:
-rw-r--r--  public/assets/forest-background.png        193K

after dx bundle:
-rw-r--r--  dist/public/assets/forest-background.png   976K
-rw-r--r--  dist/public/assets/forest-background-dxh...png 976K
```

This is surprising because the original file is already optimized and no explicit image conversion/resizing was requested.

## Build command

We build the app with:

```bash
dx bundle --platform web --release --out-dir dist
```

## Steps to reproduce

1. Create a Dioxus 0.7 web app.
2. Place an already optimized PNG file in:

```text
public/assets/forest-background.png
```

3. Reference it through `asset!()`:

```rust
use dioxus::prelude::*;

fn App() -> Element {
    rsx! {
        img {
            src: asset!("public/assets/forest-background.png")
        }
    }
}
```

4. Build the app:

```bash
rm -rf dist target/dx
dx bundle --platform web --release --out-dir dist
```

5. Compare the source and output file sizes:

```bash
ls -lh public/assets/forest-background.png
find dist/public/assets -maxdepth 1 -name "*forest-background*" -exec ls -lh {} \;
```

## Observed result

```text
public/assets/forest-background.png        193K
dist/public/assets/forest-background.png   976K
dist/public/assets/forest-background-dxh...png 976K
```

## Environment

```text
Dioxus CLI: 0.7.9
Dioxus crate: 0.7.9
Platform: web
Build command: dx bundle --platform web --release --out-dir dist
Local OS: macOS arm64
CI OS: Ubuntu GitHub Actions
```

## Why this matters

This can heavily increase production bundle size and bandwidth usage.

For games and visual apps, many PNG assets are already manually optimized. A build tool silently re-encoding them by default can make the production output much worse.

A safer default would be:

```text
asset!() = bundle/hash/copy without changing bytes
AssetOptions::image() = explicit image processing
```

## Question

Is this behavior intended?

If yes, is there a documented way to disable image re-encoding for selected PNG assets?

If no, I think this should be treated as a bug or at least a problematic default behavior.

Example image attached:
Image

singhlovepreet9:
Fixes #5642.

## Problem

`asset!("image.png")` with no options re-encodes the image by default, which can **significantly increase** the size of an already-optimized PNG (the issue reports 193K → 976K).

## Root cause

An option-less `asset!(...)` resolves to `ImageAssetOptions::default()` = `format: ImageFormat::Unknown`, `size: ImageSize::Automatic`. In `process_image`, `format() == Unknown` falls into the catch-all `(Ok(image), _) => image.save(output_path)` arm, which **decodes and re-encodes** the image with the `image` crate's default PNG encoder — no optimization, and larger output for already-optimized sources.

## Fix

Add an early guard in `process_image`: when no output format conversion and no manual resize are requested (`Unknown` + `Automatic`), copy the source bytes verbatim (the asset is still content-hashed by filename) instead of decoding and re-encoding. This matches the expected behavior in the issue — *preserve the original bytes unless image processing is explicitly requested*.

Explicit `.with_format(...)` / `.with_size(...)` paths are unchanged: those still resize/convert/optimize as before.

```rust
if image_options.format() == ImageFormat::Unknown
    && image_options.size() == ImageSize::Automatic
{
    std::fs::copy(source, output_path)?;
    return Ok(());
}
```

## Test

Added `opt::image::tests::default_options_preserve_source_bytes`: it writes a `Best`-compressed source PNG carrying a `tEXt` chunk (which a decode→re-encode drops), runs `process_image` with default options, and asserts the output is **byte-identical** to the source — so the test fails if the default path ever re-encodes again.

```
test opt::image::tests::default_options_preserve_source_bytes ... ok
```

Ran with `cargo +1.93.0 test -p dioxus-cli` (repo MSRV 1.93).

---

Note: @GPNaslund mentioned interest in the issue — happy to defer if they already have a fix in flight; posting this since no PR was up.

### DIFF
--- a/packages/cli/src/opt/image.rs
+++ b/packages/cli/src/opt/image.rs
@@ -9,6 +9,20 @@ pub(crate) fn process_image(
     source: &Path,
     output_path: &Path,
 ) -> anyhow::Result<()> {
+    // Copy verbatim if no transform is requested (#5642)
+    if image_options.format() == ImageFormat::Unknown
+        && image_options.size() == ImageSize::Automatic
+    {
+        std::fs::copy(source, output_path).with_context(|| {
+            format!(
+                "Failed to copy image from {} to {}",
+                source.display(),
+                output_path.display()
+            )
+        })?;
+        return Ok(());
+    }
+
     let mut image = image::ImageReader::new(std::io::Cursor::new(&*std::fs::read(source)?))
         .with_guessed_format()
         .context("Failed to guess image format")?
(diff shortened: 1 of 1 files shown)


packages/cli/src/opt/image.rs
```diff
@@ -9,6 +9,25 @@ pub(crate) fn process_image(
     source: &Path,
     output_path: &Path,
 ) -> anyhow::Result<()> {
+    // If no image transformation was requested — no explicit output format and no
+    // manual resize — copy the source bytes verbatim instead of decoding and
+    // re-encoding. Re-encoding an already-optimized image (e.g. a hand-tuned PNG)
+    // can significantly *increase* its size, so by default we preserve the original
+    // bytes and only process the image when the user explicitly asks for a format
+    // or a size (see https://github.com/DioxusLabs/dioxus/issues/5642).
```
WaterWhisperer: It looks like AI-generated explanatory text that was submitted without sufficient review or cleanup. Six lines of comments narrating such a simple guard in a small function is unreasonable.

singhlovepreet9: Thanks for the feedback! You're completely right—the comment was unnecessarily verbose.I've shortened the comments to a single line, squashed the commits, and updated the PR. Let me know if everything looks good now!

PR #5677, opened 2026-07-11
fix(manganis): preserve original image bytes when no transform requested

### LINKED ISSUE #5642: asset!() re-encodes PNG by default and significantly increases file size

Functionhx:
Fixes #5642.

asset!() re-encodes PNG through the image crate, inflating file size (193K→976K). Default-option images are decoded and re-compressed instead of copied verbatim.

Added early return in process_image() — fs::copy byte-for-byte when no format conversion or resize is requested. Explicit transforms still go through full decode→resize→encode pipeline.

3 observations from isolation review (non-blocking):
- APNG now preserved instead of flattened
- Corrupted-image detection moved from decode-time to copy-time
- Zero-byte files pass through instead of erroring

### DIFF
--- a/packages/cli/src/opt/image.rs
+++ b/packages/cli/src/opt/image.rs
@@ -9,6 +9,25 @@ pub(crate) fn process_image(
     source: &Path,
     output_path: &Path,
 ) -> anyhow::Result<()> {
+    // If no image transformation was requested — no explicit output format and no
+    // manual resize — copy the source bytes verbatim instead of decoding and
+    // re-encoding. Re-encoding an already-optimized image (e.g. a hand-tuned PNG)
+    // can significantly *increase* its size, so by default we preserve the original
+    // bytes and only process the image when the user explicitly asks for a format
+    // or a size (see https://github.com/DioxusLabs/dioxus/issues/5642).
+    if image_options.format() == ImageFormat::Unknown
+        && image_options.size() == ImageSize::Automatic
+    {
+        std::fs::copy(source, output_path).with_context(|| {
+            format!(
+                "Failed to copy image from {} to {}",
+                source.display(),
+                output_path.display()
+            )
+        })?;
+        return Ok(());
+    }
+
     let mut image = image::ImageReader::new(std::io::Cursor::new(&*std::fs::read(source)?))
         .with_guessed_format()
         .context("Failed to guess image format")?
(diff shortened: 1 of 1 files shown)


packages/cli/src/opt/image.rs
WaterWhisperer, line 1: Oh, this looks exactly the same as #5675

WaterWhisperer (commented):
This PR contains the exact same commit as #5675 (`e2239b8`), which was opened earlier by @singhlovepreet9 . Why was the same commit submitted again from a different account?

Issue #5656, opened 2026-07-01
dx build re-encodes .webp files in public/ to VP8L lossless, dropping animation frames


DefectingCat:
## Summary

`dx build` (0.7.9) re-encodes all `.webp` files in the `public/` directory into VP8L lossless format during the build, despite the docs stating that the `public/` folder is **copied verbatim**. For animated WebP files, **all animation frames are dropped** (only the first frame survives). For static WebP files, the VP8 lossy source is re-encoded to VP8L lossless, often resulting in **7-8x larger** files.

The build log says `INFO Copying asset (N/1304): ...`, but the bytes are actually transcoded, not copied.

## Environment

```
dioxus 0.7.9 (bfcc111)
Platform: Darwin arm64 (macOS 15.5)
Rust: rustc 1.96.0 (ac68faa20 2026-05-25)
```

## Expected behavior

Per the [0.7 Assets docs](https://dioxuslabs.com/learn/0.7/essentials/ui/assets/):

> DX will automatically **copy** any files from your app's `/public` directory into the output `/public` directory.

The `public/` folder is said to be _merged_ into the output. Image optimization is documented to apply **only** to assets referenced via the `asset!()` macro — not to the `public/` directory.

## Actual behavior

All `.webp` files under `public/` are re-encoded to VP8L lossless:
- **Animated WebP** → single-frame static WebP (animation frames discarded)
- **Static VP8 (lossy) WebP** → VP8L lossless (size increases 7-8x)
- SVG, ICO → copied verbatim (as documented ✅)

## Reproduction

1. Place an animated WebP (VP8X + ANMF frames) at `public/images/anim.webp`
2. Reference it from a component via a plain string path (NOT `asset!()`):
   ```rust
   rsx! { img { src: "/images/anim.webp" } }
   ```
3. Run `dx build --release`
4. Compare the source file against the build output

### Before/after comparison

| File | Source | Build output (`target/dx/.../public/`) |
|------|--------|----------------------------------------|
| `anim.webp` (animated) | 462,810 B · VP8X · 79 ANMF frames | 61,878 B · VP8L · 0 frames |
| `anim_transparent.webp` (animated) | 4,734,700 B · VP8X · 79 frames | 63,680 B · VP8L · 0 frames |
| `static_01.webp` (VP8 lossy) | 54,650 B | 461,932 B (↑8.5x) |
| `static_02.webp` (VP8 lossy) | 51,136 B | 353,962 B (↑6.9x) |

[...]

PR #5678, opened 2026-07-14
fix(cli): preserve WebP assets copied from public directories

luvth:
Public directory assets are documented as being copied verbatim, but WebP
files were incorrectly routed through the image optimization pipeline.

This change separates public asset copying from asset!() optimization and adds
regression coverage for animated and lossy WebP files.


luvth:
Fixes #5656

sarendipitee:
Would really love to see this merged too

ealmloff:
Can we make this the default if no options are passed to manganis/assets? The public directory should derive from the file extension which results in no new options. If manganis is not told to convert the format, we shouldn't touch the file

PR #5683, opened 2026-07-16
Fix/router child dynamic prefix 0.7

rjharmon:
Same as #5655  but on [edit] ~~main~~ 0.7 branch

Includes a resync of the Cargo.lock that seems like it was misplaced during the path to 0.7.9


rjharmon:
no let chains on these versions of the patches

PR #5697, opened 2026-07-23
Support single-dimension image resizing

Charlie284:
## Summary

- add `with_width` and `with_height` image asset options
- preserve the source aspect ratio when the CLI resizes a single dimension
- cover width- and height-constrained output dimensions with focused tests

Closes #5458

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p dioxus-cli opt::image::tests`
- `cargo check -p dioxus-cli`
- `cargo test -p manganis-core`
- `cargo clippy -p dioxus-cli --tests -- -D warnings`
- `cargo test --workspace --tests`

### DIFF
--- a/packages/cli/src/opt/image.rs
+++ b/packages/cli/src/opt/image.rs
@@ -15,8 +15,18 @@ pub(crate) fn process_image(
         .decode();
 
     if let Ok(image) = &mut image {
-        if let ImageSize::Manual { width, height } = image_options.size() {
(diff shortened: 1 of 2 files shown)


nicoburns (changes requested):
Maybe if you call both .with_width and .with_height it ought to apply Manual?

Charlie284:
Addressed this in 149c5f3: chaining .with_width() and .with_height() now selects ImageSize::Manual in either order, with tests covering both cases. The branch is also updated with main. @nicoburns, could you take another look when you get a chance?

PR #5698, opened 2026-07-23
test(native-dom): reproduce "invalid key" panic from stale node_id_mapping (#5182)

nicoburns:
## Summary

Adds `packages/native-dom/tests/stale_node_mapping.rs` with tests that reproduce the "invalid key" slab panic reported in #5182, on latest `main` (blitz-dom `0.3.0-beta.1`). **3 of the 4 tests currently fail** — they encode the desired behavior and should pass once the underlying bug is fixed.

Mechanism (confirmed by these tests):
1. `remove_node`/`replace_node_with` detaches a subtree; `node_id_mapping` entries for the subtree are intentionally kept so dioxus-core can reuse the DOM nodes.
2. `assign_node_id` → `remove_node_if_unparented` later drops the detached subtree, but only the mapping for the ElementId being reassigned is updated. Mappings for *descendants* of the dropped subtree now point at freed slab slots.
3. A freed slot gets reused for a brand-new (not-yet-parented) node, and the stale descendant ElementId is reused by dioxus-core via `AssignId` → `remove_node_if_unparented` incorrectly drops the new node → "invalid key" panic later in the same render (`DocumentMutator::node_at_path` / `add_children_to_parent`), matching the stack trace in #5182.

Tests (driven through a real `VirtualDom` + `DioxusDocument`, re-rendering via `mark_dirty(ScopeId::APP)` + `poll`):
- `template_swaps_do_not_panic`: minimal deterministic repro — panics after just 3 re-renders swapping between three simple templates (`article>div>div`, bare text, `section` with keyed list).
- `template_swaps_do_not_leave_stale_mappings`: asserts the invariant that every `node_id_mapping` entry points at a live node; detects the stale-mapping precursor at step 4.
- `fuzz_template_swaps_do_not_panic`: pseudo-random (deterministic LCG) template mix; panics with "invalid key" within a handful of steps.
- `keyed_list_grow_shrink`: keyed list grow/shrink regression guard (currently passes).

Failing output:
```
thread 'fuzz_template_swaps_do_not_panic' panicked at blitz-dom-0.3.0-beta.1/src/mutator.rs:114:42: invalid key
thread 'template_swaps_do_not_leave_stale_mappings' panicked:
  stale node_id_mapping entries (ElementId, NodeId) after step 4: [(7, 38), (10, 43)]
```

Note: I could not reproduce via the earlier proposed fix path with blitz 0.2.4 because latest main already reproduces; the panic site differs slightly from #5182's trace (`node_at_path` vs `add_children_to_parent`) but is the same root cause — a stale mapping resolving to a freed/reused slab slot.

Link to Devin session: https://app.devin.ai/sessions/07ce0796079c4833b0226b4c1f592eab
Requested by: @nicoburns

### DIFF
--- a/packages/native-dom/tests/stale_node_mapping.rs
+++ b/packages/native-dom/tests/stale_node_mapping.rs
@@ -1,0 +1,234 @@
+//! Regression tests for stale `node_id_mapping` entries after node removal.
+//! See https://github.com/DioxusLabs/dioxus/pull/5182
+//!
+//! When `remove_node` (or `replace_node_with`) detaches a node, the
+//! `node_id_mapping` entries for the node's subtree are (intentionally) not
+//! cleared, so that dioxus-core can reuse the detached DOM nodes. However,
+//! when the detached subtree is later dropped via `remove_node_if_unparented`
+//! (called from `assign_node_id`), only the mapping for the ElementId being
+//! reassigned is updated. Any *descendant* of the dropped subtree that has its
+//! own ElementId mapping is left pointing at a freed slab slot.
+//!
+//! When that slab slot is reused for a brand-new (not yet parented) node and
+//! the stale ElementId is then reused by dioxus-core via `AssignId`,
+//! `remove_node_if_unparented` incorrectly drops the new node, causing an
+//! "invalid key" slab panic in blitz-dom's `DocumentMutator` later in the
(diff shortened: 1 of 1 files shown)


nicoburns:
Landed upstream in the blitz repo

PR #5702, opened 2026-07-23
fix(cli): prevent duplicate SSG output on rebuild

### LINKED ISSUE #5608: SSG builds create duplicate contents if you don't clear the build output
Every time when I try to make a release, trying to build with this command `dx build --verbose --trace --web --fullstack true --features fullstack --release --ssg` my app breaks. Most of the time I use `dx serve --web` for development.
And I get this error on browser's console. This error breaks scrolling, one page superposes with another another page, button clicks stops working, etc. And then I have to git bisect, which wastes my entire day (might be a skill issue).

Image

How it would have been if focus_trap.js was written in rust? Are there limitations in dioxus to implement this? Take this as a question not as an issue

sanmaxdev:
## Summary

- Run SSG builds in client-first sequential order so `index.html` is refreshed before prerendering starts.
- Keep the existing parallel path for non-SSG builds.
- Correct the `--force-sequential` help text to match the actual client-first order.

## Root cause

The server build could finish first during a parallel fullstack build and start SSG against the previous client output. On repeated builds, that meant prerendering could use already-generated SSG HTML as its template and inject the page again.

Requiring sequential order for SSG makes the client rewrite the clean HTML template before the server starts prerendering.

## Verification

- `cargo check -p dioxus-cli --tests`
- Regression coverage for parallel, explicitly sequential, and SSG build-order selection
- `git diff --check`

Fixes #5608

### DIFF
--- a/packages/cli/src/cli/build.rs
+++ b/packages/cli/src/cli/build.rs
@@ -156,8 +156,7 @@ impl CommandWithPlatformOverrides<BuildArgs> {
         tracing::info!("Building project...");
 
         let ssg = self.shared.ssg;
-        let force_sequential =
-            should_build_sequentially(self.shared.force_sequential_build(), ssg);
+        let force_sequential = should_build_sequentially(self.shared.force_sequential_build(), ssg);
         let mode = match self.shared.fat_binary {
             true => BuildMode::Fat,
             false => BuildMode::Base,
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
Thanks! We could still do builds in parallel as long as we keep the ssg rendering after web finishes its index.html update, but this is a fine solution for now

PR #5704, opened 2026-07-25
fix(fullstack): allow the server URL to be set more than once

SilasMarvin:
`set_server_url` stored the root URL in a `OnceLock` and unwrapped, so a second call panicked — pinning the base URL for the life of the process. A native client that talks to more than one backend can't work with that.

Swaps the `OnceLock` for an `RwLock<Option<&'static str>>` so the most recent value wins. `fetch_inner` already reads the URL per request, so the change applies to the next server-function call with no restart. `get_server_url` still returns `""` when unset, so `launch`'s devserver-default check is unaffected. Mirrors the neighbouring `REQUEST_HEADERS`, already a re-settable `Mutex`.

Also removes a latent panic: the native and web base-path blocks in `launch` can compile into one binary and both call `set_server_url`, the second unconditionally — panicking whenever `base_path()` is `Some`.

Tested: `cargo fmt`, `check`, `clippy -D warnings`, `test --all` all pass.

### DIFF
--- a/packages/fullstack/src/client.rs
+++ b/packages/fullstack/src/client.rs
@@ -9,7 +9,7 @@ use headers::{ContentType, Header};
 use http::{Extensions, HeaderMap, HeaderName, HeaderValue, Method, StatusCode, response::Parts};
 use send_wrapper::SendWrapper;
 use serde::{Serialize, de::DeserializeOwned};
-use std::sync::{LazyLock, Mutex, OnceLock};
+use std::sync::{LazyLock, Mutex, OnceLock, RwLock};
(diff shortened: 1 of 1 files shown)


ealmloff (approved):
This seems reasonable. It lets you build re-exports of specific server functions that set and unset the url for the call. We should probably have an api to map the url for specific re-exports in the long term, but this is a fine workaround for now

SilasMarvin:
Thanks for the review and merge @ealmloff

This is a super cool project. I really appreciate the work you and the team are doing here.

Issue #5644, opened 2026-06-22
#[css_module] creates FOUC on page load


pkleineb:
**Problem**
When using `#[css_module]` for styling reloading the app using `Strg + r` flashes the screen with unstyled html where after a second or so the styles get loaded. This also happens for when you use an `<a>`-tag to redirect the user.

**Steps To Reproduce**

1. create a blank dioxus project using `dx new`
2. Use `#[css_module]` for styling all components
3. Start the app and reload the page.

This is the modified Bare-Bones example, created using `dx new test`
```rust
use dioxus::prelude::*;

#[derive(Debug, Clone, Routable, PartialEq)]
#[rustfmt::skip]
enum Route {
    #[layout(Navbar)]
    #[route("/")]
    Home {},
    #[route("/blog/:id")]
    Blog { id: i32 },
}

#[css_module("/assets/main.css")]
pub struct Styles;

const FAVICON: Asset = asset!("/assets/favicon.ico");
//const MAIN_CSS: Asset = asset!("/assets/main.css");
const HEADER_SVG: Asset = asset!("/assets/header.svg");

fn main() {
    dioxus::launch(App);
}

#[component]
fn App() -> Element {
    rsx! {
        document::Link { rel: "icon", href: FAVICON }
        //document::Link { rel: "stylesheet", href: MAIN_CSS }
        Router::<Route> {}
    }
}

#[component]
pub fn Hero() -> Element {
    rsx! {
        div {
            class: Styles::hero,
            img { src: HEADER_SVG, class: Styles::header }
            div { class: Styles::links,
                a { href: "https://dioxuslabs.com/learn/0.7/", "📚 Learn Dioxus" }
                a { href: "https://dioxuslabs.com/awesome", "🚀 Awesome Dioxus" }
                a { href: "https://github.com/dioxus-community/", "📡 Community Libraries" }
                a { href: "https://github.com/DioxusLabs/sdk", "⚙️ Dioxus Development Kit" }
                a { href: "https://marketplace.visualstudio.com/items?itemName=DioxusLabs.dioxus", "💫 VSCode Extension" }
                a { href: "https://discord.gg/XgGxMSkvUM", "👋 Community Discord" }
            }
        }
    }
}

/// Home page
#[component]
fn Home() -> Element {
    rsx! {
        Hero {}
        Echo {}
    }
}

/// Blog page
#[component]
pub fn Blog(id: i32) -> Element {
    rsx! {
        div {
            class: Styles::blog,

[...]

PR #5707, opened 2026-07-26
Fix css_module stylesheets missing from SSR head after the first request

### LINKED ISSUE #287: coponent rendering takes a while after reloading but is instantainiuos after building
im using the Card component as in the dioxus components exaples and just after compilation when openning it loads fast but when reloading the page it doesnt render but the text

here is the part of the sorce code that is involved 

[main.txt](https://github.com/user-attachments/files/30192321/main.txt)
[assets.txt](https://github.com/user-attachments/files/30192320/assets.txt)
[custom.txt](https://github.com/user-attachments/files/30192328/custom.txt)

https://github.com/user-attachments/assets/52fd4798-6faf-4b6c-818a-96ca26330036

nicoburns:
## Summary

Fixes DioxusLabs/dioxus-components#287 ("component rendering takes a while after reloading but is instantaneous after building").

The `#[css_module]`-generated `__CssIdent::deref` injected its stylesheet via `document().create_link(...)` guarded by a **process-global `static OnceLock`**. On a fullstack server the `ServerDocument` collecting head elements is per-request, so only the *first* request after the server boots got the `<link rel="stylesheet">` in the SSR `<head>`. Every subsequent request (i.e. any page reload) was served with no stylesheet link at all, and the page rendered as unstyled text until WASM hydration re-inserted the link client-side.

Reproduced with a minimal fullstack app: the first `curl` of `/` contained the stylesheet link in the head; the second did not.

The fix deduplicates per document (per `VirtualDom`) instead of per process:

- New `dioxus_document::insert_stylesheet(href)` helper that reuses the same root-scope `LinkContext` dedup (keyed `"{href}|stylesheet"`) as the `document::Link` component, then gates on `create_head_component()` so hydration doesn't insert a duplicate link when the server already rendered it:
  ```rust
  pub fn insert_stylesheet(href: &str) {
      // dedup first so repeated calls (deref runs on every render)
      // don't create/consume extra hydration entries
      if !should_insert_link(href, Some("stylesheet")) { return; }
      let document = document();
      if !document.create_head_component() { return; }
      document.create_link(/* rel="stylesheet" href={href} */);
  }
  ```
- The `css_module` macro's generated `Deref` now just calls `dioxus::document::insert_stylesheet(&ASSET.to_string())` instead of the `OnceLock` block.

With this, every SSR request emits the component stylesheet in the head, and hydration skips re-inserting it (previously the client always appended a duplicate link).

Verified: `curl`ing a fullstack repro repeatedly now returns the `<link rel="stylesheet">` in the head on every request; `cargo clippy --workspace --all-features --all-targets -D warnings` and `cargo fmt --check` pass.

Link to Devin session: https://dioxus.staging.devinenterprise.com/sessions/05fa859d90474349a5e9f0880dddbda0
Requested by: @nicoburns

---

#### Devin Review
| Status | Commit |
|---|---|
| ⚪ Not started | — |

> [Run Devin Review](https://dioxus.staging.devinenterprise.com/review/DioxusLabs/dioxus/pull/5707?analyze=true)

<a href="https://dioxus.staging.devinenterprise.com/review/DioxusLabs/dioxus/pull/5707" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-staging-dark.svg?v=1">
    Open in Devin Review (Staging)
  </picture>
</a>

### DIFF
--- a/packages/document/src/elements/link.rs
+++ b/packages/document/src/elements/link.rs
@@ -137,6 +137,31 @@ pub fn Link(props: LinkProps) -> Element {
     VNode::empty()
 }
 
+/// Insert a stylesheet `<link>` element into the head of the current document.
+///
+/// The stylesheet is deduplicated within each document: calling this multiple times with the
+/// same href (or rendering a [`Link`] component with the same href and `rel="stylesheet"`)
+/// only inserts the link once per document.
+pub fn insert_stylesheet(href: &str) {
+    // Deduplicate before creating the head component so that repeated calls with the same
+    // href (e.g. on every render) don't create or consume extra hydration entries.
+    if !should_insert_link(href, Some("stylesheet")) {
+        return;
+    }
+
+    let document = document();
+    if !document.create_head_component() {
+        return;
+    }
+
+    document.create_link(
+        LinkProps::builder()
+            .rel(Some("stylesheet".to_string()))
+            .href(Some(href.to_string()))
+            .build(),
+    );
+}
+
 #[derive(Default, Clone)]
 struct LinkContext(DeduplicationContext);
 
--- a/packages/manganis/manganis-macro/src/css_module.rs
+++ b/packages/manganis/manganis-macro/src/css_module.rs
@@ -112,23 +112,12 @@ pub(crate) fn expand_css_module_struct(
             pub struct __CssIdent { pub inner: &'static str }
 
             use std::ops::Deref;
-            use std::sync::OnceLock;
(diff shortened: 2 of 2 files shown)


ealmloff (approved):

mcmah309:
Related https://github.com/DioxusLabs/dioxus/issues/5644

Issue #5596, opened 2026-05-28
Hotpatch replay includes uncaptured workspace reverse-dependents


mirsella:
## Problem

`dx serve --hot-patch --package app` can fail on the first thin/hotpatch rebuild with:

```text
Build failed: Missing rustc args for replay: 'app_tests'
```

The initial app build succeeds and launches. The error appears after editing or touching a source file in the served package.

## Reproduction Shape

Workspace layout:

```toml
# Cargo.toml
[workspace]
members = [
    "app",
    "app_tests",
]

[workspace.dependencies]
app = { path = "app" }
app_tests = { path = "app_tests" }
```

```toml
# app_tests/Cargo.toml
[dependencies]
app = { workspace = true }
```

Run:

```sh
dx serve --hot-patch --package app
```

Then change a file under `app/src/`.

Observed with `dx 0.7.9 (3e43ffa)`.

## What Appears To Happen

The initial fat/app build captures rustc args only for crates in the served app build graph. For example, the captured args directory may contain files for the served package and its app dependencies, but not for unrelated workspace reverse-dependents:

```text
app.bin.json
app_dependency.lib.json
```

It does not contain `app_tests.lib.json` or `app_tests.bin.json`, because `app_tests` is not part of the served app build.

On the thin/hotpatch rebuild, `patch_rebuild` adds the changed crate and then walks `workspace_dependents_of`, which includes `app_tests` because it is a workspace member depending on `app`. Later `compile_workspace_hotpatch` tries to replay rustc args for every crate in that set. Since `app_tests` was not captured by the initial app build, `workspace_hotpatch_replay_args` errors with `Missing rustc args for replay: 'app_tests'`.

## Expected Behavior

Hotpatch replay should not fail because an unrelated workspace reverse-dependent was not part of the initial served app build. Possible fixes:

- restrict the workspace reverse-dependent cascade to crates that have captured rustc args / are in the served hotpatch graph, or
- gracefully skip uncaptured reverse-dependents that are not needed for the served package, or
- ensure all workspace reverse-dependents selected for replay have their rustc args captured before thin rebuilds.

rami3l:
@mirsella Many thanks for this issue and the patch!

To second this, actually it has been reported by myself once before: https://github.com/DioxusLabs/dioxus/issues/5540

mirsella:
oh sorry didn't find your previous issue for some reason ! I'll close this issue then

PR #5746, opened 2026-08-07
Fix hotpatch cascade replaying workspace crates not built for the active target

### LINKED ISSUE #5540: `dx serve --hotpatch` tries to replay uncaptured workspace crates
**Problem**

Upstreamed from https://github.com/Pangpang-Studio/monoxide/issues/158.

Our workspace setup worked last month when we were using Dioxus CLI 0.7.4, so some regression must have happened since then.

When trying to trigger a `--hotpatch` reloading `dx serve` in the `playground` "example" binary target of the workspace, `subsecond`'s DAG scan somehow thinks that the root binary target also needs to be rebuilt.

However, `playground` doesn't rely on the root binary target at all so it wasn't built. Thus, when `dx` is trying to replay the build, this will give a `[dev] Build failed: Missing rustc args for replay` error.

Manually rebuilding with `r` works, but this is against the point of enabling `--hotpatch`.

**Expected behavior**

The workspace reloads correctly.

**Actual behavior**

We see verbose error logs like so:

```console
[dev] Verbose logging is now on 
[dev] Patch rebuild: changed_crates=["monoxide_font"], modified_crates={"monoxide_font", "playground", "monoxide"} 
[dev] Changed crates dag using {"monoxide_font", "playground", "monoxide"} 
[dev] replaying crates: ["monoxide_font", "monoxide"] 
[dev] Setting builder to failed state 
[dev] Build failed: Missing rustc args for replay: 'monoxide' 
```

... and the hotpatch never succeeded.

**Steps To Reproduce**

To make the investigation simpler, I have made a minimal reproduction that mirrors the basic structure of our project `monoxide` under the `repros/issue-158-hmr` subdirectory.

- Download this reproducer: https://github.com/Pangpang-Studio/monoxide/tree/3eac434ec9bc245d59177cbc0b9e9da53155d397/repros/issue-158-hmr
- Follow the reproducing instructions in https://github.com/Pangpang-Studio/monoxide/blob/3eac434ec9bc245d59177cbc0b9e9da53155d397/repros/issue-158-hmr/README.md
- Observe logs like so:
   ```console
   Patch rebuild: changed_crates=["mini_font"], modified_crates={"mini_font", "mini_monoxide", "playground"}
   replaying crates: ["mini_font", "mini_monoxide"]
   Build failed: Missing rustc args for replay: 'mini_monoxide'
   ```

**Environment:**

- Dioxus version: 0.7.9
- Rust version: 1.95.0 stable
- OS info: macOS 26.4.1
- App platform: `subsecond`-only, not using Dioxus

humdrum00001010:
`patch_rebuild`'s cascade decides which workspace crates a hot patch must carry. It walks `workspace_dependents_of`, which reads the whole-workspace Cargo metadata graph — and that graph is target-agnostic. So the cascade could pick up a crate the active build never compiled: a native-only sibling, or a dependency behind `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` while serving wasm. No rustc invocation was captured for such a crate, so `compile_workspace_hotpatch` hard-errored with `Missing rustc args for replay: '<crate>'`, and hot-patching stayed broken until `dx serve` was restarted.

**First commit** excludes those crates on both routes into the replay set — the crates the user edited and the dependents found by walking — and prunes the walk through them, since anything reachable only via a crate this target never compiled doesn't consume the change either. A dependent that also has a compiled path back to the edited crate is still reached along that path, so excluding one route never drops a crate that genuinely needs the patch. The walk is extracted as `crates_to_replay` and covered by eight tests over a fake dependents graph, including the "cascade is unchanged when every crate is built" baseline.

**Second commit** stops the wasted work this exposes: editing an excluded crate still ran a full patch cycle — thin rebuild, relink, jump table, client reload — with an empty replay list, because the tip is always rebuilt. It now returns before starting the build, while leaving behaviour untouched when a changed file maps to no workspace crate at all.

Verified on a workspace with a cfg-gated native-only dependency (`wasm_app` → `shared_lib`, plus `native_ffi` depending on `shared_lib` and reachable from `wasm_app` only for non-wasm targets):

| edit | before | after |
|---|---|---|
| `native_ffi` (never built for wasm32) | `Missing rustc args for replay: 'native_ffi'`, builder stuck | skipped, no patch cycle |
| `shared_lib` (shared, in the graph) | patch applies | patch applies, identical cascade |

`cargo test --workspace --tests` passes (208 suites), as do `cargo fmt` and `cargo clippy --all-targets -- -D warnings`.

Same root cause as #5596, which was closed as a duplicate of #5540 — see the repro and analysis posted there.

Closes #5540

### DIFF
--- a/packages/cli/src/build/builder.rs
+++ b/packages/cli/src/build/builder.rs
@@ -389,6 +389,15 @@ impl AppBuilder {
             krate == tip_crate_name || artifacts.workspace_rustc.contains_crate(krate)
         };
 
+        // Nothing the user edited is compiled for this target, so it cannot affect the running
+        // binary. Patching anyway would rebuild, relink and reload the client for no change.
+        if edit_cannot_reach_target(&changed_crates, in_target_graph) {
+            tracing::debug!(
+                "Ignoring patch rebuild for {build_id:?} since none of {changed_crates:?} are built for this target."
+            );
+            return;
+        }
+
         // On web, our patches are fully relocatable, so we don't need to worry about ASLR, but
         // for all other platforms, we need to use the ASLR reference to know where to insert the patch.
         let aslr_reference = match self.aslr_reference {
@@ -2221,6 +2230,18 @@ impl AppBuilder {
     }
 }
 
+/// Whether an edit cannot reach the running binary: it mapped to workspace crates, and not one
+/// of them is built for this target.
+///
+/// An empty `changed_crates` means the changed files mapped to no workspace crate at all, which
+/// says nothing about relevance - those edits keep going through the normal patch path.
+fn edit_cannot_reach_target(
+    changed_crates: &[String],
+    in_target_graph: impl Fn(&str) -> bool,
+) -> bool {
+    !changed_crates.is_empty() && !changed_crates.iter().any(|krate| in_target_graph(krate))
+}
+
 /// The crates whose objects a patch must carry: the crates the user edited, plus every workspace
 /// crate that transitively depends on one of them, plus `tip_crate_name` (always rebuilt).
 ///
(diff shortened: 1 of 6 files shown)


nicoburns:
Ack. Didn't mean to send you automated review feedback. Never-the-less, I'd be interested in whether the review comments make sense to you.

humdrum00001010:
Scope note: the indexing is ~2ms of a ~250ms patch here, under 1% — it only shows past ~400 members. The skip saves a reload on an uncommon edit. Only the replay fix changes whether --hot-patch works at all, and just for workspaces with a crate outside the served target.

humdrum00001010:
Seed case is patched. The crates you edit and the dependents found while walking now pass one exclusion check, so neither route reaches replay uncaptured. Covered by tests against a real cargo workspace with a cfg-gated native-only dep.

humdrum00001010:
should be separate PR? if so, let me know.

humdrum00001010:
@nicoburns The seed case wasn't that performance critical from my reproduction, but I patched. could you review this?

PR #5750, opened 2026-08-10
Introduce a function to create a synthetic form event

hovinen:
This introduces `synthetic_form_event` analogously to `synthetic_click_event`. It allows test code to construct a `NativeFormData` which can set the values of form elements. Otherwise test code would have no easy way to generate such events.

This also increases the visibility of `ElementId::new()` so that it is possible to construct `ElementId` from outside Dioxus. The dioxus-test crate namely needs a way to construct these.

### DIFF
--- a/packages/native-dom/src/events.rs
+++ b/packages/native-dom/src/events.rs
@@ -610,6 +610,16 @@ pub fn synthetic_click_event(node: &Node, modifiers: Modifiers) -> Box<dyn Any> 
     ))
 }
 
+pub fn synthetic_form_event(
+    value: impl Into<String>,
+    values: impl Into<Vec<(String, FormValue)>>,
+) -> Box<dyn Any> {
+    Box::new(NativeFormData {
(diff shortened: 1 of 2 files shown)


packages/core/src/arena.rs
```diff
@@ -16,7 +16,8 @@ impl ElementId {
     /// The root element within a render target.
     pub const ROOT: Self = Self(0);
 
-    pub(crate) const fn new(index: usize) -> Self {
+    /// Constructs an [`ElementId`] from the given raw `index`.
+    pub const fn new(index: usize) -> Self {
```
ealmloff: can you use `from_raw` instead of making this public?

hovinen: I somehow missed that, sorry. This change is removed and I've created https://github.com/DioxusLabs/dioxus-test/pull/25 to update dioxus-test correspondingly.

ealmloff (approved):

PR #5793, opened 2026-08-28
fix(router): skip the query `?` when the query serializes empty

### LINKED ISSUE #5792: Router routes with query segments serialize a trailing `?` even when the query is empty
When a route has a query segment, its `Display` impl always writes the `?`, even when nothing follows it. With a spread segment (`?:..params`) and a params struct whose `Display` renders empty, the route serializes as `/translations?`, which then shows up in the address bar and in copied links. The same happens with a named segment on an `Option` field: `#[route("/reset?:token")]` with `token: None` gives `/reset?`, even though the `None` pair itself is correctly elided.

As best I can tell the `?` is written unconditionally in the router macro's query writer (`packages/router-macro/src/query.rs`), before the field's rendered output is known, and the same code is on main. It looks like an oversight rather than a design choice, since the pair elision (skipping `None` pairs and their `&` separators) is already there.

Repro on 0.7.10, with `dioxus = { version = "=0.7.10", features = ["router"] }`:

```rust
use dioxus::prelude::*;
use dioxus::router::FromQuery;
use std::fmt::{self, Display, Formatter};

#[derive(Debug, Clone, PartialEq, Default)]
struct Params {
    search: Option<String>,
}

impl Display for Params {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        match &self.search {
            Some(search) => write!(f, "search={search}"),
            None => Ok(()),
        }
    }
}

impl FromQuery for Params {
    fn from_query(query: &str) -> Self {
        let search = query
            .split('&')
            .filter_map(|segment| segment.split_once('='))
            .find(|(key, _)| *key == "search")
            .map(|(_, value)| value.to_owned())
            .filter(|value| !value.is_empty());
        Self { search }
    }
}

#[derive(Debug, Clone, PartialEq, Routable)]
enum Route {
    #[route("/translations?:..params")]
    Translations { params: Params },
    #[route("/reset?:token")]
    Reset { token: Option<String> },
}

#[component]
fn Translations(params: Params) -> Element {
    rsx! {}
}

#[component]
fn Reset(token: Option<String>) -> Element {
    rsx! {}
}

fn main() {
    let empty = Route::Translations { params: Params::default() };
    let filled = Route::Translations {
        params: Params { search: Some("zone".to_owned()) },
    };
    let reset_none = Route::Reset { token: None };
    println!("{empty}");      // /translations?
    println!("{filled}");     // /translations?search=zone
    println!("{reset_none}"); // /reset?
}
```

Expected: `/translations` and `/reset`. Actual: `/translations?` and `/reset?`.

Cosmetic, but visible on every navigation.

emilbonnek:
Fixes #5792.

Routes with a query segment wrote the `?` unconditionally, so a spread segment (`?:..params`) whose Display renders empty serialized as `/translations?`, and a named segment with only None options as `/reset?`.

This buffers the named pairs and writes the `?` only when something follows it. As far as I can tell that also fixes a related edge where a trailing `&` was left behind when a later pair elided (`?a=1&` for `(Some(1), None)`), since each pair wrote its own separator without knowing what came after. The spread form now checks the rendered string before writing.

Added a regression test covering both forms, round-tripping through FromStr. The existing router tests pass unchanged.

### DIFF
--- a/packages/router-macro/src/query.rs
+++ b/packages/router-macro/src/query.rs
@@ -39,10 +39,17 @@ impl QuerySegment {
             QuerySegment::Single(segment) => segment.write(),
             QuerySegment::Segments(segments) => {
                 let mut tokens = TokenStream2::new();
-                tokens.extend(quote! { write!(f, "?")?; });
-                for (i, segment) in segments.iter().enumerate() {
-                    tokens.extend(segment.write(i == segments.len() - 1));
+                // Buffer the pairs so the `?` (and each joining `&`) is only
+                // written when something actually follows it.
+                tokens.extend(quote! { let mut __query_string = String::new(); });
+                for segment in segments {
+                    tokens.extend(segment.write());
                 }
+                tokens.extend(quote! {
+                    if !__query_string.is_empty() {
+                        write!(f, "?{}", __query_string)?;
+                    }
+                });
                 tokens
             }
         }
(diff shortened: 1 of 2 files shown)


sarendipitee:
Thank you! This has actually been bugging me!

rogusdev:
Consider instead, that covers more: https://github.com/DioxusLabs/dioxus/pull/5827

PR #5798, opened 2026-08-31
feat(cli): add independent [bundle.rpm] settings

nikuda:
RPM bundling no longer reads Debian packaging settings. `[bundle.rpm]` is a first-class packager config next to `[bundle.deb]`, with RPM-native relation names. Shared Linux payload (the `.desktop` template) lives on `[linux]`.

Debian and Fedora package names are not interchangeable (e.g. `libwebkit2gtk-4.1-0` vs `webkit2gtk4.1`), so one `depends` list cannot produce a correct `.deb` and `.rpm`. Previously `bundle_linux_rpm` copied `[bundle.deb] depends` into RPM `Requires`.

## Config

```toml
[linux]
desktop_template = "linux.desktop"

[bundle.deb]
depends = ["libwebkit2gtk-4.1-0"]

[bundle.rpm]
requires = ["webkit2gtk4.1"]
recommends = ["shared-mime-info"]
provides = ["myapp"]
conflicts = ["myapp-nightly"]
obsoletes = ["myapp-legacy"]
```

`[bundle.rpm]` also keeps files and maintainer scripts (`pre_install_script`, `post_install_script`, `pre_remove_script`, `post_remove_script`).

Relation mapping:

| Deb | RPM |
|---|---|
| `depends` | `requires` |
| `recommends` | `recommends` |
| `provides` | `provides` |
| `conflicts` | `conflicts` |
| `replaces` | `obsoletes` |

`suggests` / `enhances` / `supplements` are not added here. No default WebKit/GTK runtime deps are injected; that can be a follow-up gated on webview vs native.

## Breaking changes

Hard cut, no compatibility shims.

- **RPM `Requires` only from `[bundle.rpm] requires`.** Unset means no user `Requires` (plus `rpmlib(...)` format markers). There is no fallback to `[bundle.deb] depends`.
- **RPM files and scripts only from `[bundle.rpm]`.** The RPM bundler no longer reads Debian `files` or maintainer scripts.
- **`.desktop` template only from `[linux] desktop_template`.** Used by `.deb`, `.rpm`, and AppImage. Format-level `desktop_template` on `[bundle.deb]` is gone.
- **`[linux.deb]` is removed.** It was unused (bundler never read it). Packaging stays under `[bundle.deb]` / `[bundle.rpm]`, matching `[bundle.macos]` / `[bundle.windows]`.
- **`DebianSettings` renamed to `DebSettings`** (the `.deb` packager, not the Debian distro). TOML key remains `[bundle.deb]`.

## Test plan

- [x] `bundle_rpm_settings_parse` covers `requires` / `recommends` / `provides` / `conflicts` / `obsoletes`
- [x] Linux smoke (`bundle-linux-smoke`) builds hotdog as AppImage, `.deb`, and `.rpm`
- [x] Smoke inspects `dpkg-deb -f Depends` and `rpm -qp --requires` (minus `rpmlib`)
- [x] Smoke extracts `.deb`, `.rpm`, and AppImage (`--appimage-extract`) and greps the custom `[linux]` desktop-template test id

### DIFF
--- a/packages/cli/src/bundler/linux.rs
+++ b/packages/cli/src/bundler/linux.rs
@@ -222,7 +222,7 @@ impl BundleContext<'_> {
     /// 6. Copy configured sidecar binaries into `/usr/bin` so RPM payloads match the
     ///    other Linux formats.
     /// 7. Add custom files and maintainer scripts from `[bundle.rpm]`.
-    /// 8. Attach `Requires` from `[bundle.rpm] depends`.
+    /// 8. Attach `Requires` from `[bundle.rpm] requires`.
     /// 9. Serialize the final package to disk and remove the temporary staging area.
     ///
     /// The resulting artifact is written to `project_out_directory()/bundle/rpm`.
@@ -416,9 +416,9 @@ impl BundleContext<'_> {
             builder = builder.post_uninstall_script(content);
         }
 
-        if let Some(deps) = &rpm_settings.depends {
-            for dep in deps {
-                builder = builder.requires(rpm::Dependency::any(dep));
+        if let Some(requires) = &rpm_settings.requires {
+            for require in requires {
+                builder = builder.requires(rpm::Dependency::any(require));
             }
         }
 
--- a/packages/cli/src/bundler/linux.rs
+++ b/packages/cli/src/bundler/linux.rs
@@ -222,7 +222,7 @@ impl BundleContext<'_> {
     /// 6. Copy configured sidecar binaries into `/usr/bin` so RPM payloads match the
     ///    other Linux formats.
     /// 7. Reuse Debian-style custom files and maintainer scripts where applicable.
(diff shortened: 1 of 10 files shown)


packages/cli/src/bundler/linux.rs
```diff
@@ -369,24 +370,7 @@ impl BundleContext<'_> {
         }
 
         let crate_dir = self.crate_dir();
-        for (dest_path, src_path) in &rpm_settings.files {
-            let src = if src_path.is_absolute() {
-                src_path.clone()
-            } else {
-                crate_dir.join(src_path)
-            };
-            if src.exists() {
```
nikuda, line 378: This change looks bigger than it is, `src.exists()` check was silently not passing paths that should be passed to RPM builder. Passing them to the RPM builder means the build fails with the source and destination paths shown (rather then being silently ignored)

PR #5851, opened 2026-09-20
Use the empty namespace by default on HTML attributes

hovinen:
The function `set_attribute_inner` was previuosly passing the `ns` parameter unchanged to `qual_name`. If `ns` was unset, this defaulted to `html`. That behaviour does not match the DOM spec, which says that the namespace should be empty if not specified; see https://dom.spec.whatwg.org/#dom-element-setattribute.

This affects dioxus-test when the data-testid attribute is used to match elements. The query uses the CSS selector `[data-testid="<valud>"]`, so the underying CSS selectors implementation must be able to find the attribute.

Until recently, this was harmless, because `BlitzNode::attr_matches` was not attribute-aware. That was recently
[changed](https://github.com/DioxusLabs/blitz/commit/b75cca2c3187c170e77b2a127a12c30b56d8a255) in order to match the [CSS selectors spec](https://www.w3.org/TR/selectors-4/#attrnmsp).

This fixes the problem by defaulting `ns` to the empty namespace rather than `html` in `set_attribute_inner`.


nicoburns (approved):

nicoburns:
Good catch - we may end up with a few of these namespace bugs as we start implementing SVG support in Blitz.

